Issue #18, opened 2012-10-05
Logger does not print Java 7 suppressed exceptions


dain:

electrum:
This is fixed with the new logger in 0.107.

PR #259, opened 2015-01-05
Remove dependency on SLF4J

dain:


log-manager/src/main/java/io/airlift/log/Level.java
```diff
@@ -0,0 +1,77 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+enum Level
+{
+    OFF {
+        @Override
+        protected java.util.logging.Level toJulLevel()
+        {
+            return java.util.logging.Level.OFF;
```
electrum: Why not make this a constructor parameter?

log-manager/src/main/java/io/airlift/log/Level.java
```diff
@@ -0,0 +1,77 @@
...
+            return INFO;
+        }
+        return DEBUG;
+    }
+
+    static Level parse(String name)
+    {
+        checkNotNull(name, "name is null");
+        name = name.toUpperCase(Locale.US);
+        for (Level level : values()) {
```
electrum:
```java
try {
    return Level.valueOf(name);
}
catch (IllegalArgumentException e) {
    return DEBUG;
}

Comment (martint) [inline log-manager/src/main/java/io/airlift/log/Level.java]: I don't think we should mask errors.

Comment (electrum) [inline log-manager/src/main/java/io/airlift/log/Level.java]: This is the same behavior of the logback `Level.toLevel()` method that we currently use when processing the log levels file. I was simply suggesting a simpler implementation.

Comment (electrum) [inline log-manager/src/main/java/io/airlift/log/Level.java]: Although I would agree that it would be better to fail than default to `DEBUG`.

File: log-manager/pom.xml [hunk 3]
```diff
@@ -27,50 +27,42 @@
...
-            <artifactId>jcl-over-slf4j</artifactId>
-            <scope>runtime</scope>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-core</artifactId>
         </dependency>
 
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
+            <groupId>com.google.inject</groupId>
```
electrum: Can you keep the dependencies in the same order so the diff is easier to read?

log-manager/src/main/java/io/airlift/log/Level.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+    };
+
+    abstract java.util.logging.Level toJulLevel();
+
+    static Level fromJulLevel(java.util.logging.Level level)
+    {
+        // Convert any implementation of Level based on the int value
+        int levelValue = level.intValue();
+        if (levelValue == java.util.logging.Level.OFF.intValue()) {
```
electrum: These could be added as constants so this can use a switch

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -194,50 +114,20 @@ public void setLevels(File file)
...
 
+    @SuppressWarnings("MethodMayBeStatic")
     public void setLevel(String loggerName, Level level)
     {
-        ch.qos.logback.classic.Logger logger = (ch.qos.logback.classic.Logger) LoggerFactory.getLogger(loggerName);
-        logger.setLevel(ch.qos.logback.classic.Level.toLevel(level.toString()));
+        java.util.logging.Logger.getLogger(loggerName).setLevel(level.toJulLevel());
     }
 
+    @SuppressWarnings("MethodMayBeStatic")
```
electrum: This isn't necessary -- it is called from a non-static context

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -194,50 +114,20 @@ public void setLevels(File file)
...
-                String newName = tempFile.getName().substring(0, tempFile.getName().length() - TEMP_FILE_EXTENSION.length());
-                File newFile = new File(tempFile.getParent(), newName + LOG_FILE_EXTENSION);
-                if (tempFile.renameTo(newFile)) {
-                    log.info("Recovered temp file: %s", tempFile);
-                }
-                else {
-                    log.warn("Could not rename temp file [%s] to [%s]", tempFile, newFile);
-                }
-            }
+            java.util.logging.Logger.getLogger(loggerName).setLevel(level.toJulLevel());
```
electrum: Call the above `setLevel()` here

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -194,50 +114,20 @@ public void setLevels(File file)
         processLevels(properties);
     }
 
+    @SuppressWarnings("MethodMayBeStatic")
```
electrum: If these can be static, why not make them static?

log-manager/src/main/java/io/airlift/log/LoggingMBean.java
```diff
@@ -15,58 +15,77 @@
...
 
-import ch.qos.logback.classic.Logger;
-import ch.qos.logback.classic.LoggerContext;
-import org.slf4j.LoggerFactory;
 import org.weakref.jmx.Managed;
 
+import java.util.Collections;
 import java.util.Map;
+import java.util.TreeMap;
+import java.util.logging.*;
```
electrum: Star import

log-manager/src/main/java/io/airlift/log/LoggingMBean.java
```diff
@@ -15,58 +15,77 @@
...
+import java.util.logging.Logger;
 
-import static ch.qos.logback.classic.Level.toLevel;
-import static com.google.common.collect.Maps.newTreeMap;
-import static org.slf4j.Logger.ROOT_LOGGER_NAME;
-
+@SuppressWarnings("MethodMayBeStatic")
```
electrum: Why?

log-manager/src/main/java/io/airlift/log/LoggingMBean.java
```diff
@@ -15,58 +15,77 @@
...
 public class LoggingMBean
 {
     @Managed
     public String getLevel(String loggerName)
     {
-        return getLogger(loggerName).getEffectiveLevel().toString();
+        java.util.logging.Logger logger = getLogger(loggerName);
```
electrum: Inline this

log-manager/src/main/java/io/airlift/log/LoggingMBean.java
```diff
@@ -15,58 +15,77 @@
...
     {
-        getLogger(loggerName).setLevel(toLevel(newLevel));
+        getLogger(loggerName).setLevel(julLevel(newLevel));
     }
 
     @Managed
     public String getRootLevel()
     {
-        return getLogger(ROOT_LOGGER_NAME).getEffectiveLevel().toString();
+        return getLevel("");
```
electrum: Add a constant for this

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,139 @@
...
+    @Override
+    public void publish(LogRecord record)
+    {
+        if (!isLoggable(record)) {
+            return;
+        }
+        String message;
+        try {
+            message = getFormatter().format(record);
+        } catch (Exception e) {
```
electrum: Formatting

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,139 @@
...
+        File logPathFile = new File(logPath).getParentFile();
+        File[] tempFiles = logPathFile.listFiles((dir, name) -> name.endsWith(TEMP_FILE_EXTENSION));
+
+        if (tempFiles != null) {
+            for (File tempFile : tempFiles) {
+                String newName = tempFile.getName().substring(0, tempFile.getName().length() - TEMP_FILE_EXTENSION.length());
+                File newFile = new File(tempFile.getParent(), newName + LOG_FILE_EXTENSION);
+
+                if (!tempFile.renameTo(newFile)) {
+                    reportError("Could not rename temp file [" + tempFile + "] to [" + newFile + "]", null, ErrorManager.OPEN_FAILURE);
```
electrum: Use String format

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,139 @@
...
+    private void recoverTempFiles(String logPath)
+    {
+        // logback has a tendency to leave around temp files if it is interrupted
+        // these .tmp files are log files that are about to be compressed.
+        // This method recovers them so that they aren't orphaned
+
+        File logPathFile = new File(logPath).getParentFile();
+        File[] tempFiles = logPathFile.listFiles((dir, name) -> name.endsWith(TEMP_FILE_EXTENSION));
+
+        if (tempFiles != null) {
```
electrum: Return early?

log-manager/src/main/java/io/airlift/log/Level.java
dain: This can't be a switch because JUL supports any int value for the `level`, so we must match on the defacto ranges.

log-manager/src/main/java/io/airlift/log/Level.java
dain: I vote we simply fail.  This method is only called while parsing the level file during startup or when setting a level via JMX.  Both of these operations can fail.

dain: And in that case we can just get rid of this method.

log-manager/src/main/java/io/airlift/log/Logging.java
dain: This method has been an instance method for a while.  I can change it to a static method, but it will break any current users.  Is this a change we want to make?

log-manager/src/main/java/io/airlift/log/LoggingMBean.java
dain: I guess I was lazy.  All of the methods can be static because they use `Logging` which is static.  I'll just tag the methods.

dain:
Updated

dain:
This should fix #18

log-manager/src/main/java/io/airlift/log/Level.java
electrum: Ah, I missed that the checks below are range checks.

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -48,22 +35,12 @@
...
-    private final ch.qos.logback.classic.Logger root;
-    private final static Logger log = Logger.get(Logging.class);
-    private OutputStreamAppender<ILoggingEvent> consoleAppender;
-
-    private static final String TEMP_FILE_EXTENSION = ".tmp";
-    private static final String LOG_FILE_EXTENSION = ".log";
-
+    private static final Logger log = Logger.get(Logging.class);
+    private static final java.util.logging.Logger ROOT = java.util.logging.Logger.getLogger("");
+    /* @GuardedBy("this") */
```
electrum: Why?

log-manager/src/main/java/io/airlift/log/LoggingMBean.java
```diff
@@ -15,58 +15,76 @@
...
-import static ch.qos.logback.classic.Level.toLevel;
-import static com.google.common.collect.Maps.newTreeMap;
-import static org.slf4j.Logger.ROOT_LOGGER_NAME;
+import java.util.TreeMap;
+import java.util.logging.LogManager;
+import java.util.logging.Logger;
 
 public class LoggingMBean
 {
+    private static final String ROOT_LOGGER_NAME = "";
```
electrum: This should be a public constant on `Logger`.

electrum: Or should we add "root" methods to `Logger`?

log-manager/src/main/java/io/airlift/log/LoggingMBean.java
```diff
@@ -15,58 +15,76 @@
...
-        return getLogger(loggerName).getEffectiveLevel().toString();
+        return getEffectiveLevel(getLogger(loggerName)).toString();
     }
 
     @Managed
+    @SuppressWarnings("MethodMayBeStatic")
     public void setLevel(String loggerName, String newLevel)
     {
-        getLogger(loggerName).setLevel(toLevel(newLevel));
+        getLogger(loggerName).setLevel(Level.valueOf(newLevel.toUpperCase(Locale.US)).toJulLevel());
```
electrum: Why not

```java
logging.setLevel(loggerName, Level.valueOf(newLevel.toUpperCase(Locale.US)));

Comment (electrum) [inline log-manager/src/main/java/io/airlift/log/LoggingMBean.java]: This class breaks the encapsulation of `Logging`. It seems like either everything here should go through `Logging`, or we simply say this class is part of the public interface. And it seems we already use the latter, so this is fine.

File: log-manager/src/main/java/io/airlift/log/OutputStreamHandler.java [hunk 18]
```diff
@@ -0,0 +1,69 @@
+package io.airlift.log;
+
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+import java.util.logging.Handler;
+import java.util.logging.LogRecord;
+
+import static com.google.common.base.Charsets.UTF_8;
```
electrum: Use `StandardCharsets`

log-manager/src/main/java/io/airlift/log/OutputStreamHandler.java
```diff
@@ -0,0 +1,69 @@
...
+    public void publish(LogRecord record)
+    {
+        if (!isLoggable(record)) {
+            return;
+        }
+
+        String message;
+        try {
+            message = getFormatter().format(record);
+        } catch (Exception e) {
```
electrum: This should be `RuntimeException`. `format()` doesn't throw any checked exceptions.

log-manager/src/main/java/io/airlift/log/OutputStreamHandler.java
```diff
@@ -0,0 +1,69 @@
...
+            reportError(null, e, FORMAT_FAILURE);
+            return;
+        }
+
+        try {
+            writer.write(message);
+        } catch (Exception e) {
```
electrum: `IOException`

log-manager/src/main/java/io/airlift/log/OutputStreamHandler.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    @Override
+    public void flush()
+    {
+        try {
+            writer.flush();
+        }
+        catch (Exception e) {
```
electrum: `IOException`

log-manager/src/main/java/io/airlift/log/StaticFormatter.java
```diff
@@ -0,0 +1,79 @@
...
+
+        if (record.getThrown() != null) {
+            stringWriter.append('\n');
+            record.getThrown().printStackTrace(new PrintWriter(stringWriter));
+            stringWriter.append('\n');
+        }
+
+        stringWriter.append('\n');
+        String message = stringWriter.toString();
+        return message;
```
electrum: Inline

log-manager/src/main/java/io/airlift/log/OutputStreamHandler.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    @Override
+    public void close()
+    {
+        try {
+            writer.flush();
+        }
+        catch (Exception e) {
```
electrum: `IOException`

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,143 @@
...
+import static java.util.logging.ErrorManager.WRITE_FAILURE;
+
+final class RollingFileHandler
+        extends Handler
+{
+    private static final String TEMP_FILE_EXTENSION = ".tmp";
+    private static final String LOG_FILE_EXTENSION = ".log";
+
+    // Tab-separated
+    // Time, ip, method, url, user, agent, response code, request length, response length, response time
```
electrum: Remove these copied comments

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,143 @@
...
+    public void publish(LogRecord record)
+    {
+        if (!isLoggable(record)) {
+            return;
+        }
+        String message;
+        try {
+            message = getFormatter().format(record);
+        }
+        catch (Exception e) {
```
electrum: Same comments about exception types

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,143 @@
...
+    private static final String TEMP_FILE_EXTENSION = ".tmp";
+    private static final String LOG_FILE_EXTENSION = ".log";
+
+    // Tab-separated
+    // Time, ip, method, url, user, agent, response code, request length, response length, response time
+    private final RollingFileAppender<String> fileAppender;
+
+    public RollingFileHandler(String filename, int retainDays, long maxSizeInBytes)
+    {
+        setFormatter(new StaticFormatter());
```
electrum: Extract an `AbstractHandler`?

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,143 @@
...
+        fileAppender.start();
+    }
+
+    @Override
+    public void publish(LogRecord record)
+    {
+        if (!isLoggable(record)) {
+            return;
+        }
+        String message;
```
electrum: Newline

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,143 @@
...
+
+    private final class StringEncoder
+            extends EncoderBase<String>
+    {
+        @Override
+        public void doEncode(String event)
+                throws IOException
+        {
+            outputStream.write(event.getBytes(UTF_8));
+            // necessary if ResilientFileOutputStream is buffered
```
electrum: Should this just say "necessary if output stream is buffered"

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,143 @@
...
+        public void close()
+                throws IOException
+        {
+            outputStream.flush();
+        }
+    }
+
+    private void recoverTempFiles(String logPath)
+    {
+        // logback has a tendency to leave around temp files if it is interrupted
```
electrum: These should be sentences

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,143 @@
...
+        if (tempFiles == null) {
+            return;
+        }
+
+        for (File tempFile : tempFiles) {
+            String newName = tempFile.getName().substring(0, tempFile.getName().length() - TEMP_FILE_EXTENSION.length());
+            File newFile = new File(tempFile.getParent(), newName + LOG_FILE_EXTENSION);
+
+            if (!tempFile.renameTo(newFile)) {
+                reportError(String.format("Could not rename temp file [%s] to [%s]", tempFile, newFile), null, ErrorManager.OPEN_FAILURE);
```
electrum: Static import

log-manager/src/test/java/io/airlift/log/TestLogging.java
```diff
@@ -128,14 +128,14 @@ public void testChildLevelOverridesParent()
...
-        logging.setLevel("testChildLevelOverridesParent.child", Logging.Level.ERROR);
+        logging.setLevel("testChildLevelOverridesParent", Level.DEBUG);
+        logging.setLevel("testChildLevelOverridesParent.child", Level.ERROR);
         assertFalse(logger.isDebugEnabled());
         assertFalse(logger.isInfoEnabled());
     }
 
 
-    private File createTempDir()
+    private static File createTempDir()
```
electrum: Use Guava `Files.createTempDir()`?

log-manager/src/main/java/io/airlift/log/OutputStreamHandler.java
electrum: Formatting is wrong

PR #262, opened 2015-01-10
Add java-home option to select java version to use.

sasakitoa:
When two or more java version are installed, I want to select java version to use.
We can select it to use this java-home option.


electrum:
This seems to be achievable by setting PATH to have the correct java binary before running the launcher, which is how we do it. Does this work for you?=

electrum:
This seems to be achievable by setting `PATH` to have the correct `java` binary before running the launcher, which is how we do it. Does this work for you?

sasakitoa:
Thanks for your advice, and sorry for my late reply.
It worked well to set correct PATH, and I will close this pull request, soon.

PR #267, opened 2015-01-23
Interrupt the BodyGenerator thread when writes fail

johngmyers:
We've been having outages where HttpClient stops making forward progress.

First, we see a bunch of EOFExceptions like:

```
java.io.EOFException: HttpConnectionOverHTTP@709941a1(l:/10.94.102.205:57021 <-> r:pulse-2107208448.us-east-1.elb.amazonaws.com/54.85.56.180:8080)
        at org.eclipse.jetty.client.http.HttpReceiverOverHTTP.earlyEOF(HttpReceiverOverHTTP.java:267) ~[jetty-client-9.2.6.v20141205.jar:9.2.6.v20141205]
        at org.eclipse.jetty.http.HttpParser.parseNext(HttpParser.java:1312) ~[jetty-http-9.2.6.v20141205.jar:9.2.6.v20141205]
        at org.eclipse.jetty.client.http.HttpReceiverOverHTTP.shutdown(HttpReceiverOverHTTP.java:170) ~[jetty-client-9.2.6.v20141205.jar:9.2.6.v20141205]
        at org.eclipse.jetty.client.http.HttpReceiverOverHTTP.readAndParse(HttpReceiverOverHTTP.java:117) ~[jetty-client-9.2.6.v20141205.jar:9.2.6.v20141205]
        at org.eclipse.jetty.client.http.HttpReceiverOverHTTP.process(HttpReceiverOverHTTP.java:70) ~[jetty-client-9.2.6.v20141205.jar:9.2.6.v20141205]
        at org.eclipse.jetty.client.http.HttpReceiverOverHTTP.receive(HttpReceiverOverHTTP.java:65) ~[jetty-client-9.2.6.v20141205.jar:9.2.6.v20141205]
        at org.eclipse.jetty.client.http.HttpChannelOverHTTP.receive(HttpChannelOverHTTP.java:75) ~[jetty-client-9.2.6.v20141205.jar:9.2.6.v20141205]
        at org.eclipse.jetty.client.http.HttpConnectionOverHTTP.onFillable(HttpConnectionOverHTTP.java:100) ~[jetty-client-9.2.6.v20141205.jar:9.2.6.v20141205]
        at org.eclipse.jetty.io.AbstractConnection$2.run(AbstractConnection.java:540) ~[jetty-io-9.2.6.v20141205.jar:9.2.6.v20141205]
        at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:626) ~[jetty-util-9.2.6.v20141205.jar:9.2.6.v20141205]
        at org.eclipse.jetty.util.thread.QueuedThreadPool$3.run(QueuedThreadPool.java:546) ~[jetty-util-9.2.6.v20141205.jar:9.2.6.v20141205]
        at java.lang.Thread.run(Thread.java:722) [na:1.7.0_17]
```

I have no idea what the underlying exception is, as Jetty only logs that at debug level.

Eventually the server hangs. There are ~150 threads blocked in either `JettyHttpClient$BodyGeneratorContentProvider$BodyGeneratorOutputStream.write` or `JettyHttpClient$BodyGeneratorContentProvider$BodyGeneratorOutputStream.close`, but nowhere near that number of open sockets.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
@@ -52,8 +52,6 @@ import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map.Entry;
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
@@ -61,6 +59,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -596,7 +596,7 @@ public class JettyHttpClient
         @Override
         public Iterator<ByteBuffer> iterator()
         {
-            final BlockingQueue<ByteBuffer> chunks = new ArrayBlockingQueue<>(16);
+            final ChunksQueue chunks = new ChunksQueue();
             final AtomicReference<Exception> exception = new AtomicReference<>();
 
             executor.execute(new Runnable()
@@ -604,21 +604,15 @@ public class JettyHttpClient
                 @Override
                 public void run()
                 {
+                    bodyGeneratorThread.set(Thread.currentThread());
+                    BodyGeneratorOutputStream out = new BodyGeneratorOutputStream(chunks);
                     try {
-                        bodyGeneratorThread.set(Thread.currentThread());
-                        BodyGeneratorOutputStream out = new BodyGeneratorOutputStream(chunks);
-                        try {
-                            bodyGenerator.write(out);
-                            out.close();
-                        }
-                        catch (Exception e) {
-                            exception.set(e);
-                            chunks.clear();
-                            chunks.add(EXCEPTION);
-                        }
+                        bodyGenerator.write(out);
+                        out.close();
                     }
-                    finally {
-                        bodyGeneratorThread.set(null);
+                    catch (Exception e) {
+                        exception.set(e);
+                        chunks.replaceAllWith(EXCEPTION);
                     }
                 }
             });
@@ -629,9 +623,9 @@ public class JettyHttpClient
         private final class BodyGeneratorOutputStream
                 extends OutputStream
         {
-            private final BlockingQueue<ByteBuffer> chunks;
+            private final ChunksQueue chunks;
 
-            private BodyGeneratorOutputStream(BlockingQueue<ByteBuffer> chunks)
+            private BodyGeneratorOutputStream(ChunksQueue chunks)
             {
                 this.chunks = chunks;
             }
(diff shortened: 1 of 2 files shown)


martint:
Are you seeing that issue only for POST requests? We've been trying to track down a problem where the client stops sending POSTs to specific hosts (GETs continue to work, though). I wonder if this is the same issue.

johngmyers:
We don't know if it's only POST/PUT requests, but it is known to affect discovery announcements and sending events.

johngmyers:
...and another service that uses POST with a BodyGenerator.

dain:
The blocking is likely caused by the code at https://github.com/airlift/airlift/blob/master/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java#L597
```java
            final BlockingQueue<ByteBuffer> chunks = new ArrayBlockingQueue<>(16);
```
This is an arbitrary, hard, limit on the number of pending chunks.  The Jetty ContentProvider doesn't have a "destroy" method, so we have no way to "free" blocked writers.  I think the only way around this is to make the queue unbounded, or maybe have a put with a time out and a fail the request.

johngmyers:
An unbounded queue would use excessive memory if the BodyGenerator created data faster than the connection accepted it. It is desirable to block the BodyGenerator when there is too much unconsumed data.

A put with a timeout could incorrectly fail a request that should succeed.

gwittel:
Would an async approach like Jetty's ```DeferredContentProvider```  work?

dain:
I totally agree, they are two bad approaches. What we really need is a way to tell the `ContentProvider`, that is "done" and will never be accessed again.  

My guess is what is really happening, is that the http request failed and the jetty client either never accessed the iterator or just stopped pulling from it.  Then the code writing to the output stream simply hangs.  If we had a notification when the request was "complete" we could, free the writers with an exception.

johngmyers:
That's what the commit in this pull request does.
An alternative would be to replace the `ArrayBlockingQueue` with a data structure where the consumer can cause the put call to throw an exception.

dain:
@johngmyers I think the problem still remains.  In certain scenarios, the jetty client won't even touch the body generator.  I think we need hook in to `execute` and `executeAsync` and when the request is "complete" we "destroy" the body generator.  That way we can assure writers are freed.

johngmyers:
@dain, I'm not following you. The `Runnable` to invoke the `BodyGenerator` isn't submitted until Jetty calls `BodyGeneratorContentProvider.iterator()`, whereupon that call will return the `ChunksIterator` that Jetty will by contract call `close()` on when the request is complete.

Why would we need to do a "destroy" operation on a `BodyGenerator` that has not been invoked?

johngmyers:
I added a commit that replaces the `ArrayBlockingQueue` with a bespoke class, as the previous method could interrupt the worker thread after it had returned from the `BodyGenerator`. The new commit also incorporates the fix in pull #256.

johngmyers:
With these fixes, the next thing we saw was a couple of threads in `BodyGeneratorContentProvider` blocked in `take()`, a whole lot of threads in a `ResponseHandler.handleException` blocked on the log file lock trying to log the failure, and a few in epollWait. No threads were running a `BodyGenerator`.

That suggests to me that we need a separate thread pool for running the BodyGenerator.

dain:
Sorry, I haven't had time to look at this yet.  I've been fighting fires since Saturday.

johngmyers:
Thinking about this some more, a separate pool for running the BodyGenerator isn't sufficient. If you have more open connections than max threads, then you can deadlock.

Consider a Jetty pool with max one thread plus a BodyGenerator pool with max one thread. Queue two requests, each with a BodyGenerator.  The first request will invoke `BodyGeneratorContentProvider.iterator()` which will submit the `BodyGeneratorOutputStream` Runnable to the BodyGenerator Executor. That will run the `BodyGenerator` and then block putting chunks into the queue. The Iterator thread will take some chunks from the queue, write them to the connection, then go async waiting for the socket buffer to drain.

The second request will then invoke `BodyGeneratorContentProvider.iterator()` which will submit the `BodyGeneratorOutputStream` Runnable to the BodyGenerator Executor. This Runnable will not be picked up because the BodyGenerator pool has all thread running the first request's BodyGenerator. The second request's `Iterator.computeNext()` call will block waiting to take a chunk, so the first request's iterator will never be picked up again. Deadlock.

johngmyers:
I'm thinking we might want to recode `HttpEventClient` to use a newly invented variant of `BodyGenerator` which doesn't need to be run on a different thread. Something where the new Generator type gets multiple `write()` calls and dribbles out a bit at a time.

johngmyers:
An API for a `BodyGenerator` replacement is in proofpoint/platform#388

johngmyers:
This change plus subsequent changes in Platform to replace `BodyGenerator` appear to have solved our problems with Jetty client deadlocking.

PR #270, opened 2015-02-04
New dense HyperLogLog format

martint:
This version tracks *all* overflows to avoid error compounding
when multiple HLL instances are merged.

For backward-compatibility, the implementation can still deserialize
the old representation, but will operate and serialize using the new
layout.


erichwang:
looks good

PR #273, opened 2015-02-09
Add support for binding default values for any configuration class

dain:
This change adds a configurable preprocessor between construction of the configuration class and setting of the configuration properties.  The preprocessor has full access to the configuration instance.  For example, the following sets the default `StoreConfig.ttl`:

```java
        ConfigBinder.configBinder(binder).bindConfigDefaults(StoreConfig.class, c -> {
            c.setTtl(new Duration(15, TimeUnit.MINUTES));
        });
```

Additionally, multiple `ConfigDefaults` can be bound, and each is applied in the order in which it was bound.  This allows for wrapper libraries to override global defaults, and for usages of the wrapper libraries to override the wrapper defaults.

### DIFF
--- a/configuration/pom.xml
+++ b/configuration/pom.xml
@@ -37,6 +37,11 @@
         </dependency>
 
         <dependency>
+            <groupId>com.google.inject.extensions</groupId>
+            <artifactId>guice-multibindings</artifactId>
+        </dependency>
+
+        <dependency>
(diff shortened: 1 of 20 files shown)


dain:
Rebased.  @electrum please review

electrum:
Mark the old binder as deprecated?

electrum:
Looks good

PR #279, opened 2015-02-10
Thread group name

dain:


martint:
You should document this in NEWS.

Looks good

PR #287, opened 2015-03-11
Wrap all exceptions as RejectedExecutionException

cberner:


dain:
The commit message is a bit vague

dain:
looks good

PR #288, opened 2015-03-12
Make the number of accept socket queue size configurable

komamitsu:
We occasionally run into this message in kernel log when using Presto and the presto workers got slow and unstable.

```
Mar 12 03:13:02 xxxxxxxx: [849534.116220] TCP: Possible SYN flooding on port 8081. Sending cookies.  Check SNMP counters.
Mar 12 03:36:02 xxxxxxxx: [850913.599336] TCP: Possible SYN flooding on port 8081. Sending cookies.  Check SNMP counters.
```

I guess that the accept socket queue and SYN-packet queue were full at that time. I saw the default value of `org.eclipse.jetty.server.ServerConnector#_acceptQueueSize` is 0 that is handled as 50 in `java.net.ServerSocket#bind(java.net.SocketAddress, int)`. I feel it can be too small depending on circumstances.

The number of accept socket queue size had better to be configurable, I think.


komamitsu:
Ping

martint:
Merged, thanks!

PR #289, opened 2015-03-12
Fix HTTP server test on IPv6-only machines

electrum:
InetAddress.getHostAddress() returns a scope ID for IPv6 addresses.


dain:
looks good

PR #292, opened 2015-03-26
Destroy Jetty client connections stuck in closing

dain:


electrum:
Do we want to say "multiple times" instead of "4 times"?

electrum:
Looks good

PR #294, opened 2015-04-16
Move io.airlift.log.Level back to public

wdong87:
Level is used by multiple Presto projects and should be public. Changes to Presto will be submitted subsequently (need to move to airlift 0.107-SNAPSHOT).


electrum:
Good catch, this should be public given that it is a requirement to call the public method `setLevel()` on `Logging`. However, I think the `toJulLevel()` method should be package-private.

electrum:
Merged, thanks!

PR #296, opened 2015-04-17
Replace lambda with static function reference

electrum:
Work around jmxutils / Paranamer crashing on MBean classes that
contain Java 8 lambda expressions.


dain:
looks good

PR #301, opened 2015-04-29
correct the way airlift selects the IPv6 address to use

zaynoon:
Be cool if the InetAddress.getLocalHost() returns an IPv6 address and filter out link-local IPv6 addresses from the "good" addresses.

### DIFF
--- a/node/src/main/java/io/airlift/node/NodeInfo.java
+++ b/node/src/main/java/io/airlift/node/NodeInfo.java
@@ -340,6 +340,7 @@ public class NodeInfo
     private static boolean isGoodAddress(InetAddress address)
     {
         return !address.isAnyLocalAddress() &&
+                !address.isLinkLocalAddress() &&
                 !address.isLoopbackAddress() &&
                 !address.isMulticastAddress();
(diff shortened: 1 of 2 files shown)


node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -261,7 +261,8 @@ private static InetAddress findPublicIp()
         InetAddress localAddress = null;
         try {
             localAddress = InetAddress.getLocalHost();
-            if (isV4Address(localAddress) && getGoodAddresses().contains(localAddress)) {
+            if ((isV4Address(localAddress) || isV6Address(localAddress))
```
electrum: I don't think we want this, since we prefer to choose an IPv4 address if it exists. If we do want this, these checks can be removed, and the comments and documentation for this class need to be updated.

electrum:
Let's leave the existing logic and just filter out link local addresses. Title the commit "Exclude link local addresses in NodeInfo" and update `CHANGES`

dain:
merged

PR #304, opened 2015-05-06
Set configuration factory in late bound modules

dain:
Late bound configuration aware modules will have providers without
the configuration factory set, so set it manually when modules are
registered.


martint:
looks good

PR #306, opened 2015-05-07
Cleanup single threaded tests

dain:


concurrent/src/test/java/io/airlift/concurrent/TestBoundedExecutor.java
```diff
@@ -17,14 +17,14 @@
...
-    @BeforeMethod
+    @BeforeClass
     public void setUp()
             throws Exception
     {
         executorService = Executors.newCachedThreadPool();
     }
 
-    @AfterMethod
+    @AfterClass
```
electrum: Add alwaysRun to all of these.

electrum:
Looks good

concurrent/src/test/java/io/airlift/concurrent/TestBoundedExecutor.java
dain: do we need that for `@AfterMethod` and `@AfterClass`

dain:
I added `(alwaysRun = true)` to all of the `@AfterMethod` and `@AfterClass`.  Does this look good?

electrum:
I think you need it on anything that does cleanup if you want the cleanup to happen even if the test fails. Looks good.

PR #308, opened 2015-05-08
Fix bootstrap test

electrum:
TestNG assumes every method is a test method if the class is annotated
with @Test. Unfortunately, annotating the class is the only way to
specifiy single-threaded behavior.


dain:
looks good

PR #314, opened 2015-05-18
Fixes for http client content provider

martint:


dain:
Can you try integrating the test cases @johngmyers wrote in #267 and maybe #256

dain:
@martint do we still need this one?

PR #319, opened 2015-05-27
Fix warnings

electrum:


stats/src/main/java/io/airlift/stats/cardinality/DenseHll.java
```diff
@@ -360,7 +360,7 @@ private void setDelta(int bucket, int value)
 
         // clear the old value
         byte clearMask = (byte) (BUCKET_MASK << shiftForBucket(bucket));
-        deltas[slot] &= ~clearMask;
+        deltas[slot] &= (byte) ~clearMask;
```
martint: This cast is unnecessary. `deltas` is a byte array, so it's irrelevant that the negated mask is promoted to an int. I think this is just a case of an overly aggressive inspection.

electrum: Would you prefer to suppress the inspection?

dain:
looks good

PR #320, opened 2015-05-27
Add flag to disable request stat recording

cberner:


http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -46,6 +46,19 @@
     private HostAndPort socksProxy;
     private String keyStorePath = System.getProperty(JAVAX_NET_SSL_KEY_STORE);
     private String keyStorePassword = System.getProperty(JAVAX_NET_SSL_KEY_STORE_PASSWORD);
+    private boolean recordRequestStats = true;
```
electrum: Move the field to the top, so the field and getter/setter order matches.

electrum:
This looks fine, but I don't like the idea of completely disabling stats. I'm thinking we could have a "simple" request stats object that keeps at least a request count. It could track total time and bytes, though those would eventually overflow.

dain:
I think @martint should figure out why the stats are so expensive.

cberner:
I filed https://github.com/airlift/airlift/issues/321 . I'll hold off on merging this

dain:
Superseded by #323

PR #322, opened 2015-06-02
Add more MoreFutures and AsyncQueue

dain:


concurrent/src/main/java/io/airlift/concurrent/AsyncQueue.java
```diff
@@ -0,0 +1,118 @@
...
+        if (finishedIndex >= 0) {
+            // add the finish marker back to the queue so future callers will not block indefinitely
+            queue.add(FinishedMarker.FINISHED);
+            // drop all elements after the finish marker (this shouldn't happen in a normal finish, but be safe)
+            elements = elements.subList(0, finishedIndex);
+        }
+
+        // if the queue is empty and the current future is finished, create a new one so
+        // readers can be notified when the queue has elements to read
+        if (queue.isEmpty()) {
```
erichwang: there is a race condition here:
1) reader checks queue.isEmpty() => true
2) writer adds element to the queue
3) writer sets future to complete
4) reader swaps the old future to the new future since it the old is complete
5) reader waits on the batch, but never realizes that it already has data available

The same thing can happen if this races with the finish() call, the readers can miss the FINISH signal.

dain:
@martint this is needed by https://github.com/facebook/presto/pull/3016

martint:
looks good

PR #323, opened 2015-06-06
Reduce contention when updating TimeDistribution

martint:
This change implements a simple lock striping strategy to help
reduce contention when updating TimeDistribution from multiple threads.
TimeDistribution now maintains a set of QuantileDigests that get updated
independently and merged lazily every so often.

### DIFF
--- a/stats/src/main/java/io/airlift/stats/TimeDistribution.java
+++ b/stats/src/main/java/io/airlift/stats/TimeDistribution.java
@@ -11,6 +11,7 @@ import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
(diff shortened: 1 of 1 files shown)


johngmyers:
Any particular reason `Distribution` doesn't need this treatment as well?

Consider making a `StripedQuantileDigest`?

dain:
other than John's concerns, looks good

martint:
Can't you use QuantileDigest directly? Or are you leaning on the fact that this class is thread-safe and has the convenience methods to translate the time units?

Would adding a method like "toQuantileDigest" that does the merge and returns an instance of the quantile digest be sufficient?

dain:
@martint in the mean time, how about we land this?

stats/src/main/java/io/airlift/stats/TimeDistribution.java
```diff
@@ -144,6 +164,23 @@ private double convertToUnit(long nanos)
         return nanos * 1.0 / unit.toNanos(1);
     }
 
+    private QuantileDigest mergeAndGetIfNeeded()
+    {
+        long now = System.currentTimeMillis();
```
martint: Use System.nanoTime()

stats/src/main/java/io/airlift/stats/TimeDistribution.java
```diff
@@ -46,67 +54,79 @@ public TimeDistribution(double alpha, TimeUnit unit)
...
+        this.alpha = alpha;
+        this.count = new DecayCounter(alpha);
+        this.merged = new QuantileDigest(MAX_ERROR, alpha);
+        for (int i = 0; i < SEGMENTS; i++) {
+            locks[i] = new Object();
+            partials[i] = new QuantileDigest(MAX_ERROR, alpha);
+        }
     }
 
     public synchronized void add(long value)
```
Dith3r, line 67: Can we drop `synchronized` now?

PR #326, opened 2015-06-19
Preserve underlying exception if GSS context fails to initialize

martint:


http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -152,14 +152,16 @@ public Result authenticate(Request request, ContentResponse response, HeaderInfo
 
         try {
             byte[] token = context.initSecContext(new byte[0], 0, 0);
-            if (token != null) {
-                return new SpnegoResult(
-                        UriUtil.normalizedUri(requestUri), headerInfo.getHeader(),
-                        format("%s %s", NEGOTIATE, Base64.getEncoder().encodeToString(token)));
+            if (token == null) {
+                throw new AuthenticationException(format("Failed to establish GSSContext for request %s", requestUri));
```
wdong87: How about making this exception message a little different than the below one. Suggested: No token generated from GSSContext for request %s

wdong87:
Like the change. The overall change looks good to me. We should capture the internal exception for better debugging.

dain:
Looks good

PR #330, opened 2015-06-24
Don't cache kerberos tokens

martint:
They can't actually be reused. Instead, remember that a server requires
auth and perform it preemptively when sending future requests to the
same server


dain:
looks good

PR #332, opened 2015-06-26
always set useKeytab=true

zhenyuy-fb:
this will enable the client to still use the keytab even its location is derived from kerberos configuration file rather than command line option.


martint:
Merged, thanks!

PR #342, opened 2015-08-28
Support initializer before LifeCycleManager.start

miniway:
There would be a situation that some initializers' operation must be finished before `@PostConstruct`( `LifeCycleManager.start`) methods are called. For example, some plugins must be loaded before other instances start services.

### DIFF
--- a/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
@@ -41,12 +41,14 @@ import io.airlift.log.Logging;
 import io.airlift.log.LoggingConfiguration;
 
 import java.io.PrintWriter;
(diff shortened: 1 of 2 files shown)


johngmyers:
In Platform, I created an `@AcceptRequests` lifecycle annotation which is guaranteed to be invoked after all `@PostConstruct` are complete. I then moved `HttpServer.start()` to the new annotation.

The idea is that one specifically marks the methods that cause the server to start taking external work.

martint:
I've always disliked that we conflate "post-construct" (which is part of object creation and meant to work around java memory model constraints) and "service start". I like the idea of a different annotation for this purpose.

johngmyers:
proofpoint/platform@41cafe0f0041238dc9158d0660359c5950a1d3be

miniway:
Brilliant, I also like the idea of a new annotation.

PR #344, opened 2015-09-04
Enable acceptor and selector configuration

erichwang:


http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -127,7 +128,7 @@ public HttpServer(HttpServerInfo httpServerInfo,
                 httpConfiguration.setSecurePort(httpServerInfo.getHttpsUri().getPort());
             }
 
-            httpConnector = new ServerConnector(server, new HttpConnectionFactory(httpConfiguration));
+            httpConnector = new ServerConnector(server, null, null, null, Optional.ofNullable(config.getHttpAcceptorThreads()).orElse(-1), Optional.ofNullable(config.getHttpSelectorThreads()).orElse(-1), new HttpConnectionFactory(httpConfiguration));
```
dain: That is an very long way to write `x == null ? -1 : x`

Maybe assign those two values to variables

dain: Or maybe add a method to construct `ServerConnector` instance

dain:
There is a compile error in Travis.

If that isn't caused by this, looks good.

PR #350, opened 2015-09-20
Add VM runtime ticker

dain:
This ticker only advances when the VM is not paused

### DIFF
--- a/concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
+++ b/concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
@@ -1,0 +1,152 @@
+package io.airlift.concurrent;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
+import com.google.common.base.Ticker;
+import io.airlift.log.Logger;
+import io.airlift.stats.TimeStat;
+import io.airlift.units.Duration;
+import org.weakref.jmx.Managed;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+/**
+ * A ticker that does not advance when the VM is paused.
+ *
+ * This is implemented using a simple background thread that sleeps for a small
+ * period and then measures the sleep duration.  If the duration is small also,
+ * the ticker is advanced; otherwise it is ignored.
+ *
+ * The pause durations are recorded in a TimeStat for monitoring.
+ */
+public class VmRuntimeTicker
+        extends Ticker
+{
+    private static final Logger log = Logger.get(VmRuntimeTicker.class);
+
+    private static final Duration DEFAULT_LOOP_SLEEP_MILLIS = new Duration(1, MILLISECONDS);
+    private static final Duration DEFAULT_MAX_SAFE_VM_PAUSE_NANOS = new Duration(100, MILLISECONDS);
+
+    private static final Supplier<Ticker> VM_RUNTIME_TICKER = Suppliers.memoize(() -> {
+        VmRuntimeTicker ticker = new VmRuntimeTicker(Ticker.systemTicker());
+        ticker.start();
+
+        // return a wrapper so shared ticker can not be stopped
+        return new Ticker()
+        {
+            @Override
+            public long read()
+            {
+                return ticker.read();
+            }
+        };
+    });
+
+    public static Ticker getVmRuntimeTicker()
+    {
+        return VM_RUNTIME_TICKER.get();
+    }
+
+    private final Ticker systemTicker;
+    private final long loopSleepMillis;
+    private final long maxSafeVmPauseNanos;
+
+    private final Thread vmRuntimeMonitorThread;
+
+    private final AtomicLong currentValue = new AtomicLong();
+
+    private final AtomicLong totalPauseNanos = new AtomicLong();
+    private final TimeStat pauseTime = new TimeStat();
+
+    public VmRuntimeTicker(Ticker systemTicker)
+    {
+        this(systemTicker, DEFAULT_LOOP_SLEEP_MILLIS, DEFAULT_MAX_SAFE_VM_PAUSE_NANOS);
+    }
+
+    public VmRuntimeTicker(Ticker systemTicker, Duration loopSleep, Duration maxSafeVmPause)
+    {
+        this.systemTicker = requireNonNull(systemTicker, "systemTicker is null");
+
+        this.loopSleepMillis = requireNonNull(loopSleep, "loopSleep is null").toMillis();
+        checkArgument(loopSleepMillis > 0, "loopSleep must be at least 1 millis second");
+
+        this.maxSafeVmPauseNanos = requireNonNull(maxSafeVmPause, "maxSafeVmPause is null").roundTo(NANOSECONDS);
+        checkArgument(maxSafeVmPauseNanos > MILLISECONDS.toNanos(loopSleepMillis) * 10, "maxSafeVmPause must be at least 10 times larger than loopSleep");
+
+        vmRuntimeMonitorThread = new Thread(this::updateLoop, "VM Runtime Monitor");
+        vmRuntimeMonitorThread.setDaemon(true);
+    }
+
+    public void start()
+    {
+        vmRuntimeMonitorThread.start();
+    }
+
+    public void stop()
+    {
+        vmRuntimeMonitorThread.interrupt();
+    }
+
+    @Managed
+    public long getTotalPauseMillis()
+    {
+        return NANOSECONDS.toMillis(totalPauseNanos.get());
+    }
+
+    @Managed
+    public TimeStat getPauseTime()
+    {
+        return pauseTime;
+    }
+
+    @Override
+    public long read()
+    {
+        return currentValue.get();
+    }
+
+    private void updateLoop()
+    {
+        long lastRead = systemTicker.read();
+        while (true) {
+            try {
+                Thread.sleep(loopSleepMillis);
+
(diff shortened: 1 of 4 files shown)


dain:
@martint this idea here is to measure time, excluding the time the VM was paused.  I am thinking we can use this as a better way to measure timeouts in the face of long VM pauses.

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,152 @@
...
+            try {
+                Thread.sleep(loopSleepMillis);
+
+                long currentRead = systemTicker.read();
+
+                update(currentRead - lastRead);
+
+                lastRead = currentRead;
+            }
+            catch (RuntimeException | Error e) {
```
electrum: Would it better to catch `InterruptedException`, then catch `Throwable`?

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,152 @@
...
+            }
+        }
+    }
+
+    @VisibleForTesting
+    void update(long deltaNanos)
+    {
+        if (deltaNanos < 0) {
+            // clock moved backwards ignore
+            log.warn("Possible kernel bug: system nano time moved backwards");
```
electrum: "kernel or JVM bug"

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
erichwang: agree here that we should catch InterruptedException first

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,152 @@
...
+    {
+        return currentValue.get();
+    }
+
+    private void updateLoop()
+    {
+        long lastRead = systemTicker.read();
+        while (true) {
+            try {
+                Thread.sleep(loopSleepMillis);
```
erichwang: take a look at https://github.com/giltene/jHiccup/blob/master/src/main/java/org/jhiccup/HiccupMeter.java
for ideas

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,152 @@
...
+
+                long currentRead = systemTicker.read();
+
+                update(currentRead - lastRead);
```
erichwang: this delta should not include the expected sleep time.

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,152 @@
...
+
+    public VmRuntimeTicker(Ticker systemTicker, Duration loopSleep, Duration maxSafeVmPause)
+    {
+        this.systemTicker = requireNonNull(systemTicker, "systemTicker is null");
+
+        this.loopSleepMillis = requireNonNull(loopSleep, "loopSleep is null").toMillis();
+        checkArgument(loopSleepMillis > 0, "loopSleep must be at least 1 millis second");
+
+        this.maxSafeVmPauseNanos = requireNonNull(maxSafeVmPause, "maxSafeVmPause is null").roundTo(NANOSECONDS);
+        checkArgument(maxSafeVmPauseNanos > MILLISECONDS.toNanos(loopSleepMillis) * 10, "maxSafeVmPause must be at least 10 times larger than loopSleep");
```
erichwang: 2x is probably better than 10x. or make it configurable.

erichwang:
ping @dain

dain:
@erichwang can you take a look at the changes

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -118,18 +125,37 @@ private void updateLoop()
             try {
                 Thread.sleep(loopSleepMillis);
 
-                long currentRead = systemTicker.read();
+                // Allocate an object to make sure potential allocation stalls are measured.
+                sleepObjectAllocation = lastRead;
```
dain: The field is public and volatile.  I copied this idea from Gil Tene's https://github.com/giltene/jHiccup/blob/master/src/main/java/org/jhiccup/HiccupMeter.java

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,180 @@
...
+
+                // check if clock moved backwards
+                if (sleepDuration < 0) {
+                    if (badSystemNanoTime.compareAndSet(false, true)) {
+                        log.warn("Possible kernel or JVM bug: system nano time moved backwards");
+                    }
+                    continue;
+                }
+
+                long sleepDelta = sleepDuration - loopSleepMillis;
```
erichwang: is Thread.sleep mostly guaranteed to return the same or longer than the sleep duration?

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,180 @@
...
+
+                // check if thread was woken up early
+                if (sleepDelta < 0) {
+                    continue;
+                }
+
+                update(sleepDuration);
+            }
+            catch (InterruptedException e) {
+                return;
```
erichwang: tiny nit: but maybe mark Thread as interrupted

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,180 @@
...
+            }
+            catch (RuntimeException | Error e) {
```
erichwang: Why not just catch Throwable?

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,180 @@
...
+
+    public void stop()
+    {
+        vmRuntimeMonitorThread.interrupt();
+    }
+
+    @Managed
+    public long getTotalPauseMillis()
+    {
+        return NANOSECONDS.toMillis(totalPauseNanos.get());
```
erichwang: I don't think you need the totalPauseNanos counter. If you want to export the millis out of convenience, than just get it from the AllTime counter from pauseTime.

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,180 @@
...
+        {
+            @Override
+            public long read()
+            {
+                return ticker.read();
+            }
+        };
+    });
+
+    public static Ticker getVmRuntimeTicker()
```
erichwang: instead of this way of initializing the VmRuntimeTicker, why don't we just make it a service that uses the PostConstruct and PreDestroy annotations?

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,180 @@
...
+ * the ticker is advanced; otherwise it is ignored.
+ *
+ * The pause durations are recorded in a TimeStat for monitoring.
+ */
+public class VmRuntimeTicker
+        extends Ticker
+{
+    private static final Logger log = Logger.get(VmRuntimeTicker.class);
+
+    private static final Duration DEFAULT_LOOP_SLEEP_MILLIS = new Duration(1, MILLISECONDS);
```
erichwang: 1 milli seems kind of short for sleep, maybe 10 millis default?

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,180 @@
...
+            try {
+                Thread.sleep(loopSleepMillis);
+
+                // Allocate an object to make sure potential allocation stalls are measured.
+                sleepObjectAllocation = lastRead;
+
+                long currentRead = systemTicker.read();
+
+                long sleepDuration = currentRead - lastRead;
+                lastRead = currentRead;
```
erichwang: food for thought: should we be timing the whole loop? or just the sleep call?

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @Managed
+    public long getTotalPauseMillis()
+    {
+        return NANOSECONDS.toMillis(totalPauseNanos.get());
+    }
+
+    @Managed
+    public TimeStat getPauseTime()
```
martint: TimeStat can represent any time unit, so calling it "nanos" doesn't make much sense. Also, the default unit is seconds, not nanos.

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
```diff
@@ -0,0 +1,180 @@
...
+ * A ticker that does not advance when the VM is paused.
+ *
+ * This is implemented using a simple background thread that sleeps for a small
+ * period and then measures the sleep duration.  If the duration is small also,
+ * the ticker is advanced; otherwise it is ignored.
+ *
+ * The pause durations are recorded in a TimeStat for monitoring.
+ */
+public class VmRuntimeTicker
+        extends Ticker
```
dain: This original point of this code was to us this ticker to measure durations that ignore vm pauses.  So you could say, I'm going to try something for 30 seconds regardless of VM pauses.

concurrent/src/main/java/io/airlift/concurrent/VmRuntimeTicker.java
erichwang: Ah got it

PR #351, opened 2015-09-21
fix spengo authentication

zhenyuy-fb:


zhenyuy-fb:
I'm not sure if the CI build failure is related to my change, cause i have seen exactly the same build error locally when i'm using the improper java version(1.8.0_25 rather than 1.8.0_40+)

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -58,72 +65,50 @@
...
+            .expireAfterWrite(CACHE_EXPIRE_TIME.roundTo(TimeUnit.HOURS), TimeUnit.HOURS)
+            .removalListener(notification -> {
+                try {
+                    ((LoginContext)notification.getValue()).logout();
+                }
+                catch (LoginException e) {
+                    Throwables.propagate(e);
+                }
+            })
+            .build(new CacheLoader<String, LoginContext>()
```
electrum: I think this formatting is wrong

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -58,72 +65,50 @@
...
+                   {
+                       @Override
+                       public LoginContext load(String key)
+                               throws Exception
+                       {
+                           return loadLoginContext();
+                       }
+                   }
+            );
```
electrum: This should go on the previous line

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -58,72 +65,50 @@
...
+    private final String principal;
+
+    private final LoadingCache<String, LoginContext> loginContextCache = CacheBuilder.newBuilder()
+            .expireAfterWrite(CACHE_EXPIRE_TIME.roundTo(TimeUnit.HOURS), TimeUnit.HOURS)
+            .removalListener(notification -> {
+                try {
+                    ((LoginContext)notification.getValue()).logout();
+                }
+                catch (LoginException e) {
+                    Throwables.propagate(e);
```
electrum: Where does this exception go? Probably should log a warning (or do nothing -- what would an error mean?)

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -58,72 +65,50 @@
...
+    private final File keytab;
+    private final File credentialCache;
     private final String remoteServiceName;
+    private final String principal;
+
+    private final LoadingCache<String, LoginContext> loginContextCache = CacheBuilder.newBuilder()
+            .expireAfterWrite(CACHE_EXPIRE_TIME.roundTo(TimeUnit.HOURS), TimeUnit.HOURS)
+            .removalListener(notification -> {
+                try {
+                    ((LoginContext)notification.getValue()).logout();
```
electrum: Missing space

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -58,72 +65,50 @@
...
         requireNonNull(remoteServiceName, "remoteServiceName is null");
 
+        this.keytab = keytab;
+        this.principal = principal;
+        this.credentialCache = credentialCache;
         this.remoteServiceName = remoteServiceName;
-
         System.setProperty("java.security.krb5.conf", kerberosConfig.getAbsolutePath());
-
-        try {
```
electrum: Did anything change with this "introduce method" refactoring? Please do it in a separate commit

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -147,6 +132,15 @@ public URI getURI()
             public void apply(Request request)
             {
                 String servicePrincipal = makeServicePrincipal(remoteServiceName, normalizedUri.getHost());
+                LoginContext loginContext = getLoginContext();
+                Subject subject = loginContext.getSubject();
+                Principal clientPrincipal = subject.getPrincipals().iterator().next();
```
electrum: This can throw if there are no principals. It should probably have an explicit check to throw a proper error message

electrum:
What is this fixing? Please add an explanation to the commit message.

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -193,6 +187,62 @@ public boolean matches(String type, URI uri, String realm)
...
+    }
+
+    private LoginContext getLoginContext()
+    {
+        try {
+            // this is the only entry in the cache
+            return loginContextCache.get("loginContext");
+        }
+        catch (ExecutionException e) {
+            Throwables.propagate(e);
```
electrum: This should get the cause from `ExecutionException`, since that is always a wrapper exception

zhenyuy-fb:
Address comments from @electrum and @martint.

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -58,72 +59,33 @@
         }
     }
 
-    private final LoginContext loginContext;
-    private final GSSCredential clientCredential;
+    private final File keytab;
+    private final File credentialCache;
+    private final String principal;
     private final String remoteServiceName;
+    private LoginContext loginContext;
```
martint: This class may be accessed concurrently by multiple threads, so these fields need to be properly synchonized

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -193,6 +161,58 @@ public boolean matches(String type, URI uri, String realm)
         return NEGOTIATE.equalsIgnoreCase(type);
     }
 
+    private void ensureClientCredentialUpdated()
+            throws LoginException, GSSException
+    {
+        if (clientCredential == null || clientCredential.getRemainingLifetime() <= MIN_CREDENTIAL_LIFE_TIME_SEC) {
+            if (loginContext != null) {
+                loginContext.logout();
```
martint: Why does it need to log out? Wouldn't it be sufficient to get a new client credential by calling `GSS_MANAGER.createCredential(...)`? I.e., is the login session or the client credential that expires?

zhenyuy-fb: we do need a new login session to get updated TGT and thus clientCredential, this is also what OSS hadoop does, https://github.com/apache/hadoop/blob/773c670943757681feeafb227a2d0c29d48f38f1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/UserGroupInformation.java#L1039-L1057.

martint:
Do we need the same treatment for SpnegoFilter?

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
zhenyuy-fb: I believe we don't need the treatment for SpnegoFilter. The server credential creation does not go through KDC, it's directly derived from the keytab which does not have expiration time.

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
zhenyuy-fb: Has put loginContext and clientCredential under an AtomicReference field and defined a new Class CredentialHolder for it.

martint:
merged

PR #352, opened 2015-09-30
Refresh spnego client credentials

martint:
The last commit fixes some remaining concurrency problems in #351

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
+++ b/http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
@@ -30,6 +30,7 @@ import java.security.Principal;
 import java.security.PrivilegedAction;
 import java.util.Base64;
 import java.util.Locale;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static com.google.common.base.Preconditions.checkState;
 import static java.lang.String.format;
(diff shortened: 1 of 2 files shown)


http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -193,6 +170,66 @@ public boolean matches(String type, URI uri, String realm)
...
 
+    private Session getCredential()
+            throws LoginException, GSSException
+    {
+        synchronized (this) {
+            checkState(!shutdown, "SpnegoAuthentication is already stopped");
+
+            if (clientSession == null || clientSession.getClientCredential().getRemainingLifetime() < MIN_CREDENTIAL_LIFE_TIME.getValue(TimeUnit.SECONDS)) {
+                if (clientSession != null) {
+                    clientSession.logout();
```
wdong87: One minor issue for this is that an outstanding request may fail if the clientSession object is being used in the middle while it's being logged out here. To prevent that, we may need to do reference counting. Otherwise, we are good.

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -193,6 +170,66 @@ public boolean matches(String type, URI uri, String realm)
         return NEGOTIATE.equalsIgnoreCase(type);
     }
 
+    private Session getCredential()
```
zhenyuy-fb: nit: maybe rename as getSession()?

martint:
merged

wdong87:
Awesome!!!

PR #357, opened 2015-10-22
modify synchronize get method in class LifeCycleMethodsMap

yujingwang:


martint:
What's the purpose of this change?

yujingwang:
Sorry for the silly mistake in my last commit.
I was a bootcamper and was assigned the task#8700769 about fixing lock contention. Thomas Georgiou told me that LifeCycleMethodMap is the class that has the contention. (https://github.com/airlift/airlift/blob/master/bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleModule.java#L56 registers an injection listener with guice so anytime you get an object from guice it has to go through a global lock.)
 So I change the synchronized method to synchronized block, thus the method will go through a global lock only when it needs to put new obj into the map.

electrum:
Use a ConcurrentHashMap and switch this to computeIfAbsent

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleMethodsMap.java
```diff
@@ -23,12 +23,16 @@
 {
     private final Map<Class<?>, LifeCycleMethods> map = Maps.newHashMap();
 
-    synchronized LifeCycleMethods get(Class<?> clazz)
-    {
+    public LifeCycleMethods get(Class<?> clazz) {
         LifeCycleMethods methods = map.get(clazz);
```
martint: This is not a safe operation. It's accessing a non-thread-safe mutable object (the map) without proper synchronization. It can cause a variety of issue ranging from getting an incorrect result to looping forever walking the map internal structures.

In general, the double-checked locking idiom is broken due to Java memory model semantics.

martint:
Be aware that this may not solve the underlying contention issue. One important assumption we make in Airlift is that Guice is used only for wiring the application, and it doesn't play a role after that.

tageorgiou:
@martint we use Guice assisted injection to create objects at runtime with Guice provided objects injected in. i.e. a WorkToken with some thrift client. We have multiple threads getting objects and seeing them all bottleneck on this Guice listener.

martint:
@tageorgiou, that's what I figured.

We'll get this merged, but I'm curious about whether it helps in your scenario.

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleMethodsMap.java
```diff
@@ -18,18 +18,14 @@
...
-    synchronized LifeCycleMethods get(Class<?> clazz)
-    {
-        LifeCycleMethods methods = map.get(clazz);
-        if (methods == null) {
-            methods = new LifeCycleMethods(clazz);
-            map.put(clazz, methods);
-        }
-        return methods;
+    public LifeCycleMethods get(Class<?> clazz) {
+        return map.computeIfAbsent(clazz,
```
ryandm: java 8 nicety -- this can probably be

```java
return map.computeIfAbsent(clazz, LifeCycleMethods::new);
```

martint:
#363 was merged. Closing.

PR #363, opened 2015-11-11
modify synchronize method in class LifeCycleMethodsMap

yujingwang:
task#8700769 about fixing lock contention
It's based on the pull request https://github.com/airlift/airlift/pull/357 I submitted before and modify the code as ryandm's inline comment suggested.


martint:
Merged, thanks!

PR #365, opened 2015-11-16
Add method to get max content length to HttpClient

nileema:


dain:
Looks good

johngmyers:
It might be more interesting to be able to override this setting on a per-`Request` basis.

Incidentally, the behavior of `.execute()` and `.executeAsync()` is surprisingly different when this limit is exceeded.

PR #367, opened 2015-11-18
Add finalizer to cleanup ThreadGroup

cberner:


concurrent/src/main/java/io/airlift/concurrent/Threads.java
```diff
@@ -84,18 +84,24 @@ public Thread newThread(Runnable runnable)
...
         @Override
-        public Thread newThread(Runnable runnable)
+        protected void finalize()
         {
-            return new Thread(threadGroup, runnable);
+            try {
+                threadGroup.destroy();
+            }
+            catch (RuntimeException e) {
+                log.warn(e, "Leaked thread group");
```
dain: Add the thread group name

dain:
Looks good

concurrent/src/main/java/io/airlift/concurrent/Threads.java
```diff
@@ -84,18 +84,24 @@ public Thread newThread(Runnable runnable)
...
+        @Override
+        public Thread newThread(Runnable runnable)
         {
-            this.threadGroup = threadGroup;
+            return new Thread(threadGroup, runnable);
         }
 
         @Override
-        public Thread newThread(Runnable runnable)
+        protected void finalize()
```
martint: A finalizer? seriously?

http://www.informit.com/articles/article.aspx?p=1216151

concurrent/src/main/java/io/airlift/concurrent/Threads.java
```diff
@@ -84,18 +84,24 @@ public Thread newThread(Runnable runnable)
...
         {
-            return new Thread(threadGroup, runnable);
+            try {
+                threadGroup.destroy();
```
martint: This can only be called after all threads in the group are dead. Does this guarantee that will happen?

martint: Honestly, I would just remove usage of ThreadGroup. They seem to be causing more pain that they are worth. Also, take a look at Effective Java, Item 73 for a discussion about that feature.

PR #369, opened 2015-11-30
do not throw exception in SpnegoAuthentication

zhenyuy-fb:
When authentication is enabled and the client cannot get correct kerberos credentiails(for instance, incorrect KDC settings, invalid personal credentials, etc), the exception thrown from SpnegoAuthentication's apply() method will propagate to Jetty's ResponseListener.notifyComplete() method and get caught there. This will cause the client's HttpRequest wait in an idle loop until timeout(60 seconds by default). This fix is to remove those exceptions and let the response against 401 authentication challenge go, which will make the client fail fast and loudly with detailed exception messages like below:

java.lang.RuntimeException: Error starting query at https://xxx:7778/v1/statement returned an invalid response: JsonResponse{statusCode=401, statusMessage=Unauthorized, headers={Content-Length=[0], Date=[Mon, 30 Nov 2015 19:42:07 GMT], WWW-Authenticate=[Negotiate realm="presto"]}, hasValue=false, value=null} [Error: ]
	at com.facebook.presto.client.StatementClient.requestFailedException(StatementClient.java:281)
	at com.facebook.presto.client.StatementClient.<init>(StatementClient.java:103)
	at com.facebook.presto.cli.QueryRunner.startInternalQuery(QueryRunner.java:82)
	at com.facebook.presto.cli.QueryRunner.startQuery(QueryRunner.java:77)
	at com.facebook.presto.cli.Console.process(Console.java:282)
	at com.facebook.presto.cli.Console.runConsole(Console.java:224)
	at com.facebook.presto.cli.Console.run(Console.java:130)
	at com.facebook.presto.cli.Presto.main(Presto.java:32)

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
+++ b/http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
@@ -125,14 +125,14 @@ public class SpnegoAuthentication
                         request.header(headerInfo.getHeader(), format("%s %s", NEGOTIATE, Base64.getEncoder().encodeToString(token)));
                     }
                     else {
-                        throw new RuntimeException(format("No token generated from GSS context for %s", request.getURI()));
+                        LOG.warn(format("No token generated from GSS context for %s", request.getURI()));
                     }
                 }
                 catch (GSSException e) {
-                    throw new RuntimeException(format("Failed to establish GSSContext for request %s", request.getURI()), e);
+                    LOG.warn(e, format("Failed to establish GSSContext for request %s", request.getURI()));
                 }
                 catch (LoginException e) {
-                    throw new RuntimeException(format("Failed to establish LoginContext for request %s", request.getURI()), e);
+                    LOG.warn(e, format("Failed to establish LoginContext for request %s", request.getURI()));
                 }
                 finally {
                     try {


electrum:
Commit messages should start with uppercase

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -125,14 +125,14 @@ public void apply(Request request)
                         request.header(headerInfo.getHeader(), format("%s %s", NEGOTIATE, Base64.getEncoder().encodeToString(token)));
                     }
                     else {
-                        throw new RuntimeException(format("No token generated from GSS context for %s", request.getURI()));
+                        LOG.warn(format("No token generated from GSS context for %s", request.getURI()));
```
electrum: The Airlift `Logger` class already takes format strings

electrum:
This fix seems incorrect. By ignoring the error, the user of the client gets a 401 without knowing why. We need to know why this exception is not being handled properly. Is this a bug in the Jetty client, or are we doing something wrong in the way we use it?

electrum:
While it seems like failing fast is better behavior than the current hang, it will cause confusion since the message is misleading: the server returned a 401 because the client knowingly sent a bad request.

zhenyuy-fb:
When the exception thrown out, Jetty's AuthenticationListener catches it within onComplete() method, which in turn triggers ResponseNotifier's notifyComplete(), where the registered ResponseListeners being called in sequence. However, those listeners are some jetty's default ones rather than those passed-in when built the Request in JettyHttpClient. Jetty provide an approach to change the registered listeners through HttpConversation's updateResponseListeners() method, and HttpConversation could be get from HttpRequest, but unfortunately when AuthenticationListener calls SpnegoAuthentication's apply method, it passes in an Request rather than HttpRequest. So my understand is this is a jetty bug(correct me if i'm wrong), and relying on server to return the second '401' is the workaround i can think of.

electrum:
Can you file a Jetty issue? We have a support contract and can prioritize fixing it.

zhenyuy-fb:
here is the jetty issue report: https://bugs.eclipse.org/bugs/show_bug.cgi?id=484621.

electrum:
We just upgraded to Jetty 9.3.7.RC0 which includes the fix for this.

wdong87:
Awesome! @electrum , thank you so much for the follow-up :)! @zhenyuy-fb is on paternity leave for 3 months. Just FYI.

PR #377, opened 2016-01-06
Fix TestHttpServerCipher

electrum:


dain:
Assuming Travis passes, looks good

PR #383, opened 2016-01-20
Fix thread leak in JettyHttpClient when using anonymous pool

cberner:


http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -94,6 +95,8 @@
     private static final long SWEEP_PERIOD_MILLIS = 5000;
     private static final String REALM_IN_CHALLENGE = "X-Airlift-Realm-In-Challenge";
 
+    @Nullable
```
electrum: Make this `Optional`

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -618,6 +625,9 @@ public void close()
         }
         catch (Exception ignored) {
         }
+        if (anonymousPool != null) {
```
electrum:
```java
anonymousPool.ifPresent(Closeable::close);

electrum:
Looks good

PR #387, opened 2016-02-18
Add upper and lower bound quantile queries.

DanielTing:
The current quantile approximation is always an upper bound on the true quantile.
This extends the functionality of the q-digest to query both upper and lower bounds for the quantile approximation.
This allows confidence intervals on percentiles to be computed.

The existing getQuantiles is renamed getQuantilesUpperBound. getQuantiles is now a wrapper for that function.

### DIFF
--- a/stats/src/main/java/io/airlift/stats/QuantileDigest.java
+++ b/stats/src/main/java/io/airlift/stats/QuantileDigest.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Deque;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -185,11 +186,67 @@ public void merge(QuantileDigest other)
...
      */
-    public List<Long> getQuantiles(List<Double> quantiles)
+    public List<Long> getQuantilesLowerBound(List<Double> quantiles)
+    {
+        checkArgument(Ordering.natural().isOrdered(quantiles), "quantiles must be sorted in increasing order");
+        for (double quantile : quantiles) {
+            checkArgument(quantile >= 0 && quantile <= 1, "quantile must be between [0,1]");
+        }
+
+        ArrayList<Double> reversedQ = new ArrayList(quantiles);
```
electrum: Nit: declare variable as `List` rather than implementation type

stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -185,11 +186,67 @@ public void merge(QuantileDigest other)
...
+        Collections.reverse(reversedQ);
+
+        final ImmutableList.Builder<Long> builder = ImmutableList.builder();
```
electrum: Nit: no `final` on local variables

stats/src/main/java/io/airlift/stats/QuantileDigest.java
electrum: Nit: don't abbreviate variables (so name this `reversedQuantiles` or perhaps just `reversed`)

stats/src/test/java/io/airlift/stats/TestQuantileDigest.java
```diff
@@ -174,6 +174,38 @@ public void testQuantile()
...
 
     @Test
+    public void testQuantileLowerBound()
+    {
+        QuantileDigest digest = new QuantileDigest(0.5);
+
+        addRange(digest, 1, 100);
+
+        assertEquals(digest.getQuantileLowerBound(0.0), 1);
+        for(int i=1; i <= 10; i++) {
```
electrum: Nit: formatting. Run code formatter before committing. https://github.com/airlift/codestyle

stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -185,11 +186,67 @@ public void merge(QuantileDigest other)
...
+
+    /**
+     * Get an upper bound on the quantiles for the given proportions. A returned q quantile is guaranteed to be within
+     * the q and q + maxError quantiles.
+     *
+     * The input list of quantile proportions must be sorted in increasing order, and each value must be in the range [0, 1]
+     */
+    public List<Long> getQuantilesUpperBound(List<Double> quantiles)
     {
+
```
electrum: Nit: remove these extra blank lines

stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -185,11 +186,67 @@ public void merge(QuantileDigest other)
...
+        // we finished the traversal without consuming all quantiles. This means the remaining quantiles
+        // correspond to the max known value
+        while (iterator.hasNext()) {
+            builder.add(min);
+            iterator.next();
+        }
+
+        List<Long> quantileValues = builder.build();
+        Collections.reverse(quantileValues);
```
dain: Reverse does not work on an ImutableList, which means the tests are not covering the case where you have multiple quantiles.

Instead you can call `builder.build().reverse()` since `ImmutableList` has this special method.

stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -185,11 +186,67 @@ public void merge(QuantileDigest other)
...
-    public List<Long> getQuantiles(List<Double> quantiles)
+    public List<Long> getQuantilesLowerBound(List<Double> quantiles)
+    {
+        checkArgument(Ordering.natural().isOrdered(quantiles), "quantiles must be sorted in increasing order");
+        for (double quantile : quantiles) {
+            checkArgument(quantile >= 0 && quantile <= 1, "quantile must be between [0,1]");
+        }
+
+        ArrayList<Double> reversedQ = new ArrayList(quantiles);
+        Collections.reverse(reversedQ);
```
dain: Replace all of this with `List<Double> reversedQ = ImmutableList.copyOf(quantiles).reverse()`

dain:
Merged, thanks!

PR #388, opened 2016-02-26
Replace GzipFilter with GzipHandler

martint:
GzipFilter is deprecated in recent versions of Jetty and is now a no-op.

### DIFF
--- a/http-server/pom.xml
+++ b/http-server/pom.xml
@@ -129,11 +129,6 @@
 
         <dependency>
             <groupId>org.eclipse.jetty</groupId>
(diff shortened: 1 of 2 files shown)


dain:
Build failure
```
[WARNING] Unused declared dependencies found:
[WARNING]    org.eclipse.jetty:jetty-servlets:jar:9.3.7.RC0:compile

dain:
Other than the build failure, looks good.

You should double check that we aren't activating this in Presto, since it will kill perf.

martint:
It only activates if the client passes the accept-encoding header.

I'll take a look at the failure

PR #389, opened 2016-03-08
Ignore only announce error only

kokosing:
Ignore only announce error only

Untill now any thrown exception within announce creation and its
execution in Announcer#start were ignored.

Exception during announce execution are typically logged in
ExponentialBackOff. While exceptions during announce creation are just
swallowed. This may lead to serious problem as no error is reported and
discovery mechanism is disabled.


kokosing:
I had an underscore in a hostname. Underscores are disallowed for hostname (URI). However java URI class (its default constructor) does not throw any exception, but internally sets hosts to null. When announce request was created it tried to access uri.host which thrown NPE. This NPE was swallowed by try-catch in Announcer. 

It took couple of hours to diagnose the issue. It think this PR will be helpful for others which get into the same situation.

electrum:
Merged, thanks!

PR #390, opened 2016-03-09
Http client improvements

arhimondr:
During the developing of our internal presto connector we have discovered the lack of possibility to get the already buffered data from the `Response` without extra copying. 
Also we noticed that `checkSufficientThreads` was missing for HttpClient when it is present for HttpServer.


arhimondr:
@electrum updated

arhimondr:
@electrum rebased on master

arhimondr:
No longer actual after the asynchronous response is buffered in a list of small buffers

PR #392, opened 2016-03-15
Add support for configurable event logging

nezihyigitbasi:
This PR adds support for configurable event logging, e.g., to log query completion events to a particular log file with a custom log format. Related issues are facebook/presto#2917 and facebook/presto#4642

Once this is merged we can add this new `EventLoggingModule` to `PrestoServer`.

@martint @electrum can you please take a look?
@billonahill can you also take a look to see whether this addresses your use case?

##### Sample configuration:
```
event-log.enabled=true
event-log.path=/mnt/presto-server-0.142-SNAPSHOT/query-completion-events.log
event-log.type=QueryCompletion
event-log.layout=presto.test.QueryCompletionLogLayout
```

##### Open issues:
- Do we want to provide a default log layout? If so where (in Presto or in Airlift)?

### DIFF
--- a/event/src/test/java/io/airlift/event/client/TestLoggingEventClient.java
+++ b/event/src/test/java/io/airlift/event/client/TestLoggingEventClient.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright 2010 Proofpoint, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.event.client;
+
+import ch.qos.logback.core.LayoutBase;
+import ch.qos.logback.core.rolling.RollingFileAppender;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.testng.Assert.assertEquals;
+
+public class TestLoggingEventClient
+{
+    private File temp;
+
+    @BeforeClass
+    public void setup() throws IOException
+    {
+        temp = File.createTempFile("event-logging-tests", ".log");
+    }
+
+    @AfterClass
+    public void tearDown()
+    {
+        temp.delete();
+    }
+
+    @Test
+    public void testInitWithDefaults() throws IOException {
+        EventConfig config = new EventConfig();
+        MockRollingFileAppender appender = new MockRollingFileAppender();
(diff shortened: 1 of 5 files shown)


martint:
I don't think we want to do this. See my comment in https://github.com/facebook/presto/pull/2917#issuecomment-104794587

In particular, I don't think it's a good idea to allow end-user customization for arbitrary events and picking layouts by specifying class names.

It would be ok to add infrastructure to allow applications built on top of airlift to expose the ability to write certain event types (chosen by the application) to files.

nezihyigitbasi:
> It would be ok to add infrastructure to allow applications built on top of airlift to expose the ability to write certain event types (chosen by the application) to files.

I can simply move this functionality to Presto as is since airlift currently has the necessary infrastructure, but I am trying to understand what part of this functionality can stay in airlift and be useful to other apps. 

- When you say applications choose the event types etc. how will that be different from the `EventConfig` here?

- Also how do you think the format should be customized if we don't want to specify class names for the layout?

event/src/test/java/io/airlift/event/client/TestLoggingEventClient.java
```diff
@@ -0,0 +1,146 @@
...
+        config.setEventTypeToLog("MockEvent");
+        config.setLogLayoutClass(null);
+        MockRollingFileAppender appender = new MockRollingFileAppender();
+        createLoggingEventClient(config, appender);
+    }
+
+    private LoggingEventClient createLoggingEventClient(EventConfig config, MockRollingFileAppender appender)
+    {
+        LoggingEventClient client = new LoggingEventClient(config);
+        client.setFileAppender(appender);
```
billonahill: appender gets initialized in the constructor but also passed here. Does this overwrite the one from the constructor? Why multiple ways to pass this?

event/src/test/java/io/airlift/event/client/TestLoggingEventClient.java
```diff
@@ -0,0 +1,146 @@
...
+        EventConfig config = new EventConfig();
+        config.setLogEnabled(true);
+        config.setLogLayoutClass(MockLayout.class.getName());
+        config.setLogPath(temp.getAbsolutePath());
+        config.setEventTypeToLog("MockEvent");
+        MockRollingFileAppender appender = new MockRollingFileAppender();
+        LoggingEventClient client = createLoggingEventClient(config, appender);
+        client.postEvent(new MockEvent());
+        assertEquals(appender.getLogs().size(), 1, "client should log a single event");
+        assertEquals(appender.getLogs().get(0).getClass(), MockEvent.class, "client should log a single event");
```
billonahill: Can you also assert the string format of the log entry?

billonahill:
How does this code determine where in the lifecycle of an object it should be logged? Like if I wanted a QueryInfo object logged upon completion, would we instrument the code at that point to log a QueryCompleteEvent that one could pick up on? I'm not seeing groking how that works when looking at the unit tests.

event/src/test/java/io/airlift/event/client/TestLoggingEventClient.java
nezihyigitbasi: This is for overriding it only for testing, in fact if you look at the `setFileAppender` method it is only visible for testing.

nezihyigitbasi:
When the new `EventLoggingModule` is registered in Presto and `event-log.type` set to `QueryCompletion`, this will start logging those events. 

Once @martint clarifies what exactly we want to have in airlift I will update this PR.

PR #393, opened 2016-03-17
Add JMX stats for HTTP response codes and failures

raghavsethi:


http-client/src/main/java/io/airlift/http/client/RequestStats.java
```diff
@@ -38,32 +48,118 @@
     @Inject
     public RequestStats()
     {
-        request = new CounterStat();
+        allResponse = new CounterStat();
```
dain: These can be in the field declaration.

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -1376,3 +1382,4 @@ protected void doStop()
         }
     }
 }
+
```
dain: is this needed?

http-client/src/main/java/io/airlift/http/client/RequestStats.java
```diff
@@ -38,32 +48,118 @@
...
+    @Managed
+    @Nested
+    public CounterStat getAllResponse()
+    {
+        return allResponse;
+    }
+
+    @Managed
+    @Nested
+    public CounterStat getInformationalResponse()
```
dain: name these `1xx` and so on

dain:
Looks good

dain:
Merged, thanks!

PR #396, opened 2016-03-28
Add call to get response size in JsonResponse

nileema:


dain:
looks good

cberner:
Looks like this is already merged

PR #397, opened 2016-03-28
Clanup QuantileDigest code

dain:


dain:
This change is not longer valid with the new QuantileDigest implementation.

PR #398, opened 2016-03-29
Upgrade to airbase 49

cberner:


cberner:
Abandoning this in favor of upgrading to airbase 50

PR #399, opened 2016-03-30
Upgrade to airbase 50

cberner:


electrum:
Update commit message to indicate this upgrades Jetty and add releases notes. Otherwise looks good.

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyHttpsClient.java
```diff
@@ -75,7 +76,7 @@ public void tearDownHttpClient()
         }
     }
 
-    @Test(expectedExceptions = {SSLHandshakeException.class, EOFException.class})
+    @Test(expectedExceptions = {SSLHandshakeException.class, EOFException.class, ClosedChannelException.class})
```
electrum: Maybe this should just be `IOException`?

johngmyers: eclipse/jetty.project#444

cberner: @johngmyers ya, I asked the Jetty folks about that. They said we shouldn't rely on the exception type in this case

johngmyers: Well, the exception they're throwing now is entirely inappropriate and misleading. It's unlikely anyone would be able to diagnose the condition from that. But that's a bug in Jetty.

PR #400, opened 2016-03-31
Guarantee that AsyncContext.complete() is only called once

cberner:
Previously, if suspend() was called multiple times, complete() could
also be called multiple times.


dain:
Looks good

PR #401, opened 2016-04-04
Fix IllegalStateException in ServletAsyncContextDelegateProvider

cberner:


martint:
Add a comment in the commit message for the second one explaining why we're switching

martint:
looks good

PR #403, opened 2016-04-04
Fix port already bound error in HTTP client tests

electrum:
Allow Jetty to atomically bind to an unused port rather than unsafely
trying to pick a free port ourselves.


dain:
Looks good

PR #404, opened 2016-04-06
Add HTTP/2 support

electrum:


http-client/src/main/java/io/airlift/http/client/jetty/HeadersListMultimap.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public List<String> get(@Nullable String key)
+    {
+        return delegate.get(getKey(key));
+    }
+
+    @Override
+    public boolean containsKey(@Nullable Object key)
```
dain: There are only three methods that ned to be overwritten?  I would expect a bunch more.

dain:
Looks good

http-client/src/main/java/io/airlift/http/client/jetty/HeadersListMultimap.java
electrum: I think this is all that can be done with this interface. We should probably switch to a custom one.

Any ideas on other methods to override?

dain: The problem is all derived collections will be case sensitive.  For example, `asMap` and `keySet`

electrum: Yeah, but I don't think we should deal with that here. It already violates the contract for a collection. This is probably good enough. But if you think we should add it, I can do that.

dain:
looks good

PR #405, opened 2016-04-06
Add useCanonicalHostname property to KerberosConfig

arhimondr:
This flag adds possibility to disable reverse dns resolve on the
environments where modifying reverse dns records isn't possible.
With this flag set to false service pricipal is simply being built
with the hostname specified by user in the service URL.
With this flag set to true reverse dns lookup is being made to
canonicalize service hostname.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
+++ b/http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
@@ -69,11 +69,12 @@ public class SpnegoAuthentication
     private final File credentialCache;
     private final String principal;
     private final String remoteServiceName;
+    private final boolean useCanonicalHostname;
 
     @GuardedBy("this")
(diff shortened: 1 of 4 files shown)


http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -69,11 +69,12 @@
     private final File credentialCache;
     private final String principal;
     private final String remoteServiceName;
+    private final boolean useCanonicalHostname;
```
martint: This should default to true

arhimondr: I think you mean to set it to true in `KerberosConfig`. Done.

martint: Ah, yes :)

PR #408, opened 2016-04-15
Add BoundedThreadPool

electrum:


concurrent/src/test/java/io/airlift/concurrent/TestBoundedThreadPool.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @AfterClass(alwaysRun = true)
+    public void tearDown()
+            throws Exception
+    {
+        executor.shutdownNow();
+    }
+
+    @Test
+    public void testCounter()
```
martint: Pick a name that conveys what this is testing more clearly.

concurrent/src/test/java/io/airlift/concurrent/TestBoundedThreadPool.java
```diff
@@ -0,0 +1,104 @@
...
+                    activeThreadCount.decrementAndGet();
+                }
+                finally {
+                    completeLatch.countDown();
+                }
+            });
+        }
+
+        startLatch.countDown(); // Signal go for threads
+        awaitUninterruptibly(completeLatch, 1, MINUTES); // Wait for tasks to complete
```
martint: 1 minute is way too long

concurrent/src/test/java/io/airlift/concurrent/TestBoundedThreadPool.java
```diff
@@ -0,0 +1,104 @@
...
+        AtomicInteger counter = new AtomicInteger();
+        CountDownLatch startLatch = new CountDownLatch(1);
+        CountDownLatch completeLatch = new CountDownLatch(totalTasks);
+
+        for (int i = 0; i < totalTasks; i++) {
+            executor.execute(() -> {
+                try {
+                    awaitUninterruptibly(startLatch); // Wait for the go signal
+
+                    // Intentional distinct read and write calls
```
martint: What's the purpose of this?

concurrent/src/test/java/io/airlift/concurrent/TestBoundedThreadPool.java
martint: It'd be good to assert that the expected number of tasks were executed. None of the above tests runs multiple threads *and* validates this.

martint: Oh, you're doing it with the countdown latch. Hmm.. the problem with that is that one execution is "lost", the latch fail with a "timeout error". It will take 1 minute for the test to fail, and the error won't be very clear about what happened.

concurrent/src/test/java/io/airlift/concurrent/TestBoundedThreadPool.java
electrum: I copied all these tests from TestBoundedExecutor and am not comfortable changing them.

concurrent/src/main/java/io/airlift/concurrent/BoundedThreadPool.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+public final class BoundedThreadPool
+{
+    private BoundedThreadPool() {}
+
+    public static ThreadPoolExecutor newBoundedThreadPool(int maxThreads, ThreadFactory threadFactory)
```
dain: Can you add:
```java
newBoundedThreadPool(int minThreads, int maxThreads, ThreadFactory threadFactory)
newBoundedThreadPool(int minThreads, int maxThreads, ThreadFactory threadFactory, Duration keepAlive)
```

dain: I might require the min threads parameter

martint:
looks good

PR #409, opened 2016-04-22
improved CDF based quantile estimator for q-digest

DanielTing:
This greatly improves the accuracy of the quantile estimator. The deterministic error guarantee is also preserved. The improvement works by using a good estimate of the cumulative distribution function (CDF) using linear interpolation. Existing implementations use the most extreme CDF that is consistent with the digest.


dain:
@martint is this something you want to review?

stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -285,15 +285,20 @@ public boolean process(Node node)
 
     public List<Long> getQuantiles(List<Double> quantiles)
     {
-        return getQuantilesUpperBound(quantiles);
+        List<Long> accumulator = new ArrayList<Long>(quantiles.size());
+
+        for (Double q : quantiles) {
```
martint: this can be `double q`

stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -371,6 +376,91 @@ public boolean process(Node node)
         return builder.build();
     }
 
+    public double getCDF(long x) {
```
martint: Don't abbreviate variable names.

stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -371,6 +376,91 @@ public boolean process(Node node)
...
+        final ImmutableList.Builder<Double> builder = ImmutableList.builder();
```
martint: We generally don't use `final` for local variables

stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -371,6 +376,91 @@ public boolean process(Node node)
...
+        final long value = x;
+
+        inOrderTraversal(root, new Callback() {
+            protected double sum = 0.0;
```
martint: make this `private`

stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -285,15 +285,20 @@ public boolean process(Node node)
```
martint: Please, reformat according to https://github.com/airlift/codestyle

stats/src/main/java/io/airlift/stats/QuantileDigest.java
martint: Is this being used as a way to return a single value from the callback traversal? It'd be cleaner to either:
1) Introduce a mutable value holder that the callback sets.
2) Pull the callback code into a named class, add a member variable and accessor.

stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -371,6 +376,91 @@ public boolean process(Node node)
...
+                    }
+                    return true;
+                } else if(value < lower) {
+                    // Stop if gone too far
+                    builder.add(sum);
+                    return false;
+                } else {
+                    // interpolate
+                    double ratio = (value - lower) / (double) (upper - lower);
+                    if (node.right == null && node.left != null) {
```
martint: Can you add a comment explaining this case?

stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -882,6 +972,39 @@ private boolean postOrderTraversal(Node node, Callback callback, TraversalOrder
         return callback.process(node);
     }
 
+    protected boolean inOrderTraversal(Node node, Callback callback, TraversalOrder order)
```
martint: This should be private. This class is not meant to be extended, so there's not benefit in making it protected.

stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -371,6 +376,91 @@ public boolean process(Node node)
...
+
+        List<Double> tmp = builder.build();
+        if(tmp.size() == 0) {
+            return 1.0;
+        } else {
+            return tmp.get(0) / weightedCount;
+        }
+    }
+
+    public long getQuantileFromCDF(double p) {
```
martint: This needs an explanation about how the algorithm works.

martint:
Can you add some tests?

kaschaeffer:
ping @DanielTing can you address comments?  Would be awesome to get this merge-ready.

stats/src/main/java/io/airlift/stats/QuantileDigest.java
DanielTing: CDF in stats nomenclature is almost like CPU in cs. Is that acceptable use?
The alternative is getCumulativeDistributionFunction(long x)

DanielTing:
@martint What tests did you have in mind? The usual quantile ones generally cover the modifications, and getCDF is now private.

DanielTing:
Also, tomorrow is my last day, so if there are any changes requested please let me know now!

kaschaeffer:
@DanielTing should I take this over since you're no longer at fb?  I'd still like to get this merged in eventually.

kaschaeffer:
Taking this over.

PR #413, opened 2016-05-03
Fix behavior of BoundedThreadPool

electrum:
Previously, it would behave as a fixed thread pool.


concurrent/src/main/java/io/airlift/concurrent/BoundedThreadPool.java
```diff
@@ -57,6 +67,9 @@ public static ThreadPoolExecutor newBoundedThreadPool(int minThreads, int maxThr
         @Override
         public boolean offer(T o)
         {
+            if (queue.size() < sizeSupplier.getAsInt()) {
```
dain: add javadoc to the class that mention this behavior

dain:
looks good

PR #416, opened 2016-05-05
Remove BoundedThreadPool

electrum:
The current implementation seems to have weird scheduling latencies
and potentially other issues.


martint:
looks good

PR #419, opened 2016-06-01
Add truststore configuration

amrutagokhale:
Add configuration properties needed to setup a truststore, specifically the truststore path and truststore password.

@martint Can you please take a look?


electrum:
Merged, thanks!

PR #421, opened 2016-06-14
Add CpuTimer to airlift-stats

nileema:


electrum:
"stats-" is redundant

electrum:
"airlift-" is redundant

nileema:
@martint - @dain  and @electrum  are ok with adding this. But they want a green light from you.

martint:
Yeah, go for it.

PR #423, opened 2016-06-17
Add FileBodyGenerator for HTTP client

electrum:


dain:
There are test failures, but otherwise looks good

electrum:
The test failure looks like an unrelated Travis timeout. Will restart it.

PR #426, opened 2016-06-22
Update to Jetty 9.3.11.M0

cberner:
This fixes a race when using async responses


martint:
looks good

electrum:
This should be done in airbase.

cberner:
I think it's better to have it in the airlift pom. Putting them in airbase is error prone, as it's less obvious what version is being used. imo, we should remove Jetty from airbase entirely

electrum:
I'm fine with removing it, but don't just update it here, because any projects using airbase will pull the wrong version.

PR #427, opened 2016-06-23
Add getField method to EventTypeMetadata

raghavsethi:


event/src/main/java/io/airlift/event/client/EventTypeMetadata.java
```diff
@@ -392,6 +392,14 @@ public EventFieldMetadata getHostField()
         return fields;
     }
 
+    public EventFieldMetadata getField(String field)
+    {
+        return fields.stream()
```
martint: Keep a map from field->field metadata so that lookups are O(1)

martint:
Looks good, but check the build failures

raghavsethi:
The http-related stuff is pretty flaky - event looks good. Merging.

PR #428, opened 2016-06-23
Move Jetty and Jersey dependencies from Airbase

electrum:
Depends on https://github.com/airlift/airbase/pull/88


cberner:
Looks good

PR #430, opened 2016-06-25
Remove dependencies on internal JVM classes for JMX agent

martint:
This sacrifices discoverability via jstatd, but makes the code
compatible with Java 9.


martint:
Superseded by https://github.com/airlift/airlift/pull/437

PR #431, opened 2016-06-25
Run tests in single-threaded mode

martint:
The testNonEnglishLocale tests modifies the VM-wide locale setting,
so the invocations need to execute serially to avoid race conditions.


electrum:
I think the class itself also needs `@Test(singleThreaded = true)`

electrum:
Looks good

electrum:
Actually, there could be a problem if the two test classes run concurrently. Maybe do something hacky like `synchronized (Locale.class)`

PR #432, opened 2016-06-28
Statically configure logging for Jetty HTTP client

electrum:
This removes the need for the logging properties file.
It also suppresses the "Logging initialized message.


electrum:
I skipped doing this for the HTTP server since it has many classes that reference Jetty classes that statically reference (and thus would initialize) the logging system, and it doesn't seem necessary in a server context.

cberner:
Looks good

PR #437, opened 2016-08-03
Start jmx agent via supported API

martint:


jmx/pom.xml
```diff
@@ -17,13 +17,16 @@
 
     <dependencies>
         <dependency>
-            <groupId>io.airlift</groupId>
-            <artifactId>log</artifactId>
+            <groupId>com.sun</groupId>
+            <artifactId>tools</artifactId>
+            <version>1.8</version>
+            <scope>system</scope>
+            <systemPath>${java.home}/../lib/tools.jar</systemPath>
```
dain: Add something like `TODO remove after upgrade to Java9`

dain:
Looks good

PR #438, opened 2016-08-04
Add missing javax.activation dependency

martint:
This is needed for Java 9, which doesn't expose
those annotations by default


dain:
Looks good

PR #439, opened 2016-08-04
Remove workaround for serializing LazyCompositeData

martint:
The class is internal to the JDK, so it's not available in Java 9.
It doesn't seem to be needed anymore, either.


dain:
Looks good

PR #441, opened 2016-08-05
Workaround IntelliJ mishandling of system-scoped dependencies

martint:
IntelliJ loads system dependencies in compile scope, which tools.jar
to be added to the classpath and causes problems when running with Java 9


dain:
looks good

PR #443, opened 2016-08-09
Add ConditionalModule

cberner:
This can be used to conditionally bind a module based on a config
property or properties

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConditionalModule.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConditionalModule.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.configuration;
+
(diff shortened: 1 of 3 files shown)


configuration/src/main/java/io/airlift/configuration/ConditionalModule.java
```diff
@@ -0,0 +1,53 @@
...
+        return new ConditionalModule<>(config, predicate, module);
+    }
+
+    private final Class<T> config;
+    private final Predicate<T> predicate;
+    private final Module module;
+
+    private ConditionalModule(Class<T> config, Predicate<T> predicate, Module module)
+    {
+        this.config = config;
```
electrum: Check for null

configuration/src/main/java/io/airlift/configuration/ConditionalModule.java
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+package io.airlift.configuration;
+
+import com.google.inject.Binder;
+import com.google.inject.Module;
+
+import java.util.function.Predicate;
+
+public class ConditionalModule<T>
```
electrum: Add @Beta

configuration/src/main/java/io/airlift/configuration/ConditionalModule.java
```diff
@@ -0,0 +1,53 @@
...
+        this.module = module;
+    }
+
+    @Override
+    protected void setup(Binder binder)
+    {
+        T configuration = buildConfigObject(config);
+        if (predicate.test(configuration)) {
+            if (module instanceof ConfigurationAwareModule) {
+                install((ConfigurationAwareModule) module);
```
electrum: Can you add a version in the parent that takes a normal Module? The current way of only taking ConditionalModule is confusing and leads to mistakes.

cberner:
Updated

electrum:
Looks good, update CHANGES

PR #444, opened 2016-08-10
Assert that latch waits don't time out

johngmyers:


johngmyers:
Second commit fixes #445

erichwang:
looks good

erichwang:
Merged, thanks!

PR #446, opened 2016-08-12
Parameterize the JMX agent for Java 8 vs 9

martint:
Using the VirtualMachine API requires access to tools.jar in Java 8.
This creates a number of tooling issues, including how to reliably
get the location of that file when running with JDK 8.

To sidestep those issues this change resurrects the old behavior
and introduces a dynamic switch to pick the appropriate implementation
given the version of Java being used.


dain:
Looks good

PR #448, opened 2016-08-23
Fix inconsistent TestMoreFutures.testTimeout

### LINKED ISSUE #447: TestMoreFutures.testTimeout() failures due to race
The assertion

`assertTrue(rootFuture.isDone());`

intermittently fails because `TimeoutFutureTask.run()` completes `timeoutFuture` before canceling `rootFuture`. The assertion can run and fail before rootFuture is canceled.

@dain should it be a requirement that the rootFuture be canceled before the timeoutFuture is completed?

dain:
There is a race between canceling the inner and outer future

Fixes #447


erichwang:
looks good

PR #456, opened 2016-09-01
Revert Jetty to 9.3.9.M1

electrum:
We saw sporadic issues in production with 9.3.11 where
client would send randomly corrupted requests.


dain:
Looks good

PR #457, opened 2016-09-08
Set ThreadContextClassLoader to null when loading bval

raghavsethi:
This forces bval to use getClass().getClassLoader() instead of the
ThreadContextClassLoader (which is initialized to the system class
loader).

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
@@ -64,7 +64,19 @@ import static java.util.stream.Collectors.toList;
 
 public class ConfigurationFactory
 {
-    private static final Validator VALIDATOR = Validation.byProvider(ApacheValidationProvider.class).configure().buildValidatorFactory().getValidator();
+    private static final Validator VALIDATOR;
+
+    static {
+        // this prevents bval from using the thread context classloader
+        ClassLoader currentClassLoader = Thread.currentThread().getContextClassLoader();
+        try {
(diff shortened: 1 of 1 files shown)


configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -64,7 +64,15 @@
 
 public class ConfigurationFactory
 {
-    private static final Validator VALIDATOR = Validation.byProvider(ApacheValidationProvider.class).configure().buildValidatorFactory().getValidator();
+    private static final Validator VALIDATOR;
+
+    static {
+        // this prevents bval from using the thread context classloader
+        ClassLoader threadContextClassLoader = Thread.currentThread().getContextClassLoader();
```
electrum: Put the save/restore in a try/finally

electrum:
Minor comment, otherwise looks good

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
electrum:
```java
ClassLoader currentClassLoader = Thread.currentThread().getContextClassLoader();
try {
    Thread.currentThread().setContextClassLoader(null);
    VALIDATOR = Validation.byProvider(ApacheValidationProvider.class).configure().buildValidatorFactory().getValidator();
}
finally {
    Thread.currentThread().setContextClassLoader(currentClassLoader);
}

raghavsethi:
@electrum do you want to release?

PR #458, opened 2016-09-19
Fix copy request

arhimondr:


arhimondr:
Fix has been merged to jetty. https://github.com/eclipse/jetty.project/commit/a9ef53b6d0d37190f305b5339507b642b635a878. After jetty is release we can update the version.

arhimondr:
@martint This bug is fixed in `jetty-9.3.13.v20161014`. Can we update jetty to that release instead of `jetty-9.3.13.M0` (https://github.com/airlift/airlift/commit/e21554a836f8da8e3f793ee7de03d79974861f27)

arhimondr:
Replaced with: https://github.com/airlift/airlift/pull/483

PR #460, opened 2016-09-20
Add toJsonWithLengthLimit to JsonCodec

raghavsethi:
Also added JsonLengthLimitExceededException to signal this condition.

### DIFF
--- a/json/src/main/java/io/airlift/json/JsonCodec.java
+++ b/json/src/main/java/io/airlift/json/JsonCodec.java
@@ -21,11 +21,14 @@ import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
 import com.google.common.reflect.TypeParameter;
 import com.google.common.reflect.TypeToken;
+import io.airlift.json.LengthLimitedWriter.LengthLimitExceededException;
 
 import java.io.IOException;
+import java.io.StringWriter;
 import java.lang.reflect.Type;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import static com.fasterxml.jackson.databind.SerializationFeature.INDENT_OUTPUT;
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -151,6 +154,29 @@ public class JsonCodec<T>
     {
         try {
             return mapper.writeValueAsString(instance);
+        }
+        catch (IOException e) {
+            throw new IllegalArgumentException(String.format("%s could not be converted to json", instance.getClass().getName()), e);
+        }
+    }
+
+    /**
+     * Converts the specified instance to optional json string with a length limit. Returns Optional.empty() if length limit is exceeded.
+     *
+     * @param instance the instance to convert to json
+     * @param lengthLimit the maximum length of the serialized string in bytes
+     * @return json string
+     * @throws IllegalArgumentException if the specified instance can not be converted to json
+     */
+    public Optional<String> toJsonWithLengthLimit(T instance, int lengthLimit)
+    {
+        try (StringWriter stringWriter = new StringWriter();
+                LengthLimitedWriter lengthLimitedWriter = new LengthLimitedWriter(stringWriter, lengthLimit)) {
+            mapper.writeValue(lengthLimitedWriter, instance);
+            return Optional.of(stringWriter.getBuffer().toString());
+        }
+        catch (LengthLimitExceededException e) {
+            return Optional.empty();
         }
         catch (IOException e) {
             throw new IllegalArgumentException(String.format("%s could not be converted to json", instance.getClass().getName()), e);
--- a/json/src/test/java/io/airlift/json/TestJsonCodec.java
+++ b/json/src/test/java/io/airlift/json/TestJsonCodec.java
@@ -15,15 +15,18 @@
  */
 package io.airlift.json;
 
+import com.google.common.base.Strings;
 import com.google.common.reflect.TypeToken;
 import org.testng.annotations.Test;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
 import static io.airlift.json.JsonCodec.jsonCodec;
 import static io.airlift.json.JsonCodec.listJsonCodec;
 import static io.airlift.json.JsonCodec.mapJsonCodec;
+import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.assertNull;
 
 public class TestJsonCodec
@@ -161,5 +164,30 @@ public class TestJsonCodec
         JsonCodec<Map<String, ImmutablePerson>> jsonCodec = jsonCodec(new TypeToken<Map<String, ImmutablePerson>>() {});
 
         ImmutablePerson.validatePersonMapJsonCodec(jsonCodec);
+    }
+
+    @Test
+    public void testToJsonWithLengthLimitSimple()
+    {
+        JsonCodec<ImmutablePerson> jsonCodec = jsonCodec(ImmutablePerson.class);
+        ImmutablePerson person = new ImmutablePerson(Strings.repeat("a", 1000), false);
+
+        assertFalse(jsonCodec.toJsonWithLengthLimit(person, 0).isPresent());
+        assertFalse(jsonCodec.toJsonWithLengthLimit(person, 1000).isPresent());
+        assertFalse(jsonCodec.toJsonWithLengthLimit(person, 1035).isPresent());
+        jsonCodec.toJsonWithLengthLimit(person, 1036);
+    }
+
+    @Test
+    public void testToJsonWithLengthLimitComplex()
+    {
+        JsonCodec<List<ImmutablePerson>> jsonCodec = listJsonCodec(jsonCodec(ImmutablePerson.class));
+        ImmutablePerson person = new ImmutablePerson(Strings.repeat("a", 1000), false);
+        List<ImmutablePerson> people = Collections.nCopies(10, person);
+
+        assertFalse(jsonCodec.toJsonWithLengthLimit(people, 0).isPresent());
+        assertFalse(jsonCodec.toJsonWithLengthLimit(people, 5000).isPresent());
+        assertFalse(jsonCodec.toJsonWithLengthLimit(people, 10381).isPresent());
+        jsonCodec.toJsonWithLengthLimit(people, 10382);
     }
 }
--- a/json/src/main/java/io/airlift/json/LengthLimitedWriter.java
+++ b/json/src/main/java/io/airlift/json/LengthLimitedWriter.java
@@ -1,0 +1,51 @@
+package io.airlift.json;
+
+import java.io.IOException;
+import java.io.Writer;
+
+import static java.util.Objects.requireNonNull;
+
+class LengthLimitedWriter
+        extends Writer
+{
+    private final Writer writer;
+    private final int maxLength;
+    private int count;
+
+    public LengthLimitedWriter(Writer writer, int maxLength)
+    {
+        this.writer = requireNonNull(writer, "writer is null");
+        this.maxLength = maxLength;
+    }
+
+    @Override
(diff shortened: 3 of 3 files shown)


json/src/test/java/io/airlift/json/TestJsonCodec.java
```diff
@@ -162,4 +165,45 @@ public void testImmutableTypeTokenMap()
...
+    @Test
+    public void testToJsonWithLengthLimitSimple()
+            throws JsonLengthLimitExceededException
+    {
+        JsonCodec<ImmutablePerson> jsonCodec = jsonCodec(ImmutablePerson.class);
+        ImmutablePerson person = new ImmutablePerson(Strings.repeat("a", 1000), false);
+
+        try {
+            jsonCodec.toJsonWithLengthLimit(person, 1011);
+            fail("expected IllegalArgumentException");
```
haozhun: The string is wrong.

json/src/test/java/io/airlift/json/TestJsonCodec.java
```diff
@@ -162,4 +165,45 @@ public void testImmutableTypeTokenMap()
...
+        }
+        catch (JsonLengthLimitExceededException e) {}
```
haozhun: I'm not sure. But this doesn't look like acceptable code style.

json/src/test/java/io/airlift/json/TestJsonCodec.java
```diff
@@ -162,4 +165,45 @@ public void testImmutableTypeTokenMap()
...
     }
+
+    @Test
+    public void testToJsonWithLengthLimitSimple()
+            throws JsonLengthLimitExceededException
+    {
+        JsonCodec<ImmutablePerson> jsonCodec = jsonCodec(ImmutablePerson.class);
+        ImmutablePerson person = new ImmutablePerson(Strings.repeat("a", 1000), false);
+
+        try {
```
haozhun: Abstract into `assertLengthExceeded` function

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -158,6 +162,30 @@ public String toJson(T instance)
...
+    public String toJsonWithLengthLimit(T instance, int lengthLimit)
+            throws JsonLengthLimitExceededException
+    {
+        try (StringWriter stringWriter = new StringWriter();
+                LengthLimitedWriter lengthLimitedWriter = new LengthLimitedWriter(stringWriter, lengthLimit)) {
+            mapper.writeValue(lengthLimitedWriter, instance);
+            return stringWriter.getBuffer().toString();
+        }
+        catch (LengthLimitedWriter.LengthLimitExceededException e) {
+            throw new JsonLengthLimitExceededException(String.format("%s could not be converted to json string with length limit %d", instance.getClass().getName(), lengthLimit), e);
```
haozhun: You don't need `e` here.

haozhun: I'm not sure about the usefulness of having `getClass().getName()` here. I've not seen this before.

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -198,4 +226,49 @@ public T fromJson(byte[] json)
...
+        }
+
+        @Override
+        public void close()
+                throws IOException
+        {
+            writer.close();
+        }
+
+        // this needs to extend IOException specifically so that Jackson doesn't wrap it
```
haozhun: `// This needs to extend IOException so that Jackson doesn't wrap it.`

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -158,6 +162,30 @@ public String toJson(T instance)
     }
 
     /**
+     * Converts the specified instance to json string with a maximum length limit.
+     *
+     * @param instance the instance to convert to json
+     * @param lengthLimit the maximum length of the serialized string
+     * @return json string
+     * @throws IllegalArgumentException if the specified instance can not be converted to json or if length limit is exceeded
```
haozhun: Add `@throws JsonLengthLimitExceededException ...`

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -158,6 +162,30 @@ public String toJson(T instance)
     }
 
     /**
+     * Converts the specified instance to json string with a maximum length limit.
```
haozhun: Remove `maximum`

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -17,18 +17,22 @@
 
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
```
haozhun: Unused import

haozhun:
@martint @electrum @dain 

I would like a confirmation about the signature:

```
public String toJsonWithLengthLimit(T instance, int lengthLimit)
        throws JsonLengthLimitExceededException
```

I would also like a confirmation about where this method should live. For example,
* in `JsonCodec`, or
* in a utility class (we can call it `JsonCodecUtil`) and take `JsonCodec` as an argument.

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -158,6 +161,31 @@ public String toJson(T instance)
     }
 
     /**
+     * Converts the specified instance to json string with a length limit.
+     *
+     * @param instance the instance to convert to json
```
electrum: JSON should be capitalized

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -158,6 +161,31 @@ public String toJson(T instance)
...
+     */
+    public String toJsonWithLengthLimit(T instance, int lengthLimit)
+            throws JsonLengthLimitExceededException
+    {
+        try (StringWriter stringWriter = new StringWriter();
+                LengthLimitedWriter lengthLimitedWriter = new LengthLimitedWriter(stringWriter, lengthLimit)) {
+            mapper.writeValue(lengthLimitedWriter, instance);
+            return stringWriter.getBuffer().toString();
+        }
+        catch (LengthLimitedWriter.LengthLimitExceededException e) {
```
electrum: Static import

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -158,6 +161,31 @@ public String toJson(T instance)
...
+            throw new JsonLengthLimitExceededException(String.format("%s could not be converted to json string with length limit %d", instance.getClass().getName(), lengthLimit), e);
```
electrum: Static import

electrum: Use `%s` in format strings. It's safer

json/src/main/java/io/airlift/json/JsonCodec.java
haozhun: json is not capitalized throughout this file

json/src/main/java/io/airlift/json/JsonCodec.java
haozhun: You can't do that. It's not public.

json/src/main/java/io/airlift/json/JsonCodec.java
electrum: I don't think we need to propagate an internal exception as the cause

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -158,6 +161,31 @@ public String toJson(T instance)
     }
 
     /**
+     * Converts the specified instance to json string with a length limit.
+     *
+     * @param instance the instance to convert to json
+     * @param lengthLimit the maximum length of the serialized string
```
electrum: This should say "maximum length of the JSON string in characters"

json/src/main/java/io/airlift/json/JsonCodec.java
electrum: "to JSON string" sounds weird. Maybe just say "to JSON"?

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -198,4 +226,49 @@ public T fromJson(byte[] json)
     {
         return (TypeToken<T>) TypeToken.of(type);
     }
+
+    private static class LengthLimitedWriter
```
electrum: Make this a package-private class

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -198,4 +226,49 @@ public T fromJson(byte[] json)
...
+
+        @Override
+        public void close()
+                throws IOException
+        {
+            writer.close();
+        }
+
+        // this needs to extend IOException so that Jackson doesn't wrap it
+        static class LengthLimitExceededException
```
electrum: This should be public since it's part of this class' interface.

json/src/main/java/io/airlift/json/JsonLengthLimitExceededException.java
```diff
@@ -0,0 +1,10 @@
+package io.airlift.json;
+
+public class JsonLengthLimitExceededException
+    extends Exception
```
electrum: Should be 8 spaces

json/src/test/java/io/airlift/json/TestJsonCodec.java
```diff
@@ -162,4 +165,37 @@ public void testImmutableTypeTokenMap()
...
+    @Test
+    public void testToJsonWithLengthLimitComplex()
+            throws JsonLengthLimitExceededException
+    {
+        JsonCodec<List<ImmutablePerson>> jsonCodec = listJsonCodec(jsonCodec(ImmutablePerson.class));
+        ImmutablePerson person = new ImmutablePerson(Strings.repeat("a", 1000), false);
+        List<ImmutablePerson> people = Collections.nCopies(10, person);
+
+        assertLengthLimitExceeded(jsonCodec, people, 10000);
+        jsonCodec.toJsonWithLengthLimit(people, 10382);
```
electrum: Add a failure for `10381` (assuming that is the boundary)

json/src/test/java/io/airlift/json/TestJsonCodec.java
```diff
@@ -162,4 +165,37 @@ public void testImmutableTypeTokenMap()
...
+
+    @Test
+    public void testToJsonWithLengthLimitSimple()
+            throws JsonLengthLimitExceededException
+    {
+        JsonCodec<ImmutablePerson> jsonCodec = jsonCodec(ImmutablePerson.class);
+        ImmutablePerson person = new ImmutablePerson(Strings.repeat("a", 1000), false);
+
+        assertLengthLimitExceeded(jsonCodec, person, 1011);
+        jsonCodec.toJsonWithLengthLimit(person, 1036);
```
electrum: Add a failure for `1035`

electrum:
I prefer to have this method return `Optional`.

json/src/main/java/io/airlift/json/JsonCodec.java
haozhun: in bytes

json/src/test/java/io/airlift/json/TestJsonCodec.java
haozhun: What about write a loop?

json/src/test/java/io/airlift/json/TestJsonCodec.java
haozhun: What about write a loop?

electrum:
Are you sure it's bytes? The writer seems to be counting characters.

PR #463, opened 2016-09-22
Fix flaky HTTP client test

electrum:


dain:
Changes look reasonable.  The tests failed with some random error fetching artifacts from maven central, so I restarted them.  Assuming they pass, push it.

PR #472, opened 2016-10-14
Implement BasicAuthenticationFilter

arhimondr:
Allows to pre-authenticate requests with the basic authentication

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/BasicAuthenticationFilter.java
+++ b/http-client/src/main/java/io/airlift/http/client/BasicAuthenticationFilter.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.http.client;
+
+import java.util.Base64;
+import java.util.function.Predicate;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.net.HttpHeaders.AUTHORIZATION;
+import static io.airlift.http.client.Request.Builder.fromRequest;
+import static java.nio.charset.StandardCharsets.ISO_8859_1;
+import static java.util.Objects.requireNonNull;
+
+public class BasicAuthenticationFilter
+        implements HttpRequestFilter
+{
+    private final Predicate<Request> requestMatcher;
+    private final String authenticationHeader;
+
+    public BasicAuthenticationFilter(String user, String password)
+    {
+        this((r) -> true, user, password);
+    }
+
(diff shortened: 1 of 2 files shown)


arhimondr:
@electrum We are going to implement LDAP authentication for Presto internal communication and support it in JDBC driver. We have already implemented LDAP in CLI. I think this is the best place where to put this class to be able to reuse it across the modules.

electrum:
This looks good to me, but can you add a test?

arhimondr:
@electrum 
I'm not sure if i can implement end-to-end test easily. As i know `HttpServer` from Airlift does not support basic authentication, and i'm not sure if we need to implement it now. But i can add unit tests.

electrum:
We should be able to bind a custom filter that does basic auth (just like in the Presto server). But a unit test is fine here.

arhimondr:
@electrum done

http-client/src/test/java/io/airlift/http/client/BasicAuthenticationFilterTest.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.testng.Assert.assertNull;
+import static org.testng.Assert.assertTrue;
+
+public class BasicAuthenticationFilterTest
+{
+    @Test
+    public void testBasicAuthentication()
+            throws Exception
+    {
+        BasicAuthenticationFilter filter = new BasicAuthenticationFilter((r) -> r.getUri().getPath().startsWith("/private"), "user", "password");
```
electrum: Don't put paranthesis around single argument

electrum: This line is a bit long. It might be more readable to wrap or introduce a variable for the predicate.

electrum: I think the variable can be declared as the interface type `HttpRequestFilter`

http-client/src/test/java/io/airlift/http/client/BasicAuthenticationFilterTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.testng.annotations.Test;
+
+import java.net.URI;
+import java.util.Base64;
+
+import static com.google.common.base.Charsets.UTF_8;
+import static io.airlift.http.client.HttpUriBuilder.uriBuilderFrom;
+import static io.airlift.http.client.Request.Builder.prepareGet;
+import static org.eclipse.jetty.http.HttpHeader.AUTHORIZATION;
```
electrum: Don't use this. Use Guava instead.

http-client/src/test/java/io/airlift/http/client/BasicAuthenticationFilterTest.java
```diff
@@ -0,0 +1,60 @@
...
+public class BasicAuthenticationFilterTest
+{
+    @Test
+    public void testBasicAuthentication()
+            throws Exception
+    {
+        BasicAuthenticationFilter filter = new BasicAuthenticationFilter((r) -> r.getUri().getPath().startsWith("/private"), "user", "password");
+
+        Request publicResourceRequest = createTestRequest("/public");
+        assertNull(filter.filterRequest(publicResourceRequest).getHeader(AUTHORIZATION.name()));
```
electrum: I'd use the string constant for the test.

arhimondr:
@electrum updated

electrum:
Merged, thanks!

PR #473, opened 2016-10-17
Implement HttpClientDefaultsBinder

arhimondr:
This binder helps to provide default configuration values and filters
that must be applied to all the HttpClients within the same Guice context.

This is introduced to cover the following use case:

In Presto multiple Http clients are being used by multiple modules. However
configuration parameters, such as SSL certificate, or some security filter must
be common for all of them. So, instead of setting these values to every single client,
what could be tricky and fragile, the HttpClientDefaultsBinder is introduced.

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigBinder.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigBinder.java
@@ -110,6 +110,15 @@ public class ConfigBinder
         createConfigDefaultsBinder(key).addBinding().toInstance(new ConfigDefaultsHolder<>(key, configDefaults));
     }
 
+    /**
+     * Binds default values for all the instances of given config class for the current binder
+     */
+    public <T> void bindConfigGeneralDefaults(Class<T> configClass, ConfigDefaults<T> configDefaults)
+    {
+        Key<T> key = Key.get(configClass, GeneralDefaults.class);
+        createConfigDefaultsBinder(key).addBinding().toInstance(new ConfigDefaultsHolder<>(key, configDefaults));
+    }
+
     private <T> Multibinder<ConfigDefaultsHolder<T>> createConfigDefaultsBinder(Key<T> key)
     {
         @SuppressWarnings("SerializableInnerClassWithNonSerializableOuterClass")
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
@@ -148,11 +148,23 @@ public class ConfigurationFactory
 
     private <T> ConfigDefaults<T> getConfigDefaults(Key<T> key)
     {
-        List<ConfigDefaults<T>> defaults = registeredDefaultConfigs.get(key).stream()
+        Key generalDefaultsKey = Key.get(key.getTypeLiteral(), GeneralDefaults.class);
+        List<ConfigDefaults<T>> generalDefaults = registeredDefaultConfigs.get(generalDefaultsKey).stream()
                 .map(holder -> (ConfigDefaultsHolder<T>) holder)
                 .sorted()
                 .map(ConfigDefaultsHolder::getConfigDefaults)
                 .collect(toList());
+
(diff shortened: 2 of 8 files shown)


arhimondr:
@electrum Please review.

arhimondr:
@martint Please review

http-client/src/main/java/io/airlift/http/client/HttpClientDefaultsBinder.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2010 Proofpoint, Inc.
```
martint: remove this line

configuration/src/main/java/io/airlift/configuration/ConfigBinder.java
```diff
@@ -110,6 +110,15 @@ private ConfigBinder(Binder binder)
         createConfigDefaultsBinder(key).addBinding().toInstance(new ConfigDefaultsHolder<>(key, configDefaults));
     }
 
+    /**
+     * Binds default values for all the instances of given config class for the current binder
+     */
+    public <T> void bindConfigGeneralDefaults(Class<T> configClass, ConfigDefaults<T> configDefaults)
```
martint: How is this different from `public <T> void bindConfigDefaults(Class<T> configClass, ConfigDefaults<T> configDefaults)`?

arhimondr: `bindConfigGeneralDefaults` binds defaults for all the instances of that particular configuration class. e.g. `key(ConfigClass.class, Annotation1.class)`, `key(ConfigClass.class, Annotation2.class)`,  `key(ConfigClass.class)`

`bindConfigDefaults` binds defaults for the instance with only the `Class` in key. e.g.: `key(ConfigClass.class)`

dain: Can rename `General` to `Global` everywhere?

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -148,12 +148,24 @@ public void registerConfigurationClasses(Module module)
...
                 .map(ConfigDefaultsHolder::getConfigDefaults)
                 .collect(toList());
 
+        List<ConfigDefaults<T>> keyDefaults = registeredDefaultConfigs.get(key).stream()
+                .map(holder -> (ConfigDefaultsHolder<T>) holder)
+                .sorted()
+                .map(ConfigDefaultsHolder::getConfigDefaults)
+                .collect(toList());
+
+        List<ConfigDefaults<T>> defaults = ImmutableList.<ConfigDefaults<T>>builder()
```
dain: If you declare this at the top of the method, you can change the stream processing above from `.collect(toList());` to `.forEach(defaults::add));`

http-client/src/main/java/io/airlift/http/client/HttpClientDefaultsBinder.java
```diff
@@ -0,0 +1,51 @@
...
+import com.google.inject.Binder;
+import com.google.inject.binder.LinkedBindingBuilder;
+import com.google.inject.multibindings.Multibinder;
+import io.airlift.configuration.ConfigDefaults;
+
+import static com.google.inject.multibindings.Multibinder.newSetBinder;
+import static io.airlift.configuration.ConfigBinder.configBinder;
+import static java.util.Objects.requireNonNull;
+
+public class HttpClientDefaultsBinder
```
dain: I don't think we need this class.  It is some pretty simple syntactic sugar.

dain:
For the global HTTP client filter, I think we should add these methods to `HttpClientBinder`:
```java
        public LinkedBindingBuilder<HttpRequestFilter> addGlobalFilterBinding()
        {
            ...
        }

        public HttpClientBinder bindGlobalFilter(Class<? extends HttpRequestFilter> filterClass)
        {
            ...
        }
```

arhimondr:
@dain Replaced with 
https://github.com/airlift/airlift/pull/489
https://github.com/airlift/airlift/pull/488

PR #474, opened 2016-10-17
Upgrade to Jetty 9.3.13.M0

cberner:
This fixes a race when using async responses which can result in a 500
error being returned


dain:
Looks good

PR #475, opened 2016-10-19
Add equals/hashCode to HttpClientConfig

ebd2:
This work is part of enabling the presto-jdbc driver to talk to multiple presto servers.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
@@ -28,6 +28,7 @@ import io.airlift.units.MinDuration;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
 
+import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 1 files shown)


electrum:
This seems like a bad idea, especially because these objects are mutable. Can you explain how this is used?

ebd2:
https://github.com/Teradata/presto/pull/391/files/1c18ff48c1fcabf204f3d8d7e296825fb1ec43a0#diff-309bc5b7d1d3088c40161acfd320a8d4R76

Is this something we could builder-ify and make immutable? I'd be willing to bet it's not something we actually mutate that often.

electrum:
No, the entire config system is designed around them being beans.

How does this relate to JDBC?

ebd2:
We need a key for the HTTP client cache in the presto-jdbc module. Ultimately, they HttpClients in the cache are distinguished by their configuration, which makes using HttpClientConfig the natural choice. The alternative is using something that represents the connection parameters that end up being used to set stuff in HttpClientConfig. I'm fine with using that option if this is unworkable.

PR #477, opened 2016-10-21
Add unwrapCompletionException to MoreFutures

electrum:


concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
```diff
@@ -33,6 +35,17 @@
     private MoreFutures() { }
 
     /**
+     * Unwrap a throwable that has been wrapped in a {@link CompletionException}.
```
martint: Maybe say "attempts to unwrap", given that if the cause is null it will still return a completionexception

PR #480, opened 2016-11-01
Buffer HTTP data with chunks

geraint0923:
- Avoid large byte array by buffering data with chunks
- Use ScatteringByteArrayInputStream to wrap buffered data

Fixes https://github.com/prestodb/presto/issues/5350

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
+++ b/http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
@@ -1,0 +1,116 @@
+package io.airlift.http.client;
+
+import javax.annotation.concurrent.GuardedBy;
+import javax.annotation.concurrent.ThreadSafe;
+
+import java.io.InputStream;
+import java.util.Iterator;
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkPositionIndexes;
+import static com.google.common.base.Preconditions.checkState;
+import static java.lang.Math.min;
+import static java.lang.System.arraycopy;
+import static java.util.Objects.requireNonNull;
+
+@ThreadSafe
+public class GatheringByteArrayInputStream
+        extends InputStream
+{
+    @GuardedBy("this")
+    private final Iterator<byte[]> buffers;
+    @GuardedBy("this")
+    private final byte[] singleByte = new byte[1];
+    @GuardedBy("this")
+    private byte[] currentBuffer = new byte[0];
+    @GuardedBy("this")
+    private int currentBufferPosition;
+    @GuardedBy("this")
+    private long remainingBytes;
+
+    public GatheringByteArrayInputStream(List<byte[]> buffers, long totalBytes)
+    {
+        checkArgument(totalBytes >= 0, "totalBytes should equal to or greater than 0");
+
+        this.buffers = requireNonNull(buffers, "buffers is null").iterator();
+        this.remainingBytes = totalBytes;
+    }
+
+    @Override
+    public synchronized int read(byte[] buffer)
+    {
+        return read(buffer, 0, buffer.length);
+    }
+
+    @Override
+    public synchronized int read()
+    {
+        int bytes = read(singleByte);
+        if (bytes == -1) {
+            return -1;
+        }
+        return singleByte[0] & 0xFF;
+    }
+
+    @Override
+    public synchronized long skip(long n)
+    {
+        if (n < 0) {
+            return 0;
+        }
+
+        long totalSkippedBytes = min(n, remainingBytes);
+
+        n = totalSkippedBytes;
+        while (n > 0) {
+            if (currentBufferPosition >= currentBuffer.length) {
+                advanceCurrentBuffer();
+            }
+            int skippedBytes = (int) min(n, currentBuffer.length - currentBufferPosition);
+            n -= skippedBytes;
+            currentBufferPosition += skippedBytes;
+        }
+        remainingBytes -= totalSkippedBytes;
+        return totalSkippedBytes;
+    }
+
+    @Override
+    public synchronized int read(byte[] buffer, int offset, int length)
+    {
+        requireNonNull(buffer, "buffer is null");
+        checkPositionIndexes(offset, offset + length, buffer.length);
+
+        if (remainingBytes == 0) {
+            return -1;
+        }
+
+        int totalReadBytes = (int) min(length, remainingBytes);
+
+        length = totalReadBytes;
+        while (length > 0) {
+            if (currentBufferPosition >= currentBuffer.length) {
+                advanceCurrentBuffer();
+            }
+            int readBytes = min(length, currentBuffer.length - currentBufferPosition);
+            arraycopy(currentBuffer, currentBufferPosition, buffer, offset, readBytes);
+            offset += readBytes;
+            length -= readBytes;
+            currentBufferPosition += readBytes;
+        }
+        remainingBytes -= totalReadBytes;
+        return totalReadBytes;
+    }
+
+    @Override
+    public void close() {}
+
+    private synchronized void advanceCurrentBuffer()
+    {
+        checkState(currentBufferPosition >= currentBuffer.length, "there is still un-read space in currentBuffer");
+        checkState(buffers.hasNext(), "buffers should have more data when remainingBytes is greater than 0");
+
+        currentBuffer = buffers.next();
+        currentBufferPosition = 0;
+    }
+}
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
@@ -1,6 +1,5 @@
 package io.airlift.http.client.jetty;
 
-import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 import com.google.common.collect.AbstractIterator;
 import com.google.common.collect.ImmutableList;
@@ -12,6 +11,7 @@ import com.google.common.primitives.Ints;
 import com.google.common.util.concurrent.AbstractFuture;
 import io.airlift.http.client.BodyGenerator;
 import io.airlift.http.client.FileBodyGenerator;
+import io.airlift.http.client.GatheringByteArrayInputStream;
 import io.airlift.http.client.HeaderName;
 import io.airlift.http.client.HttpClientConfig;
 import io.airlift.http.client.HttpRequestFilter;
@@ -25,6 +25,7 @@ import io.airlift.http.client.spnego.SpnegoAuthentication;
 import io.airlift.http.client.spnego.SpnegoAuthenticationStore;
 import io.airlift.log.Logger;
 import io.airlift.stats.Distribution;
+import io.airlift.units.DataSize;
 import io.airlift.units.Duration;
 import org.eclipse.jetty.client.DuplexConnectionPool;
 import org.eclipse.jetty.client.HttpClient;
@@ -59,7 +60,6 @@ import org.weakref.jmx.Nested;
 import javax.annotation.concurrent.GuardedBy;
 import javax.annotation.concurrent.ThreadSafe;
 
-import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InterruptedIOException;
@@ -67,6 +67,7 @@ import java.io.OutputStream;
 import java.net.URI;
 import java.nio.ByteBuffer;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
@@ -87,7 +88,12 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+import static io.airlift.units.DataSize.Unit.KILOBYTE;
+import static io.airlift.units.DataSize.Unit.MEGABYTE;
+import static java.lang.Math.max;
 import static java.lang.Math.min;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
@@ -1045,21 +1051,28 @@ public class JettyHttpClient
         }
     }
 
+    @ThreadSafe
     private static class BufferingResponseListener
             extends Listener.Adapter
     {
+        private static final long BUFFER_MAX_BYTES = new DataSize(1, MEGABYTE).toBytes();
+        private static final long BUFFER_MIN_BYTES = new DataSize(1, KILOBYTE).toBytes();
         private final JettyResponseFuture<?, ?> future;
         private final int maxLength;
 
         @GuardedBy("this")
-        private byte[] buffer = new byte[0];
+        private byte[] currentBuffer = new byte[0];
         @GuardedBy("this")
-        private int size;
+        private int currentBufferPosition;
+        @GuardedBy("this")
+        private List<byte[]> buffers = new ArrayList<>();
+        @GuardedBy("this")
+        private long size;
 
         public BufferingResponseListener(JettyResponseFuture<?, ?> future, int maxLength)
         {
             this.future = checkNotNull(future, "future is null");
-            Preconditions.checkArgument(maxLength > 0, "maxLength must be greater than zero");
+            checkArgument(maxLength > 0, "maxLength must be greater than zero");
             this.maxLength = maxLength;
         }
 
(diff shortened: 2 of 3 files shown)


electrum:
I don't think this is the correct approach. The length limit is a safety feature. Without this, the server can cause the client to consume unbounded amounts of memory.

electrum:
If this feature is removed, the config should also be removed (and marked defunct).

electrum:
This change also causes the client to create huge arrays. So even if it has, say, 1GB available, it might not be able to allocate that much as a contiguous array.

We either need the server to break up the response into multiple responses, or we need to redesign the HTTP client interface to support streaming. This is supported for sync request, but not async (it returns an `InputStream` which already has the entire response buffered).

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -1049,27 +1038,21 @@ public void close()
             extends Listener.Adapter
     {
         private final JettyResponseFuture<?, ?> future;
-        private final int maxLength;
 
         @GuardedBy("this")
         private byte[] buffer = new byte[0];
```
dain: Instead of using a single `byte[]` use a `List<Slice>` and use a modified version of `ChunkedSliceInput`.  Also the I would grow the buffered chunk from 1k to 1MB over time.

geraint0923:
@dain   Refactored with new SliceInputStream, and added test.

geraint0923:
Changed to use byte array instead of Slice.

http-client/src/main/java/io/airlift/http/client/ScatteringByteArrayInputStream.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright 2010 Proofpoint, Inc.
```
electrum: Remove this

http-client/src/main/java/io/airlift/http/client/ScatteringByteArrayInputStream.java
```diff
@@ -0,0 +1,90 @@
...
+import java.io.InputStream;
+import java.util.Iterator;
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.lang.Math.min;
+import static java.lang.System.arraycopy;
+import static java.util.Objects.requireNonNull;
+
+public class ScatteringByteArrayInputStream
```
electrum: This class should be named `Gathering` since it is gathering the data from multiple buffers into a single input stream. Scattering is used for reads when a single read is scattered into multiple output buffers.

http-client/src/test/java/io/airlift/http/client/TestScatteringByteArrayInputStream.java
```diff
@@ -0,0 +1,45 @@
...
+        byte[] lastStringArray = "you".getBytes();
+        byte[] lastArray = new byte[lastStringArray.length + 32];
+        arraycopy(lastStringArray, 0, lastArray, 0, lastStringArray.length);
+        ImmutableList<byte[]> arrays = ImmutableList.of(
+                "hello ".getBytes(),
+                "presto ".getBytes(),
+                "I ".getBytes(),
+                "love ".getBytes(),
+                lastArray);
+        byte[] expectedAll = "hello presto I love you".getBytes();
```
electrum: This text is a bit weird. Picking something more neutral.

geraint0923:
Addressed the comments.

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -1048,11 +1049,15 @@ public void close()
     private static class BufferingResponseListener
             extends Listener.Adapter
     {
+        private static final int LONGEST_ARRAY = 1 << 20;  // 1MB
```
dain: use `new DataSize(1, MEGABYTE).toBytes()`

dain: Rename these to BUFFER_MAX_BYTES and BUFFER_MIN_BYTES

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -1048,11 +1049,15 @@ public void close()
...
+        private static final int SHORTEST_ARRAY = 1 << 10; // 1KB
         private final JettyResponseFuture<?, ?> future;
         private final int maxLength;
+        private final List<byte[]> arrays = new ArrayList<>();
```
dain: rename to `buffers`

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -1048,11 +1049,15 @@ public void close()
...
+
+        private byte[] currentArray;
```
dain: `currentBuffer`

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -1048,11 +1049,15 @@ public void close()
...
+        private int nextBytes;
```
dain: rename `currentBufferPosition`

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
dain: all of these new fields should be annotated with `@GuradedBy("this")`

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -1070,30 +1075,26 @@ public synchronized void onHeaders(Response response)
             if (length > maxLength) {
                 response.abort(new ResponseTooLargeException());
             }
-            if (length > buffer.length) {
-                buffer = Arrays.copyOf(buffer, Ints.saturatedCast(length));
-            }
+            allocateCurrentArray();
```
dain: This is not needed here

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
dain: If you initialize this to byte[0], it will simplify the code

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -1104,7 +1105,26 @@ public synchronized void onComplete(Result result)
...
+                currentArray = null;
+                nextBytes = 0;
+                future.completed(result.getResponse(), new GatheringByteArrayInputStream(arrays, size));
+            }
+        }
+
+        // allocate a new empty array if necessary
+        private void allocateCurrentArray()
+        {
+            if (currentArray == null) {
```
dain: This section is not needed if you initialize `currentArray` to `new byte[0]`

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -1070,30 +1075,26 @@ public synchronized void onHeaders(Response response)
...
+            size += length;
+            if (size > maxLength) {
+                response.abort(new ResponseTooLargeException());
+                return;
             }
 
-            content.get(buffer, size, length);
-            size += length;
+            while (length > 0) {
+                allocateCurrentArray();
```
dain: In this case I'd prefer to see
```java
if (currentBufferPosition >= currentBuffer.length) {
     allocateNewBuffer();
}
```
This leaves the check in the main loop and puts only the allocation (i.e., the slow path) in a separate method.  Of course you should validate in the allocate method that there is not remaining space and throw and exception if not.

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,75 @@
...
+        this.arrays = requireNonNull(arrays, "arrays is null").iterator();
+        this.remainingBytes = totalBytes;
+    }
+
+    @Override
+    public synchronized int read(byte[] buffer, int offset, int length)
+    {
+        requireNonNull(buffer, "buffer is null");
+        checkArgument(offset >= 0, "offset should equal to or greater than 0");
+        checkArgument(length >= 0, "length should equal to or greater than 0");
```
dain: There is a guava preconditions method for this.... something like checkIndex

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,75 @@
...
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.lang.Math.min;
+import static java.lang.System.arraycopy;
+import static java.util.Objects.requireNonNull;
+
+public class GatheringByteArrayInputStream
+        extends InputStream
+{
+    private final Iterator<byte[]> arrays;
```
dain: rename these fields to `buffers`, `currentBuffer` and `currentBufferPosition`

Also if you initialize `currentBuffer` and `currentBufferPosition` in the constructor, it will make the code much simpler.  If `buffers` is empty, set them to `byte[0]` and `0`

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,75 @@
...
+    @Override
+    public synchronized int read(byte[] buffer, int offset, int length)
+    {
+        requireNonNull(buffer, "buffer is null");
+        checkArgument(offset >= 0, "offset should equal to or greater than 0");
+        checkArgument(length >= 0, "length should equal to or greater than 0");
+
+        int totalReadBytes = 0;
+        while (totalReadBytes < length && remainingBytes > 0 && (currentArray != null || arrays.hasNext())) {
+            if (currentArray == null) {
```
dain: once you initialize currentArray in the constructor, this can become just `if (currentPosition == currentBuffer.length)`

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,75 @@
...
+            nextByte += readBytes;
+            remainingBytes -= readBytes;
+
+            // If current array is finished, set current array to null
+            if (nextByte >= currentArray.length || remainingBytes <= 0) {
+                currentArray = null;
+            }
+        }
+
+        return totalReadBytes > 0 ? totalReadBytes : -1;
```
dain: This is invalid.  The user can request a read of zero bytes, and in that case you should return `0`. The `-1` result means end of stream.

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
dain: Also the `InputStream` java docs says these should be `IndexOutOfBoundsException`

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,75 @@
...
+        this.remainingBytes = totalBytes;
+    }
+
+    @Override
+    public synchronized int read(byte[] buffer, int offset, int length)
+    {
+        requireNonNull(buffer, "buffer is null");
+        checkArgument(offset >= 0, "offset should equal to or greater than 0");
+        checkArgument(length >= 0, "length should equal to or greater than 0");
+
```
dain: Add a check hear to handle the end of stream case
```java
if (remainingBytes == 0) {
    return -1;

File: http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java [hunk 17]
```diff
@@ -0,0 +1,75 @@
...
+    public synchronized int read(byte[] buffer)
+    {
+        return read(buffer, 0, buffer.length);
+    }
+
+    @Override
+    public synchronized int read()
+            throws IOException
+    {
+        throw new IOException("read without parameter not supported");
```
dain: This must be supported.

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public synchronized long skip(long n)
+            throws IOException
+    {
+        throw new IOException("skip not supported");
```
dain: you should implement this

http-client/src/test/java/io/airlift/http/client/TestScatteringByteArrayInputStream.java
```diff
@@ -0,0 +1,45 @@
+package io.airlift.http.client;
+
+import com.google.common.collect.ImmutableList;
+import org.testng.annotations.Test;
+
+import static java.lang.System.arraycopy;
+import static org.testng.Assert.assertEquals;
+
+public class TestScatteringByteArrayInputStream
```
dain: I think you need more extensive testing for this

geraint0923:
Addressed the comments.

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public GatheringByteArrayInputStream(List<byte[]> buffers, long totalBytes)
+    {
+        checkArgument(totalBytes >= 0, "totalBytes should equal to or greater than 0");
+        this.buffers = requireNonNull(buffers, "buffers is null").iterator();
+        this.remainingBytes = totalBytes;
+
+        currentBuffer = new byte[0];
+        currentBufferPosition = 0;
+        singleByte = new byte[1];
```
dain: you can move these initializations to the field definition

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,99 @@
...
+    private long remainingBytes;
+
+    public GatheringByteArrayInputStream(List<byte[]> buffers, long totalBytes)
+    {
+        checkArgument(totalBytes >= 0, "totalBytes should equal to or greater than 0");
+        this.buffers = requireNonNull(buffers, "buffers is null").iterator();
+        this.remainingBytes = totalBytes;
+
+        currentBuffer = new byte[0];
+        currentBufferPosition = 0;
```
dain: don't initialize a field to the java default unless required because it is `final`

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    public synchronized int read()
+    {
+        int bytes = read(singleByte);
+        if (bytes == -1) {
+            return -1;
+        }
+        return singleByte[0] & 0x000000ff;
```
dain: no need for the leading zeros.  Just do  `return singleByte[0] & 0xFF`

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    public synchronized long skip(long n)
+    {
+        if (n < 0) {
+            return 0;
+        }
+        return max(doRead(null, 0, n, false), 0);
```
dain: write an optimized version of this that just does simple math on the lengths and skips whole buffers when possible

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    public synchronized int read(byte[] buffer, int offset, int length)
+    {
+        requireNonNull(buffer, "buffer is null");
+        checkElementIndex(offset, buffer.length, "offset is out of range");
+        checkPositionIndex(length, buffer.length - offset, "length is out of range");
```
dain: I believe that `checkPositionIndexes` does this in one shot https://github.com/google/guava/wiki/PreconditionsExplained

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,99 @@
...
+
+        return (int) doRead(buffer, offset, length, true);
+    }
+
+    public void close() {}
+
+    private long doRead(byte[] buffer, int offset, long length, boolean copyData)
```
dain: Split this into one optimized method for `skip` and one for `read`.  The implementations should be in the `skip` and `read` methods respectively.

http-client/src/main/java/io/airlift/http/client/GatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,99 @@
...
+    {
+        checkArgument(buffer != null || !copyData, "buffer should be non-null when copying data is requested");
+
+        if (remainingBytes == 0) {
+            return -1;
+        }
+
+        long totalReadBytes = 0;
```
dain: I think you can get rid of this field by subtracting `readBytes` from length and adding to `offset` after every copy.  This will simplify the `while` condition  and the `min` and `arrayCopy` calls below.

http-client/src/test/java/io/airlift/http/client/TestGatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,156 @@
...
+import static org.testng.Assert.assertEquals;
+
+public class TestGatheringByteArrayInputStream
+{
+    @Test
+    public void testNormal()
+    {
+        byte[] lastStringArray = "client".getBytes();
+        byte[] lastArray = new byte[lastStringArray.length + 32];
+        arraycopy(lastStringArray, 0, lastArray, 0, lastStringArray.length);
```
dain: This can be `byte[] lastArray = Arrays.copyOf(lastStringArray, lastStringArray.length + 32`

http-client/src/test/java/io/airlift/http/client/TestGatheringByteArrayInputStream.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    @Test
+    public void testLargeData()
+    {
+        int length = 123456789;
+        byte[] expected = new byte[length];
+
+        // randomly generate the bytes
+        ThreadLocalRandom.current().nextBytes(expected);
```
dain: Use `new Random(0)` here instead so you get a consistent test.

geraint0923:
Addressed the comments.

dain:
Merged, thanks!

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
ericsun2: In my test case, the "Remote page is too large" issue sometimes can be resolved by 

`set session exchange_compression = true;`

Hope this makes sense too.

PR #483, opened 2016-11-04
Update jetty version to 9.3.13.v20161014

arhimondr:
This version contains the `requestCopy` method fix:
https://github.com/eclipse/jetty.project/pull/937

The bug in `requestCopy` method was causing failures for Kerberos authentication
in Presto CLI when the user name and the schema name was the same.


arhimondr:
False alert. The fix for the `copyRequest` method was included already in `9.3.13.M0`.

PR #486, opened 2016-11-15
Use custom ID generation for trace tokens

electrum:
Generating random UUIDs can be a performance bottleneck due to
the default implementation of SecureRandom.

### DIFF
--- a/trace-token/src/main/java/io/airlift/tracetoken/TraceTokenManager.java
+++ b/trace-token/src/main/java/io/airlift/tracetoken/TraceTokenManager.java
@@ -15,10 +15,17 @@
  */
 package io.airlift.tracetoken;
 
-import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
+
+import static java.lang.String.format;
+import static java.util.Locale.ENGLISH;
+import static java.util.UUID.randomUUID;
 
 public class TraceTokenManager
 {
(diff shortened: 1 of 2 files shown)


trace-token/src/main/java/io/airlift/tracetoken/TraceTokenManager.java
```diff
@@ -33,7 +39,7 @@ public String getCurrentRequestToken()
 
     public String createAndRegisterNewRequestToken()
     {
-        String newToken = UUID.randomUUID().toString();
+        String newToken = prefix + sequence.getAndIncrement();
```
dain: `sequence.getAndIncrement()` should be rendered in hex

trace-token/src/main/java/io/airlift/tracetoken/TraceTokenManager.java
```diff
@@ -15,10 +15,16 @@
...
 
-import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
+
+import static java.util.Locale.ENGLISH;
+import static java.util.UUID.randomUUID;
 
 public class TraceTokenManager
 {
+    private final String prefix = randomUUID().toString().toLowerCase(ENGLISH).replace("-", "") + ":";
```
dain: We don't need the `:` and it means the value can be used in any place that can tolerate hex characters

PR #487, opened 2016-11-16
Allow node internal address to be a hostname

arhimondr:
For SSL communication it is more convenient to generate certificates
for the top level host name. To do so, nodes must introduce themselves
with the hostname, instead of the IP. This change allows to either set
the internal node hostname explicitly, or to make the auto-resolve
algorithm to return node hostname instead of IP.

### DIFF
--- a/http-server/src/main/java/io/airlift/http/server/HttpServerInfo.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpServerInfo.java
@@ -16,7 +16,6 @@
 package io.airlift.http.server;
 
 import com.google.common.base.Throwables;
-import com.google.common.net.InetAddresses;
 import com.google.inject.Inject;
 import io.airlift.node.NodeInfo;
 
@@ -37,7 +36,7 @@ public class HttpServerInfo
     public HttpServerInfo(HttpServerConfig config, NodeInfo nodeInfo)
     {
         if (config.isHttpEnabled()) {
-            httpUri = buildUri("http", InetAddresses.toUriString(nodeInfo.getInternalIp()), config.getHttpPort());
+            httpUri = buildUri("http", nodeInfo.getInternalAddress(), config.getHttpPort());
             httpExternalUri = buildUri("http", nodeInfo.getExternalAddress(), httpUri.getPort());
         }
         else {
@@ -46,7 +45,7 @@ public class HttpServerInfo
         }
 
         if (config.isHttpsEnabled()) {
-            httpsUri = buildUri("https", InetAddresses.toUriString(nodeInfo.getInternalIp()), config.getHttpsPort());
+            httpsUri = buildUri("https", nodeInfo.getInternalAddress(), config.getHttpsPort());
             httpsExternalUri = buildUri("https", nodeInfo.getExternalAddress(), httpsUri.getPort());
         }
         else {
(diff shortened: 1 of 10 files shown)


node/src/main/java/io/airlift/node/NodeConfig.java
```diff
@@ -169,4 +164,21 @@ public NodeConfig setConfigSpec(String configSpec)
         this.configSpec = configSpec;
         return this;
     }
+
+    public AddressType getAddressType()
+    {
+        return addressType;
+    }
+
+    @Config("node.address-type")
```
electrum: Maybe `node.internal-address-type`? Since this is for configuring `internal-address`

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -248,14 +254,30 @@ public String toString()
...
                 .toString();
     }
 
-    private static InetAddress findPublicIp()
+    private static String findInternalAddress(AddressType addressType)
+    {
+        switch (addressType) {
+            case IP:
+                return InetAddresses.toAddrString(findInternalIp());
+            case HOSTNAME: {
```
electrum: No need for braces

arhimondr:
@electrum updated

node/src/main/java/io/airlift/node/NodeConfig.java
```diff
@@ -169,4 +164,21 @@ public NodeConfig setConfigSpec(String configSpec)
         this.configSpec = configSpec;
         return this;
     }
+
+    public AddressType getAddressType()
+    {
+        return addressType;
+    }
+
+    @Config("node.internal-address-type")
```
martint: Why does this need to be configurable? Can't it be inferred from the address itself?

arhimondr: You can specify node internal adress manually, with the `node.internal-address`. 
It this property is not specified, framework resolves the local address automatically. There are 3 types of automatically resolved adress. It can be a bare `IP` resolved by the existing algoritm. It can be a hostname. And it can be a FQDN.

arhimondr: Aparently the property name is missleading. Maybe we should call it `internal-address-resolution-method`?

martint: Maybe "internal-address-lookup-method"? Also, I can understand "find ip from interfaces" vs "use hostname", but why do we need hostname vs fqdn?

arhimondr: > internal-address-lookup-method

Yea, i like it

> but why do we need hostname vs fqdn?

We need FQDN when we want to setup HTTPS communication. When using FQDN it is possible to generate certificate for the top level domain, and use the same certificate on all the nodes.

http-server/src/main/java/io/airlift/http/server/HttpServerInfo.java
```diff
@@ -56,10 +55,10 @@ public HttpServerInfo(HttpServerConfig config, NodeInfo nodeInfo)
 
         if (config.isAdminEnabled()) {
             if (config.isHttpsEnabled()) {
-                adminUri = buildUri("https", InetAddresses.toUriString(nodeInfo.getInternalIp()), config.getAdminPort());
```
martint: toUriString wraps IPv6 addresses with square brackets. This PR makes getInternalAddress return a String that's constructed from `InetAddresses.toAddrString(ip)`, so it's not equivalent.

PR #488, opened 2016-11-16
Introduce configuration global defaults

arhimondr:
Global defaults will be applied to all of the configurations of the same class
within the same Guice context.


dain:
Merged, thanks!

PR #490, opened 2016-12-22
Set http-server blocking timeout

johngmyers:
Per recommendation in https://github.com/eclipse/jetty.project/issues/1196


electrum:
@johngmyers Thanks for sending this. I read through the Jetty issue, but the implications of this are not clear. Did you see any problems with slow requests timing out after apply this change?

johngmyers:
I have not had any problems reported. Unless the Jetty bug this works around manifests, such slow requests should trigger the idleTimeout.

johngmyers:
The documentation on this setting was misleading and it doesn't do what I expected. First, "this timeout is in addition to the getIdleTimeout()" doesn't mean the effective value is added to the idle timeout. Second, "and aplies to the total operation" doesn't mean the total blocking part of the read operation, it means the total of all reads on the ServletInputStream.

PR #503, opened 2017-02-09
Improve quantile digest copy performance

martint:


haozhun:
Commit for copy constructor looks good
Commit for benchmark looks good

PR #507, opened 2017-02-22
Convert back to ListenableFuture

dain:
The CompletableFuture interface is difficult to use correctly, and it is
easy to introduce subtle bugs that result in lost notifications. For example
CompletableFuture does not propagate cancel from wrapped futures to the
core future. Additionally, it is easy to accidentally schedule tasks on a system
wide fork join pool.

ListenableFuture on the other hand is easy to understand and reasonable, and
the transformations in Guava handle cancelation properly.

### DIFF
--- a/concurrent/src/test/java/io/airlift/concurrent/TestMoreFutures.java
+++ b/concurrent/src/test/java/io/airlift/concurrent/TestMoreFutures.java
@@ -1,6 +1,7 @@
 package io.airlift.concurrent;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
 import io.airlift.units.Duration;
 import org.testng.annotations.AfterClass;
@@ -17,6 +18,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Consumer;
 
+import static com.google.common.util.concurrent.Futures.immediateFailedFuture;
+import static com.google.common.util.concurrent.Futures.immediateFuture;
 import static io.airlift.concurrent.MoreFutures.addTimeout;
 import static io.airlift.concurrent.MoreFutures.allAsList;
 import static io.airlift.concurrent.MoreFutures.failedFuture;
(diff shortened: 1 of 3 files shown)


concurrent/src/test/java/io/airlift/concurrent/TestMoreFutures.java
```diff
@@ -530,6 +530,63 @@ public void testUnmodifiableAllAsList()
     }
 
     @Test
+    public void testListenableTimeout()
+            throws Exception
+    {
+        SettableFuture<String> rootFuture = SettableFuture.create();
+        ListenableFuture<String> timeoutFuture = addTimeout(rootFuture, () -> "timeout", new Duration(0, MILLISECONDS), executorService);
+
+        assertEquals(tryGetFutureValue(timeoutFuture, 10, SECONDS).get(), "timeout");
```
electrum: `.orElse("failed")`

concurrent/src/test/java/io/airlift/concurrent/TestMoreFutures.java
```diff
@@ -530,6 +530,63 @@ public void testUnmodifiableAllAsList()
...
+        ListenableFuture<String> timeoutFuture = addTimeout(rootFuture, () -> "timeout", new Duration(10, SECONDS), executorService);
+
+        // check timeout
+        assertEquals(tryGetFutureValue(timeoutFuture, 10, MILLISECONDS), Optional.<String>empty());
+
+        assertTrue(timeoutFuture.cancel(true));
+        assertTrue(timeoutFuture.isDone());
+        assertTrue(timeoutFuture.isCancelled());
+
+        // root exception is cancelled on a timeout
```
electrum: I think this comment is wrong. Should it be
```java
// cancellation of timeout future propagates to root future

File: http-server/src/main/java/io/airlift/http/server/AsyncResponseHandler.java [hunk 3]
```diff
@@ -46,9 +47,9 @@ private AsyncResponseHandler(AsyncResponse asyncResponse, CompletableFuture<?> f
         this.futureResponseReference = new WeakReference<>(requireNonNull(futureResponse, "futureResponse is null"));
     }
 
-    public static AsyncResponseHandler bindAsyncResponse(AsyncResponse asyncResponse, CompletableFuture<?> futureResponse, Executor httpResponseExecutor)
+    public static AsyncResponseHandler bindAsyncResponse(AsyncResponse asyncResponse, ListenableFuture<?> futureResponse, Executor httpResponseExecutor)
```
dain: In the long run, I'd like to remove all of the `CompletableFuture` code.

PR #509, opened 2017-03-06
Synchronize access to bval validators

findepi:
bval validation is not thread safe (e.g.
`org.apache.bval.jsr.ConstraintValidation#validator` is lazily, unsafely
initialized).

Fixes prestodb/presto#7512

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
@@ -72,6 +72,7 @@ import static java.util.Objects.requireNonNull;
 
 public class ConfigurationFactory
 {
+    @GuardedBy("VALIDATOR")
     private static final Validator VALIDATOR;
 
     static {
@@ -377,7 +378,7 @@ public class ConfigurationFactory
             }
         }
 
-        for (ConstraintViolation<?> violation : VALIDATOR.validate(instance)) {
+        for (ConstraintViolation<?> violation : validate(instance)) {
             String propertyFieldName = violation.getPropertyPath().toString();
             // upper case first character to match config attribute name
             String attributeName = LOWER_CAMEL.to(UPPER_CAMEL, propertyFieldName);
@@ -399,6 +400,13 @@ public class ConfigurationFactory
         problems.throwIfHasErrors();
 
         return new ConfigurationHolder<>(instance, problems);
+    }
+
+    private static <T> Set<ConstraintViolation<T>> validate(T instance)
+    {
+        synchronized (VALIDATOR) {
+            return VALIDATOR.validate(instance);
+        }
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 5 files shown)


electrum:
Can you elaborate on what exactly is unsafe? The code is doing Double Checked Locking with the initialization happening inside a synchronized block.

The only fault I can find with the code is that the `validator.initialize()` call can fail and leave `validator` non-null, but external synchronization won't fix that.

electrum:
I see the problem: other threads can see a non-null value before the initialization is complete. Please file an issue with the BVal project and reference it here, so that we can remove it later.
```java
// TODO: remove when BVAL-xxx initialization bug is fixed

martint:
> The code is doing Double Checked Locking 

That's broken. The field would need to be volatile (at a minimum) for DCL to be valid.

findepi:
This is broken both ways. 

1. No `volatile`. 
2. Field is first assigned and only then being initialized, so another thread can see _any_ state (before, during or after initialization; including inconsistent state due to lack of synchronization in actual validator's `initialize` & `isValid`)

findepi:
I see there is a newer version, `1.1.2`, which adds `volatile` keyword, but (unless I miss something) second point still applies.

findepi:
> Please file an issue with the BVal project and reference it here, so that we can remove it later.

I dont think it's a good idea. I dont trust this code. I didn't read it all, I only find one reason why our tests fail intermittently. I imagine following scenario:

1. i create an issue, add TODO here
2. they fix the class we know of
3. we upgrade and remove synchronization
4. ... only to realize our tests are unstable in some new way, because there is "Double Check" in some other place in the code. E.g. the class i found is lazy init-ing per-property validation. How is the per-class validation init-ed? 

*Unless* there is a compelling reason to do otherwise, I would assume the validator is not thread safe and use it so. 
@martint , @electrum , ok?

electrum:
Seems reasonable to me.

I missed that DCL needs volatile because other threads may never see the update or might see partial updates of the object (due to lack of ordering guarantees around the write).

findepi:
Yes, it needs the volatile and (the only) write to the volatile field must
be generally the last instruction under the lock.

Is there anything to do to have this merged?

>

electrum:
Merged, thanks!

findepi:
@electrum thanks for merging this. However, I come to have some doubts about synchronization in the `JsonMapper` class. I don't know where and how this class is used. If it's used frequently in some cases, then synchronization might become a bottleneck and some `ThreadLocal` or pooling would be needed. (Or replacing the library, o_O). What do you think?

electrum:
Maybe we should switch back to the Hibernate version. @dain?

dain:
@electrum did you file a bug report?  I don't mind converting back, but maybe we should repackage it, so I don't have to answer question about Presto using hibernate (which we don't).

findepi:
I'm not sure I follow. I'm not opting for hibernate validator, but I don't
understand  what's the problem with using it.

>

electrum:
We discussed options and decided to remove validation for jaxrs since we don't use it. We can add it back later using optional Guice bindings if anyone needs it.

PR #513, opened 2017-03-14
Update to Guava 21.0

electrum:


stats/src/test/java/io/airlift/stats/cardinality/TestDenseSerialization.java
```diff
@@ -157,7 +157,7 @@ public void testBaselineAdjusment()
         DenseHll hll = new DenseHll(4);
 
         for (int i = 0; i < 100; i++) {
-            hll.insertHash(Murmur3.hash64(i));
+            hll.insertHash(Murmur3Hash128.hash64((long) i));
```
dain: Why are the casts needed in these?

electrum: Weird, I don't know how those snuck in there. Removed. Nice catch.

PR #515, opened 2017-03-27
Bind server sockets to loopback address in unit tests

nezihyigitbasi:
Otherwise sockets can bind to arbitrary interfaces and tests may fail.


electrum:
Merged, thanks!

PR #517, opened 2017-03-28
Upgrade jetty to 9.4.2

nezihyigitbasi:
verifier run looks good.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
@@ -629,14 +629,14 @@ public class JettyHttpClient
 
     private static List<org.eclipse.jetty.client.api.Request> getRequestForDestination(Destination destination)
     {
-        PoolingHttpDestination<?> poolingHttpDestination = (PoolingHttpDestination<?>) destination;
+        PoolingHttpDestination poolingHttpDestination = (PoolingHttpDestination) destination;
         Queue<HttpExchange> httpExchanges = poolingHttpDestination.getHttpExchanges();
 
         List<org.eclipse.jetty.client.api.Request> requests = httpExchanges.stream()
                 .map(HttpExchange::getRequest)
                 .collect(Collectors.toList());
 
-        poolingHttpDestination.getConnectionPool().getActiveConnections().stream()
+        ((DuplexConnectionPool) poolingHttpDestination.getConnectionPool()).getActiveConnections().stream()
                 .filter(HttpConnectionOverHTTP.class::isInstance)
                 .map(connection -> ((HttpConnectionOverHTTP) connection).getHttpChannel().getHttpExchange())
                 .filter(exchange -> exchange != null)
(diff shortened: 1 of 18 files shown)


http-server/src/main/java/io/airlift/http/server/HashLoginServiceProvider.java
```diff
@@ -42,18 +37,9 @@ public HashLoginServiceProvider(HttpServerConfig config)
...
-            if (!isNullOrEmpty(authConfig)) {
-                HashLoginService service = new HashLoginService(HttpServerModule.REALM_NAME, authConfig);
-                service.loadUsers();
-                return service;
-            }
-            return null;
+        if (!isNullOrEmpty(authConfig)) {
+            return new HashLoginService(HttpServerModule.REALM_NAME, authConfig);
         }
-        catch (IOException e) {
```
electrum: Does this no longer throw `IOException`? Or throw at all?

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -1370,10 +1370,10 @@ public ConnectionPoolDistribution(HttpClient httpClient, Processor processor)
                 Distribution distribution = new Distribution();
                 httpClient.getDestinations().stream()
                         .filter(PoolingHttpDestination.class::isInstance)
-                        .map(destination -> (PoolingHttpDestination<?>) destination)
+                        .map(destination -> (PoolingHttpDestination) destination)
```
electrum: Do this as
```java
.map(PoolingHttpDestination.class::cast)

File: http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java [hunk 3]
```diff
@@ -1370,10 +1370,10 @@ public ConnectionPoolDistribution(HttpClient httpClient, Processor processor)
...
                         .map(PoolingHttpDestination::getConnectionPool)
                         .filter(pool -> pool != null)
-                        .forEach(pool -> processor.process(distribution, pool));
+                        .forEach(pool -> processor.process(distribution, (DuplexConnectionPool) pool));
```
electrum: Add a separate map before this to do the cast

http-server/src/main/java/io/airlift/http/server/HashLoginServiceProvider.java
nezihyigitbasi: `loadUsers()` was throwing previously and it's removed in this release, so nothing is throwing an IOE now. BTW it was already a noop in `9.3.13.M0`.

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
nezihyigitbasi: I wanted to make minimal changes for the upgrade so I didn't change many of these `map` calls (there are many `map` calls that I can change like that). Now you asked for that I will do it shortly.

nezihyigitbasi:
updated.

nezihyigitbasi:
Superseded by #528.

PR #518, opened 2017-03-29
Skip empty Optional fields in JSON serialization

rongrong:


json/src/test/java/io/airlift/json/Person.java
```diff
@@ -22,24 +22,46 @@
...
     private String name;
     private boolean rocks;
+    private Optional<String> lastName;
 
     public static void validatePersonJsonCodec(JsonCodec<Person> jsonCodec)
     {
         Person expected = new Person().setName("dain").setRocks(true);
 
         String json = jsonCodec.toJson(expected);
+        assertEquals(json.indexOf("lastName"), -1);
```
electrum: Use `contains` instead since it more directly conveys what you're doing

electrum: Add a comment before each section to make it easier to know what is being tested
```java
// create object with null lastName

// create object with missing lastName

// create object with present lastName

electrum:
Merged, thanks!

PR #519, opened 2017-04-07
Add support for monitoring jetty thread pools

nezihyigitbasi:
Currently we can only monitor the `http-worker` thread pool. This PR adds support for monitoring the other jetty pools too. Unfortunately, I couldn't use the existing `ThreadPoolExecutorMBean` as it requires a thread pool of type `ThreadPoolExecutor` while Jetty has thread pools of type `QueuedThreadPool`.

screen shot 2017-04-07 at 12 04 06 pm

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
@@ -29,8 +29,11 @@ import io.airlift.configuration.ConfigDefaults;
 import io.airlift.http.client.jetty.JettyHttpClient;
 import io.airlift.http.client.jetty.JettyIoPool;
 import io.airlift.http.client.jetty.JettyIoPoolConfig;
+import io.airlift.http.client.jetty.QueuedThreadPoolMBean;
+import io.airlift.http.client.jetty.QueuedThreadPoolMBeanProvider;
 import io.airlift.http.client.spnego.KerberosConfig;
 import io.airlift.log.Logger;
+import org.eclipse.jetty.util.thread.QueuedThreadPool;
 
 import javax.annotation.PreDestroy;
 import javax.inject.Inject;
@@ -46,6 +49,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import static com.google.inject.multibindings.Multibinder.newSetBinder;
 import static io.airlift.configuration.ConfigBinder.configBinder;
 import static java.util.Objects.requireNonNull;
+import static org.weakref.jmx.ObjectNames.generatedNameOf;
 import static org.weakref.jmx.guice.ExportBinder.newExporter;
 
 @Beta
@@ -56,6 +60,9 @@ public class HttpClientModule
     protected final String name;
     protected final Class<? extends Annotation> annotation;
     protected Binder binder;
+    // we have a single instance of the JettySharedIoPoolModule to make sure
+    // it gets installed only once as we export mbeans from that module
+    private static final JettySharedIoPoolModule JETTY_SHARED_IO_POOL_MODULE = new JettySharedIoPoolModule();
 
     public HttpClientModule(String name, Class<? extends Annotation> annotation)
     {
@@ -71,7 +78,18 @@ public class HttpClientModule
     void withPrivateIoThreadPool()
     {
         configBinder(binder).bindConfig(JettyIoPoolConfig.class, annotation, name);
-        binder.bind(JettyIoPoolManager.class).annotatedWith(annotation).toInstance(new JettyIoPoolManager(name, annotation));
+
+        // Jetty named thread pool
+        binder.bind(JettyIoPoolManager.class)
+                .annotatedWith(annotation)
+                .toInstance(new JettyIoPoolManager(name, annotation));
+        binder.bind(QueuedThreadPoolMBean.class)
+                .annotatedWith(annotation)
+                .toProvider(QueuedThreadPoolMBeanProvider.class)
+                .in(Scopes.SINGLETON);
+        newExporter(binder).export(QueuedThreadPoolMBean.class)
+                .annotatedWith(annotation)
+                .as(generatedNameOf(QueuedThreadPool.class, name));
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


nezihyigitbasi:
Test failure is unrelated, created an issue for that (#520)

nezihyigitbasi:
@electrum rebased, any comments?

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
```diff
@@ -212,6 +220,12 @@ public JettyIoPool get()
             if (pool == null) {
                 JettyIoPoolConfig config = injector.getInstance(keyFromNullable(JettyIoPoolConfig.class, annotation));
                 pool = new JettyIoPool(name, config);
+                if(injector.getExistingBinding(Key.get(MBeanExporter.class)) != null)
```
electrum: We don't need to bother with unexport. Nothing else in Airlift does.

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
```diff
@@ -203,6 +207,10 @@ public void destroy()
             if (pool != null) {
                 pool.close();
                 pool = null;
+                if(injector.getExistingBinding(Key.get(MBeanExporter.class)) != null) {
```
electrum: This class is effectively a provider. Instead of doing the export directly, we should bind the JMX export when we bind IO pool and let the export happen the normal way.

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
```diff
@@ -212,6 +220,12 @@ public JettyIoPool get()
             if (pool == null) {
                 JettyIoPoolConfig config = injector.getInstance(keyFromNullable(JettyIoPoolConfig.class, annotation));
                 pool = new JettyIoPool(name, config);
+                if(injector.getExistingBinding(Key.get(MBeanExporter.class)) != null)
+                {
+                    injector.getInstance(MBeanExporter.class)
+                            .export(generatedNameOf(QueuedThreadPoolMBean.class) + "_" + name,
```
electrum: Use `generatedNameOf(QueuedThreadPool.class, name)`

electrum:
The way to do this is to bind a provider for the MBean:
```java
// named
binder.bind(JettyIoPoolManager.class).annotatedWith(annotation).toInstance(new JettyIoPoolManager(name, annotation));
binder.bind(QueuedThreadPoolMBean.class).annotatedWith(annotation).toProvider(new QueuedThreadPoolMBeanProvider(name, annotation)).in(Scopes.SINGLETON);
newExporter(binder).export(QueuedThreadPoolMBean.class).annotatedWith(annotation).as(generatedNameOf(QueuedThreadPool.class, name));

// shared
binder.bind(JettyIoPoolManager.class).toInstance(new JettyIoPoolManager("shared", annotation)).in(Scopes.SINGLETON);
binder.bind(QueuedThreadPoolMBean.class).toProvider(new QueuedThreadPoolMBeanProvider("shared", annotation)).in(Scopes.SINGLETON);
newExporter(binder).export(QueuedThreadPoolMBean.class).as(generatedNameOf(QueuedThreadPool.class, name));
```
Then we can delete `SharedJettyIoPoolManager` since it isn't used anymore.

nezihyigitbasi:
@electrum rebased & finally got it working.

screen shot 2017-06-19 at 12 15 14 pm

nezihyigitbasi:
@electrum do you have any other comments?

http-client/src/main/java/io/airlift/http/client/jetty/QueuedThreadPoolMBeanProvider.java
```diff
@@ -0,0 +1,27 @@
...
+import io.airlift.http.client.HttpClientModule.JettyIoPoolManager;
+import org.eclipse.jetty.util.thread.QueuedThreadPool;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+import static java.util.Objects.requireNonNull;
+
+public class QueuedThreadPoolMBeanProvider
+    implements Provider<QueuedThreadPoolMBean>
```
electrum: Nit: 8 spaces

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
```diff
@@ -83,12 +99,17 @@ public void configure(Binder binder)
...
 
-        // Shared thread pool
+        // Jetty shared thread pool
         configBinder(binder).bindConfig(JettyIoPoolConfig.class);
         binder.bind(JettyIoPoolManager.class).to(SharedJettyIoPoolManager.class).in(Scopes.SINGLETON);
+        binder.bind(QueuedThreadPoolMBean.class)
+                .toProvider(QueuedThreadPoolMBeanProvider.class)
+                .in(SINGLETON);
+        newExporter(binder).export(QueuedThreadPoolMBean.class)
+                .as(generatedNameOf(QueuedThreadPool.class, "shared_" + annotation.getSimpleName()));
```
electrum: This shared pool is global, so the name here should be just `shared` and not use the annotation. This exports the same MBean under multiple names (one per client), which I don't think is correct.

electrum:
Does the second screenshot reflect the current code? It exports `org.eclipse.jetty.util.thread.QueuedThreadPool` but the screenshot shows the MBeans under `io.airlift.http.client`.

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
```diff
@@ -56,6 +60,9 @@
     protected final String name;
     protected final Class<? extends Annotation> annotation;
     protected Binder binder;
+    // we have a single instance of the JettySharedIoPoolModule to make sure
```
electrum: For multiline comments like this, it's easier to read as a sentence (so capitalize and add period). Also add a newline before to separate it from the other fields. And move above, since we normally put constants above fields.

electrum: +1 for documenting why we have this

electrum: I'll update this to mention explicitly that Guice deduplicates bindings

electrum:
Merged, thanks!

PR #524, opened 2017-05-01
Disable logging for secret config properties

amrutagokhale:
When a configuration property is annotated with @Config, add an ability to mark the property as something that should be kept as secret. Do not print properties annotated with "secret=true".

This is useful to prevent logging of properties such as passwords or secret keys.

Fixes https://github.com/prestodb/presto/issues/7635.

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationInspector.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationInspector.java
@@ -105,7 +105,7 @@ public class ConfigurationInspector
                     description = "";
                 }
 
-                builder.add(new ConfigAttribute(attribute.getName(), propertyName, defaultValue, currentValue, description));
+                builder.add(new ConfigAttribute(attribute.getName(), propertyName, defaultValue, currentValue, description, attribute.isRedact()));
             }
             attributes = builder.build();
         }
@@ -172,7 +172,7 @@ public class ConfigurationInspector
 
         // todo this class needs to be updated to include the concept of deprecated property names
 
-        private ConfigAttribute(String attributeName, String propertyName, String defaultValue, String currentValue, String description)
+        private ConfigAttribute(String attributeName, String propertyName, String defaultValue, String currentValue, String description, boolean redact)
         {
             requireNonNull(attributeName, "attributeName");
             requireNonNull(propertyName, "propertyName");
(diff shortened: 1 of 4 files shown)


johngmyers:
It would be better to log it with a redacted value, such as the proofpoint/platform version of configuration does with its `@ConfigSecuritySensitive` annotation.

Also, this fails to redact the value of secret properties in generated error and/or warning messages, so secrets can still appear in logs.

configuration/src/main/java/io/airlift/configuration/ConfigurationInspector.java
```diff
@@ -169,22 +175,25 @@ public int compareTo(ConfigRecord<?> that)
...
 
-        private ConfigAttribute(String attributeName, String propertyName, String defaultValue, String currentValue, String description)
+        private ConfigAttribute(String attributeName, String propertyName, String defaultValue, String currentValue, String description, boolean secret)
         {
             requireNonNull(attributeName, "attributeName");
             requireNonNull(propertyName, "propertyName");
             requireNonNull(defaultValue, "defaultValue");
             requireNonNull(currentValue, "currentValue");
             requireNonNull(description, "description");
+            requireNonNull(secret, "secret");
```
johngmyers: An unboxed boolean can never be null.

configuration/src/main/java/io/airlift/configuration/ConfigurationInspector.java
```diff
@@ -105,7 +105,13 @@ private ConfigRecord(ConfigurationFactory configurationFactory, ConfigurationPro
                     description = "";
                 }
 
-                builder.add(new ConfigAttribute(attribute.getName(), propertyName, defaultValue, currentValue, description));
+                boolean secret = false;
+                if (attribute.getInjectionPoint().getSetter().isAnnotationPresent(Config.class)) {
+                    Config config = attribute.getInjectionPoint().getSetter().getAnnotation(Config.class);
+                    secret = config.secret();
```
johngmyers: It would seem cleaner to add this property to the `ConfigurationMetadata`

amrutagokhale:
@johngmyers I've addressed your comments. Thanks for pointing out about the values being printed in the error messages. Also, I am looking at the Travis build failure, but it doesn't look related.

haozhun:
If this is intended for security sensitive data, maybe call it something along the line of security sensitive instead of "secret". Alternatively, maybe call it `redact`?

In the context of your original behavior (do not print the property), calling it secret might also encourage misuse where people use this on properties that they don't want others to be aware of (so that they won't use it).

In addition, it is not obvious what "secret" does. Does it hide or redact?

electrum:
Can we cherry pick the changes from Proofpoint Platform?

johngmyers:
Proofpoint Platform is Apache 2.0 licensed, so be my guest. You'll get unit tests and it will cover more of the error messages (such as the could not coerce value ones).

amrutagokhale:
Superseded by https://github.com/airlift/airlift/pull/525.

PR #525, opened 2017-05-05
Redact values of security sensitive properties

amrutagokhale:
Continuing from https://github.com/airlift/airlift/pull/524.

I have cherry-picked two commits from Proofpoint Platform, which added a separate `@ConfigSecuritySensitive` annotation. I added a commit to change it to a separate field in the existing `@Config` annotation since I think that's a better option. Also, in the long run, adding separate annotations for every new attribute of the config parameters will add more complexity.


amrutagokhale:
@johngmyers @electrum do you have any comments on this?

configuration/src/test/java/io/airlift/configuration/ConfigurationMetadataTest.java
```diff
@@ -1765,28 +1749,4 @@ public void setValue(String value)
             this.value = value;
         }
     }
-
-    public static class LegacyConfigSensitiveClass
```
cawallin: Add securitySensitive to LegacyConfig too

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -268,8 +268,8 @@ private void logConfiguration(ConfigurationFactory configurationFactory, Map<Str
         // Warn about unused properties
         if (!unusedProperties.isEmpty()) {
             log.warn("UNUSED PROPERTIES");
-            for (Entry<String, String> unusedProperty : unusedProperties.entrySet()) {
-                log.warn("%s=%s", unusedProperty.getKey(), unusedProperty.getValue());
+            for (String unusedProperty : unusedProperties.keySet()) {
+                log.warn("%s", unusedProperty);
```
johngmyers: Missed changing this line

```java
                    binder.addError("Configuration property '%s=%s' was not used", unusedProperty.getKey(), unusedProperty.getValue());
```

johngmyers:
I think you should consider cherry-picking 61e42bd89e91c9b1d23a45653e2ea765215b3553.

johngmyers:
I used a separate annotation for consistency with `@ConfigDescription`.

johngmyers:
You should also cherry-pick 2ce7182b523ed741828bc3e9c92371f352fac74d

configuration/src/test/java/io/airlift/configuration/ConfigurationMetadataTest.java
amrutagokhale: done

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
amrutagokhale: done

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -494,7 +494,11 @@ public void addConfigurationBindingListener(ConfigurationBindingListener listene
                     operativeName = fullName;
                 }
                 else if (!value.equals(operativeValue)) {
-                    problems.addError("Value for property '%s' (=%s) conflicts with property '%s' (=%s)", fullName, value, operativeName, operativeValue);
+                    if (attribute.isSecuritySensitive()) {
+                        problems.addError("Value for property '%s' (=[REDACTED]) conflicts with property '%s' (=[REDACTED])", fullName, operativeName);
+                    } else {
+                        problems.addError("Value for property '%s' (=%s) conflicts with property '%s' (=%s)", fullName, value, operativeName, operativeValue);
+                    }
```
johngmyers: I'm not a fan of the duplication here, which is why the original code used the `printableValue` approach.

amrutagokhale:
@johngmyers I see your point. I have cherry-picked https://github.com/airlift/airlift/commit/61e42bd89e91c9b1d23a45653e2ea765215b3553 as well. Do you mind taking another look please?

Also, I agree that adding a separate annotation is consistent with `@ConfigDescription`. But in the long run, it will become harder to maintain so many different annotations, if we want to add more. Instead, it would be easier to add fields to the existing annotation.

amrutagokhale:
@johngmyers do you have any more comments?

johngmyers:
I don't have any more comments. I disagree with making it a field and doubt there will ever be as many new annotations to make that a problem. But my position has been adequately stated and I'm not a module owner.

akshatnair:
@amrutagokhale check the build failure

amrutagokhale:
@akshatnair the failures look unrelated to me. Can someone confirm?

configuration/src/main/java/io/airlift/configuration/ConfigurationInspector.java
```diff
@@ -182,8 +182,18 @@ private ConfigAttribute(String attributeName, String propertyName, String defaul
...
-            this.defaultValue = defaultValue;
-            this.currentValue = currentValue;
+            if (securitySensitive && defaultValue != null) {
+                this.defaultValue = "[REDACTED]";
+            }
+            else {
+                this.defaultValue = defaultValue;
+            }
+            if (securitySensitive && currentValue != null) {
+                this.currentValue = "[REDACTED]";
```
akshatnair: Have you added a test for this being set?

akshatnair:
@amrutagokhale not sure who can confirm, check if master is building on Travis and if so check after rebasing

configuration/src/main/java/io/airlift/configuration/ConfigurationInspector.java
amrutagokhale: I have tested this manually. Not sure if it's possible to add an automated test for this in the current framework.

electrum:
@martint and I discussed and decided it's best to leave `@ConfigSecuritySensitive` as a separate annotation. It's more consistent with `@ConfigDescription` and makes the usages look better, since all the `@Config` usages look the same.

So we'll merge this without the last commit. Thanks for keeping it separate and for the rest of the work on this.

electrum:
Merged, thanks!

configuration/src/main/java/io/airlift/configuration/ConfigurationInspector.java
johngmyers: For what it's worth, Platform has a `Logging.addLogTester()` mechanism for unit tests that verify logging. Added in dfef5eea7c1d4a7b9e58e1136632fcf3a5f1f7c7

johngmyers: Or were you referring to Platform's `TestConfigurationInspector.testSecuritySensitiveConfig()`?

PR #526, opened 2017-05-09
Add isClosed method to HttpClient

haozhun:


http-client/src/main/java/io/airlift/http/client/testing/TestingHttpClient.java
```diff
@@ -139,6 +139,11 @@ public void close()
         closed.set(true);
     }
 
+    @Override
+    public boolean isClosed() {
```
electrum: Formatting is off

PR #527, opened 2017-05-15
Add jaxrs-testing module

haozhun:
It adds an implementation of TestingHttpClient.Processor that delegates to
JAX-RS resources.

`mvn dependency:tree`: https://gist.github.com/haozhun/107a0c98dda3dee64b037fce97b05e62

It looks like it's possible to put this in http-client (instead of a separate module). Should I do that instead?

### DIFF
--- a/jaxrs-testing/pom.xml
+++ b/jaxrs-testing/pom.xml
@@ -1,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>jaxrs-testing</artifactId>
+    <packaging>jar</packaging>
+    <name>jaxrs-testing</name>
+
+    <parent>
+        <groupId>io.airlift</groupId>
+        <artifactId>airlift</artifactId>
+        <version>0.147-SNAPSHOT</version>
+    </parent>
+
+    <properties>
+        <air.main.basedir>${project.parent.basedir}</air.main.basedir>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.airlift</groupId>
+            <artifactId>http-client</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.airlift</groupId>
+            <artifactId>log</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.glassfish.jersey.core</groupId>
+            <artifactId>jersey-client</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.glassfish.jersey.test-framework</groupId>
+            <artifactId>jersey-test-framework-core</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.glassfish.jersey.test-framework.providers</groupId>
+            <artifactId>jersey-test-framework-provider-inmemory</artifactId>
+        </dependency>
+
+        <!-- for testing -->
+        <dependency>
+            <groupId>io.airlift</groupId>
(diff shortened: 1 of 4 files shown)


jaxrs-testing/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.testng</groupId>
+            <artifactId>testng</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
+
```
electrum: Remove

jaxrs-testing/src/main/java/io/airlift/jaxrs/testing/JaxrsTestingHttpProcessor.java
```diff
@@ -0,0 +1,135 @@
...
+        try {
+            result = invocation.invoke(javax.ws.rs.core.Response.class);
+        }
+        catch (ProcessingException exception) {
+            // Peel out the exception we threw in jax-rs resource implementation
+            // to facilitate testing exceptional paths
+            if (trace) {
+                log.warn(exception.getCause(), "%-8s %s -> Exception", request.getMethod(), request.getUri());
+            }
+            throw (Exception) exception.getCause();
```
electrum: This can cause `ClassCastException`, since the cause can be anything. It needs to be wrapped if it doesn't extend `Exception`.

jaxrs-testing/src/main/java/io/airlift/jaxrs/testing/JaxrsTestingHttpProcessor.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+        catch (Throwable throwable) {
+            if (trace) {
+                log.warn(throwable, "%-8s %s -> Fail", request.getMethod(), request.getUri());
+            }
+            throw throwable;
+        }
+
+        // process response from jax-rs resource
+        ImmutableListMultimap.Builder<String, String> responseHeaders = ImmutableListMultimap.builder();
```
electrum: This can probably be done with a stream

jaxrs-testing/src/main/java/io/airlift/jaxrs/testing/JaxrsTestingHttpProcessor.java
```diff
@@ -0,0 +1,135 @@
...
+            }
+        }
+
+        if (trace) {
+            log.warn("%-8s %s -> OK", request.getMethod(), request.getUri());
+        }
+        return new TestingResponse(HttpStatus.OK, responseHeaders.build(), result.readEntity(byte[].class));
+    }
+}
+
```
electrum: Remove

jaxrs-testing/src/test/java/io/airlift/jaxrs/testing/TestJaxrsTestingHttpProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+import java.net.URI;
+
+import static io.airlift.http.client.Request.Builder.prepareGet;
+import static io.airlift.http.client.StringResponseHandler.createStringResponseHandler;
+import static java.lang.String.format;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.fail;
+
+@Test
+public class TestJaxrsTestingHttpProcessor {
```
electrum: Formatting

jaxrs-testing/src/test/java/io/airlift/jaxrs/testing/TestJaxrsTestingHttpProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+        @Path("fail/{message}")
+        @GET
+        @Produces(MediaType.TEXT_PLAIN)
+        public String fail(@PathParam("message") String errorMessage)
+        {
+            throw new TestingException(errorMessage);
+        }
+    }
+
+    private static class TestingException extends RuntimeException {
```
electrum: Formatting

jaxrs-testing/src/main/java/io/airlift/jaxrs/testing/JaxrsTestingHttpProcessor.java
haozhun: While it's possible, it will be quite complex:

```
        result.getStringHeaders().entrySet().stream()
                .flatMap(header -> header.getValue().stream()
                        .map(headerValue -> new Map.Entry(header.getKey(), headerValue)))
                .collect(toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
```
(In addition, `new Map.Entry` doesn't actually work. I would need to come up with something else.)

electrum:
There is a collector for immutable multimap.

PR #534, opened 2017-06-10
Provide utility methods for Futures: getDone, checkSucceess

findepi:
* `getDone` is an unchecked variant of Guava's `Futures#getDone`
* `checkSucceess` is equivalent to `getDone` but does not return value
  and conveys intent better, if the caller is interested in just that

Moved to airlift from https://github.com/prestodb/presto/pull/8166/commits/63ed2fae6020430cb68aa4e78f951b3a562691ab (prestodb/presto#8166).

### DIFF
--- a/concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
+++ b/concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
@@ -241,6 +241,26 @@ public final class MoreFutures
     }
 
     /**
+     * Returns the result of the input {@link Future}, which must have already completed.
+     * <p>
+     * Similar to Futures{@link Futures#getDone(Future)}, but does not throw checked exceptions.
+     */
+    public static <T> T getDone(Future<T> future)
+    {
+        requireNonNull(future, "future is null");
+        checkArgument(future.isDone(), "future not done yet");
+        return getFutureValue(future);
+    }
+
+    /**
+     * Checks that the completed future completed successfully.
+     */
+    public static void checkSucceess(Future<?> future)
+    {
+        getDone(future);
+    }
+
+    /**
      * Creates a future that completes when the first future completes either normally
      * or exceptionally. Cancellation of the future propagates to the supplied futures.
      */
--- a/concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
+++ b/concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
@@ -255,9 +255,17 @@ public final class MoreFutures
     /**
      * Checks that the completed future completed successfully.
      */
-    public static void checkSucceess(Future<?> future)
+    public static void checkSuccess(Future<?> future, String errorMessage)
     {
-        getDone(future);
+        requireNonNull(future, "future is null");
+        requireNonNull(errorMessage, "errorMessage is null");
+        checkArgument(future.isDone(), "future not done yet");
+        try {
+            getFutureValue(future);
+        }
+        catch (RuntimeException e) {
+            throw new IllegalArgumentException(errorMessage, e);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


findepi:
cc @losipiuk @martint

concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
```diff
@@ -241,6 +241,26 @@ public static Throwable unwrapCompletionException(Throwable throwable)
...
+    {
+        requireNonNull(future, "future is null");
+        checkArgument(future.isDone(), "future not done yet");
+        return getFutureValue(future);
+    }
+
+    /**
+     * Checks that the completed future completed successfully.
+     */
+    public static void checkSucceess(Future<?> future)
```
losipiuk: shouldn't this method take a `message` argument like `checkState`, `requireNotNull`?

findepi: I'm not sure. In case of failure, future's exception will be thrown, so it serves for the message too.

findepi:
@dain , would you like to review this?

concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
dain: I agree.  I'd change the signature to `checkSuccess(Future<?> future, String message)` and implement it like this:
```java
    public static void checkSuccess(Future<?> future, String errorMessageTemplate, Object... errorMessageArgs)
    {
        requireNonNull(future, "future is null");
        if (future.isDone()) {
            try {
                getFutureValue(future);
                return;
            }
            catch (RuntimeException ignored) {

            }
        }
        throw new IllegalArgumentException(String.format(errorMessageTemplate, errorMessageArgs));
    }
```
or maybe not error message formatting...  either way, but, I would only throw an illegal argument exception

findepi: Okay.. Still I'd like to include future's failure in the exception raised (perhaps as cause), otherwise it has high chances to be lost.

findepi:
@dain, i did it a bit differently in order to preserve future's exception
if there is one.

concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
```diff
@@ -241,6 +241,34 @@ public static Throwable unwrapCompletionException(Throwable throwable)
...
+    public static void checkSuccess(Future<?> future, String errorMessage)
+    {
+        requireNonNull(future, "future is null");
+        requireNonNull(errorMessage, "errorMessage is null");
+        checkArgument(future.isDone(), "future not done yet");
+        try {
+            getFutureValue(future);
+        }
+        catch (RuntimeException e) {
+            throw new IllegalArgumentException(errorMessage, e);
```
findepi: Including future's exception is very important. Otherwise failure cause is going to be lost in most of the cases.

findepi:
@dain , back to you. (Can't assign because it's airlift)

concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
```diff
@@ -241,6 +241,34 @@ public static Throwable unwrapCompletionException(Throwable throwable)
     }
 
     /**
+     * Returns the result of the input {@link Future}, which must have already completed.
+     * <p>
+     * Similar to Futures{@link Futures#getDone(Future)}, but does not throw checked exceptions.
+     */
+    public static <T> T getDone(Future<T> future)
```
dain: These need basic coverage tests.  Just a checks with a success, not done, and has exception futures.

findepi:
@dain , tests added

dain:
Merged, thanks!

PR #537, opened 2017-06-19
add an option to specify the etc dir path

zaynoon:
I added a new option`--etc-dir` that allows specifying a location for the `etc` directory other than `INSTALL_PATH/etc`.


electrum:
Merged, thanks!

Issue #520, opened 2017-04-10
TestHttpServerProvider::testStop is flaky


nezihyigitbasi:
If you look at the [code](https://github.com/airlift/airlift/blob/master/http-server/src/test/java/io/airlift/http/server/TestHttpServerProvider.java#L218-L239), it's possible that `server.stop()` executes after the async request completes (in ~50+s) and so the `future` is really done at that point and test fails.  Here is one failed run:

https://s3.amazonaws.com/archive.travis-ci.org/jobs/219828132/log.txt

PR #538, opened 2017-06-22
Ensure Jetty client request doesn't live past synchronous return

johngmyers:
Otherwise Jetty might later access objects in the BodyGenerator that are no longer valid.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
@@ -359,11 +359,13 @@ public class JettyHttpClient
         }
         catch (InterruptedException e) {
             stats.recordRequestFailed();
+            jettyRequest.abort(e);
             Thread.currentThread().interrupt();
             return responseHandler.handleException(request, e);
         }
         catch (TimeoutException e) {
             stats.recordRequestFailed();
+            jettyRequest.abort(e);
             return responseHandler.handleException(request, e);
         }
         catch (ExecutionException e) {


nezihyigitbasi:
@johngmyers To better understand what the real problem is I have some questions:
- Is this a problem that you started seeing after the Jetty upgrade (to 9.4.6.v20170531) in commit c8a2f0bc6f228504f4b571870cfccd3496bbe57c  ?
- Can you provide a stack trace or more info about what this is causing? You say `Jetty might later access objects in the BodyGenerator that are no longer valid.`, but `BodyGenerator`s don't access request objects they just provide content to the request bodies. What I see is the `responseHandler.handleException()` calls may access the `requests` on exception before they might be sent over, but still not sure what it may cause.

johngmyers:
I don't use airlift. This is a fix from proofpoint/platform.

What happened is the Jetty client request queue got backed up due to an overloaded destination. The server thread made a synchronous client request with a body reading from the server request's `HttpInput`. The call to `InputStreamResponseListener.get(long, TimeUnit)` threw `TimeoutException`. The server thread then returned from the servlet and Jetty recycled the request`HttpInput` object. Minutes later, Jetty picked up the client request from the queue and requested body content. The `BodyGenerator`-equivalent read from the from the now-invalid `HttpInput` and blocked forever.

nezihyigitbasi:
Thanks for the info.

- Just curious, when you observed the problem what was the `The BodyGenerator-equivalent`? Is it an instance of `InputStreamContentProvider`? 

- Depending on how different threads get scheduled and timing of various events (depending on the timeout settings and how bad the overload situation is), it's possible that the request timeouts and before the `abort` call is made in the exception handler the content provider/body generator may access the invalid input stream. So there is still a possibility that things can go bad. 

@dain this looks good. Can you restart the travis job? Seems like it's hitting #520.

electrum:
Merged, thanks!

PR #539, opened 2017-07-06
Prevent "Address already in use" for HTTP server

electrum:
When using automatic port selection, the port is bound once and kept bound
rather than unbinding and hoping the OS does not reuse the port number
before it can be explicitly bound again.


http-server/src/test/java/io/airlift/http/server/TestHttpServerProvider.java
```diff
@@ -104,13 +107,125 @@ public void teardown()
...
+
+    @Test
     public void testHttp()
             throws Exception
     {
         createServer();
         server.start();
 
         try (JettyHttpClient httpClient = new JettyHttpClient(new HttpClientConfig().setHttp2Enabled(false))) {
+            System.out.println(httpServerInfo.getHttpUri());
```
electrum: Remove before merging

electrum:
We can remove the reflection in Jetty 9.4.7

PR #541, opened 2017-07-11
Add sanity checks in HttpUriBuilder

amrutagokhale:
We saw failure of a Presto product test on an EMR cluster with the following stack trace:
```18:22:17 Jul 11, 2017 6:19:28 PM com.teradata.tempto.internal.listeners.ProgressLoggingListener onTestFailure
18:22:17 SEVERE: Failure cause:
18:22:17 java.lang.NullPointerException
18:22:17 	at com.google.common.base.CharMatcher.matchesAllOf(CharMatcher.java:635)
18:22:17 	at io.airlift.http.client.HttpUriBuilder.percentDecode(HttpUriBuilder.java:283)
18:22:17 	at io.airlift.http.client.HttpUriBuilder.<init>(HttpUriBuilder.java:58)
18:22:17 	at io.airlift.http.client.HttpUriBuilder.uriBuilderFrom(HttpUriBuilder.java:71)
18:22:17 	at com.facebook.presto.tests.querystats.HttpQueryStatsClient.getQueryStats(HttpQueryStatsClient.java:54)
18:22:17 	at com.facebook.presto.tests.hive.HivePartitioningTest.getProcessedLinesCount(HivePartitioningTest.java:53)
18:22:17 	at com.facebook.presto.tests.hive.TestTablePartitioningInsertInto.testQuerySplitsNumber(TestTablePartitioningInsertInto.java:83)
18:22:17 	at com.facebook.presto.tests.hive.TestTablePartitioningInsertInto.selectFromPartitionedNation(TestTablePartitioningInsertInto.java:54)
```
The reason I believe is, `null` value is being passed to `percentDecode()` in `HttpUriBuilder()`, but `percentDecode()` makes an assumption that the input is non-null.

I believe the checks added in this commit should fix this.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
@@ -17,6 +17,7 @@ import java.nio.charset.CharacterCodingException;
 import java.nio.charset.CodingErrorAction;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Optional;
 
 import static com.google.common.base.CharMatcher.ascii;
 import static java.lang.Character.forDigit;
@@ -55,7 +56,7 @@ public class HttpUriBuilder
         scheme = previous.getScheme();
         host = previous.getHost();
         port = previous.getPort();
-        path = percentDecode(previous.getRawPath());
+        path = Optional.ofNullable(previous.getRawPath()).map(HttpUriBuilder::percentDecode).orElse(null);
         params.putAll(parseParams(previous.getRawQuery()));
     }
 
@@ -280,6 +281,7 @@ public class HttpUriBuilder
      */
     private static String percentDecode(String encoded)
     {
+        requireNonNull(encoded, "encoded is null");
         Preconditions.checkArgument(ascii().matchesAllOf(encoded), "string must be ASCII");
 
         ByteArrayOutputStream out = new ByteArrayOutputStream(encoded.length());
--- a/http-client/src/test/java/io/airlift/http/client/TestHttpUriBuilder.java
+++ b/http-client/src/test/java/io/airlift/http/client/TestHttpUriBuilder.java
@@ -4,6 +4,7 @@ import com.google.common.net.HostAndPort;
 import org.testng.annotations.Test;
 
 import java.net.URI;
+import java.net.URISyntaxException;
 
 import static io.airlift.http.client.HttpUriBuilder.uriBuilder;
 import static io.airlift.http.client.HttpUriBuilder.uriBuilderFrom;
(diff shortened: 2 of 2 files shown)


http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
```diff
@@ -280,6 +282,7 @@ private String encode(String input, byte... allowed)
      */
     private static String percentDecode(String encoded)
     {
+        Preconditions.checkArgument(nonNull(encoded), "string must not be null");
```
electrum: This should be requireNonNull since NPE is the appropriate exception

http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
```diff
@@ -55,7 +57,7 @@ private HttpUriBuilder(URI previous)
         scheme = previous.getScheme();
         host = previous.getHost();
         port = previous.getPort();
-        path = percentDecode(previous.getRawPath());
+        path = isNull(previous.getRawPath()) ? null : percentDecode(previous.getRawPath());
```
electrum: These methods are intended for use as lambdas. Using them directly makes the code harder to read since it's not the standard way to write null checks (which have been in Java forever).

electrum: An alternative way to write this is

Optional.ofNullable(...).map(...).orElse(null)

This avoids repeating the input expression, though I'm not sure it's an improvement over traditional way.

electrum (commented):
Looks good, but can you add a test?

http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
amrutagokhale: I rewrote it as you suggested; don't know of a better option.

amrutagokhale:
@electrum Thanks for reviewing! I added a test and addressed your comments.

http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
electrum: I just looked at this code for a quick review before merging and that line confused me. So I'm going to revert to your original version which is much more obvious and straight forward :)

electrum:
I'll change the commit message to "Fix NPE for null paths in HttpUriBuilder" since this is actually a bug fix and not just a sanity check.

http-client/src/test/java/io/airlift/http/client/TestHttpUriBuilder.java
```diff
@@ -118,6 +119,15 @@ public void testAppendPathElidesSlashes()
     }
 
     @Test
+    public void testNullPath()
+            throws URISyntaxException
+    {
+        URI uri = new URI("test-server:8080");
```
electrum: We need to run this through `HttpUriBuilder`

electrum: I changed the test to
```java
URI uri = new URI("test-server:8080");
uri = uriBuilderFrom(uri).build();
```
But it fails in the `build()` method.

electrum: Making the path `""` actually causes another failure:
```
Caused by: java.net.URISyntaxException: Expected authority at index 14: test-server://
```
So this is much more involved :(

electrum (changes requested):
I think we need additional sanity checking, like checking if "scheme" is empty

PR #542, opened 2017-07-14
Fluent locks

findepi:


findepi:
Travis failure seems unrelated

```
testStop(io.airlift.http.server.TestHttpServerProvider)  Time elapsed: 1.071 sec  <<< FAILURE!
java.util.concurrent.TimeoutException
	at io.airlift.http.server.TestHttpServerProvider.testStop(TestHttpServerProvider.java:232)
```

findepi:
cc @electrum

concurrent/src/main/java/io/airlift/concurrent/Locks.java
```diff
@@ -0,0 +1,120 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.concurrent;
+
+import java.util.concurrent.locks.Lock;
+
+import static java.util.Objects.requireNonNull;
+
+public class Locks
```
electrum: Add `@Beta` since this is a new API which might change

concurrent/src/test/java/io/airlift/concurrent/TestLocks.java
```diff
@@ -0,0 +1,168 @@
...
+                        maxUnderLock.accumulateAndGet(underLock.incrementAndGet(), Integer::max);
+                        Thread.sleep(30);
+                        underLock.decrementAndGet();
+                    });
+
+                    return null;
+                }))
+                .collect(toImmutableList());
+
+        futures.forEach(MoreFutures::getFutureValue);
```
electrum: This test seems to be missing assertion

concurrent/src/test/java/io/airlift/concurrent/TestLocks.java
```diff
@@ -0,0 +1,168 @@
...
+            // main thread is under lock now
+            lockingInterruptibly(lock, () -> {
+                throw new AssertionError("test background thread got the lock");
+            });
+            throw new AssertionError("lockingInterruptibly returned, InterruptedException was expected");
+        });
+
+        locking(lock, () -> {
+            Thread testBackgroundThread = testBackgroundThreadExchanger.exchange(null, 5, SECONDS);
+            Thread.sleep(40); // give the thread chance to lock
```
electrum: However, I'm not sure there's a better way, aside from hacky ways like trying to observe the stack trace of the thread.

electrum:
Tests look good

concurrent/src/test/java/io/airlift/concurrent/TestLocks.java
findepi: I agree. However, in this case I expect it to work even if we interrupt to early. The reason is that `Lock#lockInterruptibly` throws `InterruptedException` even if the lock is available.
The sleep is here to make it "almost certain" thread is already waiting for a lock so that we test the (expectedly) most common case. However, test passes even with sleep removed.

concurrent/src/test/java/io/airlift/concurrent/TestLocks.java
```diff
@@ -162,7 +162,9 @@ public void testLockingInterruptiblyCanBeInterrupted()
             fail("ExecutionException was expected");
         }
         catch (ExecutionException expected) {
-            assertTrue(expected.getCause() instanceof InterruptedException);
+            if (!(expected.getCause() instanceof InterruptedException)) {
+                fail("Expected failure caused by interruption", expected);
+            }
```
findepi: I wasn't preserving failure exception in case where it was other than expected. That would be rogue.

concurrent/src/test/java/io/airlift/concurrent/TestLocks.java
```diff
@@ -0,0 +1,170 @@
...
+            testBackgroundThread.interrupt();
+            // we can release lock now, as Lock.lockInterruptibly throws interrupted even if lock is available
+        });
+
+        try {
+            future.get(5, SECONDS);
+            fail("ExecutionException was expected");
+        }
+        catch (ExecutionException expected) {
+            if (!(expected.getCause() instanceof InterruptedException)) {
```
electrum: We have `assertInstanceOf` in `Assertions`

findepi: Hm.. I'd like to have `expected.getCause()` or `expected` attached when check fails.

findepi:
@martint, did you have time to review the API?

martint:
Briefly. I'm trying to understand what this tries to solve that can't be achieved in other ways.

dain:
Looking at the examples, I'm concerned that this might break Intellij's static lock analysis. Does this API work with IntelliJ static analyzer?

findepi:
@martint , this is aimed at increasing readability of code using `Lock` objects. `lock try finally unlock` is quite verbose and unstructured, making it less obvious what is the actual logic besides locking.

@dain , it doesn't.

electrum:
@findepi please clarify if "it doesn't" means it doesn't work or doesn't break?

The lock analysis can find subtle bugs, so I'd argue that having it work is worth having slightly less readable code. They talked about doing bytecode anyalsis to automatically infer contracts for things like Preconditions, so in theory, this is something they could support in the future.

martint:
@findepi, have you considered using a wrapper that implements AutoCloseable so that it can be used with try-with-resources? Something like:

```
try(Locker locker = locker(lock)) {
 ...
}
```

findepi:
@electrum, "it doesn't" -- it doesn't work correctly with IntelliJ static code analysis. IntelliJ doesn't "see" locking code in a library function.

@martint , I haven't. That's cool idea! And doesn't require to use lambdas (`Checked...` interfaces). If IDE support isn't a showstopper, I can explore this option, but it's up to you to decide.

findepi:
I think there is little traction to merge this, closing.

PR #544, opened 2017-07-28
Support custom Java verison string

hquan:
Besides legacy Java version string and the new one introduced in Java 9, we still need to support Java version string like `1.8.0_131-foo-bar-baz`. This provides more flexibility for developers.

### DIFF
--- a/jmx/src/main/java/io/airlift/jmx/JavaVersion.java
+++ b/jmx/src/main/java/io/airlift/jmx/JavaVersion.java
@@ -35,7 +35,7 @@ class JavaVersion
     private static final Pattern PATTERN = Pattern.compile(VERSION_NUMBER + PRE + BUILD + OPT);
 
     // For Java 8 and below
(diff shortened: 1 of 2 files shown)


hquan:
Reopen to trigger Travis CI build.

jmx/src/main/java/io/airlift/jmx/JavaVersion.java
```diff
@@ -35,7 +35,10 @@
     private static final Pattern PATTERN = Pattern.compile(VERSION_NUMBER + PRE + BUILD + OPT);
 
     // For Java 8 and below
-    private static final Pattern LEGACY_PATTERN = Pattern.compile("1\\.(?<MAJOR>[0-9]+)(\\.(?<MINOR>(0|[1-9][0-9]*)))?(_(?<UPDATE>[1-9][0-9]*))?(?:-ea)?");
+    private static final String LEGACY_REGEX = "1\\.(?<MAJOR>[0-9]+)(\\.(?<MINOR>(0|[1-9][0-9]*)))?(_(?<UPDATE>[1-9][0-9]*))?(?:-ea)?";
```
martint: This seems overkill. We don't need two matchers (LEGACY_PATTERN vs LEGACY_CUSTOM_PATTERN). All we need to do is extend the pattern for -ea to match any combination of letters and numbers.

martint:
Merged with a minor tweak.

PR #545, opened 2017-07-30
Add RebindSafeMBeanServer to the jmx module

nezihyigitbasi:
RebindSafeMBeanServer is an MBeanServer wrapper that ignores calls to
registerMBean when there is already an MBean registered with that
name.

(Presto has five copies of this class.)

### DIFF
--- a/jmx/pom.xml
+++ b/jmx/pom.xml
@@ -51,6 +51,11 @@
             <artifactId>javax.inject</artifactId>
         </dependency>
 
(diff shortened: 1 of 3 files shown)


martint:
I don't think this is a good idea. The class only exists in Presto as a hack. The real solution is to re-namespace mbeans to avoid conflicts. It's something we've talked about doing many times but haven't had the chance to work on it.

PR #549, opened 2017-08-11
Enable detailed dumps for QueuedThreadPool

nezihyigitbasi:
This is useful when debugging Jetty with JMX dump() operations.

There won't be any overhead.


electrum:
Can you add this to changes?

nezihyigitbasi:
@electrum updated.

electrum:
Merged, thanks!

PR #550, opened 2017-08-15
Support for PostgreSQL in dbpool.

spasam:
Similar to MySql, added support for PostgreSQL.

### DIFF
--- a/dbpool/src/main/java/io/airlift/dbpool/PostgreSqlDataSource.java
+++ b/dbpool/src/main/java/io/airlift/dbpool/PostgreSqlDataSource.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2010 Proofpoint, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.dbpool;
+
+import org.postgresql.ds.PGConnectionPoolDataSource;
+import io.airlift.discovery.client.ServiceDescriptor;
+import io.airlift.discovery.client.ServiceSelector;
+
+import javax.sql.PooledConnection;
+
+import java.sql.SQLException;
+import java.util.UUID;
+
+public class PostgreSqlDataSource extends ManagedDataSource
+{
+    private final ServiceSelector serviceSelector;
+    private final int defaultFetchSize;
+    private UUID currentServer;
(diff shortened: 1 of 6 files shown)


dbpool/src/main/java/io/airlift/dbpool/PostgreSqlDataSource.java
```diff
@@ -0,0 +1,102 @@
...
+            // skip bogus announcements
+            String jdbcUrl = serviceDescriptor.getProperties().get("jdbc");
+            if (jdbcUrl == null) {
+                continue;
+            }
+
+            try {
+                PGConnectionPoolDataSource dataSource = new PGConnectionPoolDataSource();
+                dataSource.setUrl(jdbcUrl);
+                int timeout = getMaxConnectionWaitMillis() * 1000;
```
electrum: This can use `TimeUnit.MILLISECONDS.toSeconds(...)`

dbpool/src/main/java/io/airlift/dbpool/PostgreSqlDataSourceModule.java
```diff
@@ -0,0 +1,116 @@
...
+        implements Module
+{
+    private final Class<? extends Annotation> annotation;
+    private final List<Class<? extends Annotation>> aliases;
+    private final String type;
+
+    @SafeVarargs
+    public PostgreSqlDataSourceModule(String type, Class<? extends Annotation> annotation, Class<? extends Annotation>... aliases)
+    {
+        if (annotation == null) {
```
electrum: We can update this to use `requireNonNull`

dbpool/src/main/java/io/airlift/dbpool/PostgreSqlDataSourceModule.java
```diff
@@ -0,0 +1,116 @@
...
+            throw new NullPointerException("annotation is null");
+        }
+        if (type == null) {
+            throw new NullPointerException("type is null");
+        }
+        this.annotation = annotation;
+        this.type = type;
+        if (aliases != null) {
```
electrum: This can be simplified to be
```java
this.aliases = ImmutableList.copyOf(firstNonNull(aliases, ImmutableList.of()));

File: dbpool/src/main/java/io/airlift/dbpool/PostgreSqlDataSource.java (hunk 1 above)
Comment (electrum) [inline dbpool/src/main/java/io/airlift/dbpool/PostgreSqlDataSource.java]: Note that the original calculation is incorrect, since we should be dividing, not multiplying. This is why using `TimeUnit` is preferable :)

File: dbpool/src/main/java/io/airlift/dbpool/PostgreSqlDataSourceConfig.java [hunk 4]
```diff
@@ -0,0 +1,46 @@
...
+ */
+package io.airlift.dbpool;
+
+import io.airlift.configuration.Config;
+import io.airlift.configuration.DefunctConfig;
+
+/**
+ * see <a href="https://jdbc.postgresql.org/documentation/head/connect.html">PostgreSQL JDBC Driver Connecting to Database</a>
+ */
+@DefunctConfig({"db.host", "db.port", "db.database", "db.ssl.enabled"})
```
electrum: These aren't needed since this is a new class

electrum:
Merged, thanks!

PR #552, opened 2017-08-18
Add instance sizes to HyperLogLog and SparseHll

highker:
We observed 3X retained size in heap dump than the calculated one. Add
instance sizes to fill the gap.

### DIFF
--- a/stats/src/main/java/io/airlift/stats/cardinality/HyperLogLog.java
+++ b/stats/src/main/java/io/airlift/stats/cardinality/HyperLogLog.java
@@ -16,12 +16,14 @@ package io.airlift.stats.cardinality;
 import com.google.common.annotations.VisibleForTesting;
 import io.airlift.slice.Murmur3Hash128;
 import io.airlift.slice.Slice;
+import org.openjdk.jol.info.ClassLayout;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static io.airlift.stats.cardinality.Utils.indexBitLength;
 
 public class HyperLogLog
 {
+    private static final int INSTANCE_SIZE = ClassLayout.parseClass(HyperLogLog.class).instanceSize();
     private static final int MAX_NUMBER_OF_BUCKETS = 65536;
     private HllInstance instance;
 
@@ -97,7 +99,7 @@ public class HyperLogLog
 
     public int estimatedInMemorySize()
     {
-        return instance.estimatedInMemorySize();
+        return instance.estimatedInMemorySize() + INSTANCE_SIZE;
     }
 
     public int estimatedSerializedSize()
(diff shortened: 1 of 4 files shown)


stats/src/main/java/io/airlift/stats/cardinality/HyperLogLog.java
```diff
@@ -97,7 +99,7 @@ public long cardinality()
 
     public int estimatedInMemorySize()
```
nezihyigitbasi: while you are at it, it would be nice to make these size-related methods to return `long`.

stats/src/main/java/io/airlift/stats/cardinality/SparseHll.java
```diff
@@ -189,7 +191,7 @@ public long cardinality()
     @Override
     public int estimatedInMemorySize()
```
nezihyigitbasi: we should return `long` and get rid of the cast.

stats/src/main/java/io/airlift/stats/cardinality/HyperLogLog.java
highker: fixed; thanks

stats/src/test/java/io/airlift/stats/cardinality/TestSparseHll.java
```diff
@@ -57,6 +59,25 @@ public void testToDense(int prefixBitLength)
         verifyToDense(prefixBitLength, ImmutableList.of(224L, 271L));
     }
 
+    @Test
+    public void testRetainedSize()
+            throws Exception
```
nezihyigitbasi: `throws Exception` is unnecessary I think

highker:
Thanks @nezihyigitbasi for the review; could @martint or @electrum help walking through this? It is really a small patch.

martint:
Why do we need to change it to return long? HLL instances will never be that big.

stats/src/test/java/io/airlift/stats/cardinality/TestSparseHll.java
```diff
@@ -57,6 +59,24 @@ public void testToDense(int prefixBitLength)
...
+        SparseHll sparseHll = new SparseHll(10);
+        // use an empty array to mock the entries in SparseHll
+        int[] entries = new int[1];
+        long retainedSize = sizeOf(entries) + ClassLayout.parseClass(SparseHll.class).instanceSize();
+        for (int value = 0; value < 100; value++) {
+            sparseHll.insertHash(Murmur3Hash128.hash64(value));
+            if (value % 10 == 1) {
+                // we increase the capacity by 10 once full
+                entries = new int[value + 10];
+                retainedSize = sizeOf(entries) + ClassLayout.parseClass(SparseHll.class).instanceSize();
```
martint: Computing ClassLayout.parseClass(SparseHll.class).instanceSize() can be expensive. Move it out of the loop (ideally, to a constant).

highker:
@martint, comments addressed. Thanks!

PR #553, opened 2017-08-19
Update to Airbase 68

electrum:


jmx-http/pom.xml
```diff
@@ -18,6 +18,13 @@
         <dependency>
             <groupId>io.airlift</groupId>
             <artifactId>discovery</artifactId>
+            <exclusions>
+                <!-- only the binder is used, not the client -->
```
dain: maybe we should split discovery into API and client

PR #558, opened 2017-09-27
Support SSL KeyManager password

kwondw:
Airlift only allows to set SSL keystore & keymanager password as the same one.

In order to support Jetty's setKeyManagerPassword of SSLContextFactory, this patch introduce 'http-server.https.keymanager.password' config so that SSL can be set keyManager password differently from keystore password


kwondw:
http-client/TestJettyHttpsClient test failed due to connection time out 

> [ERROR] testConnectTimeout(io.airlift.http.client.jetty.TestJettyHttpsClient)  Time elapsed: 2.034 s  <<< FAILURE!
> java.lang.AssertionError: 
> unexpected exception: [java.util.concurrent.TimeoutException

In my local test, I don't have any test failure nor reproduce this issue. 

I found other request had the same issue before.
https://travis-ci.org/airlift/airlift/jobs/279651634

I wonder is this any way I can retry to run test to see whether it's intermittent issue or not.

electrum:
I restarted it.

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -183,6 +184,18 @@ public HttpServerConfig setKeystorePassword(String keystorePassword)
         return this;
     }
 
+    public String getKeyManagerPassword()
+    {
+        return keyManagerPassword;
+    }
+
+    @Config("http-server.https.keymanager.password")
+    public HttpServerConfig setKeyManagerPassword(String keyManagerPassword)
```
electrum: We should add `@ConfigSecuritySensitive` so the password isn't printed in the log

electrum:
Merged, thanks!

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
kwondw: I didn't know, I saw you added it, thanks for that, electrum!

PR #562, opened 2017-10-07
Add a concurrent timeout scheduler for Jetty Client

dain:
Use multiple scheduler instances to reduce contention Jetty client for timeout executor
when under heavy.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyIoPool.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyIoPool.java
@@ -1,7 +1,9 @@
 package io.airlift.http.client.jetty;
 
+import io.airlift.concurrent.Threads;
 import org.eclipse.jetty.io.ByteBufferPool;
 import org.eclipse.jetty.io.MappedByteBufferPool;
+import org.eclipse.jetty.util.component.AbstractLifeCycle;
 import org.eclipse.jetty.util.component.LifeCycle;
 import org.eclipse.jetty.util.thread.QueuedThreadPool;
 import org.eclipse.jetty.util.thread.ScheduledExecutorScheduler;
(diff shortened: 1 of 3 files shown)


http-client/src/main/java/io/airlift/http/client/jetty/JettyIoPool.java
```diff
@@ -101,4 +118,52 @@ public String toString()
...
+    {
+        private final int threadsPerScheduler;
+        private final ScheduledExecutorService[] schedulers;
+        private final String threadBaseName;
+
+        public ConcurrentScheduler(int schedulerCount, int threadsPerScheduler, String threadBaseName)
+        {
+            this.schedulers = new ScheduledThreadPoolExecutor[schedulerCount];
+            this.threadsPerScheduler = threadsPerScheduler;
+            this.threadBaseName = threadBaseName;
```
raghavsethi: requireNonNull

http-client/src/main/java/io/airlift/http/client/jetty/JettyIoPool.java
```diff
@@ -101,4 +118,52 @@ public String toString()
...
+        }
+
+        @Override
+        protected void doStart()
+                throws Exception
+        {
+            for (int i = 0; i < schedulers.length; i++) {
+                schedulers[i] = Executors.newScheduledThreadPool(threadsPerScheduler, Threads.daemonThreadsNamed(threadBaseName + "-%s-lock" + i));
```
raghavsethi: Why '-lock'?

http-client/src/main/java/io/airlift/http/client/jetty/JettyIoPool.java
```diff
@@ -101,4 +118,52 @@ public String toString()
...
+            for (int i = 0; i < schedulers.length; i++) {
+                schedulers[i].shutdownNow();
+                schedulers[i] = null;
+            }
+        }
+
+        @Override
+        public Task schedule(Runnable task, long delay, TimeUnit unit)
+        {
+            ScheduledExecutorService scheduler = schedulers[ThreadLocalRandom.current().nextInt(schedulers.length)];
```
raghavsethi: Why do it this way instead of round robining? RNG should be good enough to be broadly consistent, but why not just ensure equality?

dain: Random is much easier because there is no coordination required.  Also, it is a uniform random so it should be even.

PR #563, opened 2017-10-08
Improve startup time.

spasam:
30 seconds seems arbitrary and the PostConstruct method will return once the future times out. Looks like refresh continues asynchronously after timeout. This change reduces the wait time. This seems to improve startup time of Presto when running in dev environment.

### DIFF
--- a/discovery/src/main/java/io/airlift/discovery/client/CachingServiceSelector.java
+++ b/discovery/src/main/java/io/airlift/discovery/client/CachingServiceSelector.java
@@ -80,7 +80,7 @@ public class CachingServiceSelector
 
             // if discovery is available, get the initial set of servers before starting
             try {
-                refresh().get(30, TimeUnit.SECONDS);
+                refresh().get(1, TimeUnit.SECONDS);
             }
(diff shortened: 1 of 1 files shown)


spasam:
@electrum @dain Can you provide input? Thanks

electrum:
I just hit this today while developing. I'd like to debug why it doesn't refresh immediately (since it normally does). Do you notice this when running in the IDE?

spasam:
I can reproduce it every time. I am running presto using launcher. Not in IDE. Simple presto config as described in documentation with coordinator acting as worker also.

electrum:
I think this is an ordering problem. In Presto, `CachingServiceSelector` is needed by various JAX-RS resources via `DiscoveryNodeManager`, and the resources are needed by the HTTP server, so it tries to refresh before the HTTP server is started (or even constructed).

This brings up the question if we should have this wait at all, or if it's better to have application use `ServiceSelectorManager.forceRefresh()` like we do in `TestingPrestoServer` (which is why we added that class). For now, this will improve things substantially.

PR #568, opened 2017-10-25
Use async appender to log events

highker:
Logging events while holding a thread from a thread pool can quickly
drain the pool given IO is in general slow. Use async appender to fast
release the thread if applicable. Also, increase the buffer size of the
appender from 8KB to 1MB.

### DIFF
--- a/http-server/src/main/java/io/airlift/http/server/DelimitedRequestLog.java
+++ b/http-server/src/main/java/io/airlift/http/server/DelimitedRequestLog.java
@@ -15,6 +15,7 @@
  */
 package io.airlift.http.server;
 
+import ch.qos.logback.core.AsyncAppenderBase;
 import ch.qos.logback.core.ContextBase;
 import ch.qos.logback.core.rolling.RollingFileAppender;
 import ch.qos.logback.core.rolling.SizeAndTimeBasedFNATP;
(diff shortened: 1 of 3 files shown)


log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -101,23 +103,28 @@ public void close()
...
+
         @Override
-        public void doEncode(String event)
-                throws IOException
+        public byte[] encode(String event)
         {
-            outputStream.write(event.getBytes(UTF_8));
-            // necessary if output stream is buffered
-            outputStream.flush();
+            // Keep immediateFlush as true in RollingFileAppender guarantees a flush of output stream
```
electrum: I'm not sure what this means

http-server/src/main/java/io/airlift/http/server/DelimitedRequestLog.java
```diff
@@ -39,13 +39,14 @@
     private static final Logger log = Logger.get(DelimitedRequestLog.class);
     private static final String TEMP_FILE_EXTENSION = ".tmp";
     private static final String LOG_FILE_EXTENSION = ".log";
+    private static final FileSize BUFFER_SIZE_IN_BYTES = new FileSize(1 << 20);
```
electrum: Use `DataSize` to construct this

highker:
Comments addressed. We need an airbase release to make the build pass.

dain:
Merged, thanks!

PR #570, opened 2017-10-28
Add utilities for pem and certificate signing request

dain:


electrum:
Update CHANGES for these

PR #571, opened 2017-10-31
Remove LifeCycleManager shutdown hook when stop() is called

electrum:
This prevents memory leaks when used explicitly such as in tests.


bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -122,6 +120,15 @@ public void stop()
             return;
         }
 
+        Thread thread = shutdownHook.get();
```
dain: `getAndSet(null)`

PR #572, opened 2017-10-31
Add HTTP client selector thread count config

nezihyigitbasi:
Currently we hardcode `2` selectors for the HTTP clients and this may not be sufficient under high load (two selector threads cannot keep up with the rate of async events under load).


electrum:
Merged, thanks!

PR #575, opened 2017-11-07
Add support for using sync socket address resolution in HTTP client

nezihyigitbasi:
Per discussion with the Jetty team, this will reduce the load on the client selector thread pool. This will also help when IP addresses are used instead of host names, because the calls to
InetAddress.getAllByName() will still need to be dispatched to the selector threads
by the default async socket address resolver.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
@@ -60,6 +60,7 @@ public class HttpClientConfig
     private String kerberosPrincipal;
     private String kerberosRemoteServiceName;
     private int selectorCount = 2;
+    private boolean useSyncSocketAddressResolution;
 
     private boolean http2Enabled;
     private DataSize http2InitialSessionReceiveWindowSize = new DataSize(16, MEGABYTE);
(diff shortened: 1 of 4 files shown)


electrum:
The commit message title is a bit long. How about shortening to "Support using sync socket address resolution in HTTP client"

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -60,6 +60,7 @@
     private String kerberosPrincipal;
     private String kerberosRemoteServiceName;
     private int selectorThreads = 2;
+    private boolean syncSocketAddressResolutionEnabled;
```
electrum: Nit, but I think this sound better named `useSyncSocketAddressResolution`. Compare
```
http-client.sync-socket-address-resolution-enabled=true
http-client.use-sync-socket-address-resolution=true
```
Trying to break down why, "enabled" seems more appropriate for something like HTTP/2 or authentication that we enable to allow it to be used, whereas "use" is better for "always use this thing".

nezihyigitbasi:
@electrum new name sounds good, updated.

electrum:
Closing per our discussion of writing a custom resolver to short-circuit the IP address literal case.

PR #576, opened 2017-11-07
Rename HTTP client selector config

nezihyigitbasi:
We should rename this one as it's really not setting thread count it's setting the number of selector objects, which are run by the client executors.

Do we want to add a `LegacyConfig` annotation?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,7 @@
+* 0.158
+
+- Rename HTTP client selector config parameter to "http-client.selector-count".
+
 * 0.157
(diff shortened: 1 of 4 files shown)


electrum:
This config has been around for a while, so we should add `@LegacyConfig`

electrum:
Nevermind, I was looking at the wrong config. Let's not bother with `@LegacyConfig` since this was just added in 0.156

electrum:
Merged, thanks!

PR #578, opened 2017-11-16
Fix Kerberos SPNEGO handling in HTTP client

electrum:
The client ignored the Negotiate scheme if a realm was not provided,
causing it to fail with the following error:

HTTP protocol violation: Authentication challenge without WWW-Authenticate header

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthenticationProtocolHandler.java
+++ b/http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthenticationProtocolHandler.java
@@ -1,0 +1,53 @@
+package io.airlift.http.client.spnego;
+
+import org.eclipse.jetty.client.AuthenticationProtocolHandler;
+import org.eclipse.jetty.client.HttpClient;
+import org.eclipse.jetty.client.WWWAuthenticationProtocolHandler;
+import org.eclipse.jetty.client.api.Response;
+import org.eclipse.jetty.client.api.Result;
+import org.eclipse.jetty.http.HttpFields;
+import org.eclipse.jetty.http.HttpHeader;
+
+/**
+ * Jetty {@link AuthenticationProtocolHandler} requires that the
+ * {@code "WWW-Authenticate"} header contains a realm, but the
+ * {@code "Negotiate"} scheme will not have one. This hacks around that by
+ * adding a new header value for that scheme which contains a dummy realm,
+ * allowing the rest of the authentication handling to work as normal.
+ * Unfortunately, there is no easier way as all of the related code in
+ * that class is private.
+ */
+public class SpnegoAuthenticationProtocolHandler
+        extends WWWAuthenticationProtocolHandler
+{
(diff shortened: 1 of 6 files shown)


dain:
@martint this is a client side fix instead of the server side we used to have.  I'd like this to be fixed in the Jetty client itself, but a work around is good for now

martint:
The previous fix was also in the client.

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthenticationProtocolHandler.java
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return new ForwardingResponseListener(super.getResponseListener())
+        {
+            @Override
+            public void onComplete(Result result)
+            {
+                HttpHeader header = getAuthenticateHeader();
+                HttpFields headers = result.getResponse().getHeaders();
+                if (headers.getValuesList(header).stream().anyMatch(NEGOTIATE::equalsIgnoreCase)) {
+                    headers.put(header, NEGOTIATE + " realm=\"dummy\"");
```
martint: What if the header contains a spnego token?

electrum: It won’t be a 401 so this code is never reached. The token is only sent on a successful authentication response.

electrum:
The previous fix was a hack that required cooperation with the server. The client sent a magic header asking the server to include a dummy realm. This fixes the client to work with any spec-complaint server.

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthenticationProtocolHandler.java
martint: Add a comment explaining this.

electrum: The Javadoc for the class explains this

electrum: Or do you mean why it only matches without a token?

electrum:
RFC 2617 is obsoleted by 7235

martint:
> RFC 2617 is obsoleted by 7235

Yes, I'm aware. I just find it interesting that the SPNEGO RFC, which directly references the old Auth RFC, has such a glaring error.

Apparently there's errata for it, which states:

> [...] and furthermore uses syntax in the WWW-Authenticate and Authorization header fields that is in violation of the base ABNF definitions.

But, no further explanation or fixes to it.

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthenticationProtocolHandler.java
martint: > Or do you mean why it only matches without a token?

Yes, so that it's clear to whoever looks at this 3 months down the road since it's not "obvious" from the code.

dain:
Merged

PR #582, opened 2017-11-30
Add an optimized async Jetty SocketAddressResolver

nezihyigitbasi:
When IP address literals are used this new SocketAddressResolver will not dispatch
the address resolution to the client executor, which will help reduce the load on
that executor. This especially helps under high load and when the executor is shared
by multiple HTTP clients.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyAsyncSocketAddressResolver.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyAsyncSocketAddressResolver.java
@@ -1,0 +1,49 @@
+package io.airlift.http.client.jetty;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.net.InetAddresses;
+import org.eclipse.jetty.util.Promise;
+import org.eclipse.jetty.util.SocketAddressResolver;
+import org.eclipse.jetty.util.thread.Scheduler;
+
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.Executor;
+
+/**
+ * For IP address literals this SocketAddressResolver implementation does not dispatch the address resolution
+ * to the executor. Under high load this helps offloading the executor, which may be shared by multiple different
+ * HTTP clients.
+ */
+class JettyAsyncSocketAddressResolver
+        extends SocketAddressResolver.Async
+{
+    public JettyAsyncSocketAddressResolver(Executor executor, Scheduler scheduler, long timeout)
+    {
+        super(executor, scheduler, timeout);
+    }
+
+    @Override
+    public void resolve(String host, int port, Promise<List<InetSocketAddress>> promise)
+    {
+        Optional<InetAddress> address = resolve(host);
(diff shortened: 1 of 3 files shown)


http-client/src/main/java/io/airlift/http/client/jetty/JettyAsyncSocketAddressResolver.java
```diff
@@ -0,0 +1,41 @@
...
+{
+    public JettyAsyncSocketAddressResolver(Executor executor, Scheduler scheduler, long timeout)
+    {
+        super(executor, scheduler, timeout);
+    }
+
+    @Override
+    public void resolve(String host, int port, Promise<List<InetSocketAddress>> promise)
+    {
+        //TODO Update after https://github.com/google/guava/issues/3001
```
electrum: It's unclear if they will add a new API, so I'd rather remove this. If they ever add a better API, we can update

http-client/src/main/java/io/airlift/http/client/jetty/JettyAsyncSocketAddressResolver.java
```diff
@@ -0,0 +1,41 @@
...
+import java.net.InetSocketAddress;
+import java.util.List;
+import java.util.concurrent.Executor;
+
+/**
+ * For IP address literals this SocketAddressResolver implementation does not dispatch the address resolution
+ * to the executor. Under high load this helps offloading the executor, which may be shared by multiple different
+ * HTTP clients.
+ */
+public class JettyAsyncSocketAddressResolver
```
electrum: This can be package private

http-client/src/main/java/io/airlift/http/client/jetty/JettyAsyncSocketAddressResolver.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public void resolve(String host, int port, Promise<List<InetSocketAddress>> promise)
+    {
+        //TODO Update after https://github.com/google/guava/issues/3001
+        try {
+            InetAddress address = InetAddresses.forString(host);
+            promise.succeeded(ImmutableList.of(new InetSocketAddress(address, port)));
+            return;
+        }
+        catch (IllegalArgumentException ignored) {
```
electrum: We should make this try/catch as small as possible, so it doesn't catch errors from anything other than `InetAddresses.forString()`

electrum: How about
```java
@Override
public void resolve(String host, int port, Promise<List<InetSocketAddress>> promise)
{
    Optional<InetAddress> address = resolve(host);
    if (address.isPresent()) {
        promise.succeeded(ImmutableList.of(new InetSocketAddress(address.get(), port)));
        return;
    }
    super.resolve(host, port, promise);
}

private static Optional<InetAddress> resolve(String host)
{
    try {
        return Optional.of(InetAddresses.forString(host));
    }
    catch (IllegalArgumentException ignored) {
        // not an IP address literal
        return Optional.empty();
    }
}

electrum:
Merged, thanks!

PR #583, opened 2017-12-01
Add option to disable recording HTTP request complete info

highker:
By default, an HTTP request records stats on completion. However, adding
counters to stats can be expensive given the lock contention in the
stats. Add an option to disable recording the completion info.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
@@ -60,6 +60,7 @@ public class HttpClientConfig
     private String kerberosPrincipal;
     private String kerberosRemoteServiceName;
     private int selectorCount = 2;
+    private boolean recordRequestComplete = true;
 
     private boolean http2Enabled;
     private DataSize http2InitialSessionReceiveWindowSize = new DataSize(16, MEGABYTE);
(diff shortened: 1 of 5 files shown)


electrum:
Please add this to CHANGES

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -60,6 +60,7 @@
     private String kerberosPrincipal;
     private String kerberosRemoteServiceName;
     private int selectorCount = 2;
+    private boolean recordRequestCompleteEnabled = true;
```
electrum: How about naming this `recordRequestComplete`

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -359,4 +360,16 @@ public HttpClientConfig setSelectorCount(int selectorCount)
         this.selectorCount = selectorCount;
         return this;
     }
+
+    public boolean getRecordRequestCompleteEnabled()
+    {
+        return recordRequestCompleteEnabled;
+    }
+
+    @Config("http-client.record-request-complete.enabled")
```
electrum: I think we can name this `http-client.record-request-complete`

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -120,6 +120,7 @@
     private final long maxContentLength;
     private final long requestTimeoutMillis;
     private final long idleTimeoutMillis;
+    private final boolean isRecordRequestComplete;
```
electrum: Remove the "is" from the field name. That's only used for getters.

electrum:
There's a typo here: *"given the we set a small number to it"*

It should be *"given that we set it to a small number"*

highker:
Comments addressed; thanks for review.

electrum:
Can you fix the commit description?

highker:
done; thanks

electrum:
Merged, thanks!

PR #584, opened 2017-12-01
Fix sporadic failures in HTTP client testConnectTimeout

electrum:


electrum:
It appears that it's not just a flaky test. When `TimeoutException` occurs, the connect timeout was not respected:
```
[ERROR] testConnectTimeout(io.airlift.http.client.jetty.TestAsyncJettyHttpClient)  Time elapsed: 2.026 s  <<< FAILURE!
java.lang.AssertionError: expected:<2.01s> to be less than <300.00ms>
```

PR #591, opened 2017-12-13
Use colon as separator in launcher for LD_PRELOAD

wenleix:
While both colon and space are valid separators for LD_PRELOAD,
they cannot be used together.

This causes issue when launching shell command that appending other
libraries into LD_PRELOAD with colon separated.

As colon is more commonly used separator, switching to use colon
will reduce such potential issues.


electrum:
Merged, thanks!

PR #594, opened 2018-01-02
Make format exceptions for config errors impossible

twilliamson:
Instead of making it the responsibility of the caller, make `Problems` smarter:
- If it gets a format string but no params, then it can't be a format string.
- Even if the format string is bogus, preserve the original message.


electrum:
Merged, thanks!

PR #595, opened 2018-01-03
Bind a dedicated thread pool to each HTTP client 

nezihyigitbasi:


http-client/src/test/java/io/airlift/http/client/TestHttpClientBinder.java
```diff
@@ -358,56 +306,21 @@ public void configure(Binder binder)
...
         assertNotSame(fooClient, barClient);
 
-        assertPrivatePools(injector, FooClient.class, BarClient.class);
+        assertJettyIoPools(injector, FooClient.class, BarClient.class);
 
         assertPoolsDestroyProperly(injector, FooClient.class, BarClient.class);
     }
 
-    @Test
-    public void testMultiplePrivateAndSharedThreadPools()
```
nezihyigitbasi: this is now the same as `testMultipleClients`, so removed.

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
```diff
@@ -56,11 +52,6 @@
 {
     private static final Logger log = Logger.get(HttpClientModule.class);
```
electrum: This is no longer used.

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
```diff
@@ -153,106 +137,11 @@ public HttpClient get()
...
-            if (injector.getExistingBinding(Key.get(JettyIoPoolManager.class, annotation)) != null) {
-                log.debug("HttpClient %s uses private IO thread pool", name);
-                ioPoolProvider = injector.getInstance(Key.get(JettyIoPoolManager.class, annotation));
-            }
-            else {
-                log.debug("HttpClient %s uses shared IO thread pool", name);
-                ioPoolProvider = injector.getInstance(JettyIoPoolManager.class);
-            }
-
+            verify(injector.getExistingBinding(Key.get(JettyIoPoolManager.class, annotation)) != null, "binding should exist");
```
electrum: This check isn't needed anymore. We know it exists because we bound it at the same time we created this provider.

http-client/src/main/java/io/airlift/http/client/jetty/JettyIoPoolManager.java
```diff
@@ -0,0 +1,73 @@
...
+
+import javax.annotation.PreDestroy;
+import javax.inject.Inject;
+
+import java.lang.annotation.Annotation;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class JettyIoPoolManager
```
electrum: Let's refactor this to be a top-level class first, as a separate commit. That will make it easier to see the changes needed for shared pool removal.

http-client/src/main/java/io/airlift/http/client/jetty/JettyIoPoolManager.java
```diff
@@ -0,0 +1,73 @@
...
+    private final List<JettyHttpClient> clients = new ArrayList<>();
+    private final String name;
+    private final Class<? extends Annotation> annotation;
+    private final AtomicBoolean destroyed = new AtomicBoolean();
+    private JettyIoPool pool;
+    private Injector injector;
+
+    public JettyIoPoolManager(String name, Class<? extends Annotation> annotation)
+    {
+        this.name = name;
```
electrum: Name is no longer used

http-client/src/main/java/io/airlift/http/client/jetty/JettyIoPoolManager.java
```diff
@@ -0,0 +1,73 @@
...
+        this.annotation = annotation;
+    }
+
+    public void addClient(JettyHttpClient client)
```
electrum: We will only have one client

http-client/src/main/java/io/airlift/http/client/jetty/JettyIoPoolManager.java
electrum: Actually, I think this class can be replaced with a `JettyIoPoolProvider`. It will probably look similar to `DiscoveryExecutorProvider`.

http-client/src/test/java/io/airlift/http/client/TestHttpClientBinder.java
```diff
@@ -236,8 +236,8 @@ public void configure(Binder binder)
         assertSame(injector.getInstance(Key.get(HttpClient.class, FooAlias2.class)), client);
         assertSame(injector.getInstance(Key.get(HttpClient.class, FooAlias3.class)), client);
 
-        // a private pool should not be registered for these clients
-        assertNull(injector.getExistingBinding(Key.get(JettyIoPoolManager.class, FooClient.class)));
+        // a private pool should be registered for the FooClient only
```
electrum: Let's remove the word "private" since that concept no longer exists. Also change the comment to
```java
// a pool should be registered for the FooClient only (not for the aliases)

File: http-client/src/test/java/io/airlift/http/client/TestHttpClientBinder.java [hunk 8]
```diff
@@ -268,8 +268,8 @@ public void configure(Binder binder)
         assertSame(injector.getInstance(Key.get(HttpClient.class, FooAlias1.class)), client);
         assertSame(injector.getInstance(Key.get(HttpClient.class, FooAlias2.class)), client);
 
-        // a private pool should not be registered for these clients
-        assertNull(injector.getExistingBinding(Key.get(JettyIoPoolManager.class, FooClient.class)));
+        // a private pool should be registered for FooClient only
```
electrum: Same as above

http-client/src/test/java/io/airlift/http/client/TestHttpClientBinder.java
electrum: Similarly, the test method names below probably need to be updated now

http-client/src/test/java/io/airlift/http/client/TestHttpClientBinder.java
```diff
@@ -281,14 +281,9 @@ public void testMultipleClients()
...
         Injector injector = new Bootstrap(
-                new Module()
-                {
-                    @Override
-                    public void configure(Binder binder)
-                    {
-                        httpClientBinder(binder).bindHttpClient("foo", FooClient.class);
-                        httpClientBinder(binder).bindHttpClient("bar", BarClient.class);
-                    }
+                (Module) binder -> {
```
electrum: I don't think these casts are needed. The code worked when I removed them.

http-client/src/test/java/io/airlift/http/client/TestHttpClientBinder.java
```diff
@@ -298,15 +293,13 @@ public void configure(Binder binder)
         HttpClient barClient = injector.getInstance(Key.get(HttpClient.class, BarClient.class));
         assertNotSame(fooClient, barClient);
 
-        assertNull(injector.getExistingBinding(Key.get(JettyIoPoolManager.class, FooClient.class)));
-        assertNull(injector.getExistingBinding(Key.get(JettyIoPoolManager.class, BarClient.class)));
-
+        // a private pool should be registered for these clients
```
electrum: Same

http-client/src/test/java/io/airlift/http/client/TestHttpClientBinder.java
```diff
@@ -125,20 +118,16 @@ public void testBindingMultipleFiltersAndClients()
...
-                        builder.addFilterBinding().to(AnotherHttpRequestFilter.class);
-                    }
+                binder -> {
+                    httpClientBinder(binder)
+                            .bindHttpClient("foo", FooClient.class)
+                            .withFilter(TestingRequestFilter.class)
+                            .withFilter(AnotherHttpRequestFilter.class)
+                            .withTracing();
+
+                    HttpClientBindingBuilder builder = httpClientBinder(binder).bindHttpClient("bar", BarClient.class);
```
electrum: While we're changing this, let's get rid of the builder. Users of the binder shouldn't normally see the type `HttpClientBindingBuilder`

http-client/src/main/java/io/airlift/http/client/jetty/JettyIoPoolManager.java
nezihyigitbasi: `name` is passed to the `JettyIoPool` below.

http-client/src/main/java/io/airlift/http/client/jetty/JettyIoPoolManager.java
```diff
@@ -0,0 +1,71 @@
...
+            pool = null;
+        }
+        destroyed.set(true);
+    }
+
+    public JettyIoPool get()
+    {
+        if (pool == null) {
+            JettyIoPoolConfig config = injector.getInstance(keyFromNullable(JettyIoPoolConfig.class, annotation));
+            pool = new JettyIoPool(name, config);
```
nezihyigitbasi: `name` is used here.

electrum:
Merged, thanks!

PR #596, opened 2018-01-09
Makes JsonFactory used for ObjectMapperProvider configurable

hgschmie:
this allows using the ObjectMapperProvider for other formats than Json:

```java
public void configure(Binder binder) {
    binder.bind(ObjectMapper.class).annotatedWith(Json.class).toProvider(ObjectMapperProvider.class);
    binder.bind(ObjectMapper.class).annotatedWith(Smile.class).toProvider(new ObjectMapperProvider(new SmileFactory()));
}
```

while all the serializers, modules etc. are configured in the same way for all ObjectMappers independently from the output format.

### DIFF
--- a/json/src/main/java/io/airlift/json/ObjectMapperProvider.java
+++ b/json/src/main/java/io/airlift/json/ObjectMapperProvider.java
@@ -53,6 +53,7 @@ public class ObjectMapperProvider
 
     private final Set<Module> modules = new HashSet<>();
 
+    @Inject
     public ObjectMapperProvider() {
         this(new JsonFactory());
     }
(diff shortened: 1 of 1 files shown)


json/src/main/java/io/airlift/json/ObjectMapperProvider.java
```diff
@@ -43,15 +44,23 @@
 public class ObjectMapperProvider
         implements Provider<ObjectMapper>
 {
+    private final JsonFactory jsonFactory;
```
electrum: This needs to be passed to the `ObjectMapper` factory

electrum:
Merged, thanks!

PR #597, opened 2018-01-09
Update to Airbase 79

electrum:


bootstrap/src/main/java/io/airlift/bootstrap/LoggingWriter.java
```diff
@@ -59,21 +51,8 @@ public void flush()
...
-                    }
-
-                    case INFO: {
-                        if (logger.isInfoEnabled()) {
-                            logger.info(line);
-                        }
-                        break;
-                    }
+                if (logger.isInfoEnabled()) {
+                    logger.info(line);
```
dain: I don't think we need the "if" check.  Just always log at info.  The logger is going to check enabled anyway.... the "is enabled" is only useful when the log line does work

PR #599, opened 2018-01-10
Fix shutdown of HTTP client when lifecyle is stopped

electrum:


http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
```diff
@@ -113,7 +118,14 @@ public HttpClient get()
...
 
-            return new JettyHttpClient(name, config, kerberosConfig, ImmutableList.copyOf(filters));
+            client = new JettyHttpClient(name, config, kerberosConfig, ImmutableList.copyOf(filters));
+            return client;
+        }
+
+        @PreDestroy
+        public void destroy()
+        {
+            client.close();
```
nezihyigitbasi: set to null after close?

PR #600, opened 2018-01-10
Refactor JettyHttpClient to extract inner classes

nezihyigitbasi:
This class is pretty huge right now, so this PR extracts the inner classes
as top-level classes.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/jetty/CachedDistribution.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/CachedDistribution.java
@@ -1,0 +1,132 @@
+package io.airlift.http.client.jetty;
+
+import io.airlift.stats.Distribution;
+import org.weakref.jmx.Managed;
+
+import javax.annotation.concurrent.GuardedBy;
+import javax.annotation.concurrent.ThreadSafe;
+
+import java.util.Map;
+import java.util.function.Supplier;
+
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+/*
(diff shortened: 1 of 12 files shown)


http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -771,7 +734,7 @@ private static void closeQuietly(LifeCycle service)
                 service.stop();
             }
         }
-        catch (InterruptedException ignored) {
+        catch (InterruptedException ie) {
```
electrum: We can name this `e`, as we do elsewhere. I thought it needed to be `ignored` to prevent a warning, but apparently that's only a problem if the catch block is empty.

http-client/src/main/java/io/airlift/http/client/jetty/CachedDistribution.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.function.Supplier;
+
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+/*
+ * This class is needed because jmxutils only fetches a nested instance object once and holds on to it forever.
+ * todo remove this when https://github.com/martint/jmxutils/issues/26 is implemented
+ */
+@ThreadSafe
+public class CachedDistribution
```
electrum: Does this class need to be public?

nezihyigitbasi: good catch, no, I will update.

nezihyigitbasi:
@electrum updated.

electrum:
Merged, thanks!

PR #603, opened 2018-01-25
Change async appender queue size to be configurable

raghavsethi:


http-server/src/main/java/io/airlift/http/server/DelimitedRequestLog.java
```diff
@@ -50,13 +50,14 @@
     private final CurrentTimeMillisProvider currentTimeMillisProvider;
     private final AsyncAppenderBase<HttpRequestEvent> asyncAppender;
 
-    public DelimitedRequestLog(String filename, int maxHistory, long maxFileSizeInBytes, TraceTokenManager traceTokenManager, EventClient eventClient)
+    public DelimitedRequestLog(String filename, int maxHistory, int asyncBuffer, long maxFileSizeInBytes, TraceTokenManager traceTokenManager, EventClient eventClient)
```
martint: Call this "bufferSize"

PR #605, opened 2018-01-27
Add whenAnyCompleteCancelOthers

haozhun:


concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
```diff
@@ -284,6 +284,32 @@ public static void checkSuccess(Future<?> future, String errorMessage)
...
+    {
+        requireNonNull(futures, "futures is null");
+        checkArgument(!isEmpty(futures), "futures is empty");
+
+        // wait for the first task to unblock and then cancel all futures to free up resources
+        ListenableFuture<V> anyComplete = whenAnyComplete(futures);
+        anyComplete.addListener(
+                () -> {
+                    for (ListenableFuture<?> future : futures) {
+                        future.cancel(true);
```
dain: using an ExtendedSettableFuture instead of a settable future, you can know if the user canceled the outer exception with the interrupted flag set.

concurrent/src/test/java/io/airlift/concurrent/BenchmarkWhenAnyCompleteCancelOthers.java
```diff
@@ -0,0 +1,85 @@
...
+    //        with optimization in ExtendedSettableFuture.setAsync    without
+    // 300                    32.7 +/- 1.7 us                     810.6 +/- 31 us
+    // 1000                  106.2 +/- 3.0 us                    2813.5 +/- 121 us
+
+    @Param({"300", "1000"})
+    private int futureCount;
+
+    @Benchmark
+    public void benchmark()
+            throws ExecutionException, InterruptedException
```
dain: Just throw `Exception`?

concurrent/src/test/java/io/airlift/concurrent/BenchmarkWhenAnyCompleteCancelOthers.java
```diff
@@ -0,0 +1,85 @@
...
+        Semaphore semaphore = new Semaphore(futureCount);
+
+        ArrayList<SettableFuture<?>> futures = new ArrayList<>();
+        for (int i = 0; i < futureCount; i++) {
+            SettableFuture<?> future = SettableFuture.create();
+            future.addListener(() -> semaphore.release(1), directExecutor());
+            futures.add(future);
+        }
+        ListenableFuture<?> anyComplete = whenAnyCompleteCancelOthers(futures);
+        futures.get(futureCount / 2).set(null);
```
dain: did you try benchmarking with `cancel`?

haozhun: When benchmarking with `cancel`. The `anyComplete.get()` two lines below will throw an exception. This adds a ~5 us overhead to add test parameters. As a result, the ratio of improvement changes. But the overall conclusion of this benchmark stands.

concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
haozhun: I'll leave this as is, as we discussed offline.

The returned future could be completed (by calling `set`), and we need to cancel here still. `wasInterrupted` wouldn't give any valuable insight in such a case.

PR #607, opened 2018-02-15
Add JmxGcMonitor to track GC events

dain:


stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -0,0 +1,276 @@
...
+                        "Major GC: application %sms, stopped %sms:: %s -> %s",
+                        applicationRuntime,
+                        info.getDurationMs(),
+                        info.getBeforeGcTotalMb(),
+                        info.getAfterGcTotalMb());
+            }
+            else if (info.isMinorGc()) {
+                minorGc.add(info.getDurationMs(), MILLISECONDS);
+
+                // assume that minor GCs run currently
```
martint: It's not that simple. For CMS, young generation collections are stop-the-world (it uses the parallel collector). For G1, young and mixed collections have a stop-the-world behavior (evacuation pause, etc) and can be significant if the GC is not tuned correctly (target pause, young generation size and so on).

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -0,0 +1,267 @@
...
+        {
+            return usageBeforeGc;
+        }
+
+        public Map<String, MemoryUsage> getMemoryUsageAfterGc()
+        {
+            return usageAfterGc;
+        }
+
+        private long getBeforeGcTotalMb()
```
martint: Why not DataSize for these?

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -0,0 +1,267 @@
...
+
+                // assume that major GCs stop the application
+                long applicationRuntime = max(0, info.getStartTime() - lastGcEndTime);
+                lastGcEndTime = info.getEndTime();
+
+                log.info(
+                        "Major GC: application %sms, stopped %sms:: %s -> %s",
+                        applicationRuntime,
+                        info.getDurationMs(),
+                        info.getBeforeGcTotalMb(),
```
martint: We should either use DataSize for these or add the appropriate unit in the log message.

PR #608, opened 2018-02-23
Increase timeout for flakey http client test

dain:


electrum:
Typo in commit message

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -199,7 +199,7 @@ public void test100kGets()
         }
     }
 
-    @Test(timeOut = 5000)
```
electrum: We should probably keep this

pom.xml
```diff
@@ -428,4 +428,17 @@
             </dependency>
         </dependencies>
     </dependencyManagement>
+    <build>
```
electrum: Add newline before this

PR #614, opened 2018-03-01
Make launcher.py Python 3 compatible.

jthompson6:
This means loosing compatibility with versions < 2.6, as `except OSError, e:` is not valid in Python 3, while its replacement `except OSError as e:` is only valid for 2.6+.

However, Python 2.6 and earlier are no longer supported, and have been out for a very long time, so it should be time to upgrade.

The most non-trivial thing was the need to set the file descriptor for launcher.pid as inheritable, otherwise it would be lost in the switch from the python to the java process, and the status checks would not work. This is needed for Python versions 3.4+ (see https://docs.python.org/3/library/os.html#inheritance-of-file-descriptors).

### DIFF
--- a/launcher/src/main/scripts/bin/launcher.py
+++ b/launcher/src/main/scripts/bin/launcher.py
@@ -270,12 +270,11 @@ def start(process, options):
         process.write_pid(pid)
         print('Started as %s' % pid)
         return
-    try:
+
+    if hasattr(os, "set_inheritable"):
         # See https://docs.python.org/3/library/os.html#inheritance-of-file-descriptors
         # Since Python 3.4
         os.set_inheritable(process.pid_file.fileno(), True)
-    except AttributeError as e:
-        pass
 
     os.setsid()
 
--- a/launcher/src/main/scripts/bin/launcher.py
+++ b/launcher/src/main/scripts/bin/launcher.py
@@ -35,7 +35,7 @@ def makedirs(p):
     """Create directory and all intermediate ones"""
     try:
         os.makedirs(p)
-    except OSError, e:
+    except OSError as e:
         if e.errno != errno.EEXIST:
             raise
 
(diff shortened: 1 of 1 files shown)


launcher/src/main/scripts/bin/launcher.py
```diff
@@ -251,9 +251,15 @@ def run(process, options):
     os.execvpe(args[0], args, env)
 
 
+def need_set_inheritable():
+    # See https://docs.python.org/3/library/os.html#inheritance-of-file-descriptors
+    py_version = sys.version_info.major + .1 * sys.version_info.minor
+    return py_version >= 3.4
```
electrum: I think this needs to be a tuple comparison. Using a number will fail when Python 3.10 comes out.

launcher/src/main/scripts/bin/launcher.py
```diff
@@ -268,9 +274,12 @@ def start(process, options):
     pid = os.fork()
     if pid > 0:
         process.write_pid(pid)
-        print 'Started as %s' % pid
+        print('Started as %s' % pid)
         return
 
+    if need_set_inheritable():
+        os.set_inheritable(process.pid_file.fileno(), True)
```
electrum: Instead of checking for the version, what if we just always call it and catch/ignore AttributeError?

jthompson6:
Hi @electrum.

I moved the set_inheritable into a try as you suggested so the python version check is not necessary anymore. Alternatively, I could use `hasattr(os, "set_inheritable")`.

Do you want me to squash the commits?

jthompson6:
I switched to hasattr because I think it's neater.

electrum:
Before I merge this, please confirm the Git author:
```
Author: jthompson6 <4349284+jthompson6@users.noreply.github.com>
```
Is this the email address you wanted to use?

jthompson6:
The name should be "Jenny Thompson", but leave the email as it is (I don't want my real email to be public).

electrum:
Merged, thanks!

PR #616, opened 2018-03-06
Update createHttpRequestEvent to check the token header first

nezihyigitbasi:
Currently, HttpRequestEvent::createHttpRequestEvent() does not check
whether the request has the "X-Airlift-TraceToken" header. This change
first checks that HTTP header and uses it if present. Otherwise, it
gets the current token in the traceTokenManager if available.

Without this change tokens set by clients will not be logged.

### DIFF
--- a/http-server/src/main/java/io/airlift/http/server/HttpRequestEvent.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpRequestEvent.java
@@ -29,6 +29,7 @@ import java.time.Instant;
 import java.util.Enumeration;
 
 import static io.airlift.event.client.EventField.EventFieldMapping.TIMESTAMP;
+import static io.airlift.http.server.TraceTokenFilter.TRACETOKEN_HEADER;
 import static java.lang.Math.max;
 
 @EventType("HttpRequest")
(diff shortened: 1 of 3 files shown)


http-server/src/main/java/io/airlift/http/server/HttpRequestEvent.java
```diff
@@ -42,8 +43,8 @@ public static HttpRequestEvent createHttpRequestEvent(Request request, Response
             user = principal.getName();
         }
 
-        String token = null;
-        if (traceTokenManager != null) {
+        String token = request.getHeader(TRACETOKEN_HEADER);
+        if (token == null && traceTokenManager != null) {
```
dain: I would expect that the header token would be in the token manager already from the `TraceTokenFilter`

nezihyigitbasi: If requests are processed synchronously then yes the token will be available in thread locals, but for async processing threads other than the Jetty worker threads will go through this code path and the token won't be available in the thread locals.

http-server/src/main/java/io/airlift/http/server/HttpRequestEvent.java
```diff
@@ -42,8 +43,8 @@ public static HttpRequestEvent createHttpRequestEvent(Request request, Response
             user = principal.getName();
         }
 
-        String token = null;
-        if (traceTokenManager != null) {
+        String token = request.getHeader(TRACETOKEN_HEADER);
```
dain: Add a comment that this is required because async responses fire on a different thread.

johngmyers:
FYI, you might be interested in platform's `TraceTokenCopyingExecutor`

johngmyers:
Platform also extended HttpCient to propagate the trace token to BodyGenerator and ResponseHandler calls.

PR #618, opened 2018-03-07
Make TestHttpServerProvider::testStop less flaky

nezihyigitbasi:
Added some coordination with the `DummyServlet` to make the test less flaky.


http-server/src/test/java/io/airlift/http/server/TestHttpServerCipher.java
```diff
@@ -191,7 +192,7 @@ private static HttpClient createClientIncludeCiphers(String... includedCipherSui
 
     private static HttpServer createServer(NodeInfo nodeInfo, HttpServerInfo httpServerInfo, HttpServerConfig config)
     {
-        return createServer(new DummyServlet(), nodeInfo, httpServerInfo, config);
+        return createServer(new DummyServlet(Optional.empty()), nodeInfo, httpServerInfo, config);
```
dain: This is what the default constructor already does.

http-server/src/test/java/io/airlift/http/server/DummyServlet.java
```diff
@@ -15,33 +15,64 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+
+import static java.util.Objects.requireNonNull;
+import static javax.servlet.http.HttpServletResponse.SC_OK;
 
 class DummyServlet
         extends HttpServlet
 {
+    private final Optional<CountDownLatch> latch;
+    private AtomicLong requestsReceived = new AtomicLong();
```
dain: Add blank line below this and the constructor on the next line.

dain: Final?

http-server/src/test/java/io/airlift/http/server/DummyServlet.java
```diff
@@ -15,33 +15,64 @@
...
+    public DummyServlet(Optional<CountDownLatch> latch)
```
dain: Since this is only used when you have a latch, I'd change this to take a `CountDownLatch` directly.

http-server/src/test/java/io/airlift/http/server/TestHttpServerProvider.java
```diff
@@ -481,7 +493,7 @@ public void testInsufficientThreadsAdmin()
     private void createAndStartServer()
             throws Exception
     {
-        createAndStartServer(new DummyServlet());
+        createAndStartServer(new DummyServlet(Optional.empty()));
```
dain: Use the default constructor for all of these

http-server/src/test/java/io/airlift/http/server/TestHttpServerProvider.java
```diff
@@ -392,17 +394,27 @@ public void testShowStackTraceDisabled()
...
+            // wait until the request is active
+            while (servlet.getRequestsReceived() == 0) {
+                Thread.sleep(100);
+            }
+
+            // call stop when the request is still active (waiting on the latch)
             server.stop();
 
+            // let the request proceed
+            latch.countDown();
```
dain: Maybe put a `sleep(100)` before this so the stop has a chance to really kick in (since the server is multithreaded)

http-server/src/test/java/io/airlift/http/server/DummyServlet.java
```diff
@@ -15,33 +15,64 @@
...
+
+    public DummyServlet()
+    {
+        this.latch = Optional.empty();
+    }
+
     @Override
-    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
-            throws ServletException, IOException
+    protected void doGet(HttpServletRequest request, HttpServletResponse response)
```
electrum: Please move this refactoring to a separate commit.

http-server/src/test/java/io/airlift/http/server/DummyServlet.java
```diff
@@ -15,33 +15,64 @@
...
+        if (request.getUserPrincipal() != null) {
+            response.getOutputStream().write(request.getUserPrincipal().getName().getBytes());
         }
-        resp.setHeader("X-Protocol", req.getProtocol());
+        response.setHeader("X-Protocol", request.getProtocol());
 
         try {
-            if (req.getParameter("sleep") != null) {
-                Thread.sleep(Long.parseLong(req.getParameter("sleep")));
+            if (request.getParameter("sleep") != null) {
```
electrum: Is this still used?

nezihyigitbasi:
I added code to `DummyServlet` to capture more details and addressed your comments.

http-server/src/test/java/io/airlift/http/server/TestHttpServerProvider.java
```diff
@@ -392,18 +395,33 @@ public void testShowStackTraceDisabled()
...
+        createAndStartServer(servlet);
         try (HttpClient client = new JettyHttpClient()) {
             URI uri = URI.create(httpServerInfo.getHttpUri().toASCIIString() + "/?sleep=50000");
             Request request = prepareGet().setUri(uri).build();
-            HttpResponseFuture<?> future = client.executeAsync(request, createStatusResponseHandler());
+            HttpResponseFuture<StatusResponse> future = client.executeAsync(request, createStatusResponseHandler());
 
+            // wait until the servlet starts processing the request
+            while (servlet.getActiveRequests() == 0) {
+                Thread.sleep(100);
```
dain: You could use a latch here also, but I'm not sure it is worth the hassle.

http-server/src/test/java/io/airlift/http/server/TestHttpServerProvider.java
```diff
@@ -392,18 +395,33 @@ public void testShowStackTraceDisabled()
...
+            }
+
+            // call stop when the request is still active
             server.stop();
 
+            // wait until the server is stopped
+            server.join();
```
electrum: This can wait for a long time (close to but not exceeding the `@Test(timeOut)` value), which defeats the purpose of the test, which is to assure that the server shuts down quickly, even if there are long-running requests. We should have a timeout here.

http-server/src/test/java/io/airlift/http/server/TestHttpServerProvider.java
```diff
@@ -392,18 +395,33 @@ public void testShowStackTraceDisabled()
...
+
             try {
-                future.get(5, TimeUnit.SECONDS);
-                fail("expected exception");
+                StatusResponse response = future.get(5, SECONDS);
+                String stackTrace = null;
+                if (servlet.getException() != null) {
```
electrum: No need to do this. Just use
```java
throw new AssertionError("expected request to fail", servlet.getException());

electrum: The `fail()` method is just a wrapper to throw `AssertionError`

http-server/src/test/java/io/airlift/http/server/TestHttpServerProvider.java
electrum: Use a `CountDownLatch` here

http-server/src/test/java/io/airlift/http/server/TestHttpServerProvider.java
electrum: Thinking about this more... we actually don't know if `stop()` takes a long time. We expect it returns immediately, but it might not.

nezihyigitbasi: Yep. Also we have to wait anyway until the server stops, because as you say we want to make sure that the server shuts down, without joining we can't know.

http-server/src/test/java/io/airlift/http/server/DummyServlet.java
nezihyigitbasi: now yes.

PR #626, opened 2018-03-22
Monitor log queue

nezihyigitbasi:
We have seen that the new HTTP client logger can become a bottleneck under load with a large number of HTTP client threads logging events.

Before exploring more complex solutions, this PR first adds support for tuning the logger buffer size and adds monitoring support to the underlying logger queue.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,7 @@
+0.165
+
+- Add log buffer size config to HTTP client.
+
 0.164
 
 - Fix HTTP server log of trace tokens when using async responses.
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 0.165
 
 - Add log buffer size config to HTTP client.
+- Add support for monitoring the HTTP client and server logger queue size.
 
 0.164
 
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
@@ -79,6 +79,7 @@ public class HttpClientConfig
     private int logHistory = 15;
     private int logQueueSize = 10_000;
     private DataSize logMaxFileSize = new DataSize(1, GIGABYTE);
+    private DataSize logBufferSize = new DataSize(1, MEGABYTE);
 
     public boolean isHttp2Enabled()
     {
@@ -511,6 +512,21 @@ public class HttpClientConfig
     public HttpClientConfig setLogQueueSize(int logQueueSize)
     {
         this.logQueueSize = logQueueSize;
+        return this;
+    }
+
+    @NotNull
+    @MinDataSize("1MB")
+    @MaxDataSize("1GB")
+    public DataSize getLogBufferSize()
+    {
+        return logBufferSize;
+    }
+
+    @Config("http-client.log.buffer-size")
+    public HttpClientConfig setLogBufferSize(DataSize logBufferSize)
+    {
+        this.logBufferSize = logBufferSize;
         return this;
     }
 }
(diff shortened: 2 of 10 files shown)


http-client/src/main/java/io/airlift/http/client/jetty/DefaulHttpClientLogger.java
```diff
@@ -88,6 +89,12 @@ public void close()
         asyncAppender.stop();
     }
 
+    @Managed
```
electrum: This class doesn't seem to be exported. Probably add a `@Managed @Nested getLogger()` to `JettyHttpClient`.

http-client/src/main/java/io/airlift/http/client/jetty/DefaulHttpClientLogger.java
```diff
@@ -88,6 +89,12 @@ public void close()
...
+    public int getNumberOfElementsInQueue()
```
electrum: This method name seems overly verbose. How about `getQueueSize()`, since that will match the config name.

http-server/src/main/java/io/airlift/http/server/DelimitedRequestLog.java
```diff
@@ -174,6 +175,12 @@ public void removeLifeCycleListener(Listener listener)
     {
     }
 
+    @Managed
```
electrum: Similar comments

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -226,7 +226,11 @@ public JettyHttpClient(
         // configure logging
         if (config.isLogEnabled()) {
             String logFilePath = Paths.get(config.getLogPath(), format("%s-http-client.log", name)).toAbsolutePath().toString();
-            requestLogger = new DefaulHttpClientLogger(logFilePath, config.getLogHistory(), config.getLogQueueSize(), config.getLogMaxFileSize().toBytes());
+            requestLogger = new DefaulHttpClientLogger(logFilePath,
```
electrum: Nit: wrap first argument

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -513,4 +514,18 @@ public HttpClientConfig setLogQueueSize(int logQueueSize)
         this.logQueueSize = logQueueSize;
         return this;
     }
+
+    @NotNull
+    @MinDataSize("1MB")
```
electrum: Probably want a max here, even if it's huge like `1GB`

nezihyigitbasi:
I also had to set `fileAppender.setImmediateFlush(false);` in both the client and the server loggers as currently `immediateFlush=true` for the file appenders and that make `bufferSize` config useless.

http-server/src/main/java/io/airlift/http/server/DelimitedRequestLog.java
```diff
@@ -175,6 +176,12 @@ public void removeLifeCycleListener(Listener listener)
...
+    public int getQueueSize()
```
electrum: This logger class also needs to be exposed on `HttpServer`.

electrum:
Please pull the flush change into a separate commit

CHANGES
```diff
@@ -1,6 +1,7 @@
 0.165
 
 - Add log buffer size config to HTTP client.
+- Add support for monitoring the HTTP client/server logger queue size.
```
electrum: Let's say "client and server" here since they are independent components.

http-client/src/main/java/io/airlift/http/client/jetty/HttpClientLogger.java
```diff
@@ -25,6 +25,7 @@
 {
     void log(RequestInfo requestInfo, ResponseInfo responseInfo);
     void close();
+    int getLoggerQueueSize();
```
electrum: Name this `getQueueSize()`? The "logger" part seems redundant since this is the logger interface.

electrum:
Merged, thanks!

PR #627, opened 2018-03-27
Add config to disable HTTP client/server log compression

nezihyigitbasi:
Compression may cause issues with GCLocker. Disabling the compression and doing the compression in an external process will resolve that. Verified also with Presto.


CHANGES
```diff
@@ -1,3 +1,8 @@
+0,165
```
electrum: Typo

CHANGES
```diff
@@ -1,3 +1,8 @@
...
+
+- Add a config property to disable the compression of HTTP client/server
```
electrum: Don't wrap here. You could shorten to "Allow disabling the compression of ..."

nezihyigitbasi:
Updated the commit message details to give more info.

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -513,4 +514,16 @@ public HttpClientConfig setLogQueueSize(int logQueueSize)
         this.logQueueSize = logQueueSize;
         return this;
     }
+
+    public boolean isLogCompressionEnabled()
+    {
+        return logCompressionEnabled;
+    }
+
+    @Config("http-client.log.compression-enabled")
```
electrum: Should this be `compression.enabled`? Maybe we would add a `compression.format` later? (such as Zstd or LZ4)

http-client/src/main/java/io/airlift/http/client/jetty/DefaulHttpClientLogger.java
```diff
@@ -49,10 +49,15 @@
         TimeBasedRollingPolicy<HttpRequestEvent> rollingPolicy = new TimeBasedRollingPolicy<>();
 
         rollingPolicy.setContext(context);
-        rollingPolicy.setFileNamePattern(filename + "-%d{yyyy-MM-dd}.%i.log.gz");
         rollingPolicy.setMaxHistory(maxHistory);
         rollingPolicy.setTimeBasedFileNamingAndTriggeringPolicy(triggeringPolicy);
         rollingPolicy.setParent(fileAppender);
+        if (compressionEnabled) {
```
electrum: Should we avoid repeating the pattern and make it clear this is appending an extension?
```java
rollingPolicy.setFileNamePattern(filename + "-%d{yyyy-MM-dd}.%i.log");
if (compressionEnabled) {
    rollingPolicy.setFileNamePattern(rollingPolicy.getFileNamePattern() + ".gz");
}

File: http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java [hunk 5]
```diff
@@ -563,4 +564,16 @@ public HttpServerConfig setHttp2StreamIdleTimeout(Duration http2StreamIdleTimeou
         this.http2StreamIdleTimeout = http2StreamIdleTimeout;
         return this;
     }
+
+    public boolean isLogCompressionEnabled()
+    {
+        return logCompressionEnabled;
+    }
+
+    @Config("http-server.log.compression-enabled")
```
electrum: Same comment

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -563,4 +564,16 @@ public HttpServerConfig setHttp2StreamIdleTimeout(Duration http2StreamIdleTimeou
         this.http2StreamIdleTimeout = http2StreamIdleTimeout;
         return this;
     }
+
+    public boolean isLogCompressionEnabled()
```
electrum: Move these to be next to the other log-related methods.

BTW, separately, we should rename `queueSize` and the methods to be `logQueueSize`.

http-server/src/test/java/io/airlift/http/server/TestHttpServerConfig.java
```diff
@@ -71,7 +71,8 @@ public void testDefaults()
                 .setHttp2InitialSessionReceiveWindowSize(new DataSize(16, MEGABYTE))
                 .setHttp2InputBufferSize(new DataSize(8, KILOBYTE))
                 .setHttp2InitialStreamReceiveWindowSize(new DataSize(16, MEGABYTE))
-                .setHttp2StreamIdleTimeout(new Duration(15, SECONDS)));
+                .setHttp2StreamIdleTimeout(new Duration(15, SECONDS))
+                .setLogCompressionEnabled(true));
```
electrum: Move next to other log methods (same for all in this class)

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
nezihyigitbasi: Maybe, updating.

CHANGES
electrum: Let's say "client and server" here, since they are separate components.

http-client/src/main/java/io/airlift/http/client/jetty/DefaulHttpClientLogger.java
nezihyigitbasi: I am fine either way, updating.

CHANGES
```diff
@@ -1,3 +1,7 @@
+0.165
+
+- Allow disabling the compression of HTTP client and server log files.
```
electrum: Let's say "log files for HTTP client and server"

electrum:
Merged, thanks!

PR #634, opened 2018-04-18
Add SSL session timeout and cache size server configs

nezihyigitbasi:
This change adds the SSL session timeout and session cache size config
parameters to the HTTP server.

### DIFF
--- a/http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
@@ -25,6 +25,7 @@ import io.airlift.units.DataSize;
 import io.airlift.units.Duration;
 import io.airlift.units.MaxDataSize;
 import io.airlift.units.MinDataSize;
+import io.airlift.units.MinDuration;
 
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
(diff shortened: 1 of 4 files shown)


http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -70,6 +70,11 @@
     private List<String> includedCipherSuites = ImmutableList.of();
     private List<String> excludedCipherSuites = ImmutableList.of();
 
+    // For the SSL session-related config parameters
+    // a value of -1 means using the JVM default
+    private int sslSessionTimeout = -1;
```
dain: This should be a duration

dain: Oh also, lets have defaults.... 10k and 4 hours is most likely a fine place to start.

nezihyigitbasi:
@dain addressed the comments.

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -162,6 +166,30 @@ public HttpServerConfig setHttpsPort(int httpsPort)
         return this;
     }
 
+    public Duration getSslSessionTimeout()
+    {
+        return sslSessionTimeout;
+    }
+
+    @Config("http-server.https.ssl-session-timeout")
+    public HttpServerConfig setSslSessionTimeout(Duration sslSessionTimeout)
```
dain: add min duration of second

nezihyigitbasi:
That test doesn't use https and I don't see how these changes would cause it. But, it's weird that it fails. I tried reproducing it locally with 1K runs, but cannot.

nezihyigitbasi:
I added a commit with some logging to see what's going on.

dain:
And then it passed :(

nezihyigitbasi:
That test is totally flaky. I tried many times on Travis, but couldn't repro so far.

dain:
We can fix that in another PR.  Can you remove that commit and I'll land this.

nezihyigitbasi:
Removed.

PR #639, opened 2018-05-22
Add additional columns to the HTTP server logs

nezihyigitbasi:
This change adds the following columns to the HTTP server logs:

- stream id: stream identifier for HTTP/2 requests, -1 otherwise.
- protocol version: version of the HTTP protocol specified by the
request.
- request begin-to-dispatch time [ms]: time from when the request is
received to when the request is dispatched to the application.
- request begin-to-end time [ms]: time from when the request is received
to when the request has been fully consumed/parsed.
- first-to-last-content-time [ms]: time from when the first chunk of the
response content has been written to the network to when the last
chunk of the response content has been written.
- response content interarrival time stats: the summary statistics
for the interarrival time of the response content writes. For example,
if three response content chunks have been written by the HTTP server
at t=0, t=3, and t=20, then this field will contain the summary
statistics for the values {3, 17}.

/cc @martint

### DIFF
--- a/http-server/src/main/java/io/airlift/http/server/DoubleSummaryStats.java
+++ b/http-server/src/main/java/io/airlift/http/server/DoubleSummaryStats.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.http.server;
+
+import io.airlift.event.client.EventField;
+import io.airlift.event.client.EventType;
+
+import java.util.DoubleSummaryStatistics;
+
+import static java.util.Objects.requireNonNull;
+
+@EventType
+public class DoubleSummaryStats
+{
+    private final DoubleSummaryStatistics stats;
+
+    public DoubleSummaryStats(DoubleSummaryStatistics stats)
+    {
+        this.stats = requireNonNull(stats, "stats is null");
+    }
+
+    @EventField
+    public double getMin()
+    {
+        return stats.getMin();
+    }
+
+    @EventField
+    public double getMax()
+    {
+        return stats.getMax();
+    }
+
+    @EventField
+    public double getAverage()
+    {
+        return stats.getAverage();
+    }
+
+    @EventField
+    public long getCount()
+    {
+        return stats.getCount();
+    }
+}
--- a/http-server/src/main/java/io/airlift/http/server/HttpLogLayout.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpLogLayout.java
@@ -5,6 +5,7 @@ import ch.qos.logback.core.LayoutBase;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 
+import static java.lang.String.format;
 import static java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME;
 
 public class HttpLogLayout
@@ -16,6 +17,14 @@ public class HttpLogLayout
     public String doLayout(HttpRequestEvent event)
     {
         StringBuilder builder = new StringBuilder();
+
+        // format content interarrival time [ms] stats
+        String contentInterarrivalStats = null;
+        DoubleSummaryStats stats = event.getResponseContentInterarrivalStats();
+        if (stats != null) {
+            contentInterarrivalStats = format("%.2f, %.2f, %.2f, %d", stats.getMin(), stats.getAverage(), stats.getMax(), stats.getCount());
+        }
+
         builder.append(ISO_FORMATTER.format(event.getTimeStamp()))
                 .append('\t')
                 .append(event.getClientAddress())
@@ -37,6 +46,16 @@ public class HttpLogLayout
                 .append(event.getTimeToLastByte())
                 .append('\t')
                 .append(event.getTraceToken())
+                .append('\t')
+                .append(event.getProtocolVersion())
+                .append('\t')
+                .append(event.getBeginToDispatchMillis())
+                .append('\t')
+                .append(event.getBeginToEndMillis())
+                .append('\t')
+                .append(event.getFirstToLastContentTimeInMillis())
+                .append('\t')
+                .append(contentInterarrivalStats)
                 .append('\n');
 
         return builder.toString();
--- a/http-server/src/main/java/io/airlift/http/server/HttpServerChannelListener.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpServerChannelListener.java
@@ -1,0 +1,121 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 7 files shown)


http-server/src/main/java/io/airlift/http/server/HttpLogLayout.java
```diff
@@ -16,10 +17,22 @@
     public String doLayout(HttpRequestEvent event)
     {
         StringBuilder builder = new StringBuilder();
+
+        // format content interarrival time stats
+        String contentInterarrivalStats = null;
+        DoubleSummaryStats stats = event.getResponseContentInterarrivalStats();
+        if (stats != null) {
+            contentInterarrivalStats = format("%.2f, %.2f, %.2f, %d", stats.getMin(), stats.getAverage(), stats.getMax(), stats.getCount());
```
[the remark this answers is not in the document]
dain: Maybe add a `toString` that prints the way you want

http-server/src/main/java/io/airlift/http/server/HttpLogLayout.java
```diff
@@ -16,10 +17,22 @@
...
+        }
+
         builder.append(ISO_FORMATTER.format(event.getTimeStamp()))
                 .append('\t')
                 .append(event.getClientAddress())
                 .append('\t')
+                .append(event.getStreamId())
+                .append('\t')
+                .append(event.getProtocolVersion())
+                .append('\t')
```
dain: I don't think we want to reorder the columns, as it will break anyone parsing the file (like the Presto local file connector)

http-server/src/main/java/io/airlift/http/server/HttpServerChannelListener.java
```diff
@@ -0,0 +1,129 @@
...
+    public void onRequestBegin(Request request)
+    {
+        request.setAttribute(REQUEST_BEGIN_ATTRIBUTE, System.nanoTime());
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public void onBeforeDispatch(Request request)
+    {
+        long requestBeginTime = (Long) request.getAttribute(REQUEST_BEGIN_ATTRIBUTE);
```
dain: are you concerned about NPE due to out of order callbacks?

http-server/src/main/java/io/airlift/http/server/HttpServerChannelListener.java
```diff
@@ -0,0 +1,129 @@
...
+    public void onResponseBegin(Request request)
+    {
+        request.setAttribute(RESPONSE_CONTENT_TIMESTAMPS_ATTRIBUTE, new ArrayList<Long>());
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public void onResponseContent(Request request, ByteBuffer content)
+    {
+        List<Long> contentTimestamps = (List<Long>) request.getAttribute(RESPONSE_CONTENT_TIMESTAMPS_ATTRIBUTE);
```
dain: are you concerned about this growing unbounded?

http-server/src/main/java/io/airlift/http/server/HttpServerChannelListener.java
```diff
@@ -0,0 +1,129 @@
...
+        // In this case we return a DoubleSummaryStats containing only the value 0.
+        if (contentTimestamps.size() == 1) {
+            statistics.accept(0);
+            return new DoubleSummaryStats(statistics);
+        }
+
+        long previousTimestamp = contentTimestamps.get(0);
+        for (int i = 1; i < contentTimestamps.size(); i++) {
+            long timestamp = contentTimestamps.get(i);
+            statistics.accept(NANOSECONDS.toMillis(timestamp - previousTimestamp));
```
dain: what would happen if we got a negative duration due to a bad nanoTime implementation (we had this happen in the past)

nezihyigitbasi: It will skew the stats. For now I am leaving this as is, since I don't think the likelihood of that happening is high.

http-server/src/main/java/io/airlift/http/server/HttpServerChannelListener.java
nezihyigitbasi: Also never seen this becoming an issue in my investigations. We log the number of samples we have (`DoubleSummaryStats.count`) so we will see how large it may get.

http-server/src/main/java/io/airlift/http/server/HttpServerChannelListener.java
nezihyigitbasi: I have never seen this during my previous investigations where I used this listener. I will get more info about this.

http-server/src/main/java/io/airlift/http/server/HttpLogLayout.java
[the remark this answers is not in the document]
nezihyigitbasi: Those curlies are not in the format, I just used it to group those two values. I will update the commit body.

http-server/src/main/java/io/airlift/http/server/HttpLogLayout.java
nezihyigitbasi: Appended the new columns.

http-server/src/main/java/io/airlift/http/server/HttpServerChannelListener.java
sbordet: `onBeforeDispatch()` is guaranteed to happen _after_ `onRequestBegin()`.

http-server/src/main/java/io/airlift/http/server/DoubleSummaryStats.java
```diff
@@ -0,0 +1,56 @@
...
+package io.airlift.http.server;
+
+import io.airlift.event.client.EventField;
+import io.airlift.event.client.EventType;
+
+import java.util.DoubleSummaryStatistics;
+
+import static java.util.Objects.requireNonNull;
+
+@EventType
```
martint: Why does this need to be tagged with `@EventType`?

http-server/src/main/java/io/airlift/http/server/HttpLogLayout.java
```diff
@@ -37,6 +46,16 @@ public String doLayout(HttpRequestEvent event)
...
+                .append(event.getBeginToDispatchMillis())
+                .append('\t')
+                .append(event.getBeginToEndMillis())
+                .append('\t')
+                .append(event.getFirstToLastContentTimeInMillis())
+                .append('\t')
+                .append(contentInterarrivalStats)
```
martint: What are the units for the inter-arrival stats? It's not clear from looking at this method or the summary object.

http-server/src/main/java/io/airlift/http/server/HttpServerChannelListener.java
```diff
@@ -0,0 +1,129 @@
...
+        requireNonNull(contentTimestamps, "contentTimestamps is null");
+
+        // no content (HTTP 204)
+        if (contentTimestamps.size() == 0) {
+            return null;
+        }
+
+        DoubleSummaryStatistics statistics = new DoubleSummaryStatistics();
+
+        // If there was a single response chunk there won't be any interarrival times.
```
martint: Wouldn't it be more appropriate to have no inter-arrival stats in this case, then?

http-server/src/main/java/io/airlift/http/server/DoubleSummaryStats.java
nezihyigitbasi: Because, otherwise I get the below error. Is there any other way around this?

```
 Event class [class io.airlift.http.server.HttpRequestEvent] EventField method [public io.airlift.http.server.DoubleSummaryStats io.airlift.http.server.HttpRequestEvent.getResponseContentInterarrivalStats()] return type [class io.airlift.http.server.DoubleSummaryStats] is not supported
  at io.airlift.http.server.HttpServerModule.configure(HttpServerModule.java:76)
```

http-server/src/main/java/io/airlift/http/server/HttpLogLayout.java
nezihyigitbasi: It's in ms, please see `HttpServerChannelListener::processContentTimestamps`. I will add a comment.

http-server/src/main/java/io/airlift/http/server/HttpServerChannelListener.java
nezihyigitbasi: Makes sense I think. Will update.

PR #641, opened 2018-05-24
 Add additional columns to HTTP client logs

nezihyigitbasi:
I also added some code to get a client dump on timeouts (when `DEBUG` is enabled for the `JettyClientDiagnostics` class), which will help us with HTTP/2 stuff. This is a change for a particular exception, so it is a very specific change. If you have ideas about how to make it better and more generic please let me know.


http-client/src/main/java/io/airlift/http/client/jetty/JettyClientDiagnostics.java
```diff
@@ -0,0 +1,46 @@
...
+import javax.annotation.concurrent.ThreadSafe;
+
+import static java.util.Objects.requireNonNull;
+
+@ThreadSafe
+public class JettyClientDiagnostics
+{
+    private static final Logger log = Logger.get(JettyClientDiagnostics.class);
+
+    private final HttpClient httpClient;
```
martint: This doesn't need to be a field. You can just pass it into the logDiagnosticsInfo method.

If you do that, the object can be created where it's declared in JettyHttpClient and, in theory, could be shared for all clients.

http-client/src/main/java/io/airlift/http/client/jetty/JettyClientDiagnostics.java
```diff
@@ -22,20 +22,20 @@
 import static java.util.Objects.requireNonNull;
 
 @ThreadSafe
-public class JettyClientDiagnostics
```
martint: Nothing in this commit seems to depend on this. Do this in the commit that adds this class instead.

http-client/src/main/java/io/airlift/http/client/jetty/JettyClientDiagnostics.java
```diff
@@ -22,20 +22,20 @@
...
+class JettyClientDiagnostics
 {
     private static final Logger log = Logger.get(JettyClientDiagnostics.class);
 
     private final HttpClient httpClient;
-    // log information at most once per 10s
+    // log at most once per 10s
     private final RateLimiter rateLimiter = RateLimiter.create(0.1);
 
-    public JettyClientDiagnostics(HttpClient httpClient)
```
martint: Why?

http-client/src/main/java/io/airlift/http/client/jetty/JettyClientDiagnostics.java
```diff
@@ -22,20 +22,20 @@
...
+    JettyClientDiagnostics(HttpClient httpClient)
     {
         this.httpClient = requireNonNull(httpClient, "httpClient is null");
     }
 
-    public void logDiagnosticsInfo()
+    void logDiagnosticsInfo()
```
martint: Why?

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -391,7 +394,11 @@ public void onContent(Response response, ByteBuffer content)
             }
         };
 
-        RequestInfo requestInfo = RequestInfo.from(jettyRequest, System.currentTimeMillis());
+        long requestTimestamp = System.currentTimeMillis();
```
martint: System.currentTimeMillis() is sensitive to clock adjustments. We may want to use System.nanoTime() if all we're doing is measure durations.

nezihyigitbasi: Other than the request timestamp field we use other fields to measure durations, so I will update them to use `nanoTime()`.

http-client/src/main/java/io/airlift/http/client/jetty/HttpRequestEvent.java
```diff
@@ -133,6 +173,16 @@ static HttpRequestEvent createHttpRequestEvent(RequestInfo requestInfo, Response
             responseCode = response.get().getStatus();
         }
 
+        long requestTotalTime = requestInfo.getRequestQueueTime() + (responseInfo.getResponseCompleteTimestamp() - requestInfo.getRequestBeginTimestamp());
```
martint: What's the unit for these numbers?

nezihyigitbasi: I renamed these variables to clarify the units. I also found a bug which was introduced when I changed several places to use System.nanoTime() instead of System.currentTimeMillis() based on your previous comment. Will add a fixup shortly.

martint: That’s why I asked. I suspected there was a bug :)

PR #642, opened 2018-05-25
Change HTTP PEM support to fail on bad PEM files

dain:
Instead of falling back to guessing the file is a Java key store, check
if the file contains any PEM keys or certificates and either load them
or fail with a clean error message.

### DIFF
--- a/security/src/main/java/io/airlift/security/pem/PemReader.java
+++ b/security/src/main/java/io/airlift/security/pem/PemReader.java
@@ -70,6 +70,19 @@ public final class PemReader
 
     private PemReader() {}
 
+    public static boolean isPem(File file)
+            throws IOException
+    {
+        return isPem(asCharSource(file, US_ASCII).read());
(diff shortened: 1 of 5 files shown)


security/src/main/java/io/airlift/security/pem/PemReader.java
```diff
@@ -70,6 +70,26 @@
...
 
+    public static boolean isPem(File file)
+            throws IOException
+    {
+        return isPem(asCharSource(file, US_ASCII).read());
+    }
+
+    public static boolean isPem(String data)
+    {
+        if (CERT_PATTERN.matcher(data).find()) {
```
electrum: Why not
```java
return CERT_PATTERN.matcher(data).find() ||
        PUBLIC_KEY_PATTERN.matcher(data).find() ||
        PRIVATE_KEY_PATTERN.matcher(data).find();

PR #643, opened 2018-06-01
Increase default HTTP client selector count

nezihyigitbasi:
A larger number of selectors will help HTTP clients serve more traffic
by default.


electrum:
What are the downsides to doing this? Does it increase resource usage?

nezihyigitbasi:
Presto test here: https://github.com/prestodb/presto/pull/10743

nezihyigitbasi:
Presto tests look good.

nezihyigitbasi:
Will set this config elsewhere as discussed offline.

PR #646, opened 2018-06-12
Preserve User-Agent and Authorization when following redirects

electrum:


electrum:
Yes, the credentials will be sent to the target of the redirect, so the client needs to trust the server in that regard. However, if the server isn’t trustworthy, you’re already sunk, since you just sent it your credentials. And that server doesn’t need to redirect the client — it could send the credentials itself.

martint:
What if the server is trustworthy but can send redirects to any random site on the internet  (e.g., a url shortener service)?

http-client/src/main/java/io/airlift/http/client/Request.java
```diff
@@ -30,14 +30,26 @@
...
-public class Request
+public final class Request
 {
     private final URI uri;
     private final String method;
     private final ListMultimap<String, String> headers;
     private final BodyGenerator bodyGenerator;
+    private final boolean followRedirects;
 
+    @Deprecated
```
dain: Add message telling people to use the builder.

PR #647, opened 2018-06-12
Periodically flush HTTP server and HTTP client logs

electrum:
The large buffer used for logs requires logging many messages before the
buffer fills up and writes the logs to disk. Flushing periodically makes
debugging low traffic servers and clients easier, and reduces the chance
of log messages getting lost on an unclean shutdown.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/jetty/DefaultHttpClientLogger.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/DefaultHttpClientLogger.java
@@ -18,13 +18,18 @@ import ch.qos.logback.core.ContextBase;
 import ch.qos.logback.core.rolling.RollingFileAppender;
 import ch.qos.logback.core.rolling.SizeAndTimeBasedFNATP;
 import ch.qos.logback.core.rolling.TimeBasedRollingPolicy;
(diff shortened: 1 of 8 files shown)


nezihyigitbasi:
Why can't we just set `http-client.log.buffer-size` to a small value for such cases?

http-client/src/main/java/io/airlift/http/client/jetty/DefaultHttpClientLogger.java
```diff
@@ -118,4 +130,46 @@ private static void recoverTempFiles(String logPath)
...
+
+            long now = System.nanoTime();
+            long last = lastFlushed.get();
+            if (((now - last) > flushIntervalNanos) && lastFlushed.compareAndSet(last, now)) {
+                flush();
+            }
+        }
+
+        @SuppressWarnings("Duplicates")
+        private void flush()
```
dain: I feel like we are going to be getting extra flushes, since we don't know when the super class flushes.

electrum:
We default to 1MB which is ridiculously large for most people. Even the default of 8KB could be 50-100 log entries before a flush. This allows us to have both performance and good behavior out of the box.

PR #663, opened 2018-08-15
Update to Jetty 9.4.12.RC2

nezihyigitbasi:


nezihyigitbasi:
@dain Has this test ever been flaky?

```
[ERROR] testKeepAlive(io.airlift.http.client.jetty.TestJettyHttpsClientSpengo)  Time elapsed: 2.036 s  <<< FAILURE!
java.lang.AssertionError: expected [50220] but found [50222]
```

dain:
@nezihyigitbasi I don't think so

dain:
looking at the test, I think something is broken.  The test is verifying that we use the same connection for each request by checking the remote port on the server side.  It looks like we are not reusing connections.

nezihyigitbasi:
Yes, seems like the http client is opening a new connection, and we end up with a new client port. @sbordet any ideas?

sbordet:
Cloned airlift, updated main POM to reference 9.4.12.RC2, ran the test multiple times and it does not fail for me. Using JDK 10.0.2, you?

The way the test is written there should be plenty of time between the 3 requests to make it use the same connection.

The only way the test can fail is if the first request takes > 1 second to establish the connection (where "establish connection" means TCP SYN + TLS handshake + HTTP/2 preface), and the second request is sent concurrently with the first; this will trigger the opening of a second connection since the first is not established yet.

What changed recently was that when establishing the connection, we wait for the server to respond to the HTTP/2 preface, while before the client did not wait to receive the server preface. This was necessary to avoid the max concurrent streams exceeded, as during connection establishment the client and the server may not agree on what is the number of concurrent streams (the client needs to wait for the server to communicate that number).

sbordet:
Spoke too soon, HTTP/2 was not enabled in airlift.

nezihyigitbasi:
The test run on travis uses `1.8.0_151`. Locally I cannot also reproduce it.

For this particular test I don't think HTTP/2 is not enabled.

sbordet:
Tried HTTP/2 as well, cannot make it fail.
Do you have a reproducible failure in Travis? If so, can you enable DEBUG logging?

nezihyigitbasi:
I enabled Jetty DEBUG logs, will re-run the tests until we can repro it.

nezihyigitbasi:
@sbordet When I enable DEBUG for the package `org.eclipse.jetty` I cannot reproduce this issue, I have run many tens of times, no luck. However, I was able to reproduce the issue with DEBUG for `org.eclipse.jetty.client`, I hope it has some useful info: https://gist.github.com/nezihyigitbasi/655d7425ef89cf2eedd29047197d7f98  (In this log you will see that I only run the failing test 20 times, and all the other tests are disabled).

When DEBUG is enabled for `org.eclipse.jetty` I hit another issue: https://gist.github.com/nezihyigitbasi/75dbfb5e4567930c4576d3ab0fabc2d8

sbordet:
@nezihyigitbasi I'm looking at this.

joakime:
800 builds of airlift with Jetty 9.4.12.RC2 (so far) and no test failures. (linux, amd64, 24 core machine, ubuntu 16.04, maven 3.5.2, java 8u181 / 64-bit)

joakime:
We have been able to reproduce this a few times on our own Jenkins system with full debug (woo 1.3GB of logs!).

Analysis is pointing to a strange corner case related to overloaded system behavior.

Also, at the point in time that it does fail, there's about 700 HttpDestinations being tracked by the HttpClient in your testcase (one for each prior testcase that uses HttpClient).
Perhaps the HttpClient should be stopped/started for each TestCase class to be sure that you don't have actions from other Testcases impacting your current testcase?

dain:
@nezihyigitbasi should we close this one for now?

nezihyigitbasi:
Yep, let's close this for now. Once the new build is out I will create a new PR.

PR #667, opened 2018-09-02
Update to Jetty 9.4.12.v20180830

nezihyigitbasi:


joakime:
The test cipher ...

https://github.com/airlift/airlift/blob/bd1dffa179fa386b026d53ba476c39b861268dea/http-server/src/test/java/io/airlift/http/server/TestHttpServerCipher.java#L45

Is excluded by Jetty in 9.4.12.

Note: excludes always win over includes.

See:

* Issue eclipse/jetty.project#2807 - Update default excludes for Jetty
* Commit eclipse/jetty.project@5e07592a692e7400cd641e608decd8e0c942872d - for new exclusion list

nezihyigitbasi:
@joakime thank you. Since `^TLS_RSA_.*$` is excluded in the latest Jetty version I updated the test `CIPHER_1` to `TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384`.

nezihyigitbasi:
I will follow up with @joakime about the implications of this change on our production config.

nezihyigitbasi:
@dain Please review when you have a chance.

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -291,6 +291,9 @@ public HttpServer(HttpServerInfo httpServerInfo,
                 }
                 sslContextFactory.setSecureRandomAlgorithm(config.getSecureRandomAlgorithm());
                 sslContextFactory.setWantClientAuth(true);
+                // Since Jetty 9.4.12 the list of excluded cipher suites includes "^TLS_RSA_.*$" by default.
+                // We reset that list here to enable use of those cipher suites.
+                sslContextFactory.setExcludeCipherSuites();
```
nezihyigitbasi: This updates the `sslContextFactory` for the `adminConnector`. The `sslContextFactory` for the `httpsConnector` is already updated from the config on L238, so we don't need to change that.

dain: I think it would make more sense to add the exclude/include cipher calls L238 here.  My guess is we just missed the admin connector when adding that code.

haozhun:
After reading jetty docs at https://www.eclipse.org/jetty/documentation/9.4.x/configuring-ssl.html

My understanding is that, to enable the non-FS ciphers, we need to call `setExcludeCipherSuites`. Doing this would override the default exclusions. After overriding the default exclusion, we can set the inclusions to our whitelist.

The way the code is currently written today, we always override both include and exclude. As a result, our production set up will happen to work.

https://github.com/airlift/airlift/blob/bd1dffa179fa386b026d53ba476c39b861268dea/http-server/src/main/java/io/airlift/http/server/HttpServer.java#L236

It was not our intention to override the default unconditionally. However, it was natural to believe that calling `setExcludeCipherSuites` with an empty list has no effect.

We probably want to change the default of our config parameter to match Jetty's default exclusion list, add test to make sure they are always in sync, and document the effect of setting the config property. (Note that doing so will lead to production breakage if we don't change our deployment configuration.)

The issue with the test is test-specific. We didn't call `setExcludeCipherSuites` there. https://github.com/airlift/airlift/blob/bd1dffa179fa386b026d53ba476c39b861268dea/http-server/src/test/java/io/airlift/http/server/TestHttpServerCipher.java#L184

nezihyigitbasi:
@dain @haozhun added a new commit to update the default value of `http-server.https.excluded-cipher`.

CHANGES
```diff
@@ -1,5 +1,7 @@
 0.173
 - Update to Jetty 9.4.12.v20180830.
+- Update the default value of the "http-server.https.excluded-cipher" config
+  property to match Jetty's default excluded cipher suites.
```
haozhun: I think we should be more explicit here about the implications: 

All non-FS ciphers are now excluded by default. In particular, this means two things: 

1. All ciphers that uses RSA as the key exchange algorithm
2. Clients are required to support TLS1.2. Those using TLS1.0 or TLS1.1 will not establish a connection successfully.

nezihyigitbasi:
Seems like @dain has merged this. @dain Can you please release a new version?

PR #674, opened 2018-10-10
Add config to override default cipher suites for HTTP client

shixuan-fan:


http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -180,6 +180,10 @@ public JettyHttpClient(
             }
         }
         sslContextFactory.setSecureRandomAlgorithm(config.getSecureRandomAlgorithm());
+        List<String> includedCipherSuites = config.getHttpsIncludedCipherSuites();
+        List<String> excludedCipherSuites = config.getHttpsExcludedCipherSuites();
+        sslContextFactory.setIncludeCipherSuites(includedCipherSuites.toArray(new String[includedCipherSuites.size()]));
```
nezihyigitbasi: Recommended way for collection to array conversion is to use zero sized array: `includedCipherSuites.toArray(new String[0])`.

dain: Is that the convention we are using now.  If I remember there was an discussion about this, but I'm not sure how it landed.  I have been using the size style above, but we could also use the stream variant.  Here are the style I know about:
```java
        foo.toArray(new String[0]);
        foo.toArray(new String[foo.size()]);
        foo.stream().toArray(String[]::new);
```
@martint and @electrum thoughts?

nezihyigitbasi: I don't know whether this is our convention, but this is recommended for mostly performance reasons since Java 6. Intellij also recommends this. There are some nice pointers here: https://github.com/naver/pinpoint/issues/4138

shixuan-fan: I personally don't have much insights. Intellij would suggest me using `foo.toArray(new String[0])` when I tried the other two methods. If we decide to change this, we probably also want to change the one in `HttpServer`.

martint: Yes, I've been using the form @nezihyigitbasi suggested for the reasons listed in the issue he linked and we should be doing that. 

The stream form is reasonable when you have to do something more than just turn the collection into an array (e.g., filter/transform), but it's too roundabout for the simple case and probably doesn't get the performance benefits from the zero-sized array variant.

electrum: I agree with @martint. The steam way is clever and would seem to avoid object allocation, but you have to crest the steam anyway. Plus the zero sized array version is optimized by the JIT per the link above.

shixuan-fan:
Comment addressed. Please take another look and feel free to merge.

I did notice that in `HttpServer` we could also change the style
(https://github.com/airlift/airlift/blob/f99e0c08d14f5495a0aa5a60b14bf876c3a9f9c1/http-server/src/main/java/io/airlift/http/server/HttpServer.java#L238-L239). But I'm not sure if it should be a separate commit in this PR, or a separate PR. Any guidance would be appreciated :)

CHANGES
```diff
@@ -1,3 +1,7 @@
+0.174
+- Add configuration property to include and exclude HTTPS cipher suites for
```
nezihyigitbasi: `Add configuration properties to ... for the HTTP client.` (note the period at the end).

I don't know how comprehensive we want to be here, but we can also tell the config property names. 

/cc @electrum

nezihyigitbasi:
@shixuan-fan you can create a separate PR for that.

shixuan-fan:
Comment addressed

nezihyigitbasi:
Merged with minor doc changes, thank you @shixuan-fan.

PR #677, opened 2018-10-11
Use zero-sized array for array conversion in HttpServer

shixuan-fan:


nezihyigitbasi:
Merged with a minor change (added more info to the commit message detail), thanks @shixuan-fan.

PR #683, opened 2018-10-29
 Allow application to provide predefined log configuration

findepi:


findepi:
cc @electrum @dain @kokosing

findepi:
I am yet to verify this actually would work with Presto as expected. I want to know your opinion about general idea first.

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -201,6 +217,8 @@ public void configure(LoggingConfiguration config)
             disableConsole();
         }
 
+        setBuiltInLevels();
```
kokosing: shouldn't built in levels be loaded before user log levels?

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -128,12 +129,27 @@ public void setRootLevel(Level newLevel)
         setLevel(ROOT_LOGGER_NAME, newLevel);
     }
 
+    private void setBuiltInLevels()
```
kokosing: move it under the usage

log-manager/src/main/java/io/airlift/log/Logging.java
findepi: They should. Aren't they?

log-manager/src/main/java/io/airlift/log/Logging.java
findepi: Please see the whole file, not just the diff. This was IMO the most logical place.

electrum:
For the Hadoop log levels in Presto, @martint and I discussed adding an API for connectors to set the default levels.

Loading from a resource won’t work for Presto since we want to set defaults inside a plugin, but logging is configured in the main server. We wouldn’t want to ship plugin-specific defaults in the main server.

electrum:
Do we still want the first commit?

findepi:
I forgot about the first commit... Do you want it?

electrum:
Yes, let's fix it. It seems unlikely to matter in practice, but we shouldn't ever depend on the platform encoding.

findepi:
@electrum https://github.com/airlift/airlift/pull/686

PR #686, opened 2018-10-29
Fix platform-dependent loading of log.properties

findepi:


dain:
This change seems correct to me, but I'd like to seem a confirmation from @electrum or @martint.  Also, @findepi, can you check if we made similar mistakes anywhere-else in Airlift?

findepi:
@dain good point. I found one more usage of `FileReader` (fixed)
Once upon a time, i was using Findbugs's http://findbugs.sourceforge.net/bugDescriptions.html#DM_DEFAULT_ENCODING . Not sure how would that apply to airlift and nowdays.

findepi:
for the record, default encoding checks is tracked for Presto @ https://github.com/prestodb/presto/issues/9278

electrum:
We have SpotBugs (FindBugs successor) in Airbase, but haven't done the work to enable it for Airlift yet. It is enabled for other projects like Slice and Drift.

findepi:
Are you going to enable it for Presto as well?

electrum:
I have no current plans to do it, personally, but would be happy to review if anyone has time to do it. Enabling it is a combination of

1) Fix found issues
2) Add targeted suppressions
3) Disable default rules that aren’t suitable (too many suppressions relative to the value of the check)

PR #687, opened 2018-11-04
Stop http-request log on server shutdown

johngmyers:


nezihyigitbasi:
Looks like we have missed this. Merged, thanks!

PR #688, opened 2018-11-07
Make request and response HttpClient buffers configurable

sopel39:
When using GSS requests might exceed request buffer size.
Therefore those buffers needs to be configurable.


http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -218,6 +220,34 @@ public HttpClientConfig setMaxContentLength(DataSize maxContentLength)
...
     }
 
+    @NotNull
+    public DataSize getRequestBufferSize()
+    {
+        return requestBufferSize;
+    }
+
+    @Config("http-client.request-buffer-size")
+    @MaxDataSize("32MB")
```
electrum: Bean validations need to go on the getter (the validation framework validates the entire object after it is built)

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -218,6 +220,34 @@ public HttpClientConfig setMaxContentLength(DataSize maxContentLength)
...
+    }
+
+    @NotNull
+    public DataSize getResponseBufferSize()
+    {
+        return responseBufferSize;
+    }
+
+    @Config("http-client.response-buffer-size")
+    @MaxDataSize("32MB")
```
electrum: Same here

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -200,6 +200,10 @@ public JettyHttpClient(
 
         httpClient = new AuthorizationPreservingHttpClient(transport, sslContextFactory);
 
+        // request and response buffer size
+        httpClient.setRequestBufferSize((int) config.getRequestBufferSize().toBytes());
```
electrum: Use `toIntExact` instead of a cast

sopel39:
ac

PR #690, opened 2018-11-09
Install Jackson ParameterNamesModule module by default

findepi:


findepi:
cc @electrum

findepi:
@tdcmeehan indeed, the problem occurs when class has: `@JsonValue` and single-arg static `@JsonCreator`.
The question is, shouldn't we upgrade straight to newest jackson (2.9.7)? (https://github.com/airlift/airlift/pull/692)

PR #692, opened 2018-11-10
Upgrade to Jackson 2.9.7

findepi:
Upgrade to Jackson 2.9.7 to resolve deserialization failure when class
has `@JsonValue` and single-argument static factory `@JsonCreator`
method.


electrum:
We should put the dependency change in Airbase. Otherwise, dependent projects like Presto that also use Airbase will use the wrong version.

Can you put up a PR for Airbase and I’ll get everything merged and released later today?

findepi:
https://github.com/airlift/airbase/pull/166

findepi:
ac. I made airlift depend on yet-to-be-released airbase 88, so travis cannot pass until after airbase release

PR #693, opened 2018-11-12
Fix JSON serialization of null map values for Jackson 2.9

electrum:
See https://github.com/FasterXML/jackson-databind/issues/1812

### DIFF
--- a/json/src/main/java/io/airlift/json/ObjectMapperProvider.java
+++ b/json/src/main/java/io/airlift/json/ObjectMapperProvider.java
@@ -117,8 +117,9 @@ public class ObjectMapperProvider
         // use ISO dates
         objectMapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
 
-        // skip fields that are null instead of writing an explicit json null value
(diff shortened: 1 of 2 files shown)


json/src/main/java/io/airlift/json/ObjectMapperProvider.java
```diff
@@ -118,7 +118,7 @@ public ObjectMapper get()
         objectMapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
 
         // skip fields that are null instead of writing an explicit json null value
```
dain: Maybe mention that this only applies to JSON object fields and specifically not maps or arrays.

json/src/test/java/io/airlift/json/TestJsonCodec.java
```diff
@@ -95,6 +97,18 @@ public void testTypeLiteralMap()
         Person.validatePersonMapJsonCodec(jsonCodec);
     }
 
+    @Test
+    public void testMapNullValues()
```
dain: Add a list test

PR #694, opened 2018-11-13
Fix extraction of level when deserializing qdigest

tdcmeehan:
`nodeStructure` bytes need to be treated as unsigned for the purposes
of level extraction.

I added a minimal test that will manifest when this is a problem (whenever the MSB of the level is flipped to 1).


tdcmeehan:
CC: @martint

stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -203,7 +204,7 @@ public QuantileDigest(Slice serialized)
             byte nodeStructure = input.readByte();
             boolean hasRight = (nodeStructure & Flags.HAS_RIGHT) != 0;
             boolean hasLeft = (nodeStructure & Flags.HAS_LEFT) != 0;
-            byte level = (byte) (nodeStructure >>> 2);
+            byte level = (byte) (toUnsignedInt(nodeStructure) >>> 2);
```
martint: Doing it this way makes it clear that it's extracting the topmost 6 bits:

```java
(nodeStructure >> 2) & 0b111111
```

PR #698, opened 2018-11-19
Add joda-to-java-time-bridge module

haozhun:
Mixed use of joda and java.time results in two different versions of
tzdata being used at the same time. This inconsistencies can often
lead to result that is incorrect in any tzdata version. For example,
operations as simple as casting a string representation of zoned
date time to unixtime and back will not roundtrip.

This modules make it possible to mix usage of joda and java.time
without the above problem. Implementation wise, it makes joda
delegate to java.time for tz rules.

It is expected to work with any reasonable version of joda.


electrum:
We should create a new project for this, as it’s not related to the main Airlift project.

I created a new one you can fork and send this PR to: https://github.com/airlift/joda-to-java-time-bridge

PR #699, opened 2018-11-26
Prevent ClassPathResourceHandler from using default encoding

findepi:


findepi:
@dain or @electrum

findepi:
@electrum ping

http-server/src/main/java/io/airlift/http/server/ClassPathResourceHandler.java
```diff
@@ -121,16 +119,7 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
                 return;
             }
 
-            // Send the content out. Lifted straight out of ResourceHandler.java
-            OutputStream out;
-            try {
-                out = response.getOutputStream();
-            }
-            catch (IllegalStateException e) {
```
martint: What would cause this exception to occur?

findepi: I _think_ this would be when someone already asked for response Writer (presumably to write something).

Whatever the reason of this could be, we're going down with `WriterOutputStream`, which cannot be correct -- the `WriterOutputStream` does not know how to decode bytes, so that the chars are re-encoded correctly when being sent to the client over the wire.

PR #700, opened 2018-11-27
Make file name pattern configurable

sopel39:
It's useful to make logs file name pattern configurable. This way we can adjust for various cloud logs agents, e.g: AWS CloudWatch Logs agent that uploads logs to the cloud

@electrum Could you review?


electrum:
I don’t think we should make this take the raw pattern, since it has to have a specific format. Do you have examples of the different formats required?

sopel39:
@electrum Formats like this: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/AgentReference.html

```What kinds of file rotations are supported?

    The following file rotation mechanisms are supported:

        Renaming existing log files with a numerical suffix, then re-creating the original empty log file. For example, /var/log/syslog.log is renamed /var/log/syslog.log.1. If /var/log/syslog.log.1 already exists from a previous rotation, it is renamed /var/log/syslog.log.2.

        Truncating the original log file in place after creating a copy. For example, /var/log/syslog.log is copied to /var/log/syslog.log.1 and /var/log/syslog.log is truncated. There might be data loss for this case, so be careful about using this file rotation mechanism.

        Creating a new file with a common pattern as the old one. For example, /var/log/syslog.log.2014-01-01 remains and /var/log/syslog.log.2014-01-02 is created.

```

sopel39:
@electrum Should we have an enum instead?

sopel39:
Specifically I'm thinking of supporting:
`Creating a new file with a common pattern as the old one. For example, /var/log/syslog.log.2014-01-01 remains and /var/log/syslog.log.2014-01-02 is created. `

electrum:
If I’m understanding this, we need an option to drop the .log.gz suffix?

findepi:
@sopel39 will that suffice to capture logs before the log file is rotated? 
i.e. if, say, last 1h of logs is in `/var/log/server.log` will Cloud Watch have logs with 1h delay?

electrum:
I don’t know if you’re trying to solve this for a specific environment, but I’ll mention that services are often run using a “supervisor” such as daemontools or runit that automatically handle restarts and capture stdout/stderr to rotates log files. Depending on what you’re doing, this can be a good alternative. This is done with “launcher run” rather than “launcher start”.

dain:
My concern is that file patters have meaning in logging systems.  Typically, the logging system reverse engineers the pattern to determine the requested rolling period (and sometimes desired compression).   I prefer a more straight forward design where we use configs to say what rules we want.

sopel39:
We could try to rotate stdout of presto-server, but that would be another (duplicate) log files. I preferred to simply have adjustable rotation pattern within presto itself. This could be useful for other people too.

Anyway, let me explore different options too. If we wanted to customize pattern within Presto itself, would pattern enum be OK?

sopel39:
> When you use the “run” mode, it logs to stdout/stderr rather than the log files.

Ah OK, but I was thinking about simply running presto-server as a system service. In such case, logs are not outputted to stdout. Maybe they could be handler somehow by `launcher`, but that feels like not a good direction

sopel39:
I will address it with custom handler

PR #701, opened 2018-12-07
Limit the size of http-request.log by default

findepi:
This brings default configuration for `http-request.log` in line with
configuration for `server.log`, making `http-request.log` less likely to
fill up the disk space when there is a lot to be logged.


findepi:
cc @electrum

PR #702, opened 2018-12-16
Do not adjust ThreadContextClassLoader loading bval

electrum:


martint:
What’s the motivation? I suspect that code was there for a reason.

PR #705, opened 2019-01-10
Add missing JSON annotations to Announcement

nezihyigitbasi:
This caused issues while running experiments with smile encoding that I am prototyping right now.

### DIFF
--- a/discovery/src/main/java/io/airlift/discovery/client/Announcement.java
+++ b/discovery/src/main/java/io/airlift/discovery/client/Announcement.java
@@ -15,6 +15,7 @@
  */
 package io.airlift.discovery.client;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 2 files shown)


discovery/src/main/java/io/airlift/discovery/client/Announcement.java
```diff
@@ -32,7 +33,13 @@
     private final String pool;
     private final Set<ServiceAnnouncement> services;
 
-    public Announcement(String environment, String nodeId, String pool, String location, Set<ServiceAnnouncement> services)
+    @JsonCreator
+    public Announcement(
+            @JsonProperty String environment,
```
electrum: I don't think we need `@JsonProperty` on the parameter names when using the Jackson parameter names module -- just having `@JsonCreator` should be sufficient. Alternatively, maybe we should add the parameter names to the annotations to be consistent with all the other classes in Airlift?

nezihyigitbasi: I added the parameter names for consistency.

nezihyigitbasi:
> Can you provide more details on the problem you ran into? I can't find any code that actually deserializes Announcement.

It has been a while that I hit this issue, and I don't have the same code to repro it. While experimenting with jackson-smile I was tweaking both the discovery server and airlift (and probably other libraries), and I remember switching all encoding to be in smile format at some point (for all services), and got an exception looking similar to the one below. 

Even though I cannot repro it now, it's still good to have these annotations in place.

```
Caused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `io.airlift.discovery.client.Announcement` (no Creators, like default construct, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
```

nezihyigitbasi:
Closing this one as this doesn't seem to be necessary. I probably messed up my environment while testing so I needed this at that time.

electrum:
Feel free to merge this if it turns out it is needed for anything.

PR #706, opened 2019-01-24
Add support for smile encoding

nezihyigitbasi:


nezihyigitbasi:
@electrum can you please take a look when you have time?

json/src/main/java/io/airlift/json/SmileCodecFactory.java
```diff
@@ -0,0 +1,155 @@
...
+    private final Provider<ObjectMapper> objectMapperProvider;
+    private final boolean prettyPrint;
+
+    public SmileCodecFactory()
+    {
+        this(new SmileObjectMapperProvider());
+    }
+
+    @Inject
+    public SmileCodecFactory(@Named("SMILE") Provider<ObjectMapper> smileObjectMapperProvider)
```
electrum: Add a `@ForSmile` annotation. We avoid using `@Named`.

json/src/main/java/io/airlift/json/SmileCodecFactory.java
```diff
@@ -0,0 +1,155 @@
...
+                .getType();
+
+        return new JsonCodec<>(createObjectMapper(), mapType);
+    }
+
+    private ObjectMapper createObjectMapper()
+    {
+        ObjectMapper objectMapper = null;
+
+        // Work around for http://code.google.com/p/google-guice/issues/detail?id=627
```
electrum: It looks like we don't need this anymore. Can you remove it from the other one in a commit before this one?

json/src/main/java/io/airlift/json/SmileCodecFactory.java
```diff
@@ -0,0 +1,155 @@
...
+        this(new SmileObjectMapperProvider());
+    }
+
+    @Inject
+    public SmileCodecFactory(@Named("SMILE") Provider<ObjectMapper> smileObjectMapperProvider)
+    {
+        this(smileObjectMapperProvider, false);
+    }
+
+    public SmileCodecFactory(Provider<ObjectMapper> smileObjectMapperProvider, boolean prettyPrint)
```
electrum: Does Smile actually have pretty print?

json/src/test/java/io/airlift/json/TestSmileCodecFactory.java
```diff
@@ -0,0 +1,199 @@
...
+    @Test
+    public void testImmutableTypeTokenMap()
+    {
+        JsonCodec<Map<String, ImmutablePerson>> smileCodec = codecFactory.smileCodec(new TypeToken<Map<String, ImmutablePerson>>() {});
+        validateImmutablePersonMapSmileCodec(smileCodec);
+    }
+
+    private void validatePersonListSmileCodec(JsonCodec<List<Person>> smileCodec)
+    {
+        ImmutableList<Person> expected = ImmutableList.of(
```
electrum: Nit: declare these as `List`, `Map`, etc.

http-client/src/main/java/io/airlift/http/client/FullJsonResponseHandler.java
```diff
@@ -63,7 +68,7 @@ private FullJsonResponseHandler(JsonCodec<T> jsonCodec)
     {
         byte[] bytes = readResponseBytes(response);
         String contentType = response.getHeader(CONTENT_TYPE);
-        if ((contentType == null) || !MediaType.parse(contentType).is(MEDIA_TYPE_JSON)) {
+        if ((contentType == null) || !SUPPORTED_MEDIA_TYPES.contains(MediaType.parse(contentType))) {
```
electrum: This doesn't seem right. The `JsonCodec` in this handler supports either JSON or Smile, but not both, so we shouldn't accept both. Also, the error handling logic and `getJson()` API (used by error handling) assumes JSON. It wouldn't work for Smile which is binary.

electrum (commented):
While supporting multiple encodings in `JsonCodec` seems attractive, the API methods `fromJson()` and `toJson()` take `String` which is only appropriate for JSON, not Smile (or other binary formats).

I think we need to think more about the APIs and how they are used in applications. Another consideration is that the choice of encoding between client and server should ideally be handled by content negotiation, but we don't have a "JSON HTTP client" abstraction in Airlift, so we can't control the request side.

nezihyigitbasi:
> I think we need to think more about the APIs and how they are used in applications.

Agreed. This initial patch is  an easy-way-out kind of patch to add beta support for smile encoding. I also thought about having separate classes like SmileCodec, SmileCodecFactory, etc., which may be a better way to handle it.

> Another consideration is that the choice of encoding between client and server should ideally be handled by content negotiation, but we don't have a "JSON HTTP client" abstraction in Airlift, so we can't control the request side.

Currently in my tests the http clients set the content type/accept headers themselves, and the server is also updated to support both json/smile.

johngmyers:
See also `SmileMapper`, `SmileBodyGenerator`, `SmileResponseHandler`, etc. in proofpoint/platform.

nezihyigitbasi:
I updated the PR @martint, please review when you have time. I also updated the title to be "WIP" as we need to sort out API details and I need to do more testing with all these latest changes that I did.

I created different classes for smile support and that cleaned up the code quite a bit. But, at the same time it made it a bit difficult to support both json and smile on the client side (say, based on a config flag) as we have separate `SmileCodec` and `JsonCodec` classes now, and the client needs to decide which codec it needs to use up front (e.g., what should be the type of `HttpRemoteTask.taskInfoCodec`?). Maybe we should have a base class/interface for these two codec classes, so the clients can just that type.

json/src/main/java/io/airlift/json/SmileCodec.java
```diff
@@ -0,0 +1,161 @@
...
+    /**
+     * Gets the type this codec supports.
+     */
+    public Type getType()
+    {
+        return type;
+    }
+
+    /**
+     * Coverts the specified smile bytes (UTF-8) into an instance of type T.
```
johngmyers: typo

nezihyigitbasi:
@martint I updated the PR with a fixup commit (I can squash if that's what you prefer):
- fix the typo
- add a `SmileBodyGenerator`
- add a `Codec` to make clients more flexible to be able to use both json and smile codecs (say, depending on some config). 

I tested this change with Presto (with a config that enables using smile encoding and with REST endpoints that support both json and smile), so it's no longer a WIP PR.

http-client/src/main/java/io/airlift/http/client/FullSmileResponseHandler.java
```diff
@@ -0,0 +1,211 @@
...
+import java.io.UncheckedIOException;
+import java.util.List;
+
+import static com.google.common.base.MoreObjects.toStringHelper;
+import static com.google.common.io.ByteStreams.toByteArray;
+import static com.google.common.net.HttpHeaders.CONTENT_TYPE;
+import static io.airlift.http.client.ResponseHandlerUtils.propagate;
+import static java.util.Objects.requireNonNull;
+
+public class FullSmileResponseHandler<T>
```
nezihyigitbasi: Modeled after `FullJsonResponseHandler`.

http-client/src/main/java/io/airlift/http/client/SmileBodyGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.http.client;
+
+import com.google.common.annotations.Beta;
+import io.airlift.json.SmileCodec;
+
+@Beta
+public class SmileBodyGenerator<T>
```
nezihyigitbasi: Modeled after `JsonBodyGenerator`.

json/src/main/java/io/airlift/json/SmileCodec.java
```diff
@@ -0,0 +1,162 @@
...
+import java.lang.reflect.Type;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+
+@Beta
+public class SmileCodec<T>
```
nezihyigitbasi: Modeled after `JsonCodec`.

json/src/main/java/io/airlift/json/SmileCodecBinder.java
```diff
@@ -0,0 +1,97 @@
...
+import com.google.inject.internal.MoreTypes.ParameterizedTypeImpl;
+
+import java.lang.reflect.Type;
+import java.util.List;
+import java.util.Map;
+
+import static java.util.Objects.requireNonNull;
+
+@Beta
+public class SmileCodecBinder
```
nezihyigitbasi: Modeled after `JsonCodecBinder`.

json/src/main/java/io/airlift/json/SmileCodecFactory.java
```diff
@@ -0,0 +1,119 @@
...
+import javax.inject.Provider;
+
+import java.lang.reflect.Type;
+import java.util.List;
+import java.util.Map;
+
+import static java.util.Objects.requireNonNull;
+
+@Beta
+public class SmileCodecFactory
```
nezihyigitbasi: Modeled after `JsonCodecFactory`.

json/src/main/java/io/airlift/json/SmileCodecProvider.java
```diff
@@ -0,0 +1,70 @@
...
+package io.airlift.json;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+import java.lang.reflect.Type;
+
+import static java.util.Objects.requireNonNull;
+
+class SmileCodecProvider
```
nezihyigitbasi: Modeled after `JsonCodecProvider`.

nezihyigitbasi (commented):
Per our offline discussion I commented on the classes that are modeled after the json counterparts.

electrum:
I merged the first commit that removes the Guava workaround.

json/src/main/java/io/airlift/json/Codec.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.json;
+
+public interface Codec<T>
```
arhimondr: If you had a method for deserialization here as well, (e.g.: ` public T fromJson(byte[] bytes)`  -> ` public T fromBytes(byte[] bytes)` ) you can nicely abstract a lot of stuff.

http-client/src/main/java/io/airlift/http/client/FullSmileResponseHandler.java
arhimondr: This can be `A___ResponseHandler`, that accepts `Codec`, and uses `Codec#fromBytes` to decode response. So it can be used for both, Json and Smile codec

http-client/src/main/java/io/airlift/http/client/ResponseHandler.java
```diff
@@ -25,4 +28,27 @@ T handleException(Request request, Exception exception)
 
     T handle(Request request, Response response)
             throws E;
+
+    interface BaseResponse<T>
```
arhimondr: With `A_ResponseHandler` approach this won't be needed.

http-client/src/main/java/io/airlift/http/client/SmileBodyGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+        extends StaticBodyGenerator
+{
+    public static <T> SmileBodyGenerator<T> smileBodyGenerator(SmileCodec<T> smileCodec, T instance)
+    {
+        return new SmileBodyGenerator<>(smileCodec, instance);
+    }
+
+    private SmileBodyGenerator(SmileCodec<T> smileCodec, T instance)
+    {
+        super(smileCodec.toBytes(instance));
```
arhimondr: Nice

http-client/src/test/java/io/airlift/http/client/TestFullSmileResponseHandler.java
```diff
@@ -0,0 +1,160 @@
...
+import static java.lang.String.format;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNotSame;
+import static org.testng.Assert.assertNull;
+import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
+
+public class TestFullSmileResponseHandler
```
arhimondr: This can also be abstracted into an `AbstractTestA__ResponseHandler`. The Codec creation can be abstracted behind factory, and specialized in the `TestResponseHandlerSmile` and `TestResponseHandlerJson`

json/src/main/java/io/airlift/json/SmileCodec.java
arhimondr: Do we need all of these static factories? I think it should be disencouraged. Usually we should use `Factory`.

@dain ? @electrum ?

json/src/main/java/io/airlift/json/SmileCodecFactory.java
arhimondr: It can also be abstracted

json/src/main/java/io/airlift/json/SmileCodecBinder.java
arhimondr: This can be abstracted

json/src/test/java/io/airlift/json/TestSmileCodecFactory.java
```diff
@@ -0,0 +1,199 @@
...
+import org.testng.annotations.Test;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.testng.Assert.assertEquals;
+
+public class TestSmileCodecFactory
```
arhimondr: This can be abstracted

json/src/test/java/io/airlift/json/TestSmileSupport.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static io.airlift.json.JsonBinder.jsonBinder;
+import static io.airlift.json.TestJsonModule.CAR;
+import static org.testng.Assert.assertEquals;
+
+public class TestSmileSupport
```
arhimondr: This can be abstracted

http-client/src/main/java/io/airlift/http/client/ResponseHandler.java
electrum: This interface seems rather specific to put in `ResponseHandler`. Maybe a top level `ValueResponse`?

json/src/main/java/io/airlift/json/Codec.java
electrum: Why not have a `T fromBytes(byte[] bytes)`  method as well? This would allow fully generic use of the codec.

nezihyigitbasi (commented):
I defined several abstract classes to pull up the common code and simplify the PR. Thanks @arhimondr.

PR #707, opened 2019-01-30
Allow custom content listener in HttpClient.executeAsync

mbasmanova:
Extend HttpClient.executeAsync to allow callers to provide custom content listener. We'd like to use this to modify memory allocation currently happening in BufferingResponseListener.allocateCurrentBuffer to use shared pull of buffers to reduce GC activity.

CC: @oerling @nezihyigitbasi @tdcmeehan @elonazoulay @yingsu00 

@electrum @dain David, Dain, do you think this would be a reasonable addition?

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClient.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClient.java
@@ -29,6 +29,8 @@ public interface HttpClient
 
     <T, E extends Exception> HttpResponseFuture<T> executeAsync(Request request, ResponseHandler<T, E> responseHandler);
 
+    <T, E extends Exception> HttpResponseFuture<T> executeAsync(Request request, ResponseHandler<T, E> responseHandler, ResponseListener responseListener);
+
     RequestStats getStats();
 
     long getMaxContentLength();
--- a/http-client/src/main/java/io/airlift/http/client/ResponseListener.java
+++ b/http-client/src/main/java/io/airlift/http/client/ResponseListener.java
@@ -1,0 +1,27 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.http.client;
+
+import com.google.common.annotations.Beta;
+
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+@Beta
+public interface ResponseListener
+{
+    void onContent(ByteBuffer content);
+
+    InputStream onComplete();
+}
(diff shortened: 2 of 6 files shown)


mbasmanova:
@nezihyigitbasi Nezih, thank you for the feedback. We are trying to accomplish 2 things here: (1) avoid extra allocations; (2) avoid unnecessary copying. (1) might be less important than (2). More specifically, currently we are allocating new byte arrays when receiving the response and copy that data when clients access it via InputStream interface. To avoid the copying, we need clients to have access to the list of byte arrays that hold the response. Adding `allocateResponseBuffer` method to `ResponseHandler` solves (1), but not (2). 

Having (1) and (2) combined allows for 6% CPU gain for Queen of the Night query on a test cluster using ORC files compressed with GZIP and 12% CPU gain on a desktop.

The query is:

```
select count(*), sum(l.extendedprice * (1 - l.discount) - l.quantity * p.supplycost) 
from lineitem_sf1000_unpartitioned l, partsupp_sf1000_unpartitioned p 
where l.partkey = p.partkey 
   and l.suppkey = p.suppkey
   and p.availqty < 1000
```

electrum:
This seems like the wrong approach. We should add a better async API that gets rid of InputStream, which is legacy from before we had async. InputStream is hard to use and makes no sense as we buffer the full response.

Note that as general philosophy, the HTTP client interface is supposed to be simple and high level. If you want full control over all the low level details, you can use the Jetty client directly.

dain:
If this change is just for a dev branch, I'd just experiment in a branch for now.

mbasmanova:
@dain @electrum Dain, David, thanks for the feedback.

> If this change is just for a dev branch, I'd just experiment in a branch for now.

Currently, this is only needed for a dev branch, but we expect to move this to master in the next couple of months. However, even using it for a dev branch is challenging because Travis fails to download airlift artifact.  Would it be possible to publish airlift snapshot artifact to allow Travis to succeed? I don't seem to have permissions.

> We should add a better async API that gets rid of InputStream, which is legacy from before we had async. InputStream is hard to use and makes no sense as we buffer the full response.

That makes sense. Do you have an idea of what you'd like this API to look like? How about the following?

```
public interface ResponseListener
{
    void onHeaders(ListMultimap<HeaderName, String> headers);
    void onResponseChunk(ByteBuffer chunk);
    void onCompletion(int statusCode, String statusMessage);
}

HttpResponseFuture<Void> executeAsync(Request request, ResponseListener responseListener);
```

> Note that as general philosophy, the HTTP client interface is supposed to be simple and high level. If you want full control over all the low level details, you can use the Jetty client directly.

martint:
>  However, even using it for a dev branch is challenging because Travis fails to download airlift artifact. Would it be possible to publish airlift snapshot artifact to allow Travis to succeed? 

We prefer not to have snapshots in maven that don't reflect code that's in trunk. You could achieve that by publishing a snapshot to your internal maven repo with a custom version string, instead.

mbasmanova:
@martint 

> You could achieve that by publishing a snapshot to your internal maven repo with a custom version string, instead.

I do this for local builds, but this doesn't work with Travis. Am I missing something?

martint:
Ah, that's right. Travis can't see your private maven repository. Another alternative would be to publish public artifacts under `com.facebook`.

mbasmanova:
> Another alternative would be to publish public artifacts under com.facebook.

That's an option, but we were hoping to avoid going that route as it involves quite a lot of changes. Given that we'd like to put these changes into master within the next couple of months, what would be the best path forward? David suggested to modify the async API to make it truly async. Should I make  PR for that? Does the API I proposed above makes sense?

mbasmanova:
@dain @electrum @martint *ping*

dain:
@mbasmanova I published the PR as `0.179-masha-SNAPSHOT` for you.

mbasmanova:
@dain Dain, thank you. It is helpful to have a version in Maven Central, but that's not quite exactly what I wanted. This snapshot didn't work as is. It turns out that I need to update other airlift components to avoid runtime errors due to mismatch versions of Guava. I'll work on that.

In the meantime, could you suggest how to make progress on implementing the changes we need in a way that would be acceptable to merge into master?

PR #709, opened 2019-02-05
Add future callback overloads with executor

dain:
Add overloaded versions of addSuccessCallback and addExceptionCallback that take an executor.

### DIFF
--- a/concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
+++ b/concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
@@ -19,6 +19,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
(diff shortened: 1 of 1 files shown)


concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
```diff
@@ -494,6 +495,11 @@ public void onFailure(Throwable t)
     }
 
     public static <T> void addSuccessCallback(ListenableFuture<T> future, Consumer<T> successCallback)
```
nezihyigitbasi: It would be nice to add some javadoc here that mentions that direct executor will be used here, and the consequences of that.

concurrent/src/main/java/io/airlift/concurrent/MoreFutures.java
```diff
@@ -509,17 +515,27 @@ public void onSuccess(@Nullable T result)
             @Override
             public void onFailure(Throwable t) {}
         };
-        Futures.addCallback(future, callback, directExecutor());
+        Futures.addCallback(future, callback, executor);
     }
 
     public static <T> void addSuccessCallback(ListenableFuture<T> future, Runnable successCallback)
```
nezihyigitbasi: ditto

PR #712, opened 2019-02-18
Report server port when bind fails

findepi:


electrum:
Can you update CHANGES?

findepi:
@electrum updated

PR #713, opened 2019-02-18
Retry SpnegoAuthentication when Network is unreachable

kokosing:
Retry SpnegoAuthentication when Network is unreachable

PR #714, opened 2019-02-22
Add RetryDriver

kokosing:


kokosing:
@findepi @electrum Would you like to take a look?

kokosing:
More or less it is a copy of `RetryDriver` from `presto-hive` connector with changes from #713.

kokosing:
@electrum @dain ping

electrum:
I don’t think we want to add this as a generic library. The code is messy and the API is not great, especially around exception handing. We probably want some Guice and configuration integration as well.

I’d prefer to look at existing alternatives such as https://github.com/jhalterman/failsafe

For the HTTP client, let’s have a private copy of the code for now. Exposing a public library with a good, stable API is a lot of work.

(This is my opinion. Others may disagree.)

kokosing:
Thanks for your opinion. That would mean that we could (should?) migrate `RetryDriver` in Presto to something else (like `failsafe` you mentioned).

electrum:
I just played around with `failsafe` and took a quick look through the code, and I'm impressed, it looks quite nice. I think it could easily replace `RetryDriver` (though I don't know if it's worth doing so).

PR #715, opened 2019-03-05
Add QuantileDigestIterator

### LINKED ISSUE #710: Add public method to iterate over quantile digest
The QuantileDigest class has a private method [`postOrderTraversal`](https://github.com/airlift/airlift/blob/master/stats/src/main/java/io/airlift/stats/QuantileDigest.java#L962) which is used to implement:

* quantile queries
* histogram 
* min
* max

It would be nice to have a method which is similar, but has public visibility, and doesn't have access into the internal state of the digests.  This method could be used to perform custom analysis which isn't supported out of the box.  A couple of examples come to mind:

* mean/truncated mean
* approximate values over/under quantile

I could also implement these directly in the class.  But I see it being useful and flexible to have a mechanism to iterate over a quantile digest, as one may want to iterate over, for example, a histogram.

tdcmeehan:
QuantileDigestIterator will iterate over the nodes with positive count
of a qdigest tree in forwards order (with the `iterator` method) or
reverse order (with the `reverseIterator` method).  It has the 
following accessors:

* upper bound value of the node
* lower bound value of the node
* count of the node
* cumulative count of the iteration
* lower bound quantile of the node
* upper bound quantile of the node

Iteration is performed with the `next` method, which indicates whether
further iteration is possible.  To reduce garbage, there is no 
`hasNext` and `next` differentiation, where `next` returns an object; 
iteration forward is performed with `next`, and once the iterator has 
moved forward then the accessors will retrieve values at the associated
position.

There's no fail-fast behavior for concurrent modification.

Fixes #710

CC @martint

### DIFF
--- a/stats/src/test/java/io/airlift/stats/TestQuantileDigest.java
+++ b/stats/src/test/java/io/airlift/stats/TestQuantileDigest.java
@@ -10,6 +10,7 @@ import org.testng.annotations.Test;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.NoSuchElementException;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


stats/src/test/java/io/airlift/stats/TestQuantileDigest.java
```diff
@@ -815,6 +816,187 @@ public void testSerializationRandom()
...
+
+    @Test
+    public void testNoElement()
+    {
+        QuantileDigest digest = new QuantileDigest(0.1);
+        QuantileDigest.QuantileDigestIterator iterator = digest.iterator();
+        assertFalse(iterator.next(), "Expected no calls to next() to return true");
+    }
+
+    @Test(expectedExceptions = NoSuchElementException.class)
```
electrum: This is dangerous because we don't know which method threw. It might have been one of the calls we expected to succeed. When you have a complex method, it's better to assert the throws directly:
```java
assertThatThrownBy(() -> iterator.upperBoundQuantile()).isInstanceOf(NoSuchElementException.class)

tanvach:
Hi, what is the progress of this PR? I'm hoping to see #12290 merged sometime.

PR #716, opened 2019-03-22
Add service principal pattern

luohao:


electrum:
Merged, thanks!

Now that we have this change, should we deprecate `remote-service-name`, since the pattern allows overriding that?

luohao:
`remote-service-name` is still used for passing  `serviceName`, maybe find a better name for it?

PR #717, opened 2019-03-25
Allow binding resource to path ending with '/'

findepi:


http-server/src/main/java/io/airlift/http/server/ClassPathResourceHandler.java
```diff
@@ -65,7 +65,6 @@ public ClassPathResourceHandler(String baseUri, String classPathResourceBase, Li
         requireNonNull(welcomeFiles, "welcomeFiles is null");
 
         baseUri = baseUri.startsWith("/") ? baseUri : '/' + baseUri;
-        baseUri = baseUri.endsWith("/") ? baseUri.substring(baseUri.length() - 1) : baseUri;
```
findepi: This line was not correct anyway (should be `baseUri.substring(0, baseUri.length() - 1)` rather than `baseUri.substring(baseUri.length() - 1)`).
It worked only for replacing "/" → "", but "/" is explicitly handled in https://github.com/airlift/airlift/blob/4c8e464e10576280b51721a4b40ad9606b06f6ec/http-server/src/main/java/io/airlift/http/server/ClassPathResourceHandler.java#L139

electrum:
Looks good to me, but can you add a test for this? There are existing tests in `io.airlift.http.server.TestHttpServerModule#testServer`

PR #721, opened 2019-04-09
Run life cycle methods on instances from providers

electrum:


dain:
@electrum there is a build error

PR #723, opened 2019-04-17
Add GSS name type configuration in http-client

luohao:
Support configurable GSS name type.

Ref: https://github.com/prestosql/presto/pull/645

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
@@ -33,6 +33,7 @@ import javax.validation.constraints.NotNull;
 
 import java.util.List;
 
+import static io.airlift.http.client.KerberosNameType.HOSTBASED_SERVICE;
 import static io.airlift.units.DataSize.Unit.GIGABYTE;
 import static io.airlift.units.DataSize.Unit.KILOBYTE;
 import static io.airlift.units.DataSize.Unit.MEGABYTE;
@@ -77,6 +78,7 @@ public class HttpClientConfig
     private String kerberosServicePrincipalPattern = "${SERVICE}@${HOST}";
     private String kerberosPrincipal;
     private String kerberosRemoteServiceName;
+    private KerberosNameType kerberosNameType = HOSTBASED_SERVICE;
     private int selectorCount = 2;
     private boolean recordRequestComplete = true;
     private boolean connectBlocking;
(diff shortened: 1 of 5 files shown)


http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -77,6 +77,7 @@
     private String kerberosServicePrincipalPattern = "${SERVICE}@${HOST}";
     private String kerberosPrincipal;
     private String kerberosRemoteServiceName;
+    private String kerberosNameType = "NT_HOSTBASED_SERVICE";
```
kokosing: please define an enum

findepi:
@luohao what's the relationship between this and https://github.com/airlift/airlift/pull/716?

luohao:
@findepi we have a ___non-standar___  krb5 setup, therefore, we need to make a few things configurable in the krb5 server/client, including the krb5 service principal name and the GSS name type.

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -407,6 +409,25 @@ public HttpClientConfig setKerberosRemoteServiceName(String serviceName)
         return this;
     }
 
+    public enum NameType
```
luohao: @electrum is there a common place where I can put this enum? It's defined in presto-main too, but doesn't seem appropriate to include `http-client` in krb5 authenticator.

electrum: No, it's better to have two copies

electrum: Let's name this `KerberosNameType` so that it's un-ambiguous and make it a top level class. It would be fine to reuse it in `presto-main` since it already depends on `http-client`.

http-client/src/main/java/io/airlift/http/client/spnego/SpnegoAuthentication.java
```diff
@@ -241,6 +253,18 @@ private static String canonicalizeServiceHostname(String hostName)
         }
     }
 
+    private static Oid getGSSNameType(HttpClientConfig.NameType nameType)
+    {
+        switch (nameType) {
+            case HOSTBASED_SERVICE:
+                return NT_HOSTBASED_SERVICE;
```
electrum: Make this part of the enum:
```java
public class KerberosNameType
{
    HOSTBASED_SERVICE(NT_HOSTBASED_SERVICE),
    USER_NAME(NT_USER_NAME);

    private final Oid oid;

    KerberosNameType(Oid oid)
    {
        this.oid = requireNonNull(oid, "oid is null");
    }

    public Oid getOid()
    {
        return oid;
    }
}

File: http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java [hunk 4]
```diff
@@ -407,6 +409,25 @@ public HttpClientConfig setKerberosRemoteServiceName(String serviceName)
...
+    }
+
+    public NameType getKerberosNameType()
+    {
+        return kerberosNameType;
+    }
+
+    @Config("http-client.authentication.krb5.name-type")
+    @ConfigDescription("Set kerberos GSS name type")
+    public HttpClientConfig setKerberosNameType(String nameType)
```
electrum: Make the config method take the enum directly:
```java
public HttpClientConfig setKerberosNameType(KerberosNameType nameType)
```
The config system will do the conversion and give a good error message on failure.

PR #725, opened 2019-04-22
Hide HTTP Client's key/trust store passwords from logs

findepi:


CHANGES
```diff
@@ -2,6 +2,7 @@
 
 - Run life cycle methods on instances from providers.
 - Log HTTP server requests even when body is not read.
+- Hide key/trust store passwords from logs.
```
electrum: We should say "HTTP client" here, since these were already hidden for the HTTP server

findepi:
@electrum thanks, updated

PR #731, opened 2019-05-08
Make HyperLogLog indexBitLength more efficient

jagill:
Previously, it calculated the Log2 of an integer `n` known to be a power of
two by `(int) Log(n)/Log(2)`, which is two slow log ops, a float
division, and a cast.  This replaces it with `Integer.numberOfTrailingZeros(n)`,
which is one int machine instruction.

### DIFF
--- a/stats/src/main/java/io/airlift/stats/cardinality/Utils.java
+++ b/stats/src/main/java/io/airlift/stats/cardinality/Utils.java
@@ -44,7 +44,8 @@ final class Utils
     public static int indexBitLength(int numberOfBuckets)
     {
         Preconditions.checkArgument(isPowerOf2(numberOfBuckets), "numberOfBuckets must be a power of 2, actual: %s", numberOfBuckets);
-        return (int) (Math.log(numberOfBuckets) / Math.log(2));
+        // 2**N has N trailing zeros, and we've asserted numberOfBuckets == 2**N
+        return Integer.numberOfTrailingZeros(numberOfBuckets);
     }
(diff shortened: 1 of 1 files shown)


jagill:
The test error "TestMBeanServerResource.testIsInstanceOf:127 » Socket Connection reset" seems both unrelated, and transitory.

jagill:
cc @martint I think you did the initial implementation, so I'd love to hear your thoughts!

stats/src/main/java/io/airlift/stats/cardinality/Utils.java
```diff
@@ -44,7 +44,9 @@ public static boolean isPowerOf2(long value)
     public static int indexBitLength(int numberOfBuckets)
     {
         Preconditions.checkArgument(isPowerOf2(numberOfBuckets), "numberOfBuckets must be a power of 2, actual: %s", numberOfBuckets);
-        return (int) (Math.log(numberOfBuckets) / Math.log(2));
+        // This is a quick way to calculate floor(log2(numberOfBuckets)) with just machine instructions
+        // q.v. https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#numberOfLeadingZeros(int)
+        return 31 - Integer.numberOfLeadingZeros(numberOfBuckets);
```
martint: Integer.numberOfTrailingZeros is an even simpler way of doing it.

martint:
Merged, thanks!

PR #733, opened 2019-05-12
Improve JMX agent handing on Java 9

electrum:


jmx/src/main/java/io/airlift/jmx/JmxAgent9.java
```diff
@@ -43,27 +44,71 @@
...
+        }
+
+        if ((existingServerPort != null) && (config.getRmiServerPort() != null) &&
+                (!existingServerPort.equals(config.getRmiServerPort()))) {
+            throw new RuntimeException(format("System property [%s] value [%s] does match configured RMI server port [%s]",
+                    JMX_SERVER_PORT, existingServerPort, config.getRmiServerPort()));
+        }
+
         int registryPort;
-        if (config.getRmiRegistryPort() == null) {
```
martint: It's a little hard to follow the flow of this if block now. I think it might be more readable to spell out the conditions and write it as follows:

```java
int registryPort;
if (existingRegistryPort != null && config.getRmiRegistryPort() == null) {
    registryPort = existingRegistryPort;
}
else if (existingRegistryPort == null && config.getRmiRegistryPort() != null) {
    registryPort = config.getRmiRegistryPort();
}
else {
    registryPort = NetUtils.findUnusedPort();
}
```

(or some other form using firstNonNull)

jmx/src/main/java/io/airlift/jmx/JmxAgent9.java
```diff
@@ -43,27 +44,71 @@
...
-        if (config.getRmiServerPort() != null) {
+        if (existingServerPort != null) {
+            serverPort = existingServerPort;
+        }
+        else if (config.getRmiServerPort() != null) {
             serverPort = config.getRmiServerPort();
         }
 
+        // this is how the JDK JMX agent constructs its URL
+        JMXServiceURL jmxUrl = new JMXServiceURL("rmi", null, registryPort);
```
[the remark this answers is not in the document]
electrum: This can fail for environmental reasons:
```java
if (host == null) {
    InetAddress local;
    try {
        local = InetAddress.getLocalHost();
    } catch (UnknownHostException e) {
        throw new MalformedURLException("Local host name unknown: " + e);
    }
    ...

File: jmx/src/main/java/io/airlift/jmx/JmxAgent9.java [hunk 3]
```diff
@@ -44,26 +46,71 @@ public JmxAgent9(JmxConfig config)
...
-            registryPort = NetUtils.findUnusedPort();
+        Integer existingRegistryPort = Integer.getInteger(JMX_REGISTRY_PORT);
+        Integer configuredRegistryPort = config.getRmiRegistryPort();
+        if (existingRegistryPort != null) {
+            if (configuredRegistryPort != null && !existingRegistryPort.equals(configuredRegistryPort)) {
+                throw new RuntimeException(format(
+                        "System property '%s=%s' does match configured RMI registry port %s",
+                        JMX_REGISTRY_PORT, existingRegistryPort, configuredRegistryPort));
+            }
+            if (existingRegistryPort == 0) {
```
martint: This compares an Integer with 0 potentially using identity equality, depending on whether 0 boxes to Integer or existingRegistryPort unboxes to its value.

PR #735, opened 2019-05-13
Closes #734 -- QuantileDigest.add gets long-double method

### LINKED ISSUE #734: QDigest add could have a (long, double) signature
Redirecting from https://github.com/prestosql/presto/issues/758

The current `(long, long)` signature converts the second argument immediately to `double` anyway; adding this second signature would directly allow more flexibility downstream to define `double`-weighted percentiles in Presto (and potentially other downstream users for other similar use cases):

https://github.com/airlift/airlift/blob/master/stats/src/main/java/io/airlift/stats/QuantileDigest.java#L260

```
    public void add(long value, long count)
    {
        checkArgument(count > 0, "count must be > 0");

        boolean needsCompression = false;
        double weight = count;
        if (alpha > 0.0) {
            long nowInSeconds = TimeUnit.NANOSECONDS.toSeconds(ticker.read());
            if (nowInSeconds - landmarkInSeconds >= RESCALE_THRESHOLD_SECONDS) {
                rescale(nowInSeconds);
                needsCompression = true; // rescale affects weights globally, so force compression
            }

            weight = weight(nowInSeconds) * count;
        }

        max = Math.max(max, value);
        min = Math.min(min, value);

        double previousCount = weightedCount;
        insert(longToBits(value), weight);

        // When totalCount crosses the next multiple of k (compression factor), the compression
        // equation changes for every node in the tree, so we need to compress globally.
        // Otherwise, only node along the insertion path are affected -- TODO: implement this.
        int compressionFactor = calculateCompressionFactor();
        if (needsCompression || ((long) previousCount) / compressionFactor != ((long) weightedCount) / compressionFactor) {
            compress();
        }
    }
```

MichaelChirico:
Sorry I didn't see any guide on specific protocols for contributing.

I'm in over my head a bit here so please feel free to suggest improvements, especially as regards adding tests for the new method.

### DIFF
--- a/stats/src/main/java/io/airlift/stats/QuantileDigest.java
+++ b/stats/src/main/java/io/airlift/stats/QuantileDigest.java
@@ -257,12 +257,11 @@ public class QuantileDigest
     /**
      * Adds a value to this digest. The value must be {@code >= 0}
      */
-    public void add(long value, long count)
+    public void add(long value, double weight)
     {
-        checkArgument(count > 0, "count must be > 0");
+        checkArgument(weight > 0, "weight must be > 0");
 
         boolean needsCompression = false;
-        double weight = count;
         if (alpha > 0.0) {
             long nowInSeconds = TimeUnit.NANOSECONDS.toSeconds(ticker.read());
             if (nowInSeconds - landmarkInSeconds >= RESCALE_THRESHOLD_SECONDS) {
@@ -270,7 +269,7 @@ public class QuantileDigest
                 needsCompression = true; // rescale affects weights globally, so force compression
             }
 
-            weight = weight(nowInSeconds) * count;
+            weight *= weight(nowInSeconds);
         }
 
         max = Math.max(max, value);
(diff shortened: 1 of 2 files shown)


stats/src/main/java/io/airlift/stats/QuantileDigest.java
```diff
@@ -257,12 +257,11 @@ public void add(long value)
     /**
      * Adds a value to this digest. The value must be {@code >= 0}
      */
-    public void add(long value, long count)
+    public void add(long value, double weight)
     {
         checkArgument(count > 0, "count must be > 0");
```
martint: There's a compilation failure. You'll need to rename this to `weight`, also.

Line 272 also needs to change from:

```java
weight = weight(nowInSeconds) * count;
```

to 

```java
weight *= weight(nowInSeconds);
```

[ERROR] /home/travis/build/airlift/airlift/stats/src/main/java/io/airlift/stats/QuantileDigest.java:[262,22] error: cannot find symbol
  symbol:   variable count
  location: class QuantileDigest
[ERROR] /home/travis/build/airlift/airlift/stats/src/main/java/io/airlift/stats/QuantileDigest.java:[272,44] error: cannot find symbol
  symbol:   variable count
  location: class QuantileDigest
```

MichaelChirico:
Shall I squash?

martint:
Yes, please do.

MichaelChirico:
OK done

martint:
I adjusted the commit message and merged it. Thanks!

PR #739, opened 2019-07-03
Handle PostConstruct and PreDestroy Exceptions in LifeCycleManager

pettyjamesm:
Handle catch and log exceptions thrown from `@PreDestroy` methods in`LifeCycleManager::stop()` so that other stop behaviors have a chance to run.

### DIFF
--- a/bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
@@ -27,7 +27,10 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
+
+import static java.lang.String.format;
 
 /**
  * Manages PostConstruct and PreDestroy life cycles
@@ -39,6 +42,16 @@ public final class LifeCycleManager
     private final Queue<Object> managedInstances = new ConcurrentLinkedQueue<>();
     private final LifeCycleMethodsMap methodsMap;
     private final AtomicReference<Thread> shutdownHook = new AtomicReference<>();
+
+    /**
+     * Provides a mechanism to handle exceptions raised from {@link PreDestroy} methods in
+     * whatever way makes the most sense. Implementations should not raise exceptions of
+     * their own
+     */
+    private interface LifeCycleStopFailureHandler
+    {
+        void handlePreDestroyException(Class<?> klass, Method method, Exception exception);
+    }
 
     private enum State
     {
@@ -52,10 +65,10 @@ public final class LifeCycleManager
     /**
      * @param managedInstances list of objects that have life cycle annotations
      * @param methodsMap existing or new methods map
-     * @throws Exception exceptions starting instances (depending on mode)
+     * @throws LifeCycleStartException exceptions starting instances (depending on mode)
      */
     public LifeCycleManager(List<Object> managedInstances, LifeCycleMethodsMap methodsMap)
-            throws Exception
+            throws LifeCycleStartException
     {
         this.methodsMap = (methodsMap != null) ? methodsMap : new LifeCycleMethodsMap();
         for (Object instance : managedInstances) {
@@ -74,15 +87,13 @@ public final class LifeCycleManager
     }
 
     /**
-     * Start the life cycle - all instances will have their {@link javax.annotation.PostConstruct} method(s) called
-     *
-     * @throws Exception errors
+     * Start the life cycle - all instances will have their {@link PostConstruct} method(s) called
      */
     public void start()
-            throws Exception
+            throws LifeCycleStartException
     {
         if (!state.compareAndSet(State.LATENT, State.STARTING)) {
-            throw new Exception("System already starting");
+            throw new LifeCycleStartException("System already starting");
         }
         log.info("Life cycle starting...");
 
@@ -110,11 +121,53 @@ public final class LifeCycleManager
 
     /**
      * Stop the life cycle - all instances will have their {@link javax.annotation.PreDestroy} method(s) called
+     * and any exceptions raised will be collected and thrown in a wrapped {@link LifeCycleStopException} as
+     * suppressed exceptions. Those failures will not be logged and are the responsibility of the caller to
+     * handle appropriately.
      *
-     * @throws Exception errors
+     * @throws LifeCycleStopException If any failure occurs during the clean up process
+     */
+    public void stopWithoutFailureLogging()
+            throws LifeCycleStopException
+    {
+        List<Exception> failures = new ArrayList<>();
+        stop((klass, method, exception) -> failures.add(exception));
+        if (!failures.isEmpty()) {
+            LifeCycleStopException stopException = new LifeCycleStopException();
+            for (Exception e : failures) {
+                stopException.addSuppressed(e);
+            }
+            throw stopException;
+        }
+    }
+
+    /**
+     * Stop the life cycle - all instances will have their {@link PreDestroy} method(s) called
+     * and any exceptions raised will be immediately logged. If any such exceptions occur, a single
+     * {@link LifeCycleStopException} will be raised at the end of processing which will <b>not</b>
+     * contain any reference to exceptions already logged.
+     *
+     * @throws LifeCycleStopException If any failure occurs during the clean up process
      */
     public void stop()
-            throws Exception
+            throws LifeCycleStopException
+    {
+        AtomicBoolean failure = new AtomicBoolean(false);
+        stop((klass, method, exception) -> {
+            failure.set(true);
+            log.error(exception, "Exception in PreDestroy method %s::%s()", klass.getName(), method.getName());
+        });
+
+        if (failure.get()) {
+            throw new LifeCycleStopException();
+        }
+    }
+
+    /**
+     * Stop the life cycle - all instances will have their {@link PreDestroy} method(s) called and any
+     * exceptions raised will be passed to the provided {@link LifeCycleStopFailureHandler} to collect.
+     */
+    private void stop(LifeCycleStopFailureHandler handler)
     {
         if (!state.compareAndSet(State.STARTED, State.STOPPING)) {
             return;
@@ -135,12 +188,7 @@ public final class LifeCycleManager
         Collections.reverse(reversedInstances);
 
         for (Object obj : reversedInstances) {
-            log.debug("Stopping %s", obj.getClass().getName());
-            LifeCycleMethods methods = methodsMap.get(obj.getClass());
-            for (Method preDestroy : methods.methodsFor(PreDestroy.class)) {
-                log.debug("\t%s()", preDestroy.getName());
-                preDestroy.invoke(obj);
-            }
+            stopInstance(obj, handler);
         }
 
         state.set(State.STOPPED);
@@ -151,10 +199,10 @@ public final class LifeCycleManager
      * Add an additional managed instance
      *
      * @param instance instance to add
-     * @throws Exception errors
+     * @throws LifeCycleStartException errors during {@link PostConstruct} method invocation
      */
     public void addInstance(Object instance)
-            throws Exception
+            throws LifeCycleStartException
     {
         State currentState = state.get();
         if ((currentState == State.STOPPING) || (currentState == State.STOPPED)) {
@@ -168,14 +216,49 @@ public final class LifeCycleManager
         }
     }
 
+    private void stopInstance(Object obj, LifeCycleStopFailureHandler handler)
+    {
+        log.debug("Stopping %s", obj.getClass().getName());
+        LifeCycleMethods methods = methodsMap.get(obj.getClass());
+        for (Method preDestroy : methods.methodsFor(PreDestroy.class)) {
+            log.debug("\t%s()", preDestroy.getName());
+            try {
+                preDestroy.invoke(obj);
+            }
+            catch (Exception e) {
+                handler.handlePreDestroyException(obj.getClass(), preDestroy, unwrapInvocationTargetException(e));
+            }
+        }
+    }
+
     private void startInstance(Object obj)
-            throws IllegalAccessException, InvocationTargetException
+            throws LifeCycleStartException
     {
         log.debug("Starting %s", obj.getClass().getName());
         LifeCycleMethods methods = methodsMap.get(obj.getClass());
         for (Method postConstruct : methods.methodsFor(PostConstruct.class)) {
             log.debug("\t%s()", postConstruct.getName());
-            postConstruct.invoke(obj);
+            try {
+                postConstruct.invoke(obj);
+            }
+            catch (Exception e) {
+                LifeCycleStartException failure = new LifeCycleStartException(
+                        format("Exception in PostConstruct method %s::%s()", obj.getClass().getName(), postConstruct.getName()),
+                        unwrapInvocationTargetException(e));
+                stopInstance(obj, (Class<?> klass, Method method, Exception exception) -> {
+                    String message = format("Exception in PreDestroy method %1$s::%2$s() after PostConstruct failure in %1$s::%3$s()", klass.getName(), method.getName(), postConstruct.getName());
+                    failure.addSuppressed(new RuntimeException(message, exception));
+                });
+                throw failure;
+            }
         }
+    }
+
+    /**
+     * Unwraps {@link InvocationTargetException} instances to their underlying cause, otherwise returns the provided Exception
+     */
+    private static Exception unwrapInvocationTargetException(Exception e)
+    {
+        return (e instanceof InvocationTargetException && e.getCause() instanceof Exception) ? (Exception) e.getCause() : e;
     }
 }
--- a/bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleStopException.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleStopException.java
@@ -1,0 +1,10 @@
+package io.airlift.bootstrap;
+
+public class LifeCycleStopException
+        extends RuntimeException
+{
+    public LifeCycleStopException()
+    {
+        super("Exceptions occurred during lifecycle stop");
+    }
+}
--- a/bootstrap/src/test/java/io/airlift/bootstrap/DestroyExceptionInstance.java
+++ b/bootstrap/src/test/java/io/airlift/bootstrap/DestroyExceptionInstance.java
@@ -1,0 +1,27 @@
+package io.airlift.bootstrap;
+
+import javax.annotation.PreDestroy;
+import javax.inject.Inject;
+
+public class DestroyExceptionInstance
+{
+    @Inject
+    public DestroyExceptionInstance(DependentInstance ignored)
+    {
+        // the constructor exists to force injection of the dependent instance
+    }
+
+    @PreDestroy
+    public void preDestroyExceptionOne()
+    {
+        TestLifeCycleManager.note("preDestroyExceptionOne");
+        throw new IllegalStateException("preDestroyExceptionOne");
+    }
+
+    @PreDestroy
+    public void preDestroyExceptionTwo()
+    {
+        TestLifeCycleManager.note("preDestroyExceptionTwo");
+        throw new IllegalStateException("preDestroyExceptionTwo");
+    }
+}
--- a/bootstrap/src/test/java/io/airlift/bootstrap/TestLifeCycleManager.java
+++ b/bootstrap/src/test/java/io/airlift/bootstrap/TestLifeCycleManager.java
@@ -16,6 +16,7 @@
 package io.airlift.bootstrap;
 
 import com.google.common.collect.ImmutableList;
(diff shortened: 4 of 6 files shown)


pettyjamesm:
Test failures seem unrelated and likely due to a racy assertion, tests pass locally for me. This implementation chooses to swallow exceptions raised by `@PreDestroy` methods and just log them, but they could be re-thrown at the end of `LifeCycleManager::stop()` as well. If that's preferred, here's a diff of that implementation: https://github.com/airlift/airlift/compare/master...pettyjamesm:handle-predestroy-exceptions-and-rethrow

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleStopException.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2010 Proofpoint, Inc.
```
findepi: ?
If i read `src/license/LICENSE-HEADER.txt` correctly, this line is not mandatory.

(same in other new files)

bootstrap/src/test/java/io/airlift/bootstrap/DestroyExceptionInstance.java
```diff
@@ -0,0 +1,48 @@
...
+{
+    @Inject
+    public DestroyExceptionInstance(DependentInstance ignored)
+    {
+    }
+
+    @PostConstruct
+    public void postDestroyException()
+    {
+        TestLifeCycleManager.note("postDestroyException");
```
findepi: - `postDestroyException` -> `post*Construct*...` ?
- did you intent to have `throw ` here? Is this method needed at all?

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -133,17 +133,31 @@ public void stop()
...
             LifeCycleMethods methods = methodsMap.get(obj.getClass());
             for (Method preDestroy : methods.methodsFor(PreDestroy.class)) {
                 log.debug("\t%s()", preDestroy.getName());
-                preDestroy.invoke(obj);
+                try {
+                    preDestroy.invoke(obj);
+                }
+                catch (Exception e) {
+                    Throwable t = (e instanceof InvocationTargetException) ? ((InvocationTargetException) e).getTargetException() : e;
+                    log.error("Exception in PreDestroy method %s::%s()", obj.getClass().getName(), preDestroy.getName(), t);
```
findepi: Is `t` being logged? (fmt string has 2 `%s`s only)

If you intent to log the stacktrace in full, `t` should be the first param... but I don't think it's a good idea, since `t` gets propagated eventually.

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleStopException.java
electrum: Yes, we shouldn't add this copyright line to new files, since it isn't correct

bootstrap/src/test/java/io/airlift/bootstrap/TestLifeCycleManager.java
```diff
@@ -212,6 +214,44 @@ public void testJITInjection()
...
+        try {
+            lifeCycleManager.stop();
+            fail("Expected exception to be thrown");
+        }
+        catch (LifeCycleStopException e) {
+            assertEquals(e.getSuppressed().length, 2, "Expected two suppressed exceptions");
+            Set<String> suppressedMessages = Arrays.stream(e.getSuppressed())
+                    .map(Throwable::getMessage)
+                    .collect(Collectors.toSet());
+            assertEquals(suppressedMessages, Sets.newHashSet("preDestroyExceptionOne", "preDestroyExceptionTwo"));
```
electrum: Nit: use `ImmutableSet.of`

bootstrap/src/test/java/io/airlift/bootstrap/TestLifeCycleManager.java
```diff
@@ -212,6 +214,44 @@ public void testJITInjection()
...
+        }
+        catch (Exception e) {
+            fail("Expected LifeCycleStopException to be throw, found: " + e.getClass());
+        }
+
+        assertEquals(new HashSet<>(stateLog), Sets.newHashSet(
```
electrum: `ImmutableSet.copyOf` and `ImmutableSet.of`

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -133,17 +133,31 @@ public void stop()
...
             log.debug("Stopping %s", obj.getClass().getName());
             LifeCycleMethods methods = methodsMap.get(obj.getClass());
             for (Method preDestroy : methods.methodsFor(PreDestroy.class)) {
                 log.debug("\t%s()", preDestroy.getName());
-                preDestroy.invoke(obj);
+                try {
+                    preDestroy.invoke(obj);
+                }
+                catch (Exception e) {
+                    Throwable t = (e instanceof InvocationTargetException) ? ((InvocationTargetException) e).getTargetException() : e;
```
electrum: This can be simplified to
```java
Throwable t = (e instanceof InvocationTargetException) ? e.getCause() : e;
```
Per the documentation, `getCause()` is the preferred way.

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
findepi: @electrum if there are multiple components failing, logging one line immediately can improve log output readability. 
IMO it's better to log here, but just `t` `toString()` instead of full stacktrace.

dain: The general rule is "log or throw, not both".

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
pettyjamesm: Easy fix, will change.

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleStopException.java
pettyjamesm: Sure, will remove the header from new files.

bootstrap/src/test/java/io/airlift/bootstrap/DestroyExceptionInstance.java
pettyjamesm: Not strictly, was just following the convention of other instances. I'll remove.

bootstrap/src/test/java/io/airlift/bootstrap/TestLifeCycleManager.java
pettyjamesm: Easy fix, will change.

bootstrap/src/test/java/io/airlift/bootstrap/TestLifeCycleManager.java
pettyjamesm: Will change.

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
pettyjamesm: Re: the first comment, I didn't look hard enough at the logger API used here. Throwable should indeed be the first argument if being logged. Will do that.

Regarding the log ^ throw debate, I'm torn about how to proceed. If the lifecycle is stopping because the JVM is exiting, it definitely makes sense to log errors immediately because the logger infrastructure may be shutting down or the JVM may force-exit after some timeout. Logging immediately should be clearly preferable.

However, lifecylce stops initiated by an application that will continue to run will probably prefer to handle the exceptions all at once, probably via a checked exception type so there's only one type of exception to handle the problematic (or "dirty") shutdown.

These two use cases seem at odds with one another, so we can either:
- pick one at the detriment of the other use case
- break the log ^ throw convention in this particular case
- make these two use cases call distinct methods and support both modes.

I'm going to pursue the two method approach and update the CR, but I'm happy to implement either of the other two options if you guys prefer. Let me know.

pettyjamesm:
I've added a new commit to put the exception handling mechanism during `stop()` behind an interface so that log / suppressing exceptions can be done in whatever mechanism makes sense. This also enables calling `@PreDestroy` methods on instances that fail during their `@PostConstruct` phase, which I added as well.

findepi:
Re: "log ^ throw" -- I already expressed my opinion in https://github.com/airlift/airlift/pull/739#discussion_r300422597
I am deferring to @electrum and @dain on a decision.

pettyjamesm:
@findepi any feedback / opinions on the new commit added? I was trying to strike a balance for how to handle that "log ^ throw" problem and saw the opportunity to clean up instances that fail to `@PostConstruct` cleanly

findepi:
@pettyjamesm from engineering perspective this is well done. I am not convinced yet, however, that we need this flexibility.
I would prefer to settle on one simple option and revisit the decision when some real-life situation demands this.

pettyjamesm:
@findepi still waiting on a review follow up from @dain or @electrum, but regarding your concern about it being over complicated- I both agree and unfortunately don't see a way to accomplish this more simply. Ignoring the fact that this enables "arbitrary" implementations of `LifeCycleStopFailureHandler` to be provided by the caller, we have 3 use cases internal to this class alone:
1. Stop and log errors immediately, throw a `LifeCycleStopException` at the end if any occurred but without any reference to the already logged exception(s)
1. Stop and attach all errors as suppressed to a parent `LifeCylceStopException` without logging
1. Stop a single instance that failed to fully `@PostConstruct` successfully and attach all exceptions as suppressed to the exception raised from start-up logic.

I played around trying to find a simpler way to handle the cases mentioned above without resorting to a full blown interface type, but was ultimately unsuccessful. Whether we choose to make the `LifeCycleStopFailureHandler` interface and `stop(LifeCycleStopFailureHandler)` method public is another matter- I'll gladly make them private if that's preferable.

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -110,14 +110,52 @@ public void start()
...
+            LifeCycleStopException e = new LifeCycleStopException();
+            for (Throwable t : failures) {
+                e.addSuppressed(t);
+            }
+            throw e;
+        }
+    }
+
+    /**
+     * Stop the life cycle - all instances will have their {@link javax.annotation.PreDestroy} method(s) called
```
electrum: `PreDestroy` probably doesn't need to be qualified, since it's already imported. Same for other usages.

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleStopException.java
```diff
@@ -0,0 +1,10 @@
+package io.airlift.bootstrap;
+
+public class LifeCycleStopException
+        extends Exception
```
electrum: Let's make this extend `RuntimeException`. No point in forcing the caller to handle it, since they don't have any reasonable way to do so. I did a quick survey of the ~20 instances in Presto and they all either log or simply rethrow it.

bootstrap/src/test/java/io/airlift/bootstrap/DestroyExceptionInstance.java
```diff
@@ -0,0 +1,26 @@
...
+
+import javax.annotation.PreDestroy;
+import javax.inject.Inject;
+
+public class DestroyExceptionInstance
+{
+    @Inject
+    public DestroyExceptionInstance(DependentInstance ignored)
+    {
+    }
```
electrum: Let's add a comment
```java
{
    // the constructor exists to force injection of the dependent instance
}

File: bootstrap/src/test/java/io/airlift/bootstrap/TestLifeCycleManager.java [hunk 10]
```diff
@@ -212,6 +215,76 @@ public void testJITInjection()
...
+            fail("Expected exception to be thrown");
+        }
+        catch (LifeCycleStopException e) {
+            assertEquals(e.getSuppressed().length, 2, "Expected two suppressed exceptions");
+            Set<String> suppressedMessages = Arrays.stream(e.getSuppressed())
+                    .map(Throwable::getMessage)
+                    .collect(Collectors.toSet());
+            assertEquals(ImmutableSet.copyOf(suppressedMessages), ImmutableSet.of("preDestroyExceptionOne", "preDestroyExceptionTwo"));
+        }
+        catch (Exception e) {
```
electrum: Let's remove this catch block. Better to let the exception propagate and fail the test, since that's the normal behavior, and it will provide the full stack trace of the unexpected exception.

bootstrap/src/test/java/io/airlift/bootstrap/TestLifeCycleManager.java
```diff
@@ -212,6 +215,76 @@ public void testJITInjection()
...
+        lifeCycleManager.start();
+        assertEquals(stateLog, ImmutableList.of("postDependentInstance"));
+        try {
+            lifeCycleManager.stop();
+            fail("Expected exception to be thrown");
+        }
+        catch (LifeCycleStopException e) {
+            assertEquals(e.getSuppressed().length, 0, "Suppressed exceptions list should be empty");
+        }
+        catch (Exception e) {
```
electrum: Same as above

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -133,18 +171,33 @@ public void stop()
...
+                try {
+                    preDestroy.invoke(obj);
+                }
+                catch (Exception e) {
+                    Throwable t = (e instanceof InvocationTargetException) ? e.getCause() : e;
+                    if (logFailuresImmediately) {
+                        log.error(t, "Exception in PreDestroy method %s::%s()", obj.getClass().getName(), preDestroy.getName());
+                    }
+                    else {
+                        log.debug(t, "Exception in PreDestroy method %s::%s()", obj.getClass().getName(), preDestroy.getName());
```
electrum: We shouldn't log here at all. Even at debug level, log-and-throw is still confusing.

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -221,14 +238,49 @@ public void addInstance(Object instance)
...
+                    cause.addSuppressed(exception);
+                });
+                throw cause;
+            }
         }
     }
+
+    /**
+     * Unwraps {@link InvocationTargetException} instances to their underlying cause, otherwise returns the provided
+     * Exception
```
electrum: Nit: we don't have a hard line length limit, so either put this all on one line, or balance the text onto two lines (having a single word on a line looks strange)

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleStopException.java
```diff
@@ -1,7 +1,7 @@
 package io.airlift.bootstrap;
 
 public class LifeCycleStopException
-        extends Exception
+        extends RuntimeException
```
electrum: Move this change to the previous commit (where the class is introduced)

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -40,6 +40,16 @@
     private final LifeCycleMethodsMap methodsMap;
     private final AtomicReference<Thread> shutdownHook = new AtomicReference<>();
 
+    /**
+     * Provides a mechanism to handle exceptions raised from {@link PreDestroy} methods in
+     * whatever way makes the most sense. Implementations should not raise exceptions of
+     * their own
+     */
+    public interface LifeCycleStopFailureHandler
```
electrum: Let's make this private for now. We can make it public later if there is a use case.

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -119,13 +129,26 @@ public void start()
...
+
+            @Override
+            public void handlePreDestroyException(Class<?> klass, Method method, Exception exception)
+            {
+                log.debug(exception, "Exception in PreDestroy method %s::%s()", klass.getName(), method.getName());
+                failures.add(exception);
+            }
+        }
+
+        CollectForSuppressedExceptionsHandler handler = stop(new CollectForSuppressedExceptionsHandler());
```
electrum: We can simplify this with a lambda
```java
List<Exception> failures = new ArrayList<>();
stop((klass, method, exception) -> failures.add(exception));

File: bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java [hunk 17]
```diff
@@ -140,22 +163,34 @@ public void stopWithoutFailureLogging()
...
+
+            @Override
+            public void handlePreDestroyException(Class<?> klass, Method method, Exception exception)
+            {
+                log.error(exception, "Exception in PreDestroy method %s::%s()", klass.getName(), method.getName());
+                this.exceptionOccurred = true;
+            }
+        }
+
+        if (stop(new LogEachAndDiscardFailureHandler()).exceptionOccurred) {
```
electrum: Similarly, this seems simpler as a lambda:
```java
AtomicBoolean failure = new AtomicBoolean();
stop((klass, method, exception) -> {
    log.error(exception, "Exception in PreDestroy method %s::%s()", klass.getName(), method.getName());
    failure.set(true);
});
if (failure.get()) {
    throw new LifeCycleStopException();
}

File: bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java [hunk 18]
```diff
@@ -140,22 +163,34 @@ public void stopWithoutFailureLogging()
...
     /**
      * Stop the life cycle - all instances will have their {@link javax.annotation.PreDestroy} method(s) called
-     * and any exceptions raised will be returned.
+     * and any exceptions raised will be passed to the provided {@link LifeCycleStopFailureHandler} to collect.
      *
-     * @return All exceptions caught during cleanup processing
+     * @return The same {@link LifeCycleStopFailureHandler} instance passed to this method
      */
-    private List<Throwable> stopInternal(boolean logFailuresImmediately)
+    public <T extends LifeCycleStopFailureHandler> T stop(T handler)
```
electrum: After changing to lambda, we don't need this to return the handler -- it can be `void`. The caller already has the handler if they need it.

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -221,14 +238,49 @@ public void addInstance(Object instance)
...
         for (Method postConstruct : methods.methodsFor(PostConstruct.class)) {
             log.debug("\t%s()", postConstruct.getName());
-            postConstruct.invoke(obj);
+            try {
+                postConstruct.invoke(obj);
+            }
+            catch (Exception e) {
+                Exception cause = unwrapInvocationTargetException(e);
+                log.debug(cause, "Exception in PostConstruct method %s::%s()", obj.getClass().getName(), postConstruct.getName());
+                stopInstance(obj, (Class<?> klass, Method method, Exception exception) -> {
```
electrum: I don't think the lambda parameter types are needed here. It can be
```java
stopInstance(obj, (klass, method, exception) -> {

File: bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java [hunk 20]
```diff
@@ -221,14 +238,49 @@ public void addInstance(Object instance)
...
         LifeCycleMethods methods = methodsMap.get(obj.getClass());
         for (Method postConstruct : methods.methodsFor(PostConstruct.class)) {
             log.debug("\t%s()", postConstruct.getName());
-            postConstruct.invoke(obj);
+            try {
+                postConstruct.invoke(obj);
+            }
+            catch (Exception e) {
+                Exception cause = unwrapInvocationTargetException(e);
+                log.debug(cause, "Exception in PostConstruct method %s::%s()", obj.getClass().getName(), postConstruct.getName());
```
electrum: Rather than logging here and suffering from log-and-throw, what if we enhance the thrown exception:
```java
LifecycleStartException failure = new LifecycleStartException(
        format("Exception in PostConstruct method %s::%s()", obj.getClass().getName(), postConstruct.getName()),
        unwrapInvocationTargetException(e));
stopInstance(obj, (klass, method, exception) -> {
    String message = format("Exception in PreDestroy method %1$s::%2$s() after PostConstruct failure in %1$s::%3$s()", klass.getName(), method.getName(), postConstruct.getName());
    failure.addSuppressed(new RuntimeException(message, exception));
});
throw failure;
```
With this change, we'd introduce a new unchecked `LifecycleStartException` and change the `start()` method to throw that for all failures. This aligns it with `stop()` for the same reasons -- callers don't have any reasonable way to handle the underlying exception.

pettyjamesm:
@electrum thanks for the review, I've combined the two commits into one and addressed all your CR feedback.

electrum:
This is great! Thanks a lot for improving this and putting up with several rounds of review.

PR #741, opened 2019-07-10
 Update Travis config to use OpenJDK 8 

Praveen2112:


Praveen2112:
In prestosql repo we have specified it so added it. Will remove it.

electrum:
Merged, thanks!

PR #742, opened 2019-07-24
Allow BoundedExecutor to recover from failures

electrum:
Previously submitted tasks will eventually run if the underlying
executor recovers and additional tasks are submitted.

### DIFF
--- a/concurrent/src/main/java/io/airlift/concurrent/BoundedExecutor.java
+++ b/concurrent/src/main/java/io/airlift/concurrent/BoundedExecutor.java
@@ -8,10 +8,9 @@ import javax.annotation.concurrent.ThreadSafe;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
-import static com.google.common.base.Preconditions.checkState;
+import static java.lang.Math.max;
 import static java.util.Objects.requireNonNull;
 
 /**
@@ -36,7 +35,6 @@ public class BoundedExecutor
 
     private final Queue<Runnable> queue = new ConcurrentLinkedQueue<>();
     private final AtomicInteger queueSize = new AtomicInteger(0);
-    private final AtomicBoolean failed = new AtomicBoolean();
 
     private final Executor coreExecutor;
     private final int maxThreads;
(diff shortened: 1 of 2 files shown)


concurrent/src/main/java/io/airlift/concurrent/BoundedExecutor.java
```diff
@@ -52,35 +50,42 @@ public BoundedExecutor(Executor coreExecutor, int maxThreads)
...
         int size = queueSize.incrementAndGet();
         if (size <= maxThreads) {
-            // If able to grab a permit (aka size <= maxThreads), then we are short exactly one draining thread
             try {
                 coreExecutor.execute(this::drainQueue);
             }
             catch (Throwable e) {
-                failed.set(true);
-                log.error("BoundedExecutor state corrupted due to underlying executor failure");
+                decrementAndGetQueueSize();
```
findepi: if `coreExecutor.execute(this::drainQueue)` throws immediately, you can end up having 1 task in queue and queueSize=0.

To me, it looks like you want something like "current thread count" instead of "queue size". Question is how to do this without losing an incoming task.

concurrent/src/main/java/io/airlift/concurrent/BoundedExecutor.java
```diff
@@ -52,35 +50,42 @@ public BoundedExecutor(Executor coreExecutor, int maxThreads)
...
+                    task.run();
+                }
+                catch (Throwable e) {
+                    log.error(e, "Task failed");
+                }
             }
-            catch (Throwable e) {
-                log.error(e, "Task failed");
+
+            if ((decrementAndGetQueueSize() == 0) && (task == null)) {
```
findepi: unnecessary parenthesis

concurrent/src/main/java/io/airlift/concurrent/BoundedExecutor.java
```diff
@@ -52,35 +50,42 @@ public BoundedExecutor(Executor coreExecutor, int maxThreads)
...
+                return;
```
findepi: With maxThreads=1 consider

1. thread A calls execute
   a. adds task
   b. queueSize := 1
   c. starts a thread A'
1. thread A' removes the task from the queue and executes it
1. thread A' calls `decrementAndGetQueueSize`, i.e. queueSize := 0, but does not exit because `task` was not null
1. thread B (or A) calls execute 
   a. adds task
   b. queueSize := 1
   c. starts a thread B', because queueSize <= maxThreads

At this point, 2 threads A' and B' are executing within coreExecutor even though maxThreads is 1.

findepi: (updated)

Praveen2112: But if we could rearrange the `if` condition to `task == null && decrementAndGetQueueSize() == 0` then threadB' won't be created as `queueSize := 0` only if the `task == null` or maybe we can peek into the queue to check for the next task 
`decrementAndGetQueueSize == 0 && queue.peek() == null`

erichwang:
To achieve the desired behavior, we really only have 3 possible options here:
1. Make the bounded executor fully synchronized for all task submissions (only allowing one submission at a time) -- but this might add unwanted overhead in high performance cases since none of the existing code or dependencies use any locks.
2. Add a scheduled background thread pool whose job is to keep retrying failed task submission retries forever until success
3. Consume the calling thread as one of the draining tasks in the event of thread pool failure -- while technically this wouldn't corrupt the data structure, this would be terrible for callers and so this isn't really a desirable path.

There is no way to make this magically work without one of these three options. For me, I would suggest (2).

PR #746, opened 2019-08-17
Add t-digest data structure

martint:
Based on the paper at https://github.com/tdunning/t-digest/blob/master/docs/t-digest-paper/histo.pdf

### DIFF
--- a/stats/src/main/java/io/airlift/stats/DecayTDigest.java
+++ b/stats/src/main/java/io/airlift/stats/DecayTDigest.java
@@ -1,0 +1,154 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.stats;
+
+import com.google.common.base.Ticker;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static io.airlift.stats.ExponentialDecay.weight;
+
+public class DecayTDigest
+{
+    private static final long RESCALE_THRESHOLD_SECONDS = 50;
+    private static final double ZERO_WEIGHT_THRESHOLD = 1e-5;
+
+    private final TDigest digest;
+
+    private final Ticker ticker;
+    private final double alpha;
+    private long landmarkInSeconds;
+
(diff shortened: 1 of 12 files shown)


stats/src/test/java/io/airlift/stats/TestTDigest.java
```diff
@@ -0,0 +1,246 @@
...
+            assertTrue(value >= previous);
+            previous = value;
+        }
+    }
+
+    @Test
+    public void testBigJump()
+    {
+        TDigest digest = new TDigest(100);
+        for (int i = 1; i < 20; i++) {
```
dain: consider big jump with enough values to result in a merge?

stats/src/test/java/io/airlift/stats/TestTDigest.java
```diff
@@ -0,0 +1,246 @@
...
+
+    @Test
+    public void testSerializationComplex()
+    {
+        TDigest digest = new TDigest();
+        addAll(digest, asList(0, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 4, 5, 6, 7));
+
+        TDigest deserialized = TDigest.deserialize(digest.serialize());
+
+        assertEquals(deserialized.getMin(), digest.getMin());
```
dain: this equals for two TDigest is repeated a number of times... add a method

stats/src/main/java/io/airlift/stats/DecayTDigest.java
```diff
@@ -0,0 +1,149 @@
...
+    private final Ticker ticker;
+    private final double alpha;
+    private long landmarkInSeconds;
+
+    public DecayTDigest(double compression, double alpha)
+    {
+        this(compression, alpha, alpha == 0.0 ? noOpTicker() : Ticker.systemTicker());
+    }
+
+    public DecayTDigest(double compression, double alpha, Ticker ticker)
```
dain: should this be package protected?

stats/src/main/java/io/airlift/stats/DecayTDigest.java
```diff
@@ -0,0 +1,149 @@
...
+    {
+        digest = new TDigest(compression);
+
+        this.alpha = alpha;
+        this.ticker = ticker;
+        landmarkInSeconds = TimeUnit.NANOSECONDS.toSeconds(ticker.read());
```
dain: maybe add a method to read ticker seconds, or wrap in a `LongSupplier`

stats/src/main/java/io/airlift/stats/TDigest.java
```diff
@@ -0,0 +1,475 @@
...
+
+        needsMerge = true;
+    }
+
+    public void mergeWith(TDigest other)
+    {
+        if (centroidCount + other.centroidCount > means.length) {
+            // first, try to compact the digests to make room
+            merge(internalCompressionFactor(compression));
+            other.merge(internalCompressionFactor(compression));
```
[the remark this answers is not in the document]
martint: No, because if other.compression is much larger than this.compression we may end up growing this's capacity beyond what we want to keep around.

PR #750, opened 2019-08-22
Match enums case-insensitively in configuration

findepi:


martint:
I think this is a valuable change

PR #752, opened 2019-08-28
Remove checked exceptions from Bootstrap.initialize()

electrum:


bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -200,19 +201,16 @@ private void stop(LifeCycleStopFailureHandler handler)
...
         State currentState = state.get();
-        if ((currentState == State.STOPPING) || (currentState == State.STOPPED)) {
-            throw new IllegalStateException();
-        }
-        else {
-            startInstance(instance);
-            if (methodsMap.get(instance.getClass()).hasFor(PreDestroy.class)) {
-                managedInstances.add(instance);
-            }
+        checkState((currentState != State.STOPPING) && (currentState != State.STOPPED), "life cycle is stopped");
```
findepi: remove parens

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -181,7 +175,7 @@ public void setLevel(String loggerName, Level level)
         return levels.build();
     }
 
-    private void processLevels(Properties properties)
+    private void processLevels(Map<Object, Object> properties)
```
findepi: Shall we expect key/values to be Strings rather than being unnecessarily flexible with .toString()?

electrum: Good idea. Switched the caller to use `Maps.fromProperties()`

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
electrum: I find this more readable

PR #756, opened 2019-09-13
Add @SlatedForRemoval

findepi:
`@Config` property annotated with `@SlatedForRemoval` is automatically
renamed with addition of a suffix indicating removal date. This feature
is similar to `@LegacyConfig` and `@Deprecated` methods, with important
differences.

`@LegacyConfig` is suitable when configuration property is renamed, but
functionality is retained.

`@Deprecated` is suitable when configuration property is considered
deprecated, but is not being removed yet. Using `@Deprecated` properties
generates a warning.

`@SlatedForRemoval` is suitable when configuration property has been
deprecated for some time and is being removed. Provides additional step
in the deprecation process during which the deprecated functionality is
still accessible, but unlocking it requires manual intervention. This
enforces awareness of the fact that configuration property is deprecated
and being removed (warning generated by use of a `@Deprecated` property
can be ignored) and also awareness of the date after which the property
is no longer guaranteed to exist.

For example for config annotations:

    @Config("some-option")
    @SlatedForRemoval(notEarlierThan = "Jan 2020")

When configuration uses `some-option` property, an error is generated
informing that `some-option` has been replaced with
`some-option.slated-for-removal-2020-01`.

When configuration uses `some-option.slated-for-removal-2020-01`, the
configured value is accepted, but still a warning is generated.

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/SlatedForRemoval.java
+++ b/configuration/src/main/java/io/airlift/configuration/SlatedForRemoval.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.configuration;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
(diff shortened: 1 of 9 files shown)


configuration/src/main/java/io/airlift/configuration/SlatedForRemoval.java
```diff
@@ -0,0 +1,45 @@
...
+public @interface SlatedForRemoval
+{
+    DateTimeFormatter VALUE_DATE_FORMAT = DateTimeFormatter.ofPattern("MMM yyyy", Locale.US);
+    String PROPERTY_SUFFIX = ".slated-for-removal-";
+    DateTimeFormatter PROPERTY_DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM", Locale.US);
+
+    /**
+     * Describes date until which config is guaranteed to be retained.
+     * <p>
+     * Must be in the format accepted by {@link #VALUE_DATE_FORMAT}, for example {@code "Oct 2019"}.
```
martint: I would use the ISO format for months (YYYY-MM), too. It's unambiguous and consistent with how we treat dates/times everywhere else.

findepi: I thought this way will be more readable.

findepi:
rebased

findepi:
updated to `after=`.

Issue #738, opened 2019-07-01
Unexpected value parsed in configuration when line contains trailing spaces


findepi:
For config file like this:
```
hive.non-managed-table-writes-enabled=true 
```
(I guess you easily noticed the space character at the of the value)

The property `hive.non-managed-table-writes-enabled` is assigned a `false` value.
Direct cause is that `Boolean.valueOf("true ") == false`.
This causes bad user experience.

(Side note: when `launcher` reads the properties, it doesn't fall into this trap, as it `string.strip()` all the values (https://github.com/airlift/airlift/issues/633, https://github.com/airlift/airlift/blob/master/launcher/src/main/scripts/bin/launcher.py#L46-L48))

findepi:
This also affects non-boolean values. Eg
https://groups.google.com/d/msg/presto-users/6TcIcS6CucI/RMOeQK3mAAAJ

Maybe we should migrate from .properties to .yaml. This is much better supported in other languages (so maybe no problems like #633) and has saner semantics if all you need is key/value pairs.
Also yaml would enable better configuration options in the future (no need for `;`-delimited properties, since you can declare something as a list)

findepi:
cc @stagraqubole

PR #757, opened 2019-09-17
Convert to boolean strictly

findepi:
Change how booleans are parsed from configuration / configuration files

Case Name | Input | Before Change | After Change | Comment
--------- | ----- | ------------- | ------------ | -------
plain | `"true"` | `true` | `true` | _no change_
plain | `"false"` | `false` | `false` | _no change_
mixed case | `"tRuE"` | `true` | `true` | _no change_
whitespace | `"true "` | `false` | _error_; `true` after #758 | This case is a common configuration mistake, as trailing whitespace is not stripped when reading `.properties`
whitespace | `"false "` | `false` | _error_; `false` after #758 | _no change_ when combined with #758
unrecognized | `"yes"` | `false` | _error_
unrecognized | `"no"` | `false` | _error_

Addresses major pain point of https://github.com/airlift/airlift/issues/738 / https://github.com/prestosql/presto/issues/1546

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
@@ -555,28 +555,28 @@ public class ConfigurationFactory
         }
 
         try {
-            if (String.class.isAssignableFrom(type)) {
+            if (String.class == type) {
                 return value;
             }
-            else if (Boolean.class.isAssignableFrom(type) || Boolean.TYPE.isAssignableFrom(type)) {
+            else if (Boolean.class == type || boolean.class == type) {
                 return Boolean.valueOf(value);
             }
-            else if (Byte.class.isAssignableFrom(type) || Byte.TYPE.isAssignableFrom(type)) {
+            else if (Byte.class == type || byte.class == type) {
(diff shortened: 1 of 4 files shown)


findepi:
@willmo83 what do you think?

dain:
I'd alway strip whitespace from from the values.

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -559,7 +559,14 @@ private static Object coerce(Class<?> type, String value)
...
             if (Boolean.class == type || boolean.class == type) {
-                return Boolean.valueOf(value);
+                // Boolean.valueOf returns `false` when called with `"true "` argument
+                if ("true".equalsIgnoreCase(value)) {
+                    return Boolean.TRUE;
+                }
+                if ("false".equalsIgnoreCase(value)) {
+                    return Boolean.FALSE;
+                }
+                return null;
```
martint: We may want to fail in this case

findepi: This will be turned into a failure later.

PR #758, opened 2019-09-18
Trim whitespace when loading configuration

### LINKED ISSUE #738: Unexpected value parsed in configuration when line contains trailing spaces

findepi:
Fixes #738

### DIFF
--- a/configuration/src/test/java/io/airlift/configuration/TestConfigurationLoader.java
+++ b/configuration/src/test/java/io/airlift/configuration/TestConfigurationLoader.java
@@ -69,12 +69,8 @@ public class TestConfigurationLoader
     {
         final File file = File.createTempFile("config", ".properties", tempDir);
         PrintStream out = new PrintStream(new FileOutputStream(file));
-        try {
-            out.print("test: foo");
-        }
-        catch (Exception e) {
-            out.close();
-        }
+        out.print("test: foo");
+        out.flush();
 
         System.setProperty("config", file.getAbsolutePath());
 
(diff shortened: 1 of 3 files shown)


configuration/src/test/java/io/airlift/configuration/TestConfigurationLoader.java
```diff
@@ -60,46 +61,55 @@ public void testLoadsFromSystemProperties()
...
+        System.setProperty("config", file.getAbsolutePath());
+        System.setProperty("trim-whitespace-key2", " \t key2-value \t ");
+
+        Map<String, String> properties = loadProperties();
+
+        // config files are often human-managed and we need to be permissive, stripping trailing whitespace if any
+        assertEquals(properties.get("trim-whitespace-key1"), "key1-value");
+
+        // trailing whitespace in JVM property value is not so likely to be unintentional, so preserve it
+        assertEquals(properties.get("trim-whitespace-key2"), " \t key2-value \t ");
```
findepi: I put it here so that it's "documented".
I'm not sure I want this behavior. Maybe we should trim always.

@martint @willmo83

PR #759, opened 2019-09-24
Allow Enum to have fromString conversion method

findepi:
This restore behavior from before
6bdf92c96e772da7200c0e3b5771b57eeb4edd8e.

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
@@ -585,6 +585,22 @@ public class ConfigurationFactory
             return null;
         }
 
+        // Look for a static fromString(String) method. This is used in preference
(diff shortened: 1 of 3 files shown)


configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -624,6 +605,22 @@ else if (Double.class.isAssignableFrom(type) || Double.TYPE.isAssignableFrom(typ
         catch (Throwable ignored) {
         }
 
+        if (type.isEnum()) {
+            // case-sensitive match is handled by generic "valueOf(String)" case
```
electrum: How about

> case-sensitive or other transformations are handled by static factories above

electrum:
This seems reasonable to me: if we have a factory method, always use it. 

I agree with adding a test.

findepi:
AC

PR #760, opened 2019-09-26
Use AtomicIntegerFieldUpdater instead of AtomicInteger

pettyjamesm:
Avoids one level of indirection when performing atomic
updates on the queueSize counter.


pettyjamesm:
After an offline discussion, I've decided that the risk to maintainability doesn't justify the slight performance improvement here. Closing.

PR #761, opened 2019-10-04
Validate validation

findepi:
Validation annotation must be placed on a getter. When placed on a
setter, it is silently ignored. After this change, a misplaced validation
annotation in a Config class will be treated as an error.


configuration/src/main/java/io/airlift/configuration/ConfigurationMetadata.java
```diff
@@ -198,6 +210,23 @@ else if (arrayEntry.equals(config.value())) {
         return isValid;
     }
 
+    private static Class<? extends Annotation> getAnnotationClass(Annotation annotation)
```
electrum: I think this can simply be `annotation.annotationType()`

findepi:
AC

PR #762, opened 2019-10-05
Test with newer JDKs

findepi:


.travis.yml
```diff
@@ -2,6 +2,14 @@ language: java
...
 jdk:
   - openjdk8
+  - openjdk9
+  - openjdk10
+  - openjdk11
+  - openjdk-ea
+
+#matrix:
+#  allow_failures:
+#    - jdk: openjdk-ea
```
findepi: need to check if it is needed (uncomment) or not (remove)

martint: We may need this because EA versions of the jdk may be buggy. We shouldn’t fail the build in such cases

findepi:
Airlift tests currently won't pass on >8. https://travis-ci.org/airlift/airlift/builds/594032306
This is a much longer story than just travis configuration.

PR #764, opened 2019-10-13
Add background file resource loader

dain:
This is a proposal based on the many resource loaders in Presto.  The loader details:
* Value is always present
* Failed loads retain existing value
* Initial load is in constructor, which if fails throws directly
* Manual reload with `refreshNow` method
* Optional timed loading in background thread
* Optional exception handler for background loads (default logs)
* Builder syntax for future extensions

### DIFF
--- a/concurrent/src/main/java/io/airlift/concurrent/ResourceLoader.java
+++ b/concurrent/src/main/java/io/airlift/concurrent/ResourceLoader.java
@@ -1,0 +1,224 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.concurrent;
+
+import com.google.common.annotations.VisibleForTesting;
+import io.airlift.log.Logger;
+import io.airlift.units.Duration;
+
+import javax.annotation.PreDestroy;
+import javax.annotation.concurrent.GuardedBy;
+
+import java.io.File;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static io.airlift.concurrent.Threads.daemonThreadsNamed;
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.Executors.newSingleThreadScheduledExecutor;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+public final class ResourceLoader<T>
+{
+    private static final Logger log = Logger.get(ResourceLoader.class);
+
+    private final File file;
+    private final FileLoader<T> fileLoader;
+    private final Duration refreshPeriod;
+    private final Consumer<Exception> exceptionHandler;
+    private final AtomicReference<Runnable> stopRefresh = new AtomicReference<>();
+
+    @GuardedBy("this")
+    private T value;
+
+    private static <T> ResourceLoader<T> createResourceLoader(
+            File file,
+            FileLoader<T> fileLoader,
+            Duration refreshPeriod,
+            SimpleRecurringTaskRunner recurringTaskRunner,
+            Consumer<Exception> exceptionHandler)
+            throws Exception
+    {
+        requireNonNull(recurringTaskRunner, "recurringTaskRunner is null");
+        ResourceLoader<T> resourceLoader = new ResourceLoader<>(
+                fileLoader,
+                file,
+                refreshPeriod,
+                exceptionHandler);
+        resourceLoader.start(recurringTaskRunner);
+        return resourceLoader;
+    }
+
+    private ResourceLoader(FileLoader<T> fileLoader, File file, Duration refreshPeriod, Consumer<Exception> exceptionHandler)
+            throws Exception
+    {
+        this.fileLoader = requireNonNull(fileLoader, "fileLoader is null");
+        this.file = requireNonNull(file, "file is null");
+        this.refreshPeriod = requireNonNull(refreshPeriod, "refreshPeriod is null");
+        this.exceptionHandler = requireNonNull(exceptionHandler, "exceptionHandler is null");
+        refreshNow();
+    }
+
+    private void start(SimpleRecurringTaskRunner recurringTaskRunner)
+    {
+        stopRefresh.set(recurringTaskRunner.scheduleTask(this::refreshInternal, refreshPeriod));
+    }
+
+    @PreDestroy
+    public void stop()
+    {
+        Runnable stopRefresh = this.stopRefresh.getAndSet(null);
+        if (stopRefresh != null) {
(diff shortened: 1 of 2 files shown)


martint:
It’d be even more useful if it allowed arbitrary actions in the background loader beyond just loading a file. E.g., it could be used to refresh configuration or a list of nodes or other resource that originates from a remote service.

dain:
I guess we could allow for any action, but I was thinking that this can be tuned to be smarter about files over time. For example, only call loader when file modified time or size changes.  Additionally, we could hook into the OS file change events.

findepi:
> Failed loads retain existing value

At some point you may want to do something about that.
(rg When you fail to refresh authorization rules 10 times straight, it's not necessarily safe to continue to operate the system.)

concurrent/src/main/java/io/airlift/concurrent/ResourceLoader.java
```diff
@@ -0,0 +1,224 @@
...
+
+    public interface FileLoader<T>
+    {
+        T loadFile(File file)
+                throws Exception;
+    }
+
+    public interface SimpleRecurringTaskRunner
+    {
+        Runnable scheduleTask(Runnable runnable, Duration fixedDelay);
```
findepi: Document, it's not obvious what it's the return value

concurrent/src/main/java/io/airlift/concurrent/ResourceLoader.java
```diff
@@ -0,0 +1,224 @@
...
+
+        private InternalScheduledRecurringTaskRunner(String name)
+        {
+            this.name = requireNonNull(name, "name is null");
+        }
+
+        @Override
+        public Runnable scheduleTask(Runnable runnable, Duration fixedDelay)
+        {
+            ScheduledExecutorService executor = newSingleThreadScheduledExecutor(daemonThreadsNamed(name + "-background-loader"));
```
findepi: This might be too simple. If we have 50 refreshable configurations, we will have 50 threads idling.
Maybe a cached thread pool?

concurrent/src/main/java/io/airlift/concurrent/ResourceLoader.java
```diff
@@ -0,0 +1,224 @@
...
+    {
+        return new ResourceLoaderBuilder<>(fileLoader);
+    }
+
+    public static final class ResourceLoaderBuilder<T>
+    {
+        private final FileLoader<T> fileLoader;
+        private String name;
+        private File file;
+        private Duration refreshPeriod = new Duration(0, NANOSECONDS);
```
findepi: 0 is a treated as a special value. I'd use Optional

concurrent/src/main/java/io/airlift/concurrent/ResourceLoader.java
```diff
@@ -0,0 +1,224 @@
...
+import static io.airlift.concurrent.Threads.daemonThreadsNamed;
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.Executors.newSingleThreadScheduledExecutor;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+public final class ResourceLoader<T>
+{
+    private static final Logger log = Logger.get(ResourceLoader.class);
+
+    private final File file;
```
findepi: Maybe generify `File` into S (source). This could be Path, URL or id of an auxiliary resource.

concurrent/src/main/java/io/airlift/concurrent/ResourceLoader.java
```diff
@@ -0,0 +1,224 @@
...
+    @PreDestroy
+    public void stop()
+    {
+        Runnable stopRefresh = this.stopRefresh.getAndSet(null);
+        if (stopRefresh != null) {
+            stopRefresh.run();
+        }
+    }
+
+    public synchronized T getValue()
```
findepi: just `get()` ?

concurrent/src/test/java/io/airlift/concurrent/TestResourceLoader.java
```diff
@@ -0,0 +1,255 @@
...
+
+            assertEquals(loader.getValue(), "data");
+
+            Files.write("new-data".getBytes(UTF_8), tempFile.file());
+            assertEquals(loader.getValue(), "data");
+
+            loader.refreshNow();
+            assertEquals(loader.getValue(), "new-data");
+
+            loader.stop();
```
findepi: consider making `ResourceLoader` closeable and use t-w-r

concurrent/src/test/java/io/airlift/concurrent/TestResourceLoader.java
```diff
@@ -0,0 +1,255 @@
...
+
+    @Test
+    public void testBuiltinRefresh()
+            throws Exception
+    {
+        try (TempFile tempFile = new TempFile()) {
+            Files.write("data".getBytes(UTF_8), tempFile.file());
+
+            LoaderWaiter loaderWaiter = new LoaderWaiter();
+            Duration refreshDuration = new Duration(10, MILLISECONDS);
```
findepi: 1 milli?

concurrent/src/test/java/io/airlift/concurrent/TestResourceLoader.java
```diff
@@ -0,0 +1,255 @@
...
+            // verify again
+            loaderWaiter.waitForLoad();
+            long secondWaitValue = loader.getValue();
+            assertGreaterThan(secondWaitValue, firstWaitValue);
+
+            loader.stop();
+        }
+    }
+
+    private static class LoaderWaiter
```
findepi: move below all test methods

concurrent/src/test/java/io/airlift/concurrent/TestResourceLoader.java
```diff
@@ -0,0 +1,255 @@
...
+    public void testRefreshException()
+            throws Exception
+    {
+        try (TempFile tempFile = new TempFile()) {
+            Files.write("initial".getBytes(UTF_8), tempFile.file());
+
+            // exception from construction throws directly
+            assertThrows(
+                    SecurityException.class,
+                    () -> ResourceLoader.builder(file -> { throw new SecurityException(); })
```
findepi: use RuntimException (not Security) and verify message

```
assertThatThrownBy(...)
  .inInstanceOf
 .hasMessage
```

findepi (approved):
looks nice.

@dain can you post a PR in Presto too to verify it indeed fits all current needs?

concurrent/src/main/java/io/airlift/concurrent/ResourceLoader.java
```diff
@@ -0,0 +1,224 @@
...
+        public ResourceLoaderBuilder<T> setExceptionHandler(Consumer<Exception> exceptionHandler)
+        {
+            this.exceptionHandler = exceptionHandler;
+            return this;
+        }
+
+        public ResourceLoader<T> build()
+                throws Exception
+        {
+            checkArgument(name != null, "name not set");
```
findepi: `checkState`?

findepi:
cc @kokosing

concurrent/src/main/java/io/airlift/concurrent/ResourceLoader.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    public static final class ResourceLoaderBuilder<T>
+    {
+        private final FileLoader<T> fileLoader;
+        private String name;
+        private File file;
+        private Duration refreshPeriod = new Duration(0, NANOSECONDS);
+        private SimpleRecurringTaskRunner recurringTaskRunner;
+        private Consumer<Exception> exceptionHandler = exception -> log.error("Error refreshing " + name, exception);
```
kokosing: I think by default we should propagate the failure to user.

PR #765, opened 2019-10-19
Generate annotation implementation

findepi:


discovery/pom.xml
```diff
@@ -50,6 +50,12 @@
             <artifactId>node</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>io.joj</groupId>
+            <artifactId>joj-reflect</artifactId>
+            <version>1.1.1</version>
```
findepi: I'm the owner. This can be merged into airlift.

discovery/src/test/java/io/airlift/discovery/client/TestServiceTypes.java
```diff
@@ -1,69 +0,0 @@
...
- * limitations under the License.
- */
-package io.airlift.discovery.client;
-
-import org.testng.annotations.Test;
-
-import static io.airlift.testing.EquivalenceTester.equivalenceTester;
-import static org.testng.Assert.assertEquals;
-
-public class TestServiceTypes
```
findepi: The test was passing (https://travis-ci.org/airlift/airlift/builds/600146887).
However, technically, there is no need to test a library.

electrum:
This library is cool (and the implementation is very clever), but adding several new dependencies (including PCollections) to avoid a single generated class seems like the wrong way.

findepi:
@electrum  i kind of sensed that, but still wanted to get the feedback.

PR #767, opened 2019-10-26
Validate experimental property is not enabled by default

findepi:


findepi:
We will have to do something with `experimental.reserved-pool-enabled`: the flag can be understood as reserved pool is experimental, whereas lack of the pool is.

electrum:
I don’t think we should be basing this checks on property names, especially a simple contains check.

This actually brings up a discussion around the naming of properties. Giving them the name “experimental” leads to use pain around configuration if we expect many users will enable them. Perhaps we need a different way of signaling the status.

configuration/src/main/java/io/airlift/configuration/testing/ConfigAssertions.java
```diff
@@ -98,6 +101,14 @@ private ConfigAssertions()
             Object expectedAttributeValue = expectedAttributeValues.get(attribute.getName());
 
             assertEquals(expectedAttributeValue, actualAttributeValue, attribute.getName());
+
+            String property = attribute.getInjectionPoint().getProperty();
+            if (property != null && property.contains("experimental")) {
+                Object value = actualAttributeValue;
+                if (value != null && !FALSE.equals(value) && !Optional.empty().equals(value)) {
```
sopel39: This will prevent any numerical `experimental` properties (e.g companion configs).

sopel39:
I think such name would just make property naming more sophisticated, e.g people would use `experimental.not-feature` instead `experimental.feature`

configuration/src/main/java/io/airlift/configuration/testing/ConfigAssertions.java
findepi: > e.g companion configs

I am aware. I don't see what's the drawback. Please elaborate.

findepi:
> I think such name would just make property naming more sophisticated, e.g people would use `experimental.not-feature` instead `experimental.feature`

@sopel39 
Can you provide an example?

findepi:
> This actually brings up a discussion around the naming of properties. Giving them the name “experimental” leads to use pain around configuration if we expect many users will enable them. Perhaps we need a different way of signaling the status.

@electrum i tend to agree.

We may drop the `experimental.` convention and leverage documentation/description for this.
Note that we we *don't* use `experimental_` prefix for session toggles, which means we're inconsistent today.
cc @martint @kokosing @dain 

If we agree to change the convention, that's fine.
If we don't change it, I'd still want this PR to be merged.

configuration/src/main/java/io/airlift/configuration/testing/ConfigAssertions.java
sopel39: Example configs:
```
experimental.dynamic-filtering-max-per-driver-row-count
experimental.dynamic-filtering-max-per-driver-size
```
they have reasonable defaults.

sopel39:
> @sopel39
> Can you provide an example?

`experimental.enable-dynamic-filtering` vs `experimental.disable-dynamic-filtering`.

findepi:
> `experimental.enable-dynamic-filtering` vs `experimental.disable-dynamic-filtering`.

It's disabled by default, so in line with expecations introduced here.

> 
> or better
> 
> `experimental.spill-window-operator` (now defaults to `true`) would change to `experimental.spill-window-operator-disabled` (which defaults to `false`)
> 
> I think it's confusing

I like this example.
We enabled something by default while marking this "experimental" at the same time.
It _is_ confusing.

This is exactly the kind of situation I want to avoid in the future.

sopel39:
> We enabled something by default while marking this "experimental" at the same time.
It is confusing.

Yes, but there is master spill switch. Others control minor aspects of spill (including numerical ones).

dain:
I'm not a fan of requiring property names to be stated in a positive or negative form.  This change makes the `FALSE` value a special value and tries to understand the semantic meaning of that value.  In many cases it is clearer to have state properties in a positive form and then have the default value be `TRUE`... one of the big reasons we allow default values is so we didn't require magic values like `0` or `-1` and then automatically assign them meaning.

I would prefer that we let authors decide property names and default values that are natural to them.

martint:
I don't think airlift should be in the business of dictating how configuration options are named or policies around what default values are allowed. It's conceivable some application might actually want to enable experimental features by default.

findepi:
I'm closing.

Let's have the discussion the next time we have experimental functionality which _appears_ to be enabled by default.
(Like we did with `experimental.reserved-pool-enabled`).

PR #770, opened 2019-10-31
Allow setting config values from environment variables

electrum:
Config values may be of the form `${ENV:XXX}` where `XXX` is the name
of an environment variable. The value is replaced in `Bootstrap` with
the value of that environment variable.

### DIFF
--- a/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
@@ -16,6 +16,7 @@
 package io.airlift.bootstrap;
 
 import com.google.common.annotations.Beta;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableList.Builder;
@@ -40,15 +41,19 @@ import io.airlift.log.LoggingConfiguration;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.UncheckedIOException;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
-import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.function.BiConsumer;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static io.airlift.configuration.ConfigurationLoader.getSystemProperties;
 import static io.airlift.configuration.ConfigurationLoader.loadPropertiesFrom;
+import static java.lang.String.format;
 
 /**
  * Entry point for an application built using the platform codebase.
@@ -63,6 +68,8 @@ import static io.airlift.configuration.ConfigurationLoader.loadPropertiesFrom;
  */
 public class Bootstrap
 {
+    private static final Pattern ENV_PATTERN = Pattern.compile("\\$\\{ENV:([a-zA-Z][a-zA-Z0-9_]*)}");
+
     private final Logger log = Logger.get("Bootstrap");
     private final List<Module> modules;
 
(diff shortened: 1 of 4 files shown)


bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -291,4 +306,28 @@ private static ColumnPrinter makePrinterForConfiguration(ConfigurationFactory co
...
+    {
+        Map<String, String> replaced = new HashMap<>();
+        properties.forEach((propertyKey, propertyValue) -> {
+            Matcher matcher = ENV_PATTERN.matcher(propertyValue);
+            if (!matcher.matches()) {
+                replaced.put(propertyKey, propertyValue);
+                return;
+            }
+            String envName = matcher.group(1);
+            String envValue = environment.get(envName);
```
martint: Do empty environment variables show up as empty string or nulls?

bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrap.java
```diff
@@ -58,6 +66,42 @@ public void testDoesNotAllowCircularDependencies()
...
+                .put("apple", "apple-value")
+                .put("grape", "${ENV:GRAPE}")
+                .put("peach", "${ENV:PEACH}")
+                .put("grass", "${ENV:!!!}")
+                .put("pear", "${ENV:X_PEAR}")
+                .put("cherry", "${ENV:X_CHERRY}")
+                .put("orange", "orange-value")
+                .build();
+
+        ImmutableMap<String, String> environment = ImmutableMap.<String, String>builder()
```
martint: Use Map for the declared type

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
electrum: It should be possible for them to be empty, which is legal as a property value. Do you see any problems with allow this? It seems fine to me, although probably not useful.

bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrap.java
electrum: Thanks, missed that

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
martint: No, no problem with that. Just wanted to be sure they wouldn't be presented as `null` by `environment.get()`, which would result in them being considered missing.

bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrap.java
```diff
@@ -58,6 +66,42 @@ public void testDoesNotAllowCircularDependencies()
...
+    @Test
+    public void testEnvironmentVariableReplacement()
+    {
+        Map<String, String> original = ImmutableMap.<String, String>builder()
+                .put("apple", "apple-value")
+                .put("grape", "${ENV:GRAPE}")
+                .put("peach", "${ENV:PEACH}")
+                .put("grass", "${ENV:!!!}")
+                .put("pear", "${ENV:X_PEAR}")
+                .put("cherry", "${ENV:X_CHERRY}")
```
findepi: Add a test case with
```
.put("cherry", "${ENV:X_CHERRY}x")
```

expected value will be `"${ENV:X_CHERRY}x"`, since we replace only whole values, no substring replacement.

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -183,18 +192,29 @@ public Injector initialize()
...
+        // combine property sources
+        Map<String, String> properties = new HashMap<>();
         if (optionalConfigurationProperties != null) {
             properties.putAll(optionalConfigurationProperties);
         }
         properties.putAll(requiredProperties);
         properties.putAll(getSystemProperties());
+
+        // replace environment variables in property values
+        properties = replaceEnvironmentVariables(properties, System.getenv(), (key, error) -> {
```
kokosing: It would be nice to have an option to provide (override) the source of environment variables (`Bootstrap::setEnvironmentVarabilesSupplier(Supplier<Map<String,String>>)`)

PR #773, opened 2019-11-08
Validate HTTPS configuration settings

aalbu:
If `http-server.https.enabled` is true, but `http-server.https.keystore.path` is not specified, `HttpServer` blows up with a non-descriptive NPE at startup. Adding a bit of validation to help diagnose the misconfiguration.

### DIFF
--- a/http-server/pom.xml
+++ b/http-server/pom.xml
@@ -193,6 +193,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-core</artifactId>
         </dependency>
--- a/http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
@@ -27,6 +27,7 @@ import io.airlift.units.MaxDataSize;
 import io.airlift.units.MinDataSize;
 import io.airlift.units.MinDuration;
 
+import javax.validation.constraints.AssertTrue;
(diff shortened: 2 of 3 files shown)


http-server/pom.xml
```diff
@@ -192,6 +192,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
```
findepi: please move this test dep below `<artifactId>logback-core</artifactId>` (which is a non-test dep)

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -225,6 +226,12 @@ public HttpServerConfig setKeystorePassword(String keystorePassword)
         return this;
     }
 
+    @AssertTrue(message = "PEM keystore path/password must be provided when HTTPS is enabled")
```
findepi: I think the keystore may be in different format

```suggestion
    @AssertTrue(message = "keystore path/password must be provided when HTTPS is enabled")
```

findepi:
> This fix is fine, but we should really split up the config class so that the HTTPS config part is bound conditionally (and then the keystore could be required)

i understand this is out of scope for this PR, right?

findepi:
@aalbu 
```
[ERROR] src/test/java/io/airlift/http/server/TestHttpServerConfig.java:[26] (imports) ImportOrder: 'java.util.Arrays' should be separated from previous imports.
[ERROR] src/test/java/io/airlift/http/server/TestHttpServerConfig.java:[36] (imports) ImportOrder: Wrong order for 'io.airlift.testing.ValidationAssertions.assertFailsValidation' import.
```

also, please squash commits

findepi:
Merged, thanks!

PR #774, opened 2019-11-10
Add method for duplicating a Distribution object

martint:


stats/src/main/java/io/airlift/stats/DecayCounter.java
```diff
@@ -38,9 +38,20 @@ public DecayCounter(double alpha)
 
     public DecayCounter(double alpha, Ticker ticker)
     {
+        this(0, alpha, ticker, TimeUnit.NANOSECONDS.toSeconds(ticker.read()));
```
findepi:
```suggestion
        this(0, alpha, requireNonNull(ticker, "ticker is null"), TimeUnit.NANOSECONDS.toSeconds(ticker.read()));
```

stats/src/main/java/io/airlift/stats/DecayTDigest.java
```diff
@@ -33,16 +33,20 @@
 
     public DecayTDigest(double compression, double alpha)
     {
-        this(compression, alpha, alpha == 0.0 ? noOpTicker() : Ticker.systemTicker());
+        this(new TDigest(compression), alpha, alpha == 0.0 ? noOpTicker() : Ticker.systemTicker());
     }
 
-    DecayTDigest(double compression, double alpha, Ticker ticker)
+    DecayTDigest(TDigest digest, double alpha, Ticker ticker)
```
findepi: `private`? `@VisibleForTesting`?

stats/src/main/java/io/airlift/stats/DecayTDigest.java
```diff
@@ -151,4 +155,9 @@ public long read()
             }
         };
     }
+
+    public DecayTDigest duplicate()
+    {
+        return new DecayTDigest(TDigest.copyOf(digest), alpha, ticker, landmarkInSeconds);
```
findepi: `TDigest.copyOf(digest)` or `digest.duplicate()`?

stats/src/main/java/io/airlift/stats/DecayCounter.java
martint: I think that muddles the line too much. The call would fail with NPE due to `ticker.read()`, anyway.

stats/src/main/java/io/airlift/stats/DecayTDigest.java
martint: There's no `digest.duplicate()`. I went the other way with that class when I wrote it, but I may change it. For Distribution, it's a lot more natural (and almost necessary) due to the synchronization requirements.

PR #779, opened 2019-11-25
Improve Dense HLL merge performance

martint:


stats/src/test/java/io/airlift/stats/cardinality/BenchmarkDenseHll.java
```diff
@@ -63,6 +72,27 @@ public void initialize()
...
+    public static class MergeData
+    {
+        public DenseHll base;
+        public DenseHll toMerge;
+
+        @Setup(Level.Iteration)
+        public void initialize()
+        {
+            base = new DenseHll(12);
+            for (int i = 0; i < 1_000_000; i++) {
```
sopel39: Is it important that here are `1_000_000` and below `100` only? Should that be parametrized?

stats/src/main/java/io/airlift/stats/cardinality/DenseHll.java
```diff
@@ -249,16 +249,14 @@ public void insert(int bucket, int value)
...
-            if (!setOverflow(bucket, overflow)) {
-                // grow overflows arrays if necessary
-                overflowBuckets = Ints.ensureCapacity(overflowBuckets, overflows + 1, OVERFLOW_GROW_INCREMENT);
-                overflowValues = Bytes.ensureCapacity(overflowValues, overflows + 1, OVERFLOW_GROW_INCREMENT);
-
-                overflowBuckets[overflows] = bucket;
-                overflowValues[overflows] = (byte) overflow;
-                overflows++;
+            int overflowEntry = findOverflowEntry(bucket);
+            if (overflowEntry != -1) {
```
sopel39: Why this makes it faster? It introduces branching

stats/src/main/java/io/airlift/stats/cardinality/DenseHll.java
```diff
@@ -272,28 +270,24 @@ public void insert(int bucket, int value)
         }
     }
 
-    private int getOverflow(int bucket)
+    private int findOverflowEntry(int bucket)
     {
         for (int i = 0; i < overflows; i++) {
             if (overflowBuckets[i] == bucket) {
```
sopel39: should that be some hash map instead?

sopel39:
@martint did you profile HLL after your changes? What takes the most time?

stats/src/main/java/io/airlift/stats/cardinality/BucketListener.java
```diff
@@ -0,0 +1,19 @@
+/*
```
sopel39: Do you have results from before?

stats/src/test/java/io/airlift/stats/cardinality/BenchmarkDenseHll.java
martint: No special meaning other than having a "large" and a "small" set for the comparison with sparse hll in the next commit to be meaningful (sparse can store only a small set).

stats/src/main/java/io/airlift/stats/cardinality/DenseHll.java
martint: `delta > MAX_DELTA` is statistically unlikely, so this branch doesn't hit often. In any case, the old code already had a branch `if (!setOverflow(...))`. This just moves things around a bit.

stats/src/main/java/io/airlift/stats/cardinality/DenseHll.java
martint: Possibly, but overflows are statistically infrequent and there's usually only a few of them. Scanning a few items may be faster than computing a hashcode.

stats/src/main/java/io/airlift/stats/cardinality/BucketListener.java
martint: Before, we'd convert the sparse HLL into a dense HLL and then merge it. So the performance is approximately what the first benchmark showed (there's some extra cost in the conversion that we're not measuring):

`21.557 ± 1.116 us/op` for current implementation,
`11.677 ± 0.788  us/op` after Dense HLL merging is optimized,
`0.320 ± 0.013 us/op` after dedicated merge implementation for Dense vs Sparse

stats/src/main/java/io/airlift/stats/cardinality/HyperLogLog.java
```diff
@@ -84,6 +84,9 @@ public void mergeWith(HyperLogLog other)
             ((SparseHll) instance).mergeWith((SparseHll) other.instance);
             instance = makeDenseIfNecessary((SparseHll) instance);
         }
+        else if (instance instanceof DenseHll && other.instance instanceof SparseHll) {
```
dain: What about the other way around?

martint: In that case, the sparse needs to be converted to dense before the merge can take place. That’s handled by the fall-through case

stats/src/main/java/io/airlift/stats/cardinality/DenseHll.java
sopel39: Right, could you just use array with number of buckets length? Would that help?

martint: That would be a lot of wasted space. Out of, say, 4096 buckets, there might only < 10 that overflow at any given point.

PR #780, opened 2019-11-29
Add TestingClock

findepi:


testing/src/main/java/io/airlift/testing/TestingClock.java
```diff
@@ -0,0 +1,70 @@
...
+import java.time.Clock;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.util.concurrent.TimeUnit;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.util.Objects.requireNonNull;
+
+public final class TestingClock
```
kokosing: Can you please add javadoc?

testing/src/main/java/io/airlift/testing/TestingClock.java
```diff
@@ -0,0 +1,70 @@
...
+        extends Clock
+{
+    private final ZoneId zone;
+    private Instant instant;
+
+    public TestingClock()
+    {
+        this(ZoneOffset.UTC);
```
kokosing: Why do you care about time zone? Is it going to be stored or compared between different time zones?

testing/src/main/java/io/airlift/testing/TestingClock.java
findepi: I don't think it's necessary. What would i write beyond "Testing Clock"?

testing/src/main/java/io/airlift/testing/TestingClock.java
findepi: See `Clock` "interface" (ie what methods are abstract)

PR #783, opened 2019-12-04
Fix sporadic failure in testHttpRequestEvent

electrum:


http-server/src/test/java/io/airlift/http/server/TestHttpServerModule.java
```diff
@@ -364,4 +367,21 @@ protected void service(HttpServletRequest request, HttpServletResponse response)
...
+
+    private static class SingleUseEventClient
+            extends AbstractEventClient
+    {
+        private final SettableFuture<Object> future = SettableFuture.create();
+
+        @Override
+        protected <T> void postEvent(T event)
+        {
+            future.set(event);
```
dain: Throw if already set

PR #785, opened 2019-12-04
Update to Jetty 9.4.24.v20191120

electrum:


CHANGES
```diff
@@ -1,3 +1,7 @@
+0.188
+
+- Remove status message support from HTTP client.
```
martint: What's the motivation?

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -717,7 +717,7 @@ public void testCompressionIsDisabled()
         assertEquals(body, "");
         assertFalse(servlet.getRequestHeaders().containsKey(HeaderName.of(ACCEPT_ENCODING)));
 
-        String json = "{\"foo\":\"bar\",\"hello\":\"world\"}";
+        String json = "{\"fuite\":\"apple\",\"hello\":\"world\"}";
```
martint: Unrelated change?

CHANGES
dain: It isn't really supported anymore in the Jetty clients, and for HTTP/2 you just get `null`.  Also we never use it.

martint: Ok, maybe we should add that to the commit message

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
electrum: No, the minimum size for gzip increased.

PR #786, opened 2019-12-04
Close http client input stream after response handling

### LINKED ISSUE #781: StatusResponseHandler leaks connections
The StatusResponseHandler does not close the input streams on responses. This causes the connection to be held open until the request ultimately times out. If the number of concurrent requests reaches the client's configured limit for connections per host, this can cause requests to queue up.

dain:
Fixes #781

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
@@ -531,6 +531,14 @@ public class JettyHttpClient
             value = responseHandler.handle(request, jettyResponse);
         }
         finally {
+            if (jettyResponse != null) {
+                try {
+                    jettyResponse.getInputStream().close();
+                }
+                catch (IOException ignored) {
+                    // ignore errors closing the stream
+                }
+            }
             if (recordRequestComplete) {
                 recordRequestComplete(stats, request, requestStart, jettyResponse, responseStart);
             }


http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -181,11 +181,12 @@ public void abstractTeardown()
         }
     }
 
-    @Test(enabled = false, description = "This takes over a minute to run")
```
dain: This test takes ~45minutes in CI, but only 10s for each of the 6 sub classes on my laptop, so I think we should disable this before merging.

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -531,6 +533,15 @@ else if ((cause instanceof NoClassDefFoundError) && cause.getMessage().endsWith(
             value = responseHandler.handle(request, jettyResponse);
         }
         finally {
+            if (jettyResponse != null && jettyResponse.getStatusCode() == 200) {
```
electrum: This should be done regardless of the status, i.e. for any response that has a body.

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -531,6 +533,15 @@ else if ((cause instanceof NoClassDefFoundError) && cause.getMessage().endsWith(
...
+                try (InputStream inputStream = jettyResponse.getInputStream()) {
+                    ByteStreams.exhaust(inputStream);
+                }
+                catch (IOException ignored) {
+                    // if the response processing didn't throw, don't throw
```
electrum: This comment is a bit confusing, since it says "if ... didn't throw", but this also prevents overwriting the exception if it did throw. Might be better to simply say
```java
// ignore errors while draining and closing the stream

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
findepi: Are you saying CI is 45x slower?

does it make sense to run the test for at least one of the subclasses?

dain: Yes maybe.  I think the only important one is plain old http/1.  We could also lower the number to like 30k to make the test faster.

dain:
BTW, I'm not sure we want to do the drain at all. For HTTP/2, it doesn't seem to be needed at all.  I'd guess that they kill the internal stream automatically.  For HTTP/1, it could be very expensive to drain the stream.

Regardless, I do think we should add the close code.

dain:
I updated this so it only closes the stream and does not drain.

PR #788, opened 2019-12-10
Limit total size of server.log

s2lomon:
Users expect & need the following semantics of config properties:

- `log.max-history` to limit number of log files
- `log.max-size` to limit the size of log files (compression aside)

However, these get passed to logback where max history is treated as
"number of logging periods to keep". The logging period is derived
from log file name pattern, i.e. day in our case.

This commit introduces additional limit, bringing the actual behavior to
be in line with common expectations.

Documentation reference: https://logback.qos.ch/manual/appenders.html#tbrpTotalSizeCap

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,6 @@
 0.188
 
+- Limit the total size of server.log by log.max-history * log.max-size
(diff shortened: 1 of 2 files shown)


findepi:
Merged, thanks!

PR #789, opened 2019-12-26
Get jquery script from https instead of http

tooptoop4:
Get jquery script from https instead of http

https://github.com/prestodb/presto/issues/13693


findepi:
Merged as 1235245bd8c875f2924cc77cc5d78b81e721c9cd, thanks!

PR #799, opened 2020-01-08
Support AutoCloseable in Closeables

kokosing:
Support AutoCloseable in Closeables

### DIFF
--- a/testing/src/main/java/io/airlift/testing/Closeables.java
+++ b/testing/src/main/java/io/airlift/testing/Closeables.java
@@ -1,7 +1,9 @@
 package io.airlift.testing;
 
-import java.io.Closeable;
-import java.io.IOException;
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
+import java.util.Optional;
 
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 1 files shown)


testing/src/main/java/io/airlift/testing/Closeables.java
```diff
@@ -1,8 +1,5 @@
 package io.airlift.testing;
 
-import java.io.Closeable;
-import java.io.IOException;
-
 import static java.util.Objects.requireNonNull;
 
 public final class Closeables
```
kokosing: Should I rename the class?

testing/src/main/java/io/airlift/testing/Closeables.java
```diff
@@ -1,8 +1,5 @@
 package io.airlift.testing;
```
kokosing: BTW in Presto there are 8 such methods defined:
```
cox@Grzegorzs-MacBook-Pro:~/presto$ git grep 'void closeQuietly' | wc -l
       8
```

testing/src/main/java/io/airlift/testing/Closeables.java
```diff
@@ -11,36 +8,36 @@ private Closeables()
...
         }
-        for (Closeable closeable : closeables) {
+        for (AutoCloseable closeable : closeables) {
             try {
                 if (closeable != null) {
                     closeable.close();
                 }
             }
-            catch (IOException | RuntimeException ignored) {
+            catch (Exception ignored) {
```
findepi: you shouldn't ignore InterruptedException

kokosing: Thanks. Updated.

testing/src/main/java/io/airlift/testing/Closeables.java
electrum: Nah, we can keep it the same

testing/src/main/java/io/airlift/testing/Closeables.java
```diff
@@ -11,36 +8,44 @@ private Closeables()
...
+        for (AutoCloseable closeable : closeables) {
             try {
                 if (closeable != null) {
                     closeable.close();
                 }
             }
-            catch (IOException e) {
+            catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw e;
```
electrum: I think we should still continue to close everything, since these methods are used to prevent resource leaks. For `InterruptedException`, it probably makes sense to make this one the root exception. You could change the method to collect the exceptions into a list, then grab the first `InterruptedException` as the root one (if any).

testing/src/main/java/io/airlift/testing/Closeables.java
```diff
@@ -91,18 +96,22 @@ else if (rootCause != e) {
...
-        for (Closeable closeable : closeables) {
+        for (AutoCloseable closeable : closeables) {
             try {
                 if (closeable != null) {
                     closeable.close();
                 }
             }
+            catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException(e);
```
electrum: Same here, need to close everything

kokosing:
Updated. Thanks for the review!

testing/src/main/java/io/airlift/testing/Closeables.java
```diff
@@ -11,98 +13,92 @@ private Closeables()
     {
     }
 
-    public static void closeQuietly(Closeable... closeables)
+    public static void closeQuietly(AutoCloseable... closeables)
```
findepi, line 16: let's try to avoid this method -- https://prestosql.slack.com/archives/CP1MUNEUX/p1586372796137000

findepi:
What about letting JVM do the heavy lifting with exception handling?

```
public static void closeQuietly(AutoCloseable... closeables)
        throws Exception
{
    if (closeables == null) {
        return;
    }
    closeQuietly(closeables, 0);
}

private static void closeQuietly(AutoCloseable[] closeables, int currentIndex)
        throws Exception
{
    if (currentIndex >= closeables.length) {
        return;
    }
    try (AutoCloseable current = closeables[currentIndex]) {
        closeQuietly(closeables, currentIndex + 1);
    }
}
```

PR #801, opened 2020-01-09
Deprecate findUnusedPort

findepi:


findepi:
it's still used. I want to deprecate this as a reminder for ourselves.
Otherwise the code looks like everything is all right.

electrum:
This is used to attempt to find an unused port on which to start the JMX agent (if no port is configured). According to @martint, if you start the agent without giving it a port, it will correctly choose an unused port, but there is no way to find this port later.

It only makes sense to deprecate this if the JDK had a better way of doing this, as deprecation means it is going away and should be replaced by something else. I'd be fine if you want to rename the method to something else to make the behavior more obvious to readers.

PR #804, opened 2020-01-14
Use Servlet filter for static resources

dain:
Replace Jetty specific Handlers with generic Servlet filters to serve static resources


http-server/src/main/java/io/airlift/http/server/ClassPathResourceFilter.java
```diff
@@ -40,8 +41,8 @@
  * Intended to serve a couple of static files e.g. for javascript or HTML.
  */
 // Forked from https://github.com/NessComputing/components-ness-httpserver/
-public class ClassPathResourceHandler
-        extends AbstractHandler
+public class ClassPathResourceFilter
+        extends HttpFilter
```
martint: Why is this a filter and not a servlet?

dain: It can be a servlet because some times we register other servlets with conflicting paths. We have resources act as an overlay currently.

PR #805, opened 2020-01-17
Allow HTTP client PRESERVE_AUTHORIZATION_KEY to be unset

dain:


http-client/src/main/java/io/airlift/http/client/jetty/AuthorizationPreservingHttpClient.java
```diff
@@ -45,6 +46,6 @@ public static void setPreserveAuthorization(Request request, boolean preserveAut
 
     private static boolean isPreserveAuthorization(Request request)
     {
-        return (boolean) request.getAttributes().get(PRESERVE_AUTHORIZATION_KEY);
+        return request.getAttributes().get(PRESERVE_AUTHORIZATION_KEY) == TRUE;
```
electrum: I'd use `equals` to make it clear we don't care about identity semantics (readers won't have to verify the value that is being set), but it's fine either way
```java
true.equals(request.getAttributes().get(PRESERVE_AUTHORIZATION_KEY));

PR #807, opened 2020-01-21
Do not require keystore password for HTTPS

electrum:
PEM files used as the keystore do not require a password.

The validation introduced in https://github.com/airlift/airlift/pull/773 was too strict.

### DIFF
--- a/http-server/src/test/java/io/airlift/http/server/TestHttpServerConfig.java
+++ b/http-server/src/test/java/io/airlift/http/server/TestHttpServerConfig.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Map;
 
 import static io.airlift.testing.ValidationAssertions.assertFailsValidation;
+import static io.airlift.testing.ValidationAssertions.assertValidates;
 import static io.airlift.units.DataSize.Unit.GIGABYTE;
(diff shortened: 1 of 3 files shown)


http-server/src/test/java/io/airlift/http/server/TestHttpServerConfig.java
```diff
@@ -182,15 +183,19 @@ public void testExplicitPropertyMappings()
...
     {
-        assertFailsValidation(
+        assertValidates(
                 new HttpServerConfig()
                         .setHttpsEnabled(true)
-                        // keystore path not set
-                        .setKeystorePassword("keystore password"),
+                        .setKeystorePath("/test/keystore"));
+
+        // keystore path not set
```
findepi: technically  this is redundant

http-server/src/test/java/io/airlift/http/server/TestHttpServerConfig.java
```diff
@@ -182,15 +183,19 @@ public void testExplicitPropertyMappings()
     }
 
     @Test
-    public void testInvalidHttpsConfiguration()
+    public void testHttpsConfigurationValid()
```
findepi:
```suggestion
    public void testHttpsConfigurationValidation()
```

(you're testing both valid and invalid cases)

PR #809, opened 2020-01-23
Require host to be set in HttpUriBuilder

electrum:


CHANGES
```diff
@@ -1,6 +1,7 @@
 0.191
 
 - Do not require keystore password for HTTPS.
+- Require host to be set in HttpUriBuilder.
```
findepi: do you want to indicate this is backwards-incompatible change?

http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
```diff
@@ -68,6 +68,8 @@ public static HttpUriBuilder uriBuilder()
     public static HttpUriBuilder uriBuilderFrom(URI uri)
     {
         requireNonNull(uri, "uri is null");
+        checkArgument(uri.getScheme() != null, "URI does not have a scheme: %s", uri);
+        checkArgument(uri.getHost() != null, "URI does not have a host: %s", uri);
```
findepi: Is there much advantage of verifying early, not in `build()`?

usually, we don't validate in builders' setters, deferring until final call.
This works since builders are typically short-lived.

http-client/src/test/java/io/airlift/http/client/TestHttpUriBuilder.java
```diff
@@ -348,10 +349,28 @@ public void testEncodesPath()
...
+    }
+
+    @Test
+    public void testVerifyOnCreate()
+    {
+        assertThatThrownBy(() -> uriBuilderFrom(URI.create("./foo")))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessage("URI does not have a scheme: ./foo");
+
+        assertThatThrownBy(() -> uriBuilderFrom(URI.create("http://test_foo/abc")))
```
findepi: Add a comment why `test_foo` is not considered a host

findepi:
`URI` doesn't "support" hostnames with underscore, but it still "kind of supports" it --

```java
URI uri = URI.create("http://test_foo/abc");
System.out.println(uri.getHost()); // null
System.out.println(uri.getAuthority()); // test_foo

uri = URI.create("http://test_foo:123/abc");
System.out.println(uri.getHost()); // null
System.out.println(uri.getAuthority()); // test_foo:123
```

Is this a viable solution?

CHANGES
electrum: We don't typically go into that much detail for small changes. I think it's clear from the description that it will now fail in this (edge) case.

http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
electrum: The reason to verify here is that we can print the invalid URI object. We couldn't think of a valid use case to create an `HttpUriBuilder` from a URI where the scheme or host is not set (as that would be something like `http://` or a completely different type of URI ... the JDK bug not withstanding).

http-client/src/test/java/io/airlift/http/client/TestHttpUriBuilder.java
electrum: Done

electrum:
I see two options to handle this:

1. Parse the raw authority into host and port in `HttpUriBuilder`. This is complicated because you have to handle decoding, username and password, etc. Getting this right would be quite complex (take a look at the code in `URI`).

2. Change `HttpUriBuilder` to preserve authority. This is what the Jersey implementation of the JAX-RS `UriBuilder` does. (The Jersey behavior doesn't seem to be required according to the Javadoc, but it might be covered by the JAX-RS TCK.)

I'm inclined to call this patch good enough, given it's a bug in `URI`. For Presto, we can change most/all usages to the JAX-RS `UriBuilder`.

electrum:
Looks like there is another potential solution according to [JDK-8221675](https://bugs.openjdk.java.net/browse/JDK-8221675):
```java
String host = uri.getHost();
if (host == null) {
    host = uri.toURL().getHost()
}
```
But I don't know the implications of doing this.

findepi:
> I'm inclined to call this patch good enough, given it's a bug in `URI`. For Presto, we can change most/all usages to the JAX-RS `UriBuilder`.

Then maybe we should deprecate the class.
It's correct from RFC perspective (undescores aren't valid) but sometimes fails (underscroes happen)

PR #813, opened 2020-02-06
Add Scale function to QuantileDigest

TrongDucNguyen:
This function scale all the counts by a given factor.


martint:
Thanks for submitting this change. Can you describe the use case and motivation?

TrongDucNguyen:
I withdraw this, as I commit it to another folk

PR #814, opened 2020-02-10
Provide clear error messaging when Java is not installed

alexjo2144:
If Java is not found `launcher.py` fails with an error message `[Errno 2] No such file or directory`, this adds an explicit error message for that case.

### DIFF
--- a/launcher/src/main/scripts/bin/launcher.py
+++ b/launcher/src/main/scripts/bin/launcher.py
@@ -3,6 +3,7 @@
 import errno
 import os
 import platform
+import subprocess
 import sys
 import traceback
 
@@ -188,6 +189,11 @@ def build_java_execution(options, daemon):
         raise Exception('Launcher config file is missing: %s' % options.launcher_config)
     if options.log_levels_set and not exists(options.log_levels):
         raise Exception('Log levels file is missing: %s' % options.log_levels)
+
+    try:
+        subprocess.check_output(['java', '-version'], stderr=subprocess.STDOUT)
+    except (OSError, subprocess.CalledProcessError):
+        raise Exception('Java is not installed')
 
     properties = options.properties.copy()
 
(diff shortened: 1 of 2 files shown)


electrum:
Thanks. Did you test this on both Python 2 and 3?

launcher/src/main/scripts/bin/launcher.py
```diff
@@ -5,6 +5,7 @@
 import platform
 import sys
 import traceback
+import subprocess
```
electrum: Sort imports

launcher/src/main/scripts/bin/launcher.py
```diff
@@ -189,6 +190,11 @@ def build_java_execution(options, daemon):
     if options.log_levels_set and not exists(options.log_levels):
         raise Exception('Log levels file is missing: %s' % options.log_levels)
 
+    try:
+        subprocess.check_call(['java -version'])
```
electrum: I don't see how this works, since it's trying to run a binary named `java -version` (as `shell` defaults to `False`). It should be
```python
subprocess.check_call(['java', '-version'])

electrum: Also, `check_call` doesn't handle the output, so it will be passed through. While showing the Java version at startup might be useful, I'd rather not make a behavioral change like that as part of this change.

We can handle it like this:
```python
subprocess.check_output(['java', '-version'], stderr=subprocess.STDOUT)

alexjo2144: Sorry that was dumb. I tested that the error was thrown when java wasn't installed but didn't make sure it wasn't thrown when it is. Will fix.

alexjo2144: Thanks. Updated and tested both cases on python 2 and 3.

electrum:
Thanks!

PR #818, opened 2020-02-14
Fix DecayTDigest#rescale handling of max of means

BlakeMesdag:
Discovered this while investigating another issue, seems to have just been overlooked from the original implementations review.


martint:
Good catch and thanks for the fix! Can you fix the checkstyle issues before I merge this?

> [ERROR] src/test/java/io/airlift/stats/TestDecayTDigest.java:[8] (imports) ImportOrder: Wrong order for 'java.util.concurrent.TimeUnit' import.
> [ERROR] src/test/java/io/airlift/stats/TestDecayTDigest.java:[10,31] (blocks) LeftCurly: '{' at column 31 should be on a new line.
> [ERROR] src/test/java/io/airlift/stats/TestDecayTDigest.java:[12,37] (blocks) LeftCurly: '{' at column 37 should be on a new line.

BlakeMesdag:
@martint style issues should be fixed, the tests pass locally without checkstyle violations now.

tooptoop4:
ship it!

PR #821, opened 2020-03-05
Fix launcher for Python < 2.7

electrum:


launcher/src/main/scripts/bin/launcher.py
```diff
@@ -190,10 +190,11 @@ def build_java_execution(options, daemon):
     if options.log_levels_set and not exists(options.log_levels):
         raise Exception('Log levels file is missing: %s' % options.log_levels)
 
-    try:
-        subprocess.check_output(['java', '-version'], stderr=subprocess.STDOUT)
-    except (OSError, subprocess.CalledProcessError):
-        raise Exception('Java is not installed')
+    with open(os.devnull, 'w') as devnull:
```
findepi: Technically this part (and `stdout=devnull` below) seems unrelated, so this could be 2 commits.

electrum: How is this unrelated? The old method returns the output. This one just passes it through, so we need to discard the output to have the same behavior.

PR #822, opened 2020-03-09
Add CacheStatsMBean for reporting Guava cache statistics

alexjo2144:
Taken from Presto https://github.com/prestosql/presto/blob/master/presto-main/src/main/java/io/prestosql/sql/gen/CacheStatsMBean.java for availability in more places.


findepi:
Merged, thanks!

PR #824, opened 2020-03-20
Forbid direct binder.install()

findepi:
Child classes need access to `binder`, but they should not use its
`install` method. Instead, the `AbstractInvocationHandler#install`
should be used, to make sure configuration is bound properly.

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/AbstractConfigurationAwareModule.java
+++ b/configuration/src/main/java/io/airlift/configuration/AbstractConfigurationAwareModule.java
@@ -14,8 +14,12 @@
 package io.airlift.configuration;
 
 import com.google.common.annotations.Beta;
+import com.google.common.reflect.AbstractInvocationHandler;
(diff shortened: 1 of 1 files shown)


findepi:
cc @kokosing @electrum

configuration/src/main/java/io/airlift/configuration/AbstractConfigurationAwareModule.java
```diff
@@ -68,4 +72,45 @@ protected synchronized void install(Module module)
...
+            catch (NoSuchMethodException e) {
+                throw new RuntimeException(e);
+            }
+        }
+
+        static Binder proxy(Binder binder)
+        {
+            return (Binder) Proxy.newProxyInstance(
+                    ForbidInstallBinder.class.getClassLoader(),
+                    new Class[] {Binder.class},
```
electrum, line 94: This needs `Class<?>`

PR #825, opened 2020-03-23
Allow passing additional JVM options in launcher

findepi:


launcher/src/main/scripts/bin/launcher.py
```diff
@@ -363,6 +363,7 @@ def create_parser():
     parser.add_option('--pid-file', metavar='FILE', help='Defaults to DATA_DIR/var/run/launcher.pid')
     parser.add_option('--launcher-log-file', metavar='FILE', help='Defaults to DATA_DIR/var/log/launcher.log (only in daemon mode)')
     parser.add_option('--server-log-file', metavar='FILE', help='Defaults to DATA_DIR/var/log/server.log (only in daemon mode)')
+    parser.add_option('-J', action='append', metavar='OPT', dest='jvm_options', help='Pass Java options')
```
electrum, line 366: Help should be `Set a JVM option` since this is for a single option

launcher/src/main/scripts/bin/launcher.py
```diff
@@ -440,6 +441,7 @@ def main():
     o.launcher_log = realpath(options.launcher_log_file or pathjoin(o.data_dir, 'var/log/launcher.log'))
     o.server_log = realpath(options.server_log_file or pathjoin(o.data_dir, 'var/log/server.log'))
 
+    o.jvm_options = options.jvm_options or ()
```
electrum, line 444: Let's move this above around line 428, since this code block is currently just for properties, and this is unrelated.

electrum: This needs to be `[]` or it fails with
```
can only concatenate list (not "tuple") to list

electrum:
Merged, thanks!

findepi:
thanks @electrum!

PR #830, opened 2020-04-07
New rolling log file handler

dain:


log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,579 @@
...
+            LegacyRollingFileHandler.recoverTempFiles(filename);
+        }
+        catch (IOException e) {
+            reportError(null, e, OPEN_FAILURE);
+        }
+        if (Files.exists(link)) {
+            try {
+                // if existing link file is a legacy log file, rename it to so link file can be recreated as a symlink
+                BasicFileAttributes attributes = Files.readAttributes(link, BasicFileAttributes.class);
+                if (attributes.isRegularFile()) {
```
martint, line 159: We may also want to check if it's a directory and just fail in that case.

log-manager/src/main/java/io/airlift/log/LogFileName.java
```diff
@@ -0,0 +1,247 @@
...
+    private static final DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter.ofPattern("-yyyyMMdd.HHmmss");
+
+    private final String fileName;
+    private final LocalDateTime dateTime;
+    private final OptionalInt index;
+    private final OptionalInt legacyIndex;
+    private final Optional<String> slug;
+    private final boolean compressed;
+
+    public static Optional<LogFileName> forName(String baseName, String fileName)
```
martint, line 46: This needs an explanation -- it's not clear what's the purpose of this method. What's `baseName` vs `fileName`? What does a result of `Optional.empty()` mean? The method is way too complex to expect someone to reverse engineer those answers from it.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,579 @@
...
+                else {
+                    continue;
+                }
+                if (LogFileName.forName(masterLogFile.getFileName().toString(), fileNameWithoutPrefix).isPresent()) {
+                    // this is our temp or "to be deleted' file, so try to delete it
+                    try {
+                        Files.deleteIfExists(file);
+                    }
+                    catch (IOException ignored) {
+                    }
```
martint, line 551: Do we want to `reportError` here?

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,579 @@
...
+                }
+            }
+        }
+        catch (IOException ignored) {
+        }
```
martint, line 556: Do we want to `reportError` here?

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,579 @@
...
+
+    @GuardedBy("this")
+    private Path currentOutputFile;
+    @GuardedBy("this")
+    private LogFileName currentOutputFileName;
+    @GuardedBy("this")
+    private OutputStream currentOutputStream;
+    @GuardedBy("this")
+    private long currentFileSize;
+    private final LogHistoryManager logHistoryManager;
```
martint, line 106: I'd name this just `historyManager`. It's understood that it's for logs from the context.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,579 @@
...
+
+    private synchronized void rollFile()
+            throws IOException
+    {
+        // carefully update the stream, so if there is a problem logging can continue
+
+        LogFileName newFileName = null;
+        Path newFile = null;
+        OutputStream newOutputStream = null;
+        for (int i = 0; i < 100; i++) {
```
martint, line 393: Is the purpose of this loop to retry if we see a `FileAlreadyExistsException`? In that case, think it will be ineffective. It will run all the iterations in an instant, so if the intent is to wait until the file "magically" disappears, there won't be enough time for anything meaningful to happen in between iterations. 

Also, is there any significance to the `100` limit? Maybe make it a constant for self-documentation purposes.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,579 @@
...
+            try {
+                newFileName = LogFileName.newLogFileName(link, compressionType.getExtension());
```
martint, line 395: Oh, now I think I see what's going on (I'll leave my comment above for context). I think the problem is that it's not at all clear that `newLogFileName` returns a randomized name. 

Two things contribute to that misunderstanding: 
* the name doesn't suggest that the method does anything other than compose the name and the extension. 
* the method is static, so it's not clear that what state (if any) goes into that randomization.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,579 @@
...
+        }
+
+        // If a new file can not be opened, abort and continue using the existing file
+        if (newOutputStream == null) {
+            throw new IOException("Could not create new a unique log file");
+        }
+
+        // The new file is open, so we will always switch to this new output stream
+        // If any error occurs, with the cleanup steps, we add them to this exception as suppressed and throw at the end
+        IOException exception = new IOException(String.format("Unable to %s log file", currentOutputStream == null ? "setup initial" : "roll"));
```
martint, line 411: Maybe add the file name to the exception message.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,579 @@
...
+
+        if (currentOutputStream != null) {
+            try {
+                currentOutputStream.close();
+            }
+            catch (IOException e) {
+                exception.addSuppressed(new IOException("Unable to close old output stream", e));
```
martint, line 418: Add `currentOutputFile` to the exception message.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,579 @@
...
+            }
+            logHistoryManager.addFile(currentOutputFile, currentOutputFileName, currentFileSize);
+            if (compressionExecutor != null) {
+                Path originalFile = currentOutputFile;
+                LogFileName originalLogFileName = currentOutputFileName;
+                long originalFileSize = currentFileSize;
+                compressionExecutor.submit(() -> compressInternal(originalFile, originalLogFileName, originalFileSize));
```
martint, line 425: It's probably ok for now, but I think it'd be cleaner and easier to follow the code if this was done unconditionally. The `compressInternal` method would skip if compression is not needed.

BTW, the fact that this method (`rollFile`) is being called in the constructor before `compressionExecutor` is set, and that the behavior of this method depends on whether `currentOutputStream` is set (which doesn't happen on the first run) makes it hard to reason about the interaction between all the parts. I don't have better ideas right now -- hopefully will develop some as I get a better picture of how everything fits together -- but you may want to think about whether that could be simplified.

dain: in the case of the constructor the `currentOutputStream` will be `null`. I added comment explaining that.

dain: I agree, but I don't want the unused compressor when we are not compressing.  Initially, I tried to put the compression in a separate manager, but the concurrency of this code makes adding that abstraction nearly impossible.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
dain: I wasn't sure.  This method gets called every time we compress, so if there is a problem that gets fixed we cleanup.

log-manager/src/main/java/io/airlift/log/LogFileName.java
dain: I added comments and renamed the method. Let me know what you think.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
dain: I renamed the method to `generateNextLogFileName` and added the constant `MAX_OPEN_NEW_LOG_ATTEMPTS`

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
dain: Added a constant `MAX_OPEN_NEW_LOG_ATTEMPTS`

log-manager/src/main/java/io/airlift/log/LogFileName.java
```diff
@@ -0,0 +1,255 @@
...
+            if (parts.size() != 4) {
+                return Optional.empty();
+            }
+
+            LocalDateTime dateTime;
+            try {
+                dateTime = LocalDateTime.of(parseInt(parts.get(0)), parseInt(parts.get(1)), parseInt(parts.get(2)), 0, 0);
+            }
+            catch (NumberFormatException | DateTimeException e) {
+                return Optional.empty();
```
findepi, line 88: a single regex 
```
(?<date>\d\d\d\d-\d\d-\d\d).\d.log(?<compressed>.gz)?
```

would yield to shorter & more readable code.

Is this piece considered performance critical?

log-manager/src/main/java/io/airlift/log/LogHistoryManager.java
```diff
@@ -0,0 +1,207 @@
...
+    public synchronized void pruneLogFilesIfNecessary(long otherDataSize)
+    {
+        while (totalSize + otherDataSize > maxTotalSize) {
+            LogFile logFile = files.poll();
+            if (logFile == null) {
+                break;
+            }
+            totalSize -= logFile.getSize();
+            try {
+                Files.deleteIfExists(logFile.getPath());
```
findepi, line 92: Add a comment why
- you allow file not to exist (eg file could be deleted externally and we should not fail on this)
- you subtract from `totalSize` even if file did not exist (eg files deleted externally still count towards size quota, until we decide to remove them)

log-manager/src/main/java/io/airlift/log/LogFileName.java
```diff
@@ -0,0 +1,255 @@
...
+        implements Comparable<LogFileName>
+{
+    private static final DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter.ofPattern("-yyyyMMdd.HHmmss");
+    private static final int MAX_GENERATED_INDEX = 1000;
+
+    private final String fileName;
+    private final LocalDateTime dateTime;
+    private final OptionalInt index;
+    private final OptionalInt legacyIndex;
+    private final Optional<String> slug;
```
findepi, line 44: what the slug is for? 
if it's for differentiating different log streams, then i think it should come at the beginning, not close to the end?

log-manager/src/main/java/io/airlift/log/LoggingConfiguration.java
```diff
@@ -16,19 +16,28 @@
...
+import static io.airlift.units.DataSize.Unit.GIGABYTE;
 import static io.airlift.units.DataSize.Unit.MEGABYTE;
 
+@DefunctConfig("log.max-size-in-bytes")
 public class LoggingConfiguration
 {
     private boolean consoleEnabled = true;
     private String logPath;
     private DataSize maxSize = new DataSize(100, MEGABYTE);
+    private DataSize maxTotalSize = new DataSize(1, GIGABYTE);
```
findepi, line 36: for maxSize 100M and maxTotalSize 1G, shouldn't maxHistory be initialized to 10?

if the rollover is time and/or size -based (whatever triggers first), maybe just add this as a comment

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -1,89 +1,231 @@
...
 {
-    private static final String TEMP_FILE_EXTENSION = ".tmp";
-    private static final String LOG_FILE_EXTENSION = ".log";
-    private static final FileSize BUFFER_SIZE_IN_BYTES = new FileSize(new DataSize(1, MEGABYTE).toBytes());
+    private static final int MAX_OPEN_NEW_LOG_ATTEMPTS = 100;
+
+    public enum CompressionType
+    {
+        NONE(Optional.empty()),
+        GZIP(Optional.of(".gz"));
```
findepi, line 73: or maybe

NONE("")
GZIP(".gz")

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -92,74 +234,352 @@ public void publish(LogRecord record)
...
+            putUninterruptibly(queue, message);
         }
         catch (Exception e) {
             // catch any exception to assure logging always works
             reportError(null, e, WRITE_FAILURE);
         }
+
+        // if closed while queueing, try to remove the message just to clean up
+        if (closed.get()) {
+            queue.remove(message);
```
findepi, line 246: `queue.clear()` ?

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -92,74 +234,352 @@ public void publish(LogRecord record)
...
+    public synchronized void flush()
     {
+        if (currentOutputStream != null) {
+            try {
+                currentOutputStream.flush();
+            }
+            catch (Exception e) {
+                reportError(null, e, FLUSH_FAILURE);
+            }
+        }
```
findepi, line 260: should this flush `queue` too?

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -92,74 +234,352 @@ public void publish(LogRecord record)
...
         try {
-            asyncAppender.stop();
+            thread.join();
         }
-        catch (Exception e) {
-            // catch any exception to assure logging always works
-            reportError(null, e, CLOSE_FAILURE);
+        catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        }
```
findepi, line 277: this is erroneous situation and not handled correctly by the following code, so
log (stderr) that things are going wrong

(or add a comment why not doing so)

findepi: same for the other catch below

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -1,89 +1,231 @@
...
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
 
-        // Limit total log files occupancy on disk. Ideally we would keep exactly
-        // `maxHistory` files (not logging periods). This is closest currently possible.
-        rollingPolicy.setTotalSizeCap(new FileSize(LongMath.saturatedMultiply(maxSizeInBytes, maxHistory)));
+        // convert from legacy logger
+        try {
+            LegacyRollingFileHandler.recoverTempFiles(filename);
```
findepi, line 153: do we need this?

log-manager/src/main/java/io/airlift/log/LogFileName.java
dain: The slug is allowed sometime present when there are too man name collisions.  This should never actually happen, but we need to fail safe.  It goes at the end because it is last in the sort order.

log-manager/src/main/java/io/airlift/log/LoggingConfiguration.java
dain: maxHistory is not enforced for the new logger.  It is only present for the legacy logger.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
dain: I added this for testing, and then didn't need it (is it called internal). The problem is since it is called internally, trying to flush the queue during a write, would be a recursion problem. I don't think it is worth implementing.

The current logger simply ignores this method (it is empty).

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
dain: The rules for interruption is you should (with very few exception), throw an InterruptedException or set the interrupted flag on the thread.

The code here attempts to do a clean shutdown unless the interrupted flag is set, and in that case we cleanup as quickly as possible.

log-manager/src/main/java/io/airlift/log/LogFileName.java
```diff
@@ -0,0 +1,255 @@
...
+    private final boolean compressed;
+
+    /**
+     * Attempts to parse a log file name that is part of the history files for the master log file.
```
electrum, line 48: Add examples of what is being parsed

log-manager/src/main/java/io/airlift/log/LogHistoryManager.java
```diff
@@ -0,0 +1,207 @@
...
+                    .map(this::createLogFile)
+                    .filter(Optional::isPresent)
+                    .map(Optional::get)
+                    .forEach(files::add);
+            totalSize = files.stream()
+                    .mapToLong(LogFile::getSize)
+                    .sum();
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException("Unable to list existing log files", e);
```
electrum, line 66: Maybe include the path here

log-manager/src/main/java/io/airlift/log/LogHistoryManager.java
```diff
@@ -0,0 +1,207 @@
...
+            }
+            totalSize -= logFile.getSize();
+            try {
+                Files.deleteIfExists(logFile.getPath());
+            }
+            catch (IOException ignored) {
+            }
+        }
+
+        // just to be safe, re-calculate the total closed size
```
electrum, line 98: What is "closed"?

log-manager/src/main/java/io/airlift/log/LogHistoryManager.java
```diff
@@ -0,0 +1,207 @@
...
+        totalSize = files.stream()
+                .mapToLong(LogFile::getSize)
+                .sum();
+    }
+
+    private Optional<LogFile> createLogFile(Path path)
+    {
+        Optional<LogFileName> logFileName = LogFileName.parseHistoryLogFileName(masterLogFile.getFileName().toString(), path.getFileName().toString());
```
electrum, line 106: Static import

log-manager/src/main/java/io/airlift/log/LoggingConfiguration.java
electrum: We could split the config and bind conditionally based on legacy/new, but that's probably overkill.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -1,89 +1,231 @@
...
+        }
+
+        public Optional<String> getExtension()
+        {
+            return extension;
+        }
+    }
+
+    private static final int MAX_BATCH_COUNT = 1024;
+    private static final int MAX_BATCH_BYTES = toIntExact(new DataSize(1, MEGABYTE).toBytes());
```
electrum, line 89: This is deprecated, use `DataSize.of(1, MEGABYTE)`

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,579 @@
...
+    {
+        requireNonNull(filename, "filename is null");
+        requireNonNull(maxFileSize, "maxFileSize is null");
+        requireNonNull(maxTotalSize, "maxTotalSize is null");
+        requireNonNull(compressionType, "compressionType is null");
+
+        this.maxFileSize = maxFileSize.toBytes();
+        this.compressionType = compressionType;
+
+        link = Paths.get(filename);
```
electrum, line 135: Maybe name this `symlink`

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,579 @@
...
+        // open initial log file
+        try {
+            rollFile();
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+
+        if (compressionType != CompressionType.NONE) {
+            compressionExecutor = Executors.newSingleThreadExecutor(new ThreadFactoryBuilder()
```
electrum, line 189: Static import

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -0,0 +1,579 @@
...
+
+    private synchronized void logMessageBatch(List<byte[]> batch)
+    {
+        for (byte[] message : batch) {
+            if (currentFileSize + message.length > maxFileSize) {
+                try {
+                    rollFile();
+                }
+                catch (IOException e) {
+                    // it is possible the roll worked, but there was a problem cleaning up, and it is possible it failed
```
electrum, line 369: Make these complete sentences

log-manager/src/main/java/io/airlift/log/LogHistoryManager.java
```diff
@@ -0,0 +1,207 @@
...
+import java.util.Optional;
+import java.util.PriorityQueue;
+import java.util.Set;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static java.util.Objects.requireNonNull;
+
+@ThreadSafe
+final class LogHistoryManager
```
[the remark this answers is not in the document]
findepi: I assumed this is a conscious decision (https://github.com/airlift/airlift/pull/830#discussion_r405335786)

Keeping the state in memory and not refreshing should lead only to "overestimating" when files are deleted externally.
Overestimating here isn't harmful though (we're just going to still delete old log files even though newer log files were deleted externally... which doesn't sound wrong)

log-manager/src/main/java/io/airlift/log/LogHistoryManager.java
```diff
@@ -0,0 +1,207 @@
...
+            }
+            catch (IOException ignored) {
+            }
+        }
+
+        // just to be safe, re-calculate the total closed size
+        totalSize = files.stream()
+                .mapToLong(LogFile::getSize)
+                .sum();
+    }
```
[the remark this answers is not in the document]
dain: I believe this has been explained above

log-manager/src/main/java/io/airlift/log/LoggingConfiguration.java
dain: Yes, I think it is overkill.  In a few months we can remove the old logging system and the config.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
dain: I actually like the optional better as it communicates that there may not be an extension.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
dain: We support upgrading from the old code, and the old code may have left over temp files that have to be cleaned up.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
dain: Can't clear the whole queue here, there could be messages in the queue that still need to be logged.

log-manager/src/main/java/io/airlift/log/RollingFileHandler.java
```diff
@@ -1,89 +1,235 @@
...
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
 
-        rollingPolicy.start();
-        triggeringPolicy.start();
-        fileAppender.start();
-        asyncAppender.start();
+        if (compressionType != CompressionType.NONE) {
+            compressionExecutor = newSingleThreadExecutor(new ThreadFactoryBuilder()
```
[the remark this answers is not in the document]
dain: It can't since concurrent uses logging

PR #831, opened 2020-04-08
Publish Maven BOM

electrum:


electrum:
I realized that the `airlift` root POM actually works as a BOM, but it is different from an explicit BOM in that the import includes everything in `dependencyManagement` (such as things used only for specific modules or testing) and in Airbase.

PR #832, opened 2020-04-15
Increase buffer size for t-digest

martint:


stats/src/test/java/io/airlift/stats/TestTDigest.java
```diff
@@ -257,6 +260,25 @@ public void testMerge()
         assertEquals(merged.valueAt(1), 8.0);
     }
 
+    @Test
+    public void testUnmergeable()
+            throws IOException
+    {
+        // Tests an edge case where centroids cannot be merged and an attempting to add another value fails because of lack of room in the buffer
+        byte[] serialized = Resources.toByteArray(Resources.getResource("io/airlift/stats/unmergeable-tdigest"));
```
electrum, line 268: This can be
```java
getResource(getClass(), "unmergeable-tdigest")

File: stats/src/test/java/io/airlift/stats/TestTDigest.java [hunk 2]
```diff
@@ -257,6 +260,25 @@ public void testMerge()
         assertEquals(merged.valueAt(1), 8.0);
     }
 
+    @Test
+    public void testUnmergeable()
+            throws IOException
```
electrum, line 265: We normally just do `throws Exception` for test methods

martint: I generally try to use the most specific exception the test can throw (plus, that's what intellij adds when it's missing the required declaration)

stats/src/test/java/io/airlift/stats/TestTDigest.java
martint: I want to be explicit in case this class moves. Also, it's easier to "find" the file by copying and pasting the path. In reality, I could just put it at the top-level, since this is local to the module.

PR #833, opened 2020-04-15
Disable verification of algorithms not supported by bouncy

dain:


security/src/test/java/io/airlift/security/csr/TestSignatureAlgorithmIdentifier.java
```diff
@@ -32,7 +32,14 @@ public void test()
             SignatureAlgorithmIdentifier signatureAlgorithmIdentifier = entry.getValue();
             assertEquals(signatureAlgorithmIdentifier.getName(), entry.getKey());
 
-            AlgorithmIdentifier algorithmIdentifier = new DefaultSignatureAlgorithmIdentifierFinder().find(entry.getKey());
+            AlgorithmIdentifier algorithmIdentifier;
+            try {
+                algorithmIdentifier = new DefaultSignatureAlgorithmIdentifierFinder().find(entry.getKey());
+            }
+            catch (IllegalArgumentException e) {
+                // Bouncy is missing some algorithms the JVM supports
```
electrum, line 40: Make sure this is the correct error
```java
assertEquals(e.getMessage(), "Unknown signature type requested: " + entry.getKey());

electrum:
Maybe also add a counter to make sure that we verified some minimum number of signatures (in case there is a bug where we skip all of them).

PR #834, opened 2020-04-17
Fix launcher pid file lock inheritance issue (PEP446 workaround)

antonymayi:
I am offering this PR to fix a bug in the current `launcher.py` affecting all python 3.4+.

Problem
-------
Since PEP446 (Python 3.4+), file descriptors created by most of the Python API calls are **non-inheritable** which means they get closed on calls like `os.exec*`.

The `launcher.py` currently detects the liveness of the service (ie when calling `launcher.py status`) process by checking the _flock_ on its PID file created as part of the `launcher.py run` invocation. If however, the PID file descriptor created during the _run_ stage is not inheritable, the lock gets lost when the PID file is auto-closed upon doing the `os.exec` of the java process.

This means the service process invoked by `launcher.py` on Python 3.4+ cannot be further managed by the launcher as it doesn't see it alive (moreover it allows launching multiple instances).

Solution
--------
The PEP446 provided a new method for setting the file descriptors inheritable (`os.set_inehritable`). ~Using this in the launcher would however be backward-incompatible for older Python versions. I propose to implement the workaround using the `os.dup2` call which is (still after PEP446) creating inheritable descriptors and is fully compatible even for Python 2+.~

Update: going for the conditional use of `os.set_inheritable()` as discussed in the reviews.

### DIFF
--- a/launcher/src/main/scripts/bin/launcher.py
+++ b/launcher/src/main/scripts/bin/launcher.py
@@ -69,16 +69,21 @@ def try_lock(f):
         return False
 
 
-def open_read_write(f, mode):
+def open_pidfile(f, mode):
     """Open file in read/write mode (without truncating it)"""
-    return os.fdopen(os.open(f, O_RDWR | O_CREAT, mode), 'r+')
+    fd = os.open(f, O_RDWR | O_CREAT, mode)
+    if hasattr(os, 'set_inheritable'):
+        # See https://docs.python.org/3/library/os.html#inheritance-of-file-descriptors
+        # Since Python 3.4
+        os.set_inheritable(fd, True)
+    return os.fdopen(fd, 'r+')
 
 
 class Process:
     def __init__(self, path):
         makedirs(dirname(path))
         self.path = path
-        self.pid_file = open_read_write(path, 0o600)
+        self.pid_file = open_pidfile(path, 0o600)
         self.refresh()
 
     def refresh(self):
@@ -277,11 +282,6 @@ def start(process, options):
         process.write_pid(pid)
         print('Started as %s' % pid)
         return
-
-    if hasattr(os, "set_inheritable"):
-        # See https://docs.python.org/3/library/os.html#inheritance-of-file-descriptors
-        # Since Python 3.4
-        os.set_inheritable(process.pid_file.fileno(), True)
 
     os.setsid()
 


launcher/src/main/scripts/bin/launcher.py
```diff
@@ -71,7 +71,9 @@ def try_lock(f):
 
 def open_read_write(f, mode):
     """Open file in read/write mode (without truncating it)"""
-    return os.fdopen(os.open(f, O_RDWR | O_CREAT, mode), 'r+')
+    fd = os.open(f, O_RDWR | O_CREAT, mode)
+    os.dup2(os.dup(fd), fd)  # backward compatible approach to make it inheritable (PEP 446)
```
findepi, line 75: I would prefer following, but let's wait for @electrum opinion too
```
if hasattr(os, 'set_inheritable'):
    # File descriptor returned from os.open is not inheritable since Python 3.4
    os.set_inheritable(fd, True)
```

i would also rename `open_read_write` to `open_pidfile`, because the mechanics of the function are adjusted for this use-case

findepi:
@antonymayi thanks for your PR and especially for describing the underlying problem so well

launcher/src/main/scripts/bin/launcher.py
findepi: @antonymayi would that work for you?

antonymayi: sure, I agree it is much more readable, perhaps I was bit too minimalistic...

antonymayi: PR updated according to this suggestion

launcher/src/main/scripts/bin/launcher.py
```diff
@@ -69,16 +69,20 @@ def try_lock(f):
         return False
 
 
-def open_read_write(f, mode):
+def open_pidfile(f, mode):
     """Open file in read/write mode (without truncating it)"""
-    return os.fdopen(os.open(f, O_RDWR | O_CREAT, mode), 'r+')
+    fd = os.open(f, O_RDWR | O_CREAT, mode)
+    if hasattr(os, 'set_inheritable'):
```
electrum, line 75: I think we can remove this same logic from the `start` method, now that we are setting it here.

antonymayi: ah, good point, didn't realize this has already been dealt with elsewhere...
updating the PR to remove the same logic from the `start()` handler.

findepi:
Merged, thanks!

findepi:
@antonymayi FYI on [Presto slack](https://prestosql.io/slack.html) there is `#airlift` channel.
See you there!

PR #835, opened 2020-04-18
Fix compilation with Java 11

findepi:


.github/workflows/main.yml
```diff
@@ -5,11 +5,19 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java-version: [
+          8,
+          11,
+          13,
```
martint, line 14: 14 is the latest Java version. 13 is no longer supported.

.github/workflows/main.yml
```diff
@@ -5,11 +5,19 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java-version: [
```
electrum, line 11: I’d prefer to format this on one line. It’s short and more readable that way.

stats/pom.xml
```diff
@@ -42,6 +42,11 @@
             <artifactId>guava</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>javax.annotation</groupId>
+            <artifactId>javax.annotation-api</artifactId>
```
electrum, line 47: Where is this used?

jmx/pom.xml
```diff
@@ -95,5 +95,17 @@
...
         </profile>
     </profiles>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <argLine>-Djdk.attach.allowAttachSelf=true</argLine>
```
electrum, line 105: This will overwrite all the other options already specified. Maybe that’s ok here.

discovery/src/test/java/io/airlift/discovery/client/TestServiceTypes.java
```diff
@@ -48,7 +53,12 @@ public void testValue()
     @Test
     public void testToString()
     {
+        if (System.getProperty("java.version").startsWith("1.8.")) {
+            throw new SkipException("Incompatible Java version: " + System.getProperty("java.version"));
```
electrum, line 57: We should test this on both versions.

.github/workflows/main.yml
findepi: removed

.github/workflows/main.yml
findepi: i prefer have this on separate lines. It allows to track changes better (like in enums)

stats/pom.xml
findepi: it's eg `@PostConstruct` annotation

jmx/pom.xml
findepi: Right.. Alternatively we can disable `testJava9Agent` on CI.
Other ideas?

discovery/src/test/java/io/airlift/discovery/client/TestServiceTypes.java
findepi: This test verifies `ServiceTypeImpl#toString()` is consistent with a toString for JVM-created annotation, but this baseline changed.
Not sure what you actually expect here.

findepi:
AC

.github/workflows/main.yml
```diff
@@ -5,11 +5,18 @@ on: [push, pull_request]
...
```
electrum, line 11: This can use standard list syntax
```yaml
java-version:
  - 8
  - 11

File: discovery/src/test/java/io/airlift/discovery/client/TestServiceTypes.java (hunk 5 above)
Comment (electrum) [reply discovery/src/test/java/io/airlift/discovery/client/TestServiceTypes.java:57]: Ah, I see what you mean. We'd need to make the `toString()` version dependent, which doesn't seem worthwhile.

File: jmx/pom.xml (hunk 4 above)
Comment (electrum) [reply jmx/pom.xml:106]: Maybe we should add this option unconditionally in Airbase? For now, I think overriding here is fine.

Comment (findepi) [reply jmx/pom.xml:106]: https://github.com/airlift/airbase/pull/216

File: .github/workflows/main.yml (hunk 6 above)
Comment (findepi) [reply .github/workflows/main.yml]: updated

File: discovery/src/test/java/io/airlift/discovery/client/TestServiceTypes.java (hunk 5 above)
Comment (findepi) [reply discovery/src/test/java/io/airlift/discovery/client/TestServiceTypes.java:57]: BTW we will need to update this again for Java 14
```
AssertionError: expected [@io.airlift.discovery.client.ServiceType("apple")] but found [@io.airlift.discovery.client.ServiceType(value="apple")]
```

....

PR #837, opened 2020-04-20
Add @FileExists validation

wendigo:


configuration/pom.xml
```diff
@@ -21,6 +21,13 @@
             <artifactId>validation-api</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>javax.el</artifactId>
+            <version>3.0.0</version>
+            <scope>runtime</scope>
+        </dependency>
```
findepi, line 29: I am concerned about adding this dependency. `configuration` module is widely used and there is always a risk we will need to do dependency management cleanup because of this, also on downstream libraries.

This seems to be needed to produce a better error message when validation fails,
ie one including the validated value.
While I recognize the benefit, I think we should not add a new dependency for this.

configuration/src/main/java/io/airlift/configuration/validation/FileExistsValidator.java
```diff
@@ -0,0 +1,47 @@
...
+    public void initialize(FileExists ignored)
+    {
+        // Annotation has no properties
+    }
+
+    @Override
+    public boolean isValid(Object path, ConstraintValidatorContext context)
+    {
+        if (path == null) {
+            return true;
```
findepi, line 26:
```suggestion
            // @NotNull's responsibility
            return true;
```

configuration/src/main/java/io/airlift/configuration/validation/FileExistsValidator.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        if (path instanceof String) {
+            return Files.exists(Paths.get((String) path));
+        }
+
+        if (path instanceof Path) {
+            return ((Path) path).toFile().exists();
```
findepi, line 34:
```suggestion
            return Files.exists(((Path) path));
```

configuration/src/main/java/io/airlift/configuration/validation/FileExists.java
```diff
@@ -0,0 +1,27 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Target({ FIELD, METHOD, PARAMETER, ANNOTATION_TYPE })
```
findepi, line 16: TYPE_USE, to allow `Optional<@FileExists File>`

configuration/src/main/java/io/airlift/configuration/validation/FileExistsValidator.java
```diff
@@ -0,0 +1,47 @@
...
+
+        if (path instanceof File) {
+            return ((File) path).exists();
+        }
+
+        context.disableDefaultConstraintViolation();
+        context.buildConstraintViolationWithTemplate(format("could not validate value of type %s", path.getClass().getSimpleName()))
+                .addConstraintViolation();
+
+        return false;
```
findepi, line 45: Can we just throw here?
```
throw new IllegalStateException("Unsupported type: " + path.getClass().getName());
```

configuration/src/test/java/io/airlift/configuration/TestConfigurationFactory.java
```diff
@@ -277,6 +281,49 @@ public void testFailedCoercion()
                 ".*Invalid value 'abc %s xyz' for type int \\(property 'int-value'\\).*BeanValidationClass.*");
     }
 
+    @Test
+    public void testSuccessfulFileExistsValidation()
+    {
+        Map<String, String> properties = new HashMap<>();
+        properties.put("file1", "./pom.xml");
+        properties.put("file2", "./pom.xml");
+        properties.put("file3", "./pom.xml");
```
findepi, line 290: use `ImmutableMap.of` just like in the other test

configuration/src/test/java/io/airlift/configuration/TestConfigurationFactory.java
```diff
@@ -644,6 +691,73 @@ public void setIntValue(int value)
...
+        @Config("file3")
+        public void setFile3(File value)
+        {
+            this.file3 = value;
+        }
+    }
+
+    public static class FileExistsValidationClass2
+    {
+        @NotNull
```
findepi, line 744: redundant?

configuration/src/test/java/io/airlift/configuration/TestConfigurationFactory.java
```diff
@@ -644,6 +691,73 @@ public void setIntValue(int value)
...
 
+    public static class FileExistsValidationClass
+    {
+        @NotNull
+        private String file1;
+
+        @NotNull
+        private Path file2;
+
+        @NotNull
```
findepi, line 702: redundant?

configuration/src/test/java/io/airlift/configuration/TestConfigurationFactory.java
wendigo: This was taken from other test :)

configuration/pom.xml
wendigo: Sure, can do without it :)

configuration/src/main/java/io/airlift/configuration/validation/FileExists.java
```diff
@@ -0,0 +1,29 @@
...
+import static java.lang.annotation.ElementType.TYPE_USE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Target({ FIELD, METHOD, PARAMETER, ANNOTATION_TYPE, TYPE_USE })
+@Retention(RUNTIME)
+@Constraint(validatedBy = FileExistsValidator.class)
+@Documented
+public @interface FileExists
+{
+    // This message is ignored as FileExistsValidator builds constraint violation message
```
electrum, line 23: Change this to Javadoc so it's visible to users of the annotation.

configuration/src/main/java/io/airlift/configuration/validation/FileExists.java
```diff
@@ -0,0 +1,29 @@
...
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.ElementType.TYPE_USE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Target({ FIELD, METHOD, PARAMETER, ANNOTATION_TYPE, TYPE_USE })
```
electrum, line 17: Nit: remove spaces around brackets

configuration/src/main/java/io/airlift/configuration/validation/FileExistsValidator.java
```diff
@@ -0,0 +1,57 @@
...
+        if (path == null) {
+            // @NotNull's responsibility
+            return true;
+        }
+
+        boolean fileExists = exists(path);
+
+        if (!fileExists) {
+            context.disableDefaultConstraintViolation();
+            context.buildConstraintViolationWithTemplate(format("file '%s' does not exist", path.toString()))
```
electrum, line 34: It's better to put the variable part at the end:
```java
"File does not exist: " + path
```
`toString()` not required in a format or concatenation context

configuration/src/main/java/io/airlift/configuration/validation/FileExists.java
```diff
@@ -0,0 +1,29 @@
...
+@Retention(RUNTIME)
+@Constraint(validatedBy = FileExistsValidator.class)
+@Documented
+public @interface FileExists
+{
+    // This message is ignored as FileExistsValidator builds constraint violation message
+    String message() default "file does not exist";
```
electrum, line 24: We could make this empty and validate that it's empty inside the validator, in case someone changes it and expects it to work.

configuration/src/main/java/io/airlift/configuration/validation/FileExistsValidator.java
```diff
@@ -0,0 +1,57 @@
...
+
+        if (path instanceof Path) {
+            return Files.exists(((Path) path));
+        }
+
+        if (path instanceof File) {
+            return ((File) path).exists();
+        }
+
+        throw new IllegalStateException("Unsupported type: " + path.getClass().getName());
```
electrum, line 55: This should be `IllegalArgumentException` since the argument is invalid. `IllegalStateException` means the class is in an incompatible state for the current method invocation.

electrum: How about
```java
throw new IllegalArgumentException("Unsupported type for @FileExists: " + path.getClass().getName());

File: configuration/src/main/java/io/airlift/configuration/validation/FileExistsValidator.java [hunk 14]
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private static boolean exists(Object path)
+    {
+        if (path instanceof String) {
+            return Files.exists(Paths.get((String) path));
+        }
+
+        if (path instanceof Path) {
+            return Files.exists(((Path) path));
```
electrum, line 48: Remove redundant parentheses

configuration/src/test/java/io/airlift/configuration/TestConfigurationFactory.java
```diff
@@ -277,6 +282,46 @@ public void testFailedCoercion()
                 ".*Invalid value 'abc %s xyz' for type int \\(property 'int-value'\\).*BeanValidationClass.*");
     }
 
+    @Test
+    public void testSuccessfulFileExistsValidation()
```
electrum, line 286: This test class is for testing that `ConfigurationFactory` works and that it correctly uses bean validation for config classes. It's not for testing that specific validators work -- they should be tested directly using the validation framework.

Look at the tests for https://github.com/airlift/units to see how to write a validator. You can import the `test-jar` for `units` to reuse `ConstraintValidatorAssert`.

wendigo: I can't use that. ConstraintValidatorAssert is package private

wendigo:
ping @electrum

configuration/src/main/java/io/airlift/configuration/validation/FileExistsValidator.java
```diff
@@ -0,0 +1,58 @@
...
+import static com.google.common.base.Verify.verify;
+
+public class FileExistsValidator
+        implements ConstraintValidator<FileExists, Object>
+{
+    @Override
+    public void initialize(FileExists ignored)
+    {
+        // Annotation has no properties
+        verify(ignored.message().isEmpty(), "annotation.message is not empty");
```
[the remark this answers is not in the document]
findepi: (unless the current wording has been already agreed upon)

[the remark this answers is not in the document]
wendigo: Wasn't. I'm ok with suggestion.

wendigo:
@electrum can you review?

electrum:
Thanks!

PR #838, opened 2020-04-23
Add method to clear a set log level

Randgalt:
Java Logging supports clearing override log levels previously set by
calling setLevel(). Expose an Airlift method for doing this.

### DIFF
--- a/log-manager/src/main/java/io/airlift/log/Logging.java
+++ b/log-manager/src/main/java/io/airlift/log/Logging.java
@@ -27,10 +27,10 @@ import java.io.OutputStream;
 import java.io.PrintStream;
 import java.io.UncheckedIOException;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Handler;
 import java.util.logging.LogManager;
 
@@ -51,7 +51,8 @@ public class Logging
     private static Logging instance;
 
     // hard reference to loggers for which we set the level
-    private final Map<String, java.util.logging.Logger> loggers = new ConcurrentHashMap<>();
+    @GuardedBy("this")
+    private final Map<String, java.util.logging.Logger> loggers = new HashMap<>();
 
     @GuardedBy("this")
     private OutputStreamHandler consoleHandler;
@@ -160,7 +161,15 @@ public class Logging
         return Level.fromJulLevel(level);
     }
 
-    public void setLevel(String loggerName, Level level)
+    public synchronized void clearLevel(String loggerName)
+    {
+        java.util.logging.Logger logger = loggers.remove(loggerName);
+        if (logger != null) {
+            logger.setLevel(null);
+        }
+    }
+
+    public synchronized void setLevel(String loggerName, Level level)
     {
         loggers.computeIfAbsent(loggerName, java.util.logging.Logger::getLogger)
                 .setLevel(level.toJulLevel());
--- a/log-manager/src/test/java/io/airlift/log/TestLogging.java
+++ b/log-manager/src/test/java/io/airlift/log/TestLogging.java
@@ -135,4 +135,17 @@ public class TestLogging
         assertFalse(logger.isDebugEnabled());
         assertFalse(logger.isInfoEnabled());
     }
+
+    @Test
+    public void testClearLevel()
+            throws Exception
+    {
+        Logging logging = Logging.initialize();
+        Logger logger = Logger.get("testClearLevel");
+
+        logging.setLevel("testClearLevel", Level.DEBUG);
+        assertTrue(logger.isDebugEnabled());
+        logging.clearLevel("testClearLevel");
+        assertFalse(logger.isDebugEnabled());
+    }
 }


log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -160,6 +160,12 @@ private static Level getEffectiveLevel(java.util.logging.Logger logger)
         return Level.fromJulLevel(level);
     }
 
+    public void clearLevel(String loggerName)
+    {
+        loggers.computeIfAbsent(loggerName, java.util.logging.Logger::getLogger)
```
martint, line 165: We may not need to hold on to the logger once the level is cleared

Randgalt: So, instead this?

```
        java.util.logging.Logger logger = loggers.get(loggerName);
        if (logger == null) {
            logger = java.util.logging.Logger.getLogger(loggerName);
        }
        logger.setLevel(null);

```

Randgalt: We could also clear the map but that's not really implied by the method contract.

martint: Yeah, that's better

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -160,6 +160,15 @@ private static Level getEffectiveLevel(java.util.logging.Logger logger)
         return Level.fromJulLevel(level);
     }
 
+    public void clearLevel(String loggerName)
```
findepi, line 163: is `loggerName` `@Nullable`?

log-manager/src/main/java/io/airlift/log/Logging.java
findepi: Providing clearLevel method suggests that calling `setLevel + clearLevel` (with varying names) is a supported thing.
So we should try to indeed support it, eg avoid memory leak. 
I suggest that in this method
```
java.util.logging.Logger logger = loggers.get(loggerName);
```
be replaced with
```
java.util.logging.Logger logger = loggers.remove(loggerName);
```

After that, there still is possibility that we have a logger in `loggers`, which has null level (inconsistency), so
I would certainly consider synchronizing `setLevel` and `clearLevel` methods to avoid this.
(then no `loggers` becomes a non-concurrent map)

Randgalt: TBH - I'm not sure why `Logging` caches the logger in `setLevel()`. `java.util.logging.Logger.getLogger()` has its own cache internally. The logger cached in `loggers` would only be referenced again if the level is changed (or now cleared). But, I'm new to this code base. Maybe we should remove `loggers` altogether?

Randgalt: I bisecting and see that the cache in `Logging` was added here:

```
commit 117e90fbe3ae7ef2443ab658e671197b7d0e7cfb
Author: David Phillips <david@acz.org>
Date:   Fri May 15 16:34:21 2015 -0700

    Fix log levels being reset by garbage collection
```

So, maybe @electrum can comment/review

martint: The JDK keeps a weak reference to the logger. That's why we have to keep one in this class.

log-manager/src/main/java/io/airlift/log/Logging.java
Randgalt: It's the same contract as `setLevel`. The `loggers` ConcurrentHashMap already checks for `null` (per the Javadoc: `NullPointerException if the specified key is null`).

log-manager/src/main/java/io/airlift/log/Logging.java
martint: I don't think it's a big deal in practice, but let's remove it for completeness and consistency. It's unlikely anyone would be using these methods very "dynamically", or for large numbers of loggers, so any apparent leak would be generally bounded in size. In fact, that would happen anyway if they go around this class and clear the log levels on their own by calling the JDK methods directly.

Randgalt: done

Randgalt: > if they go around this class and clear the log levels on their own by calling the JDK methods directly

Which is what I'm having to do now BTW without this method.

findepi: what about synchronization (https://github.com/airlift/airlift/pull/838#discussion_r414422707)?

- thread A calls setLevel, entering it into the map
- thread B calls clearLogger, removing it from the map
- thread B calls logger.setLevel(null)
- thread A calls logger.setLevel(DEBUG)

Since both calls are "concurrent", it's whatever whether the end state will be DEBUG or null.
But, the end state  is "DEBUG... until GC, then reset" -- that doesn't seem acceptable.

@martint any reason not to make the methods synchronized?

Randgalt: A simpler solution is to leave it in the map as it was in the original version of this PR.

Randgalt: ping @martint

Randgalt:
@martint and @findepi - last push uses syncronization. Note: there is now an implied contract that you can only clear a level that was previously set via `setLevel()`.

PR #844, opened 2020-05-03
Deprecate closeQuietly

findepi:
See discussion on [Presto Slack](https://prestosql.io/slack.html)
https://prestosql.slack.com/archives/CP1MUNEUX/p1586372796137000

### DIFF
--- a/testing/src/main/java/io/airlift/testing/Closeables.java
+++ b/testing/src/main/java/io/airlift/testing/Closeables.java
@@ -11,6 +11,12 @@ public final class Closeables
     {
     }
 
+    /**
+     * @deprecated Usage of this method is discouraged, as it may hide real problems.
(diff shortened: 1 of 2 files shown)


testing/src/main/java/io/airlift/testing/Closeables.java
```diff
@@ -11,6 +11,11 @@ private Closeables()
     {
     }
 
+    /**
+     * @deprecated Usage of this method is discouraged, as it may hide real problems.
+     * Use {@link #closeAll(Closeable...)} instead and, if necessary, suppress exceptions
+     * explicitly.
+     */
     public static void closeQuietly(Closeable... closeables)
```
dain, line 19: Shouldn't this also have an `@deprecated` annotation also?

findepi: Right.

PR #846, opened 2020-05-13
Add support for Forwarded and X-Forwarded headers

dain:
Support Forwarded and X-Forwarded headers in proxied environments


http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -351,6 +353,19 @@ public HttpServerConfig setLogMaxFileSize(DataSize logMaxFileSize)
         return this;
     }
 
+    public boolean isForwardedEnabled()
+    {
+        return forwardedEnabled;
+    }
+
+    @Config("http-server.forwarded.enabled")
```
findepi, line 361: "process-forwarded", "handle-forwaded"?

http-server/src/test/java/io/airlift/http/server/ForwardedServlet.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright 2010 Proofpoint, Inc.
```
findepi, line 2: :)

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
electrum: I like `http-server.process-forwarded`

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -351,6 +353,19 @@ public HttpServerConfig setLogMaxFileSize(DataSize logMaxFileSize)
...
+    @ConfigDescription("Support Forwarded and X-Forwarded headers in proxied environments")
```
electrum, line 362: Support -> Process

http-server/src/test/java/io/airlift/http/server/ForwardedServlet.java
electrum: Yep, I think we can remove this

PR #850, opened 2020-06-07
Improve configuration error reporting

dain:


bootstrap/src/main/java/io/airlift/bootstrap/ConfigurationException.java
```diff
@@ -0,0 +1,63 @@
...
+    {
+        this.errors = ImmutableSet.copyOf(errors);
+        this.warnings = ImmutableSet.copyOf(warnings);
+
+        ImmutableList.Builder<Message> allMessages = ImmutableList.builder();
+        errors.stream()
+                .map(error -> new Message(error.getSources(), "Error: " + error.getMessage(), error.getCause()))
+                .forEach(allMessages::add);
+        warnings.stream()
+                .map(warning -> new Message(warning.getSources(), "Warning: " + warning.getMessage(), warning.getCause()))
```
findepi, line 42: Since this flies as an exception, a "warning" may be a misunderstood. (When i see a warning, i usually expect the problem will try to continue).

Maybe instead "error, warning" naming use "fatal, error" for discerning them?

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -224,30 +222,46 @@ public Injector initialize()
...
         // Log effective configuration
         if (!quiet) {
-            logConfiguration(configurationFactory, unusedProperties);
+            logConfiguration(configurationFactory);
+        }
+
+        // Log any warnings
+        if (warnings.isEmpty()) {
+            StringBuilder message = new StringBuilder();
+            message.append("Warning configuration should be updated:\n");
```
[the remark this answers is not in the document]
electrum: > Warning: configuration should be updated:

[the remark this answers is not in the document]
dain: I guess it doesn't need the word warning, since it is logged with warning?

bootstrap/src/main/java/io/airlift/bootstrap/ConfigurationException.java
electrum: This will show errors first, then warnings. We could add a validation to show this always has errors.
```java
checkArgument(!errors.isEmpty(), "no errors present");

File: bootstrap/src/main/java/io/airlift/bootstrap/ConfigurationException.java [hunk 3]
```diff
@@ -0,0 +1,63 @@
...
+    private final Set<Message> warnings;
+    private final List<Message> allMessages;
+
+    public ConfigurationException(Collection<Message> errors, Collection<Message> warnings)
+    {
+        this.errors = ImmutableSet.copyOf(errors);
+        this.warnings = ImmutableSet.copyOf(warnings);
+
+        ImmutableList.Builder<Message> allMessages = ImmutableList.builder();
+        errors.stream()
```
electrum, line 38: Probably cleaner as a for each loop
```java
for (Message error : errors) {
    allMessages.add(new Message(error.getSources(), "Error: " + error.getMessage(), error.getCause()));
}

File: bootstrap/src/main/java/io/airlift/bootstrap/ConfigurationException.java [hunk 4]
```diff
@@ -0,0 +1,63 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.inject.internal.Messages;
+import com.google.inject.spi.Message;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Set;
+
+public final class ConfigurationException
```
electrum, line 25: This name collides with `com.google.inject.ConfigurationException`. Should we name it `BootstrapConfigurationException` or `ApplicationConfigurationException`? Or maybe it doesn't matter.

bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrap.java
```diff
@@ -111,8 +110,8 @@ public void testStrictConfig()
                 .setRequiredConfigurationProperty("test-required", "foo");
 
         assertThatThrownBy(bootstrap::initialize)
-                .isInstanceOfSatisfying(CreationException.class, e ->
-                        assertThat(e.getErrorMessages()).containsExactly(
+                .isInstanceOfSatisfying(io.airlift.bootstrap.ConfigurationException.class, e ->
```
electrum, line 113: This shows the name collision I mentioned above. But this is test code, maybe doesn't matter in real usage.

bootstrap/src/main/java/io/airlift/bootstrap/ConfigurationException.java
dain: I went with `ApplicationConfigurationException`

PR #851, opened 2020-06-08
Ensure underlying tdigest weights are >= 1 

martint:


stats/pom.xml
```diff
@@ -99,6 +99,13 @@
             <artifactId>jmh-generator-annprocess</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>3.13.2</version>
```
electrum, line 106: We don't need the version here as Airbase already has a newer version

stats/src/main/java/io/airlift/stats/DecayTDigest.java
```diff
@@ -138,7 +145,7 @@ private void rescale(long newLandmarkInSeconds)
         for (int i = 0; i < digest.centroidCount; i++) {
             double weight = digest.weights[i] / factor;
 
-            if (weight < ZERO_WEIGHT_THRESHOLD) {
+            if (weight < 1) { // values with a scaled weight below 1 are effectively below the ZERO_WEIGHT_THRESHOLD
```
electrum, line 148: Nit: put comment above statement

PR #852, opened 2020-06-09
Move combine method to ConfigurationAwareModule

electrum:


configuration/src/main/java/io/airlift/configuration/ConfigurationAwareModule.java
```diff
@@ -16,11 +16,41 @@
...
+
+import static com.google.common.base.Preconditions.checkArgument;
+
 @Beta
 public interface ConfigurationAwareModule
         extends Module
 {
     void setConfigurationFactory(ConfigurationFactory configurationFactory);
+
+    static Module combine(Module... modules)
```
kokosing, line 33: I don't like interfaces poluted with utility static methods? What is the motivation behind this?

electrum: Your thinking is old-fashioned. Static methods were added to interfaces so that separate utility classes are not required. If you look at all the new JDK APIs, the static utility methods go on the interfaces.

Issue #848, opened 2020-05-20
Customizing SslContextFactory in HttpServer and HttpClient


csunwold:
We have a requirement to support frequent certificate rotation for all SSL requests. To achieve this today, we have an internal fork of airlift where we've inserted custom code that instantiates an instance of the [SslContextFactory.Client](https://www.eclipse.org/jetty/javadoc/9.4.24.v20191120/org/eclipse/jetty/util/ssl/SslContextFactory.Client.html) or [Server](https://www.eclipse.org/jetty/javadoc/9.4.24.v20191120/org/eclipse/jetty/util/ssl/SslContextFactory.Server.html) and invokes [reload](https://www.eclipse.org/jetty/javadoc/9.4.24.v20191120/org/eclipse/jetty/util/ssl/SslContextFactory.html#reload(java.util.function.Consumer)) when needed. This isn't ideal since no one else can use it and it adds some overhead from maintaining a fork. 

Ideally, I'd like a way to either:
1. Inject an SslContextFactory.Client or .Server instance that the [HttpServerProvider](https://github.com/airlift/airlift/blob/418b0adea11560b81330bf750f4b42a173848d97/http-server/src/main/java/io/airlift/http/server/HttpServerProvider.java#L43) or [HttpClientProvider](https://github.com/airlift/airlift/blob/7aff37d613461b5b54d47fa2fba5605e5dd9ce71/http-client/src/main/java/io/airlift/http/client/HttpClientModule.java#L86) can use for building a client or server.
2. Have a way of accessing the instance of the sslContextFactory 

Spring Boot solves this by giving access to Jetty's APIs: https://github.com/spring-projects/spring-boot/issues/20381. 

Introducing a method of injecting an sslContextFactory is my preference.

There are a few problems I've run into trying to achieve this
1. Airlift doesn't seem to have any existing pattern of passing optional components like this to follow.
2. [Guice is not friendly to "optional" injections.](https://github.com/google/guice/wiki/Injections#optional-injections) It requires a default of some kind. An SslContextFactory shouldn't be mandatory for clients to pass in, but if one is given I would imagine it would be best for usability to skip all of the configuration code that would [override settings in the SslContextFactory](https://github.com/airlift/airlift/blob/663119bbb3a50196e10141c30152330635b47c80/http-server/src/main/java/io/airlift/http/server/HttpServer.java#L219).

I'm happy to submit a pull request that achieves this, but I'd love to hear thoughts from the community + maintainers on the best approach here before doing so.

[...]

PR #853, opened 2020-06-12
Allow Customization of SslContextFactory and add ReloadableSslContextFactory

csunwold:
Implements changes proposed in https://github.com/airlift/airlift/issues/848.

This change does two things:  
1. It allows an SslContextFactory to be injected via Guice.
2. If one is not provided, a ReloadableSslContextFactory will be used. Reload schedule is configurable.


electrum:
The commit title is too long. We follow this guide: https://chris.beams.io/posts/git-commit/

How something like

> Allow customizing SslContext for HTTP client

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -97,8 +93,8 @@
     private final DelimitedRequestLog requestLog;
     private ConnectionStats httpConnectionStats;
     private ConnectionStats httpsConnectionStats;
-
-    private final Optional<ZonedDateTime> certificateExpiration;
+    private final ScheduledExecutorService scheduledExecutorService = new ScheduledThreadPoolExecutor(1);
```
electrum, line 96: Nit: the canonical way is the `newScheduledThreadPool(1)` factory method from `Executors`

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -575,29 +508,24 @@ void join()
...
+        Supplier<SslContextFactory.Server> supplier;
+        if (maybeSslContextFactory.isPresent()) {
+            supplier = new SingletonSslContextFactorySupplier(maybeSslContextFactory.get());
         }
-        return certificates.build();
+        else {
+            supplier = new ReloadableSslContextFactorySupplier(config, scheduledExecutorService);
+        }
+
+        return supplier.get();
```
electrum, line 528: The purpose of `ReloadableSslContextFactorySupplier` is that it returns a supplier where the `get()` method can return a different factory whenever the keys change. Since we only call the `get()` method once, it seems like this won't work. I think we need to tell Jetty to reload the factory.

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
```diff
@@ -65,6 +68,9 @@ public void configure(Binder binder)
         configBinder(binder).bindConfig(KerberosConfig.class);
         configBinder(binder).bindConfig(HttpClientConfig.class, annotation, name);
 
+        // bind an optional SslContextFactory.Client
+        OptionalBinder.newOptionalBinder(binder, SslContextFactory.Client.class);
```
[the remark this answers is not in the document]
csunwold: Addressed.

http-server/src/main/java/io/airlift/http/server/HttpServer.java
csunwold: The SslContextFactory is responsible for always returning an SslContext that has been built using the latest keys. The Jetty APIs that depend on SslContextFactory will only take a single instance, they won't invoke this supplier again to get new instances. Instead they call the factory to get what is assumed to be the correct SslContext when one is needed. 

As an example: The SslConnectionFactory calls [SslContextFactory.newSSLEngine](https://github.com/eclipse/jetty.project/blob/81c46638cd63e34e06f1c99b514d9d15713a8e7b/jetty-server/src/main/java/org/eclipse/jetty/server/SslConnectionFactory.java#L150) every time a connection is established. This captures uses whatever the latest provided [SslContext](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/SslContextFactory.java#L1788) is at this time. 

The reload will be invoked on a schedule setup within the provider here: https://github.com/airlift/airlift/pull/853/files#diff-06f4702cd9c78a864276eb020c71882aR32

csunwold:
@electrum Thank you for the feedback. I believe I've addressed all of your comments so far. Please have another look.

http-server/src/main/java/io/airlift/http/server/HttpServer.java
electrum: Thanks for explaining, I missed that part. In that case, there's no need for this to be a supplier -- it can be a utility method that returns the `SslContextFactory`. Having it be a supplier seems misleading and more complex.

With that change, this method can be eliminated and done similar to the one for the client:
```java
this.sslContextFactory = maybeSslContextFactory
        .orElseGet(() -> createReloadingSslContextFactory(config));
```
`SingletonSslContextFactorySupplier` will also no longer be needed.

http-server/src/main/java/io/airlift/http/server/HttpServerModule.java
```diff
@@ -68,6 +70,7 @@ public void configure(Binder binder)
         newSetBinder(binder, Filter.class, TheServlet.class);
         newSetBinder(binder, Filter.class, TheAdminServlet.class);
         newSetBinder(binder, HttpResourceBinding.class, TheServlet.class);
+        OptionalBinder.newOptionalBinder(binder, SslContextFactory.Server.class);
```
electrum, line 73: Nit: static import `newOptionalBinder`

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -467,53 +418,34 @@ private static DelimitedRequestLog createDelimitedRequestLog(HttpServerConfig co
...
 
-    private static Optional<KeyStore> tryLoadPemKeyStore(HttpServerConfig config)
+    private Set<X509Certificate> getCertificates()
     {
-        File keyStoreFile = new File(config.getKeystorePath());
-        try {
-            if (!PemReader.isPem(keyStoreFile)) {
-                return Optional.empty();
+        ImmutableSet.Builder<X509Certificate> certificates = ImmutableSet.builder();
+        if (this.sslContextFactory.isPresent()) {
```
electrum, line 424: Can this be switched to `sslContextFactory.ifPresent(...)`

http-server/src/main/java/io/airlift/http/server/HttpServer.java
electrum: By misleading, I mean that `Supplier` is typically used so that consumers can repeatedly call `get()` to fetch a potentially different instance. Here, we only need to create it once, so there's no need for a supplier instance if we create it directly via a factory method.

electrum:
When updating the PR, please squash everything into a single commit, then I can merge it directly.

http-server/src/main/java/io/airlift/http/server/HttpServer.java
csunwold: I'm not attached to using a Supplier, but I do think there are readability advantages to keeping the implementation of the class that manages the reloading of an SslContextFactory isolated from the implementation of HttpServer. 

The issue comes in how to implement the `reload` callback that the scheduled executor service invokes periodically. 

Here are a few examples I considered:

### Private instance method
There are a few subtle problems here. 

First, HttpServer has three allowed SslContextFactory states. It could be none/empty because Https isn't used, it could be the custom injected instance, or it could be the instantiated on construction SslContextFactory. Only the third one should be reloaded. This leaves open the possibility that reload will be invoked for either of the first two scenarios where it shouldn't be.  I don't _think_ this is that bad of a bug since reloading should generally be safe if it did happen. It is weird to read because if you aren't familiar with the complete context in which it's used it makes you wonder: "Why are we reloading when the SslContextFactory is not present?" 
```java
    private synchronized void reload()
    {
        try {
            SslContextFactory.Server updatedFactory = buildContextFactory(this.config);
            updatedFactory.start();
            if (this.sslContextFactory.isPresent()) {
                this.sslContextFactory.get().reload(factory -> factory.setSslContext(updatedFactory.getSslContext()));
            }
        }
        catch (Exception e) {
            log.warn(e, "Unable to reload SslContext.");
        }
    }
```

### Anonymous function
This one seems slightly better because it avoids the ambiguity around when reload should be invoked and what the allowable states of sslContextFactory are. It does bring up the question of how to best keep this thread safe and my inexperience with writing Java leaves me unsure of how best to handle it. I believe it requires some object to synchronize on, but I'm not sure which would be the best to use in this scenario. 
```java
 this.scheduledExecutorService.scheduleWithFixedDelay(() -> {
            synchronized (on something?) {
                try {
                    SslContextFactory.Server updatedFactory = buildContextFactory(config);
                    updatedFactory.start();
                    sslContextFactory.reload(factory -> factory.setSslContext(updatedFactory.getSslContext()));
                }
                catch (Exception e) {
                    log.warn(e, "Unable to reload SslContext.");
                }
            }
        }, refreshTime, refreshTime, MILLISECONDS);
```

My personal preference is to leave the the concern of reloading an SslContextFactory in a separate class and avoid this altogether. Changing the type to not implement Supplier may help a bit and then we are just left with a naming problem. If you feel strongly that its best to keep this inside the implementation of HttpServer, please let me know your thoughts on the above two suggestions. 

In the mean time, I'll work on addressing your other comments.

electrum: I agree that having this in a separate class makes sense. My only concern is with implementing `Supplier`, so it's mostly just a naming (and usage) problem.

csunwold: I took a stab at a change, I now have it called a ReloadableSslContextFactoryProvider and added a few comments to explain it. It no longer implements Supplier. Still feels a bit off to me in naming but its the best I've come up with after sleeping on it. I'm open to suggestions if you can think of anything better.

csunwold: @electrum Pinging to see if you had any other feedback on this.

ShashwatArghode:
Hi @csunwold, @electrum,
#848 also talks about automatic reload of SslContext when the keys or certs change. Are we not implementing that in this PR (I was not able to see any changes related to that)?

csunwold:
@ShashwatArghode Reloading is handled by this class: https://github.com/airlift/airlift/pull/853/files#diff-6c4825ac5b62245f23a44373b1c9fd19R21

ShashwatArghode:
From the implementation, I can see this class handles the scheduled refresh of sslcontext. I was talking more on the lines of automatic refresh when keys or cert file changes. Eg. old certs which were about to expire overwritten by new cert files.

csunwold:
It will read the keystore and truststore files that have been configured on the refresh schedule. It will not trigger a refresh when the file itself changes. If you have extremely frequently changing certificates (say every 10 minutes) you could configure the reload schedule to be every 1 minute.

ShashwatArghode:
In our case, the certificate refresh is not that frequent. That's why we were thinking of an automatic refresh of sslcontext on cert changes to avoid scheduled refresh operations when they are not needed.
This approach will also get the work done. Thank You.

http-server/src/main/java/io/airlift/http/server/HttpServer.java
csunwold: @electrum Pinging again for any further feedback

electrum: Sorry for the delay. This looks good to me.

csunwold:
@electrum I've rebased from master and resolved conflicts. It may be worth giving this another once over.

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -97,8 +92,8 @@
     private final DelimitedRequestLog requestLog;
     private ConnectionStats httpConnectionStats;
     private ConnectionStats httpsConnectionStats;
-
-    private final Optional<ZonedDateTime> certificateExpiration;
+    private final ScheduledExecutorService scheduledExecutorService = newScheduledThreadPool(1);
```
electrum, line 95: This can use `newSingleThreadExecutor()` since there's only one thread. Also, we should name the thread and use a daemon thread:
```java
newSingleThreadExecutor(daemonThreadsNamed("HttpServerScheduler"));

electrum: We should only start this if needed -- we don't need the executor running if we're just doing HTTP.

csunwold: Updated.

electrum: Did you forget to push? I don't see any change here.

I'm also fine merging this as-is and addressing this in a follow up.

csunwold: 🤦 Yes. Fixed.

electrum:
Thanks!

PR #858, opened 2020-06-29
added config property to enable/disable ssl hostname verification

gugalnikov:
Embedded Jetty allows enabling / disabling SSL hostname verification through config mechanisms; actually looks like the default behaviour is for this to be disabled as SslContextFactory has no default endpoint identification algorithm (ref.: https://github.com/eclipse/jetty.project/issues/2520)

Within Airlift, the "setEndpointIdentificationAlgorithm" method is hardcoded with an "HTTPS" value, which ensures hostname verification will always be performed. As a result of this, any system which uses airlift as a foundation for internal communication won't have any control over this, even if it configures its own trust manager or uses the available JVM property on startup (-Djdk.internal.httpclient.disableHostnameVerification)

This is a very simple change adding a config property which allows airlift users to leverage embedded Jetty's configuration possibilities.

The justification is that there are use cases which require disabling the hostname verification for SSL (or don't really need it), without bypassing the whole certificate chain trust process. A good example is a system or application which subscribes to the SPIFFE / SPIRE standard, where certificate chain trust is enforced but vanilla hostname verification (based on certificate CN) is not necessary as the SPIFFE URI in the certificate's SAN is a much more effective and flexible way to prevent the man-in-the-middle attack.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
@@ -102,6 +102,20 @@ public class HttpClientConfig
     private DataSize logBufferSize = new DataSize(1, MEGABYTE);
     private Duration logFlushInterval = new Duration(10, SECONDS);
     private boolean logCompressionEnabled = true;
+    private boolean verifyHostname = true;
+
+    public boolean isVerifyHostname()
+    {
+        return verifyHostname;
+    }
+
+    @Config("http-client.https.hostname-verification")
+    @ConfigDescription("Verify that server hostname matches the server certificate")
+    public HttpClientConfig setVerifyHostname(boolean verifyHostname)
+    {
+        this.verifyHostname = verifyHostname;
+        return this;
+    }
 
     public boolean isHttp2Enabled()
     {
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
@@ -158,7 +158,8 @@ public class JettyHttpClient
         creationLocation.fillInStackTrace();
 
         SslContextFactory.Client sslContextFactory = new SslContextFactory.Client();
-        sslContextFactory.setEndpointIdentificationAlgorithm("HTTPS");
+        sslContextFactory.setEndpointIdentificationAlgorithm(config.isVerifyHostname() ? "HTTPS" : null);
+
         if (config.getKeyStorePath() != null) {
             Optional<KeyStore> pemKeyStore = tryLoadPemKeyStore(config);
             if (pemKeyStore.isPresent()) {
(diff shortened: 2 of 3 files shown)


gugalnikov:
BTW, I have tested this implementation with a service-mesh enabled prestosql and it works as intended

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -102,6 +102,20 @@
...
     private boolean logCompressionEnabled = true;
+    private boolean hostnameVerificationEnabled = true;
+
+    public boolean isHostnameVerificationEnabled()
+    {
+        return hostnameVerificationEnabled;
+    }
+
+    @Config("http-client.https.hostname-verification")
+    @ConfigDescription("Enable / Disable Hostname Verification")
```
electrum, line 113: How about
> Verify that server hostname matches the server certificate

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -102,6 +102,20 @@
     private DataSize logBufferSize = new DataSize(1, MEGABYTE);
     private Duration logFlushInterval = new Duration(10, SECONDS);
     private boolean logCompressionEnabled = true;
+    private boolean hostnameVerificationEnabled = true;
```
electrum, line 105: We could shorten to `verifyHostname`

Then the getter becomes `isVerifyHostname()`

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -158,7 +158,13 @@ public JettyHttpClient(
         creationLocation.fillInStackTrace();
 
         SslContextFactory.Client sslContextFactory = new SslContextFactory.Client();
-        sslContextFactory.setEndpointIdentificationAlgorithm("HTTPS");
+        if (config.isHostnameVerificationEnabled()) {
```
electrum, line 161: We can simplify to
```java
sslContextFactory.setEndpointIdentificationAlgorithm(config.isVerifyHostname() ? "HTTPS" : null);

gugalnikov:
hi @electrum , thanks for the review!!, the changes are done, so this should be ready for merge.

BTW, while testing with Presto with this compiled version of http-client, i found out that all of the following properties can be set (in config.properties) to control the hostname verification behaviour throughout the system's lifecycle:

for coordinator:

discovery.http-client.https.hostname-verification=false
dynamic.http-client.https.hostname-verification=false
failure-detector.http-client.https.hostname-verification=false
memoryManager.http-client.https.hostname-verification=false
node-manager.http-client.https.hostname-verification=false
exchange.http-client.https.hostname-verification=false
scheduler.http-client.https.hostname-verification=false
workerInfo.http-client.https.hostname-verification=false

for worker:

discovery.http-client.https.hostname-verification=false
node-manager.http-client.https.hostname-verification=false
exchange.http-client.https.hostname-verification=false

Once this PR is merged, it could be useful to have such properties documented somewhere, but I don't know which is the best place to do so, maybe: https://prestosql.io/docs/current/admin/properties.html ?

electrum:
If this is something we want to broadly support in Presto, we'd want to add a config in `InternalCommunicationConfig` that sets this for all HTTP clients.

electrum:
Thanks!

PR #863, opened 2020-08-05
Add 0.199 changes

dain:


electrum:
Mention that certificates are no longer requested by default.

PR #868, opened 2020-09-07
Support AutoCloseable in closeAll

findepi:


losipiuk (approved):

testing/src/main/java/io/airlift/testing/Closeables.java
```diff
@@ -35,12 +36,25 @@ public static void closeQuietly(Closeable... closeables)
...
+            closeAll((AutoCloseable[]) closeables);
+        }
+        catch (Exception e) {
+            propagateIfPossible(e, IOException.class);
+            // Not possible
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static void closeAll(AutoCloseable... closeables)
```
electrum, line 50: The usages of `IOException` need to be changed

findepi: This is why i retained `closeAll(Closeable...)`

findepi: BTW I'd rather remove `closeAll(Closeable...)` if we can afford breakage.
(there is no reasonable gradual deprecation path I can think of)

martint: In the code below, we shouldn't wrap the exception with an IOException for this method:

```java
 catch (Throwable e) {
    if (rootCause == null) {
        rootCause = new IOException(e);
    }
```

findepi: good catch

findepi:
AC

martint (approved):

PR #871, opened 2020-09-17
Add Java 15 to CI

martint:


findepi:
Annotation's tostring changed.

discovery/src/test/java/io/airlift/discovery/client/TestServiceTypes.java
```diff
@@ -50,17 +49,6 @@ public void testValue()
         assertEquals(ServiceTypes.serviceType("type").value(), "type");
     }
 
-    @Test
-    public void testToString()
```
martint, line 54: Java doesn't make any guarantees about the format of an annotation's toString and it seems to change from version to version, so I chose to remove the test altogether.

findepi: rel https://github.com/google/guice/issues/1312

we do not need to make the toString match JVM provided one, but it's still valuable to test toString is something reasonable (fixed const value)

findepi (approved):

discovery/src/test/java/io/airlift/discovery/client/TestServiceTypes.java
electrum: I merged this so we can move forward with testing Java 15 in the CI. I filed #902 for follow up on this.

PR #872, opened 2020-09-20
Support compound environment variables in configurations.

choiwaiyiu:
Related to https://github.com/prestosql/presto/issues/4345
Currently, compound environment variables are treated as literal.
This PR allows configuration properties to reference multiple environment variables.

- Example

```
HOST=localhost
PORT=8080

Current behavior:
example.uri=${ENV:HOST}:${PORT} => ${ENV:HOST}:${PORT}

With this PR:
example.uri=${ENV:HOST}:${PORT} => localhost:8080
```

### DIFF
--- a/bootstrap/pom.xml
+++ b/bootstrap/pom.xml
@@ -71,5 +71,24 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.gaul</groupId>
+                <artifactId>modernizer-maven-plugin</artifactId>
+                <configuration>
+                    <exclusions>
+                        <!-- TODO: remove after upgrading to Java 9 -->
+                        <!-- Modernizer detects uses of StringBuffer and requires to use StringBuilder instead.
+                             In Bootstrap.replaceEnvironmentVariables, StringBuffer is used in Matcher.appendReplacement and Matcher.appendTail.
+                             To replace StringBuffer with StringBuilder, Java 9 or above is required.
+                             This violation exclusion allows us to use StringBuffer before upgrading to Java 9. -->
+                        <exclusion>java/lang/StringBuffer."&lt;init&gt;":()V</exclusion>
+                    </exclusions>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
 </project>
 
--- a/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
@@ -53,6 +53,7 @@ import java.util.regex.Pattern;
 import static io.airlift.configuration.ConfigurationLoader.getSystemProperties;
 import static io.airlift.configuration.ConfigurationLoader.loadPropertiesFrom;
 import static java.lang.String.format;
+import static java.util.regex.Matcher.quoteReplacement;
 
 /**
  * Entry point for an application built using the platform codebase.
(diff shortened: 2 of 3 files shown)


bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -322,17 +322,24 @@ private static ColumnPrinter makePrinterForConfiguration(ConfigurationFactory co
...
-                onError.accept(propertyKey, format("Configuration property '%s' references unset environment variable '%s'", propertyKey, envName));
-                return;
+            String replacedPropertyValue = propertyValue;
+            for (String envName : envNames) {
+                String envValue = environment.get(envName);
+                if (envValue == null) {
+                    onError.accept(propertyKey, format("Configuration property '%s' references unset environment variable '%s'", propertyKey, envName));
+                    return;
+                }
+                replacedPropertyValue = replacedPropertyValue.replaceAll("\\$\\{ENV:" + envName + "}", envValue);
```
findepi, line 340: Add a test case with

input
```
${ENV:FIRST}, ${ENV:SECOND}
```
where
FIRST is `first:${ENV:SECOND}:`
SECOND is `second:${ENV:FIRST}`

my expected results would be
```
first:${ENV:SECOND}:, second:${ENV:FIRST}
```

while current code would produce probably something like
```
first:second:${ENV:FIRST}:, second:${ENV:FIRST}
```

-- FIRST was first, so was apparently processed "recursively", while SECOND was not processed recursively

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -322,17 +322,24 @@ private static ColumnPrinter makePrinterForConfiguration(ConfigurationFactory co
         Map<String, String> replaced = new HashMap<>();
         properties.forEach((propertyKey, propertyValue) -> {
             Matcher matcher = ENV_PATTERN.matcher(propertyValue);
-            if (!matcher.matches()) {
+            List<String> envNames = new ArrayList<>();
+            while (matcher.find()) {
+                envNames.add(matcher.group(1));
```
findepi, line 327: For finding multiple matches and replacing them all it would more idiomatic to use `java.util.regex.Matcher#appendReplacement` and `java.util.regex.Matcher#appendTail` methods.

It yields for simpler code and also prevents "recursive" processing inconsistencies

The final code could similar to

```java
StringBuffer replacedPropertyValue = new StringBuffer();
Matcher matcher = ENV_PATTERN.matcher(propertyValue);
while (matcher.find()) {
    String envName = matcher.group(1);
    String envValue = environment.get(envName);
    if (envValue == null) {
        onError.accept(propertyKey, format("Configuration property '%s' references unset environment variable '%s'", propertyKey, envName));
        return;
    }
    matcher.appendReplacement(replacedPropertyValue, quoteReplacement(envValue));
}
matcher.appendTail(replacedPropertyValue);
replaced.put(propertyKey, replacedPropertyValue.toString());
```

findepi: @electrum the code i suggested above fails at modernizer
```
[INFO] --- modernizer-maven-plugin:1.7.1:modernizer (modernizer) @ bootstrap ---
[ERROR] /Users/xxx/repos/airlift/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java:325: Prefer java.lang.StringBuilder
```

thoughts?

choiwaiyiu: @findepi Thank you for the review. I will fix the "recursive replacement" problem as suggested.

I think the modernizer violation is caused by using appendReplacement with StringBuffer?
Using StringBuilder instead of StringBuffer can resolve the error but it requires language level 9 and StringBuilder is not thread-safe.

Should we exclude this violation? I am not familiar with the modernizer plugin. I have tried using @SuppressModernizer annotation and adding exclusion configuration in boostrap/pom.xml but still get the same error when running the modernizer plugin.

choiwaiyiu: @findepi I realized the problem is not caused by using appendReplacement with StringBuffer.
Instantiating a StringBuffer causes the error while only declaring a StringBuffer does not.

choiwaiyiu: Using `StringBuilder` instead of `StringBuffer` can resolve the error but it requires language level 9 to use `appendReplacement` and `appendTail`.

Should we exclude this violation?
The error is eliminated after adding the following in `bootstrap/pom.xml`.
```
<build>
    <plugins>
        <plugin>
            <groupId>org.gaul</groupId>
            <artifactId>modernizer-maven-plugin</artifactId>
            <configuration>
                <exclusions>
                    <exclusion>java/lang/StringBuffer."&lt;init&gt;":()V</exclusion>
                </exclusions>
            </configuration>
        </plugin>
    </plugins>
</build>
```

electrum: Let's exclude it for now, with TODO comment to remove after updating to Java 9.

bootstrap/pom.xml
```diff
@@ -71,5 +71,19 @@
...
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.gaul</groupId>
+                <artifactId>modernizer-maven-plugin</artifactId>
+                <configuration>
+                    <exclusions>
+                        <exclusion>java/lang/StringBuffer."&lt;init&gt;":()V</exclusion>
```
findepi, line 87:
```suggestion
                        <!-- TODO remove after upgrading to Java 9 -->
                        <exclusion>java/lang/StringBuffer."&lt;init&gt;":()V</exclusion>
```

findepi (approved):

findepi:
@choiwaiyiu can you please squash the commits into one?

choiwaiyiu:
@findepi I have moved the TODO comment as suggested and squashed the commits into one.

findepi:
Merged, thanks!

choiwaiyiu:
@findepi Thank you!

PR #875, opened 2020-09-30
Fix use of injected SslContextFactory

csunwold:
The SslContextFactory that was injected was always being ignored and the default behavior of creating a new ReloadableSslContextFactory was always in effect. 

In later lines you can see that the way the sslContextFactory was used was by using whichever was there or else creating a new ReloadableSslContextFactory. Since `this.sslContextFactory` was defaulted to Empty and the injected value `maybeSslContextFactory` was ignored this value would never be used.

Example:
```java
this.sslContextFactory = Optional.of(this.sslContextFactory.orElseGet(() -> createReloadingSslContextFactory(config, clientCertificate)));
```

I validated that existing tests would catch that `sslContextFactory` is not set after removing the default value.

### DIFF
--- a/http-server/src/main/java/io/airlift/http/server/HttpServer.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpServer.java
@@ -98,7 +98,7 @@ public class HttpServer
     private ConnectionStats httpConnectionStats;
     private ConnectionStats httpsConnectionStats;
     private ScheduledExecutorService scheduledExecutorService;
-    private Optional<SslContextFactory.Server> sslContextFactory = Optional.empty();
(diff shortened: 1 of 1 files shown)


electrum (approved):

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -134,6 +134,8 @@ public HttpServer(HttpServerInfo httpServerInfo,
         server = new Server(threadPool);
         registerErrorHandler = config.isShowStackTrace();
 
+        this.sslContextFactory = maybeSslContextFactory;
```
electrum, line 137: Add `requireNonNull`

PR #878, opened 2020-10-13
Remove @Beta annotation from ConfigurationAwareModule

kokosing:
Remove `@Beta` annotation from ConfigurationAwareModule

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationAwareModule.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationAwareModule.java
@@ -15,7 +15,6 @@
  */
 package io.airlift.configuration;
 
-import com.google.common.annotations.Beta;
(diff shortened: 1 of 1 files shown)


kokosing:
I am not sure the purpose of `@Beta` here. This interface is widely used and is does not seem like something that would be dropped.

kokosing:
`@Beta` causes a warning in Intellij.

findepi:
> `@Beta` causes a warning in Intellij.

is the warning emitted by default?
at least in Presto, we are accepting code using eg guava `@Beta`, so if this is a default inspection, we should recommend disabling it in https://github.com/prestosql/presto/blob/master/DEVELOPMENT.md#additional-ide-configuration

(i am not saying this change should or should not be merged)

electrum (approved):

PR #879, opened 2020-10-13
Add ConfigurationAwareModule#override

kokosing:
Add ConfigurationAwareModule#override


kokosing:
It does not work with configuration. It does not pass `configurationFactory` in the way airlift expects.

PR #890, opened 2020-12-18
Fix JettyHttpClient JMX stats

sopel39:


sopel39:
cc @electrum @findepi

findepi (approved):
Can you please add a note to CHANGES file?

PR #898, opened 2021-03-04
Upgrade Jersey to 2.31 for CWE-776

itay:
More info the CWE here:
* https://cwe.mitre.org/data/definitions/776.html
* https://app.snyk.io/vuln/SNYK-JAVA-ORGGLASSFISHJERSEYMEDIA-595972

It also seems that the Jersey version hasn't been changed since 2017 at least.

### DIFF
--- a/jaxrs/pom.xml
+++ b/jaxrs/pom.xml
@@ -56,11 +56,27 @@
         <dependency>
             <groupId>org.glassfish.jersey.core</groupId>
             <artifactId>jersey-server</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.glassfish.hk2.external</groupId>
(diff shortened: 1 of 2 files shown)


findepi:
XML Entity Expansion is cool.

However, I don't know of any place in Trino where we accept XML. 
Of course, doesn't mean we shouldn't upgrade.

BTW the build is red.

itay:
It's possible this change is beyond my ability to do, given I'm not quite sure how the project wants to handle dependencies.

Specifically, it seems like new versions of Jersey bring in Jakarta in some way, and that likely causes a need for some more exclusions that need to be added.

wendigo:
This can be closed @martint @electrum

PR #899, opened 2021-03-11
Don't fail silently when symlink source doesn't exist

nineinchnick:
Failing silently makes it hard to debug why some things stop working when changing paths.

There's a second commit formatting the whole file with [black](https://github.com/psf/black).

### DIFF
--- a/launcher/src/main/scripts/bin/launcher.py
+++ b/launcher/src/main/scripts/bin/launcher.py
@@ -16,7 +16,7 @@ from signal import SIGTERM, SIGKILL
 from stat import S_ISLNK
 from time import sleep
 
-COMMANDS = ['run', 'start', 'stop', 'restart', 'kill', 'status']
+COMMANDS = ["run", "start", "stop", "restart", "kill", "status"]
 
 LSB_NOT_RUNNING = 3
 LSB_STATUS_UNKNOWN = 4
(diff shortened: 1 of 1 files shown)


launcher/src/main/scripts/bin/launcher.py
```diff
@@ -16,19 +16,23 @@
 from stat import S_ISLNK
 from time import sleep
 
-COMMANDS = ['run', 'start', 'stop', 'restart', 'kill', 'status']
+COMMANDS = ["run", "start", "stop", "restart", "kill", "status"]
```
findepi, line 19: please move this to separate PR

launcher/src/main/scripts/bin/launcher.py
```diff
@@ -169,8 +169,11 @@ def create_symlink(source, target):
     """Create a symlink, removing the target first if it is a symlink"""
     if symlink_exists(target):
         os.remove(target)
-    if exists(source):
-        os.symlink(source, target)
+    if not exists(source):
```
electrum, line 172: I think this is the wrong behavior:

* we shouldn't remove the target if we're not going to create the symlink
* the `plugin` directory is optional, so this will break any application not using it

What is the purpose of this change?

nineinchnick: The first point is a good one, this kind of validation should be done as the first thing in `create_symlink`.

As for the `plugin` dir, I did not realize it's optional. This change is intended to avoid a silent failure, where the `etc` option points to an invalid dir. The service would start but would fail to operate in an expected way, as there are a few paths pointing to etc in the data dir by default.

How about I'd add an explicit check before calling this for the etc dir only? Or even better, validate the etc dir option explicitly?

findepi: > Or even better, validate the etc dir option explicitly?


sounds reasonable

PR #900, opened 2021-03-15
[launcher.py] Respect JAVA_HOME

bdw:
Ensure that the `JAVA_HOME` environment variable, if set, is respected; this helps users (like myself) who have multiple JDK variants installed

### DIFF
--- a/launcher/src/main/scripts/bin/launcher.py
+++ b/launcher/src/main/scripts/bin/launcher.py
@@ -195,11 +195,6 @@ def build_java_execution(options, daemon):
     if options.log_levels_set and not exists(options.log_levels):
         raise Exception('Log levels file is missing: %s' % options.log_levels)
 
-    with open(os.devnull, 'w') as devnull:
-        try:
-            subprocess.check_call(['java', '-version'], stdout=devnull, stderr=devnull)
-        except (OSError, subprocess.CalledProcessError):
-            raise Exception('Java is not installed')
 
     properties = options.properties.copy()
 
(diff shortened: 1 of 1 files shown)


electrum:
Thanks for PR. However, I’m not sure we want this change.

We tell users to put the desired JDK first in `PATH`. Using `JAVA_HOME` would be a behavior change and could break people relying on the current behavior.

Does changing `PATH` work for you?

bdw:
In many operating systems (in particular, in fedora, but I think also debian) the `java` executable in the `PATH` is really a symlink to something installed elsewhere, `/usr/lib/jvm/java-11-openjdk-11.0.10.0.9-0.fc33.x86_64/bin/java` - and it is also common to have an application which runs on a particular version of the JDK (say, 8), and another that runs on JDK11, or JDK14, or whatever. `JAVA_HOME` is a easy and commonly supported way to override the path to the java executable; far less invasive than updating `PATH`.

Matter of fact quite a few systems have the opposite 'problem' - they won't work without setting `JAVA_HOME` :smile: 

Unfortunately write once run everywhere didn't really work out, otherwise we wouldn't be having this discussion.

The way I see it - respecting that you may see this differently - this add a feature that is fairly commonly expected of java services and doesn't affect anybody who doesn't care or doesn't know about it.

findepi:
What i've seen OS-es doing is to put `java` executable on `PATH`, which is not a dumb wrapper. Instead this "`$PATH/java`"  would inspect `JAVA_HOME` env (if set), and forward appropriately.
Obviously this isn't the case in your case, otherwise you wouldn't want this change here.

electrum:
This would be a breaking change if people are relying on the current `PATH` behavior but also have `JAVA_HOME` set to a location that is inconsistent with `PATH`. I'm hesitant to make this change, even if this scenario seems uncommon or unlikely, since the launcher is invoked from various places, and this sort of breakage might not be obvious when upgrading.

Would it make sense to enable this using a config flag for the launcher? Or does this scenario seem so unlikely as to not be worth considering?

bdw:
Yeah, it's true that this could be breaking in some edge cases.

Then again, could be *your* edge case, in which case, you'd probably rather not have a production outage.

It's unfortunate but I think I'll close this ticket then.

pan3793:
It's rare that a Java software silently ignores the `JAVA_HOME`, I suppose usually that `JAVA_HOME` takes higher priority than `java` under `PATH`, is it possible to provide a new property to allow it to respect `JAVA_HOME` while keeping backward-compatibility by default?

PR #903, opened 2021-03-22
Replace cglib with Byte Buddy for Java 16 compatibility

electrum:
Fixes #901

PR #904, opened 2021-03-22
Update Airbase to 107

wendigo:


electrum:
Unfortunately, it won’t be possible to update until we drop Java 8 support, since we can’t compile the JMX code on newer JDKs.

wendigo:
Yeah, I've noticed that @electrum and added a commit that drops JDK8 support and not it compiles cleanly on JDK11

electrum:
I don't think we want to drop JDK 8 support, since we use several of the modules for the Trino JDBC driver which is still JDK 8. We could use older versions for those, but that would be a problem if we need to fix a bug, plus confusion around multiple versions.

Another option is to see if we can rewrite that code in `JmxAgent8` to use reflection, so that it can compile on newer JDKs.

wendigo:
Now it should be ready for a review @electrum. PTAL

electrum:
We could also target JDK 8 for the library modules like `concurrent` but require JDK 11 for service modules like `jmx`.

wendigo:
@electrum what's the benefit? Trino is now [JDK11, +)

electrum:
I think you missed my previous comment:

> we use several of the modules for the Trino JDBC driver which is still JDK 8

electrum:
I merged the code style fixes in #906

electrum:
Done in #903. I figured out how to remove the `<release>` parameter from the `maven-compiler-plugin` configuration.

wendigo:
Nice @electrum. Thanks!

PR #905, opened 2021-03-23
Use ByteBuddy instead of cglib

wendigo:


findepi:
https://github.com/airlift/airlift/pull/903 ?

PR #914, opened 2021-05-11
Fix SSL context factory reloading

lukasz-walkiewicz:
Setting the SSL context while reloading was actually disabling the actual SSL
context factory reload. org.eclipse.jetty.util.ssl.SslContextFactory#load method
skips the most important part of the reloading process if the SSL context is set.
Added a missing test.

### DIFF
--- a/http-server/pom.xml
+++ b/http-server/pom.xml
@@ -180,6 +180,13 @@
         </dependency>
 
         <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcpkix-jdk15on</artifactId>
+            <version>1.58</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>org.testng</groupId>
             <artifactId>testng</artifactId>
             <scope>test</scope>
--- a/http-server/src/main/java/io/airlift/http/server/HttpServer.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpServer.java
@@ -88,7 +88,8 @@ import static java.util.concurrent.Executors.newSingleThreadScheduledExecutor;
 
 public class HttpServer
 {
-    public enum ClientCertificate {
+    public enum ClientCertificate
+    {
         NONE, REQUESTED, REQUIRED
     }
 
--- a/http-server/src/main/java/io/airlift/http/server/HttpServer.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpServer.java
@@ -21,6 +21,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import io.airlift.event.client.EventClient;
 import io.airlift.http.server.HttpServerBinder.HttpResourceBinding;
+import io.airlift.log.Logger;
 import io.airlift.node.NodeInfo;
 import io.airlift.tracetoken.TraceTokenManager;
 import org.eclipse.jetty.http2.server.HTTP2CServerConnectionFactory;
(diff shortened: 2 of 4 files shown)


electrum:
Without this change, `factory.getKeyStore()` returns null, so something is broken. However, if we don't call `buildContextFactory()`, then we won't actually reload the keystore. Thus, if using PEM files, which are loaded by that method, the PEM files would never be reloaded. So neither the original nor the fix is correct.

electrum:
I'm not sure the test is actually valid -- it is testing a side effect, rather than what certificates the `SSLContext` is actually using. It calls `HttpServer.getCertificates()` which calls `factory.getKeyStore()` which returns null after reloading. This means `getCertificates()` is broken, but the `getKeyStore()` method is not otherwise used. The internal `SSLContext` should have the correct `KeyManager`.

I'm not saying there is not a problem here -- based on https://github.com/trinodb/trino/issues/7845 and from various user reports on Slack, there does seem to be a problem. But we need to test this directly, either by using `SSLContext`, or better yet, making an HTTPS request to the server before and after reload.

It's important to understand that Jetty `SslContextFactory` is actually two things: a holder for `javax.net.ssl.SSLContext` which is what is actually used and some Jetty specific code for loading and inspecting `KeyStore` objects used as Jetty's KeyStore and TrustStore.

electrum:
We can fix `getCertificates()` and thus test by changing the reload code to do
```java
this.sslContextFactory.reload(factory -> {
    factory.setKeyStore(updatedFactory.getKeyStore());
    factory.setTrustStore(updatedFactory.getTrustStore());
    factory.setSslContext(updatedFactory.getSslContext());
});
```

electrum:
If the problem is actually due to reloading (i.e. it works before reload and fails after), then focus on the code in `org.eclipse.jetty.util.ssl.SslContextFactory#load` inside the `if (context == null)` block, since for JKS that will be called on the initial load, then not used the second time.

electrum:
I think I found the problem, which I now realize is what you explained in the commit description. After reload, we are skipping this block inside `load()`:
```java
// Look for X.509 certificates to create alias map
if (keyStore != null)
```
This sets `_aliasX509`, `_certHosts`, and `_certWilds` which are used for SNI.

We need to change the design of `ReloadableSslContextFactoryProvider` a bit. In the constructor, we create the factory with the global settings that won't change (we can remove the fields for these since they're only used in the constructor):
```java
sslContextFactory = new SslContextFactory.Server();
sslContextFactory.setIncludeCipherSuites(includedCipherSuites.toArray(new String[0]));
sslContextFactory.setExcludeCipherSuites(excludedCipherSuites.toArray(new String[0]));
sslContextFactory.setSecureRandomAlgorithm(secureRandomAlgorithm);
switch (clientCertificate) {
...
}
sslContextFactory.setSslSessionTimeout(sslSessionTimeoutSeconds);
sslContextFactory.setSslSessionCacheSize(sslSessionCacheSize);
```
Then we change `buildContextFactory()` to be `loadContextFactory(SslContextFactory)` and call the methods on that. So the constructor becomes
```
loadContextFactory(sslContextFactory);
```
And the `reload()` method is simply
```java
sslContextFactory.reload(this::loadContextFactory);
```
Technically, the Jetty `reload(Consumer<SslContextFactory>)` is simply passing back `this` for the instance, while inside a lock, but it's better not to rely on that (and to call the methods on the instance provided).

electrum:
Replaced by #914

kokosing:
> Replaced by #914

Replaced by https://github.com/airlift/airlift/pull/915

PR #921, opened 2021-05-26
Support multilingual encoding in property files

songgane:
I'm a Trino user, I'm reading the value from the property file for LDAP password authentication .

To read the property value, the airlift ConfigurationLoader is being used.

However if a value of Property file is not an English value, there is a bug that can not be read properly.

To support various countries and usecase, multilingual encoding support is required.

ex)
property value : ldap.user-bind-pattern=OU=사용자,DC=example,DC=com
compiled value : OU=ì�¬ì�©ì��,DC=example,DC=com

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
@@ -22,7 +22,6 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
-import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
@@ -30,6 +29,7 @@ import java.util.TreeMap;
 
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.Maps.fromProperties;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 public final class ConfigurationLoader
 {
@@ -60,8 +60,7 @@ public final class ConfigurationLoader
             throws IOException
     {
         Properties properties = new Properties();
-        try (Reader reader = new BufferedReader(new InputStreamReader(new FileInputStream(path),
-            StandardCharsets.UTF_8))) {
+        try (Reader reader = new BufferedReader(new InputStreamReader(new FileInputStream(path), UTF_8))) {
             properties.load(reader);
         }
 
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
@@ -17,9 +17,12 @@ package io.airlift.configuration;
 
 import com.google.common.collect.ImmutableSortedMap;
 
+import java.io.BufferedReader;
 import java.io.FileInputStream;
 import java.io.IOException;
-import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
```diff
@@ -57,8 +60,9 @@ private ConfigurationLoader() {}
             throws IOException
     {
         Properties properties = new Properties();
-        try (InputStream inputStream = new FileInputStream(path)) {
-            properties.load(inputStream);
+        try (Reader reader = new BufferedReader(new InputStreamReader(new FileInputStream(path),
+            StandardCharsets.UTF_8))) {
```
kokosing, line 64: Please do not break the line here. Also you can use static import for UTF_8

kokosing (approved):

configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
findepi: that's a backward incompatible change, isn't it?

what encoding does `Properties.load(InputStream)` use?

songgane: Properties.load(InputStream) uses ISO 8859-1 Character Encoding, so it does not support multilingual encoding.

> Reads a property list (key and element pairs) from the input byte stream. The input stream is in a simple line-oriented format as specified in load(Reader) and is assumed to use the ISO 8859-1 character encoding; that is each byte is one Latin1 character. Characters not in Latin1, and certain special characters, are represented in keys and elements using Unicode escapes as defined in section 3.3 of The Java™ Language Specification.

With your help,  I checked that the multilingual value in the property file was correctly read after converting values using native2ascii.
But I think it is a good alternative to help reading multilingual encoding values in properties files without a separate conversion process.

Because it is a situation that reads the properties file, I think the backward compatibility issue will not be a problem.
How about you?

-  links
https://stackoverflow.com/questions/652161/how-do-i-convert-between-iso-8859-1-and-utf-8-in-java
https://www.joelonsoftware.com/2003/10/08/the-absolute-minimum-every-software-developer-absolutely-positively-must-know-about-unicode-and-character-sets-no-excuses/

findepi: Contents of .properties file can be any `ISO 8859-1`-encoded text. 
It cannot be interpreted as UTF-8-encoded text. This may fail, or may return incorrect results.

martint:
Java property files are expected to be ISO-8859-1, and historically, we've said that Airlift configuration files are Java property files. We can change this, but we need to think about the implications around backward compatibility. It may not be an real issue in practice if people don't actually use non-ISO-8859-1 characters (which would be encoded using `\uxxxx`)

electrum:
The JDK seems to be moving to UTF-8 for properties files: https://openjdk.java.net/jeps/226

electrum:
The only compatibility issue is if people are using un-escaped, non-ACSII characters, which seems unlikely given the usage of properties file for configuration. I think this is a good change as UTF-8 is clearly a better choice and aligns with the direction of the JDK.

Note that you can use `\uxxxx` today, which is obviously not as convenient or readable, but does work.

findepi:
https://openjdk.java.net/jeps/226 talks about `ResourceBundle` (which builds upon Properties), but not Properties themselves.
second, it also describes detection and fallback mechanism. I am not convinced yet fallback is safe thing to do, but for such 
backwards incompatible change, a detection feels like must-have.

Also, since configuration files rarely contain non-ASCII anyway, there is not a ton of value pursuing this project.

songgane:
This PR is began for the following reasons:

> Trino supports LDAP by external password authentication method.
> However if ldap.user-bind-pattern value of the LDAP configuration file contains non-ascii characters, it can not be read properly.
> (OpenLDAP/Active Directory supports non-ascii charaters. So some legacy systems use multilingual entry)
> Trino is used in various countries and support external authentications. So multilingual encoding support is required.

This PR will close because there is an alternative to native2ascii and a backward compatibility issue may occur.

PR #925, opened 2021-06-06
Disable admin port by default

electrum:


http-server/src/test/java/io/airlift/http/server/DummyServlet.java
```diff
@@ -50,8 +52,8 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response)
         }
     }
 
-    public CountDownLatch getLatch()
+    public Future<?> getSleeping()
     {
-        return latch;
+        return sleeping;
```
erichwang, line 57: not a huge deal since this is for test, but you technically should wrap this in a non cancellable wrapper

erichwang (approved):

http-server/src/test/java/io/airlift/http/server/DummyServlet.java
electrum: Hah, I was thinking that, but didn't bother for the test

electrum: Fixed

PR #928, opened 2021-06-11
Migrate to ListenableFuture<Void>

erichwang:
ListenableFuture<Void> provides a cleaner programming experience for signaling workloads than the commonly used in ListenableFuture<?>. In this PR, we migrate some of the libraries to using ListenableFuture<Void> and provide some useful utilities for converting to it.

### DIFF
--- a/concurrent/src/test/java/io/airlift/concurrent/TestAsyncSemaphore.java
+++ b/concurrent/src/test/java/io/airlift/concurrent/TestAsyncSemaphore.java
@@ -52,7 +52,7 @@ public class TestAsyncSemaphore
 
         AtomicInteger count = new AtomicInteger();
 
-        List<ListenableFuture<?>> futures = new ArrayList<>();
+        List<ListenableFuture<Void>> futures = new ArrayList<>();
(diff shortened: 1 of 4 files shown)


electrum (approved):
Nice! Please update the CHANGES file.

concurrent/src/test/java/io/airlift/concurrent/TestAsyncSemaphore.java
```diff
@@ -52,7 +52,7 @@ public void testInlineExecution()
 
         AtomicInteger count = new AtomicInteger();
 
-        List<ListenableFuture<?>> futures = new ArrayList<>();
+        List<ListenableFuture<Void>> futures = new ArrayList<>();
```
electrum, line 55: Technically, none of these test changes are required, since this should be a backwards compatible change? (though we should update them)

erichwang: Yea, you don't need this, but it's what we would have done if we started with this pattern -- so might as well update it.

PR #937, opened 2021-07-08
Add support for json logging

mattstep:


log-manager/src/main/java/io/airlift/log/LoggingConfiguration.java
```diff
@@ -104,4 +105,15 @@ public LoggingConfiguration setLevelsFile(String levelsFile)
         this.levelsFile = levelsFile;
         return this;
     }
+
+    public boolean useJson()
+    {
+        return useJson;
+    }
+
+    @Config("log.json")
```
martint, line 114: I would model this as a setting that controls the logging format, instead, with possible values "TEXT" and "JSON"

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
```diff
@@ -0,0 +1,47 @@
...
+        JsonRecord jsonRecord = new JsonRecord(
+                TIMESTAMP_FORMATTER.format(timestamp),
+                Level.fromJulLevel(record.getLevel()).name(),
+                Thread.currentThread().getName(),
+                record.getLoggerName(),
+                record.getMessage(),
+                stackTraceWriter.toString());
+
+        try {
+            return mapper.writeValueAsString(jsonRecord);
```
martint, line 40: Will this produce multi-line json records? We should consider producing a single entry per line for easier downstream processing, especially since the file itself is not valid json -- just a sequence of json objects.

It turns out there are a couple of popular "formats" that work in a similar way: 
* `JSONL`: https://jsonlines.org/
* ndjson: http://ndjson.org/
* JSON text sequences: https://datatracker.ietf.org/doc/html/rfc7464

https://en.wikipedia.org/wiki/JSON_streaming#Record_separator-delimited_JSON

log-manager/src/main/java/io/airlift/log/StaticFormatter.java
```diff
@@ -23,9 +23,9 @@
 class StaticFormatter
         extends Formatter
 {
-    private static final ZoneId SYSTEM_ZONE = ZoneId.systemDefault().normalized();
```
electrum, line 26: We can leave these private if we use `Instant` for JSON

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
electrum: As long as the JSON itself is not "pretty", the values should not contain strings, since JSON strings are not permitted to have newlines or other control characters.

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
```diff
@@ -0,0 +1,47 @@
...
+        extends Formatter
+{
+    private static ObjectMapper mapper = new ObjectMapper();
+
+    @Override
+    public String format(LogRecord record)
+    {
+        ZonedDateTime timestamp = ZonedDateTime.ofInstant(Instant.ofEpochMilli(record.getMillis()), SYSTEM_ZONE);
+
+        StringWriter stackTraceWriter = new StringWriter();
```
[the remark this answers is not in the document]
mattstep: I've pulled out message and class name for the throwable as well as using Throwables.getStackTraceAsString. If this is used with the typical fluent->elk logging pipeline, the stackTraceAsString should be a fully indexed and searchable field, which should include the chain at least. As a first cut, would this be sufficent?

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
mattstep: It doesn't look like it should according to the jackson docs, I'll dig through the code some more to confirm. It seems like it'll use \r\n instead of a newline within values

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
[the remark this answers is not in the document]
electrum: No, because the message for the nested exceptions is an arbitrary string, and thus it would be impossible to parse the exception correctly.

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
```diff
@@ -0,0 +1,35 @@
...
+import io.airlift.json.ObjectMapperProvider;
+
+import java.time.Instant;
+import java.util.logging.Formatter;
+import java.util.logging.LogRecord;
+
+public class JsonFormatter
+        extends Formatter
+{
+    private static ObjectMapper mapper = new ObjectMapperProvider().get();
```
electrum, line 14: This can be
```java
private static final JsonCodec<JsonRecord> CODEC = new JsonCodecFactory(new ObjectMapperProvider()).jsonCodec(JsonRecord.class);

File: log-manager/src/main/java/io/airlift/log/JsonFormatter.java [hunk 6]
```diff
@@ -0,0 +1,35 @@
...
+                record.getLoggerName(),
+                record.getMessage(),
+                record.getThrown());
+
+        try {
+            return mapper.writeValueAsString(jsonRecord);
+        }
+        catch (JsonProcessingException e) {
+            // Emit a log line that is at least json parseable and indicates things are broken
+            return "{\"timestamp\": \"" + jsonRecord.getTimestamp() + "\", \"message\":\"unable to log json record\", \"level\": \"" + Level.ERROR.name() + "\"}";
```
electrum, line 32: Nit: capitalize message

> Unable to create JSON log record

log-manager/src/main/java/io/airlift/log/JsonRecord.java
```diff
@@ -0,0 +1,99 @@
...
+import static java.util.Objects.requireNonNull;
+
+public class JsonRecord
+{
+    private final Instant timestamp;
+    private final Level level;
+    private final String thread;
+    private final String loggerName;
+    private final String message;
+    private final Throwable thrown;
```
electrum, line 18: Should we call this (and the getters) `throwable`? This better matches the logger API which logs a throwable, although `java.util.logging.Logger` and `LogRecord` calls it `thrown`.

log-manager/pom.xml
```diff
@@ -86,6 +101,10 @@
             <artifactId>testng</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
electrum, line 104: Nit: move to upper part. We put `test` dependencies at the end

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
```diff
@@ -0,0 +1,45 @@
...
+                Level.fromJulLevel(record.getLevel()),
+                Thread.currentThread().getName(),
+                record.getLoggerName(),
+                record.getMessage(),
+                record.getThrown());
+
+        try {
+            return CODEC.toJson(jsonRecord);
+        }
+        catch (IllegalArgumentException outerIllegalArgumentException) {
```
martint, line 30: The variable could be named just `outer` for conciseness. Same for `inner` below

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class JsonFormatter
+        extends Formatter
+{
+    private static final JsonCodec<JsonRecord> CODEC = new JsonCodecFactory(new ObjectMapperProvider()).jsonCodec(JsonRecord.class);
+
+    @Override
+    public String format(LogRecord record)
+    {
+        JsonRecord jsonRecord = new JsonRecord(
```
martint, line 19: The variable could be name just `record` for conciseness.

mattstep: There is already a `record` in that scope

martint: Oops. Missed that

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
```diff
@@ -0,0 +1,45 @@
...
+                return CODEC.toJson(new JsonRecord(Instant.ofEpochMilli(record.getMillis()),
+                        Level.fromJulLevel(record.getLevel()),
+                        Thread.currentThread().getName(),
+                        record.getLoggerName(),
+                        outer.getMessage(),
+                        outer));
+            }
+            catch (IllegalArgumentException inner) {
+                // Emit a log line that is at least json parseable and indicates things are broken
+                return "{\"timestamp\": \"" + jsonRecord.getTimestamp() + "\", \"message\":\"" + inner.getMessage() + "\", \"level\": \"" + Level.ERROR.name() + "\"}";
```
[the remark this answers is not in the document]
electrum: Message could contain quotes, newlines, etc. You can do
```java
private static final JsonCodec<String> STRING_CODEC = jsonCodec(String.class);

STRING_CODEC.toJson(...);

**Review by electrum — APPROVED** [review]
Can you add some tests that directly invoke `JsonFormatter`?

File: log-manager/src/main/java/io/airlift/log/JsonFormatter.java (hunk 11 above)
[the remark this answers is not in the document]
Comment (electrum) [reply log-manager/src/main/java/io/airlift/log/JsonFormatter.java:41]: Or you could use the Jackson streaming API to write the object: https://www.baeldung.com/jackson-streaming-api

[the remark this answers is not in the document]
Comment (mattstep) [reply log-manager/src/main/java/io/airlift/log/JsonFormatter.java]: Fixed

File: log-manager/src/test/java/io/airlift/log/TestJsonFormatter.java [hunk 12]
```diff
@@ -0,0 +1,70 @@
...
+{
+    @Test
+    public void testMinimalJsonErrorLogLine()
+    {
+        JsonRecord originalJsonRecord = new JsonRecord(Instant.now(), Level.DEBUG, "thread-0", "TestLogger", "Test Log Message", new Exception("Test Exception 1"));
+
+        RuntimeException exception = new RuntimeException("Test Exception 2");
+        String minimalJsonErrorLogLine = (new JsonFormatter()).minimalJsonErrorLogLine(originalJsonRecord, exception);
+
+        assertTrue(minimalJsonErrorLogLine.endsWith("\n"), "Log lines should end with newline");
```
electrum, line 26: Use AssertJ for this, which we're using in new tests:
```java
assertThat(minimalJsonErrorLogLine).as("Log lines should end with newline").endsWith("\n");
```
This produces a better error message since it shows the actual input. Generally, avoid using `assertTrue()` with a complex expression.

log-manager/src/test/java/io/airlift/log/TestJsonFormatter.java
```diff
@@ -0,0 +1,70 @@
...
+
+        assertEquals(
+                jsonCodec(JsonRecord.class).fromJson(minimalJsonErrorLogLine),
+                new JsonRecord(originalJsonRecord.getTimestamp(), Level.ERROR, null, null, exception.getMessage(), null));
+    }
+
+    @Test
+    public void testRoundTrip()
+    {
+        JsonRecord originalJsonRecord = new JsonRecord(Instant.now(), Level.DEBUG, "thread-0", "TestLogger", "Test Log Message", new Exception("Test Exception 1"));
```
electrum, line 36: I'd shorten this to `original` so the assertion is easier to read

log-manager/src/test/java/io/airlift/log/TestJsonFormatter.java
```diff
@@ -0,0 +1,70 @@
...
+    public void testLogFormatting()
+    {
+        Exception testException = new RuntimeException("Test Exception");
+        LogRecord record = new LogRecord(Level.DEBUG.toJulLevel(), "Test Log Message");
+        record.setLoggerName("TestLogger");
+        record.setThrown(testException);
+
+        String logMessage = (new JsonFormatter()).format(record);
+
+        Map<String, String> jsonMap = jsonCodec(Map.class).fromJson(logMessage);
```
electrum, line 54: I'm surprised this compiles, since `Map` is a raw type, but it does with an unchecked assignment warning. You can make this type safe with
```java
mapJsonCodec(String.class, String.class)

File: log-manager/src/test/java/io/airlift/log/TestJsonFormatter.java [hunk 15]
```diff
@@ -0,0 +1,70 @@
...
+        JsonRecord jsonRecord = jsonCodec(JsonRecord.class).fromJson(logMessage);
+
+        assertTrue(logMessage.endsWith("\n"), "Log lines should end with newline");
```
electrum, line 57: Use AssertJ

log-manager/src/test/java/io/airlift/log/TestJsonFormatter.java
```diff
@@ -0,0 +1,70 @@
...
+        assertEquals(
+                jsonCodec(JsonRecord.class).fromJson(minimalJsonErrorLogLine),
+                new JsonRecord(originalJsonRecord.getTimestamp(), Level.ERROR, null, null, exception.getMessage(), null));
+    }
+
+    @Test
+    public void testRoundTrip()
+    {
+        JsonRecord originalJsonRecord = new JsonRecord(Instant.now(), Level.DEBUG, "thread-0", "TestLogger", "Test Log Message", new Exception("Test Exception 1"));
+        JsonCodec<JsonRecord> jsonRecordJsonCodec = jsonCodec(JsonRecord.class);
```
electrum, line 37: Maybe inline this?

electrum (approved):
A few nits about the tests, otherwise looks good

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
```diff
@@ -0,0 +1,75 @@
...
+                Thread.currentThread().getName(),
+                record.getLoggerName(),
+                record.getMessage(),
+                record.getThrown());
+
+        try {
+            return new StringWriter().append(CODEC.toJson(jsonRecord)).append("\n").toString();
+        }
+        catch (IllegalArgumentException outer) {
+            try {
```
findepi, line 36: Why try almost exactly same thing (json-serialize JsonRecord with almost same fields) twice?

PR #938, opened 2021-07-12
Only bind HTTPS config for HTTP server when HTTPS is enabled

electrum:


bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -139,9 +139,15 @@ public Bootstrap quiet()
...
     }
 
+    @Deprecated
     public Bootstrap strictConfig()
     {
-        this.strictConfig = true;
+        return strictConfig(true);
+    }
+
+    public Bootstrap strictConfig(boolean strictConfig)
```
dain, line 148: Maybe add a method to make it "not strict", and make that deprecated also, so we can remove non-strictness in the future.

dain (approved):
Looks good

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
electrum: I'll make this a separate PR

PR #940, opened 2021-07-15
Bootstrap cleanup

electrum:


bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -268,6 +268,7 @@ public Injector initialize()
         moduleList.add(Binder::disableCircularProxies);
         moduleList.add(Binder::requireExplicitBindings);
         moduleList.add(Binder::requireExactBindingAnnotations);
+        moduleList.add(Binder::requireAtInjectOnConstructors);
```
dain, line 271: This one is likely more annoying than it is worth.  IIRC this only applies when the class has explicit constructors with arguments.

dain (approved):
I'd drop the last commit

PR #948, opened 2021-09-14
Introduce SwitchModule

kokosing:
Introduce SwitchModule

Next iteration of ConditionalModule, that is suited for installing
different module depending on different enum value.


electrum (approved):

electrum:
Nice, I like this abstraction.

PR #949, opened 2021-09-18
Use `Objects.deepEquals` to address array type parameters

reugn:


findepi:
I'd rather discourage array as config values. Why not use List instead?

reugn:
Application specifics. I could use a List but then would need to convert the value twice. There are several examples of built-in array support in the configuration, for instance, the Spring framework. Why not supporting that?

findepi:
Arrays are mutable. While Config objects are by nature mutable too, it's recommended that config values are immutable. Otherwise using code needs to make defensive copy anyway, which doesn't avoid "convert the value twice" concern.

reugn:
That makes sense. If you think that supporting arrays in assertions will harm the framework, please close this PR.

PR #954, opened 2021-10-19
Add more @FormatMethod annotation in Logger

ksobolew:
They may not look like format methods, but in fact they are, just without parameters. The reason to treat them this way is that then Error Prone can detect mis-use of these methods, like passing a String concatenation instead of a format string plus arguments.

The caveat is that from now on the message on these methods will be actually treated as format string, so any `%` characters will be interpreted. This will break messages like

```
Frobnicity increased by 10%!
```

which would have to be changed to

```
Frobnicity increased by 10%%!
```

This is a breaking change, but it makes treatment of the messages consistent across all variants of these log methods. For compatibility with existing code, if there is `IllegalFormatException` thrown by `String#format`, we fall back to treating the string as a literal message.

This PR includes dogfooding of this change on Airlift itself. There is a simple Maven profile provided which enables the relevant check (and only that).

### DIFF
--- a/log/src/main/java/io/airlift/log/Logger.java
+++ b/log/src/main/java/io/airlift/log/Logger.java
@@ -18,6 +18,7 @@ package io.airlift.log;
 import com.google.errorprone.annotations.FormatMethod;
 
 import java.util.IllegalFormatException;
+import java.util.function.Supplier;
 
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
@@ -59,6 +60,17 @@ public class Logger
     }
 
     /**
+     * Logs a message, provided by the given supplier, at DEBUG level.
+     *
+     * @param exception an exception associated with the debug message being logged
+     * @param messageSupplier a {@link Supplier} of the pre-formatted message
+     */
+    public void debug(Throwable exception, Supplier<String> messageSupplier)
+    {
+        logger.log(FINE, exception, messageSupplier);
+    }
+
+    /**
      * Logs a message at DEBUG level.
      *
      * @param exception an exception associated with the debug message being logged
@@ -67,6 +79,16 @@ public class Logger
     public void debug(Throwable exception, String message)
     {
         logger.log(FINE, message, exception);
+    }
+
+    /**
+     * Logs a message, provided by the given supplier, at DEBUG level.
+     *
+     * @param messageSupplier a {@link Supplier} of the pre-formatted message
+     */
+    public void debug(Supplier<String> messageSupplier)
+    {
+        debug(null, messageSupplier);
     }
 
     /**
@@ -115,9 +137,17 @@ public class Logger
     @FormatMethod
     public void debug(Throwable exception, String format, Object... args)
     {
-        if (logger.isLoggable(FINE)) {
-            logger.log(FINE, formatMessage(format, "DEBUG", args), exception);
-        }
+        debug(exception, () -> formatMessage(format, "DEBUG", args));
+    }
+
+    /**
+     * Logs a message, provided by the given supplier, at INFO level.
+     *
+     * @param messageSupplier a {@link Supplier} of the pre-formatted message
+     */
+    public void info(Supplier<String> messageSupplier)
+    {
+        logger.log(INFO, messageSupplier);
     }
 
     /**
@@ -146,9 +176,18 @@ public class Logger
     @FormatMethod
     public void info(String format, Object... args)
     {
-        if (logger.isLoggable(INFO)) {
-            logger.log(INFO, formatMessage(format, "INFO", args));
-        }
+        info(() -> formatMessage(format, "INFO", args));
+    }
+
+    /**
+     * Logs a message, provided by the given supplier, at WARN level.
+     *
+     * @param exception an exception associated with the warning being logged
+     * @param messageSupplier a {@link Supplier} of the pre-formatted message
+     */
+    public void warn(Throwable exception, Supplier<String> messageSupplier)
+    {
+        logger.log(WARNING, exception, messageSupplier);
     }
 
     /**
@@ -160,6 +199,16 @@ public class Logger
     public void warn(Throwable exception, String message)
     {
         logger.log(WARNING, message, exception);
+    }
+
+    /**
+     * Logs a message, provided by the given supplier, at WARN level.
+     *
+     * @param messageSupplier a {@link Supplier} of the pre-formatted message
+     */
+    public void warn(Supplier<String> messageSupplier)
+    {
+        warn(null, messageSupplier);
     }
 
     /**
@@ -189,9 +238,7 @@ public class Logger
     @FormatMethod
     public void warn(Throwable exception, String format, Object... args)
     {
-        if (logger.isLoggable(WARNING)) {
-            logger.log(WARNING, formatMessage(format, "WARN", args), exception);
-        }
+        warn(exception, () -> formatMessage(format, "WARN", args));
     }
 
     /**
@@ -214,6 +261,17 @@ public class Logger
     }
 
     /**
+     * Logs a message, provided by the given supplier, at ERROR level.
+     *
+     * @param exception an exception associated with the error being logged
+     * @param messageSupplier a {@link Supplier} of the pre-formatted message
+     */
+    public void error(Throwable exception, Supplier<String> messageSupplier)
+    {
+        logger.log(SEVERE, exception, messageSupplier);
+    }
+
+    /**
      * Logs a message at ERROR level.
      *
      * @param exception an exception associated with the error being logged
@@ -222,6 +280,16 @@ public class Logger
     public void error(Throwable exception, String message)
     {
         logger.log(SEVERE, message, exception);
+    }
+
+    /**
+     * Logs a message, provided by the given supplier, at ERROR level.
+     *
+     * @param messageSupplier a {@link Supplier} of the pre-formatted message
+     */
+    public void error(Supplier<String> messageSupplier)
+    {
+        error(null, messageSupplier);
     }
 
     /**
@@ -251,9 +319,7 @@ public class Logger
     @FormatMethod
     public void error(Throwable exception, String format, Object... args)
     {
-        if (logger.isLoggable(SEVERE)) {
-            logger.log(SEVERE, formatMessage(format, "ERROR", args), exception);
-        }
+        error(exception, () -> formatMessage(format, "ERROR", args));
     }
 
     /**
@@ -268,9 +334,7 @@ public class Logger
      */
     public void error(Throwable exception)
     {
-        if (logger.isLoggable(SEVERE)) {
-            logger.log(SEVERE, exception.getMessage(), exception);
-        }
+        error(exception, exception::getMessage);
     }
 
     /**
--- a/log/src/main/java/io/airlift/log/Logger.java
+++ b/log/src/main/java/io/airlift/log/Logger.java
@@ -32,6 +32,8 @@ public class Logger
 {
     private final java.util.logging.Logger logger;
 
+    private static final Object[] NO_ARGS = {};
+
     Logger(java.util.logging.Logger logger)
     {
         this.logger = requireNonNull(logger, "logger is null");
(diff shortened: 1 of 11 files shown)


log/src/main/java/io/airlift/log/Logger.java
```diff
@@ -74,6 +75,7 @@ public void debug(Throwable exception, String message)
      *
      * @param message a literal message to log
      */
+    @FormatMethod
```
findepi, line 78: does `log.debug("%%")` log single percent or two?

findepi: is the caller allows to call `log.debug("%...a malformed format string")`?

ksobolew: Ah, I haven't thought of that :/ No, the `%`s won't be interpreted by these methods, but Error Prone will treat them as format strings regardless. So, back to the drawing board I guess.

ksobolew: I pushed an update, with the following caveat:

```
(...) from now on the message on these methods will be actually treated as format string (...) This is a breaking change, but it makes treatment of the messages consistent across all variants of these log methods.
```

Are we willing to accept that?

electrum (changes requested):
These are not format methods. Making them so is a breaking change to an API that has existed for over 10 years. I like the idea of catching misuse, but not at the expense of breaking logging which is often not tested or only happens in exceptional circumstances.

electrum:
Do you have any examples of misuse seen in practice?

I’m wondering if we can do something clever here. There are two kinds of callers:

- Static message. These can be detected with Error Prone and updated.
- Dynamic message:  `log.error(e.getMessage())`

Maybe we change so that that the message is now a format string, but if it’s malformed, we simply log the original (old behavior). This would only be for the no-arg version.

It makes usage slightly uglier, but makes the API consistent and allows checkers.

cc @martint

ksobolew:
I think there are about 30 in Trino. They are in four general categories:

* Logging using concatenation instead of the format string:
  ```java
   log.warn("Switching coordinator affinity from " + coordinatorId + " to " + assignments.getCoordinatorId());
  ```
* Logging a pre-formatted String:
  ```java
  log.info(format("Copying resource dir '%s' to %s", resourceName, tempDir));
  ```
* Preparing a message beforehand so that it can be used to log and as an exception message:
  ```java
  String message = format("Failure when checking behavior against %s", driver);
  // log immediately since further tests may take more time; "log and rethrown" is not harmful in tests
  log.error(e, message);
  failures.add(new AssertionError(message, e));
  ```
* Non-static-string (like you mentioned):
  ```java
  log.error(message.toString());
  ```

Overall, nothing serious, "just" a code quality issue.

I proposed to mark these methods as format methods, but I am aware that this is an indirect way to make Error Prone do what we want. There are other tools we could use, for example SpotBugs (with some external checks) can detect mis-use of logger methods, but I don't think any of them knows that these are actually logger methods. So we still have to annotate them *somehow*, and this seems like it's as good as any other way to do that.

So given the above the question now is: is it OK that we mark these methods as format methods even though they're not really? :)

ksobolew:
And one more thing - I haven't actually found any cases where the message contains something that might be misinterpreted as a format string. Not in Trino, at least. And not in static strings. It may still happen when we log a result of some invocation, though.

ksobolew:
One reason to treat no-param logs as format strings is when we want to merge subsequent logs into one, so that they are both logged atomically:

```diff
-        log.info("trying to give consent");
-        log.info("waiting for openId checkbox");
+        log.info("trying to give consent%nwaiting for openId checkbox");
```

The `%n` format specifier is the portable way to print newline. It could be replaced with `\n`, but that would not be ideal.

electrum:
This seems reasonable to me. I do like the consistency, and the fallback of catching `IllegalFormatException` seems like a good approach.

Can you add a test for including valid placeholders like `%s` and `%d` in the no-args version?

electrum:
To expand on the motivation for this change, it allows Error Prone to flag code like this: 

```java
 log.debug("Task scheduled " + taskId); 
```

```
trino/core/trino-main/src/main/java/io/trino/execution/executor/TaskExecutor.java:266:18
java: [FormatStringAnnotation] Format strings must be either literals or variables. Other expressions are not valid.
  Invalid format string: "Task scheduled " + taskId
    (see https://errorprone.info/bugpattern/FormatStringAnnotation)
```

electrum (approved):
Can you also update the `CHANGES` file in the commit that makes the behavior change?

ksobolew:
@electrum AC, thanks!

martint:
I'm not yet 100% convinced this is something we want to change. The choice of making the no-args version a regular string (vs a format string) was intentional. It is true that "they may not look like format methods, but in fact they are, just without parameters". That was the original intention (except for the implications around how naked `%` are interpreted). 

It makes it more cumbersome to log naked strings that contain '%'. With this change, callers are forced to do `log.debug('%s', string)` or manually escape `%` characters beforehand. As a consequence, there is no safe way to use the no-arg method unless you know it doesn't contain `%` -- and every time you call it, you have to think about that fact. We might as well just remove the no-arg method, in that case.

Also, I don't like the implications and constraints this forces.  In particular, while things like `log.debug("xxx " + taskId)` are *generally* not desirable, they are also not catastrophic. Sometimes they are even necessary. E.g., you might want to log the result of a function call, so "either literals or variables" is too constraining, and forcing the caller to create an artificial variable is an anti-pattern.

findepi:
> It makes it more cumbersome to log naked strings that contain '%'. With this change, callers are forced to do `log.debug('%s", string)` or manually escape `%` characters beforehand. As a consequence, there is no safe way to use the no-arg method unless you know it doesn't contain `%` -- and every time you call it, you have to think about that fact. We might as well just remove the no-arg method, in that case.

@martint see also https://github.com/airlift/airlift/pull/954#issuecomment-947418282



> Also, I don't like the implications and constraints this forces. In particular, while things like `log.debug("xxx " + taskId)` are _generally_ not desirable, they are also not catastrophic. Sometimes they are even necessary. E.g., you might want to log the result of a function call, so "either literals or variables" is too constraining, and forcing the caller to create an artificial variable is an anti-pattern.

I agree with "not catastrophic".
i don't agree with "necessary". Concatenation in `log.debug(...)`  is a bad/undesired code which we want to eliminate, because we pay concatenation cost even if debug log is disabled. `log.debug(foo(...))` falls into same cathegory.
If the _intent_ is to call `foo` regardless of log level settings, and log the result if logging enabled, `log.debug(foo(...))` is a pretty nice way to hide that intent.


Anyway, the problem we're solving is: "how to find `log.debug(...)` calls with concatenation".
We found a disputable solution: mark this as a `@FormatMethod` and let errorprone do the job. 
I'd hope a less contented solution exists. Perhaps we can add an additional checker to errorprone. Similar to how `@FormatMethod` check works, but without misrepresenting `log.debug` as a formatting.

electrum:
Related, we should add supplier versions of all the no-arg logger methods, for times when you need to log the result of a method call or other expression. I think that would largely eliminate the objection, since it's very rare you need to log an arbitrary pre-constructed string.

While I understand @martint's argument, I think it's rare enough that the API consistency and Error Prone check makes the extra verbosity in those cases worthwhile.

electrum:
@ksobolew There are a few places in Airlift to update before this change, which are examples of arbitrary strings passed to the no-arg methods:

* `io.airlift.http.client.jetty.JettyClientDiagnostics`
* `io.airlift.bootstrap.LoggingWriter`
* `io.airlift.discovery.client.ExponentialBackOff`
* `io.airlift.log.LoggingOutputStream`
* `io.airlift.bootstrap.Bootstrap` using `log::warn`

martint:
> log.debug(foo(...)) falls into same category. If the intent is to call foo regardless of log level settings, and log the result if logging enabled, log.debug(foo(...)) is a pretty nice way to hide that intent.

What about:

```
String result = foo(...);
log.debug(foo);
```

That a seems pretty legit usage to me, and forcing `log.debug("%s", foo)` is very unnatural and likely for people to forget to do that. Unless the result from `foo()` properly escapes `%`, that is a buggy line.

In fact, such a usage now would become quite dangerous. If, for instance, `result` contains a sequence like "%1000000000%", `String.format` will expand that into a 1GB string. So innocent usages like the one above that would be easy to overlook during code review and hard to catch with ErrorProne would be potential crashes or serious performance issues waiting to happen -- way worse than the effect of string concatenation.

> Related, we should add supplier versions of all the no-arg logger methods, for times when you need to log the result of a method call or other expression. I think that would largely eliminate the objection, since it's very rare you need to log an arbitrary pre-constructed string.

Can you elaborate? Not sure how it solves the problem. Also, using capturing lambdas can have performance implications, as they are constructed eagerly before the method gets called. It'd be trading one problem (string concatenation) for another one (creating capturing lambdas before the call).

> I think it's rare enough

I personally use that often enough during development or debugging that I wouldn't want it to get in the way. It's not just about the code that will end up merged, but throwaway code you write/build/run while prototyping and developing.

martint (changes requested):
Per comment above, code like this has worse problems that those caused string concatenation that this change is attempting to catch.

```
String result = foo(...); // function which returns a string containing "%1000000000%"
log.debug(result);
```

ksobolew:
> Per comment above, code like this has worse problems that those caused string concatenation that this change is attempting to catch.
> 
> ```
> String result = foo(...); // function which returns a string containing "%1000000000%"
> log.debug(result);
> ```

I agree that this is potentially dangerous. On the other hand, this will definitely be flagged by Error Prone. On the third hand, Error Prone is not a silver bullet as it takes effort to enable it, and in projects like Trino it's not even enabled by default. So, yeah, this is a hard balance. Personally I'm biased towards more static analysis, even at the cost of some inconvenience, but I understand the hesitation.

ksobolew:
> > Related, we should add supplier versions of all the no-arg logger methods, for times when you need to log the result of a method call or other expression. I think that would largely eliminate the objection, since it's very rare you need to log an arbitrary pre-constructed string.
> 
> Can you elaborate? Not sure how it solves the problem. Also, using capturing lambdas can have performance implications, as they are constructed eagerly before the method gets called. It'd be trading one problem (string concatenation) for another one (creating capturing lambdas before the call).

The result of the lambda call is not interpreted in any way, including not as a format pattern, so this is one way to fix the problem of potentially dangerous strings. As an added benefit, it reduces overhead in some cases, as the message construction can be encapsulated and executed only when needed.

Regarding the performance concerns, the use case for lambda-taking methods is in places where creating the string is way more costly than setting up a lambda (which I don't believe have any measurable cost anyway, but that's just an aside). It's not for cases where you have a string that you'll want to use for other purposes as well, only for those where the message is constructed solely for the logger call. As this PR demonstrates, there are two such places in Airlift alone.

martint:
> I agree that this is potentially dangerous. On the other hand, this will definitely be flagged by Error Prone

How will Error Prone know to flag that? How would it ever be able to tell that the output from `foo(...)` might contain such a dangerous pattern?

ksobolew:
> > I agree that this is potentially dangerous. On the other hand, this will definitely be flagged by Error Prone
> 
> How will Error Prone know to flag that? How would it ever be able to tell that the output from `foo(...)` might contain such a dangerous pattern?

Well, it's not smart enough to track "tainted" (or "unsanitized") values, but it sees that the `debug(String)` method is a `@FormatMethod` (assuming this PR is merged) and requires the argument to be a String literal, exactly because otherwise you don't know if the string is potentially dangerous.

martint:
Well, that sucks :)

Also, one big issue is that this assumes you *are* using ErrorProne. Many projects that depend on Airlift don't, and it's unreasonable to force it on them. Without checks that prevent using the method with anything other than a literal string, this new reimagining of the method is more dangerous, as I described above.

ksobolew:
You're right, you have a responsibility for not breaking projects using Airlift. This means it's not possible to merge it in the current form, and it's fine. Backward compatibility is difficult :) I still think that detecting more issues is a worthy goal, so I'm thinking of a way to enable it. My first thought was to add methods like `debugFormatted`, but that's just silly. (Why would anyone use it? And more, why would anyone switch existing `debug` calls to it?) Or I can add a custom Error Prone check (like [the one I'm experimenting with in Trino](trinodb/trino#10473)) specifically for this, which users can opt in to if they want.

PR #958, opened 2021-11-08
Add system property override for hidden config

electrum:


configuration/src/main/java/io/airlift/configuration/ConfigurationInspector.java
```diff
@@ -109,7 +109,7 @@ private ConfigRecord(ConfigurationFactory configurationFactory, ConfigurationPro
                     description = "";
                 }
 
-                if (attribute.isHidden()) {
+                if (attribute.isHidden() && !Boolean.getBoolean("config.includeHidden")) {
```
dain, line 112: This requires `-Dconfig.includeHidden=true`

dain (approved):

configuration/src/main/java/io/airlift/configuration/ConfigurationInspector.java
electrum: I think that is fine for this.

PR #962, opened 2021-11-24
Add annotation support to logging.

mattstep:


log-manager/src/main/java/io/airlift/log/StaticFormatter.java
```diff
@@ -60,8 +60,15 @@ public StaticFormatter()
...
         return new StringWriter()
-                .append("System Context Summary")
-                .append("\t")
+                .append(TIMESTAMP_FORMATTER.format(ZonedDateTime.now()))
+                .append('\t')
+                .append(Level.INFO.name())
+                .append('\t')
+                .append(Thread.currentThread().getName())
+                .append('\t')
+                .append(StaticFormatter.class.getCanonicalName())
```
dain, line 69: You could just use "SystemContext" for the logger name, and then you wouldn't have to repeat it in the log messsage.  This is just an idea... not sure if it is better or worse.

log-manager/src/main/java/io/airlift/log/LoggingConfiguration.java
```diff
@@ -155,4 +156,16 @@ public LoggingConfiguration setLegacyLoggerImplementationEnabled(boolean enabled
         this.legacyLoggerImplementationEnabled = enabled;
         return this;
     }
+
+    public String getSystemContextFile()
```
martint, line 160: What's a "system context"?

log-manager/src/main/java/io/airlift/log/StaticFormatter.java
mattstep: That makes sense

log-manager/src/main/java/io/airlift/log/LoggingConfiguration.java
mattstep: It's a file that we can pass values in from kubernetes, either through the helm chart or environment variables in the container, that we add onto our logs so we can eliminate the need for fluentbit or fluentd to instrument the logs with data we know when we start the application. In particular, things like service name, pod name, ip address, etc. are valuable for debugging.

docs/add_logging.md
```diff
@@ -38,3 +38,45 @@ public class MyClass
...
+
+* `log.output-file` can be set to the desired output for logging and also provides a mechanism
+  to stream logs over tcp. Setting this to the format `tcp://<host>:<port>` will enable the socket
+  logger for tcp streaming.
+
+* `log.format` can be set to either `text` or `json`. When set to `json` all fields are encapsulated
+  in potentially multi-line `json` fields. This allows for capturing and indexing exceptions as singular fields in
+  a logging search system.
+
+* `log.system-context-file` allows a file containing fields to be set at startup that can be
```
martint, line 55: log.annotation-file

docs/add_logging.md
```diff
@@ -38,3 +38,45 @@ public class MyClass
...
+  interpolated from environment variables and will be included in all log output (per-line for
+  the `json` formatter and per-file for the `text` formatter).
```
martint, line 57: both "per line"

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
```diff
@@ -3,20 +3,27 @@
...
 
 import java.io.IOException;
 import java.io.StringWriter;
-import java.util.logging.Formatter;
+import java.util.Map;
 import java.util.logging.LogRecord;
 
 public class JsonFormatter
-        extends Formatter
+        extends java.util.logging.Formatter
```
martint, line 17: Why fully qualified?

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
```diff
@@ -3,20 +3,27 @@
...
 {
     private static final JsonCodec<JsonRecord> CODEC = new JsonCodecFactory(new ObjectMapperProvider()).jsonCodec(JsonRecord.class);
     private static final JsonFactory jsonFactory = new JsonFactory();
+    private final Map<String, String> logAnnotations;
+
+    public JsonFormatter(Map<String, String> logAnnotations)
+    {
+        this.logAnnotations = logAnnotations == null ? null : ImmutableMap.copyOf(logAnnotations);
```
martint, line 25: Instead of using nulls, why not just use empty maps everywhere. It will make it easier and less error prone.

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
```diff
@@ -72,4 +81,9 @@ String minimalJsonErrorLogLine(JsonRecord jsonRecord, Exception exception)
         }
         return stringWriter.append("\n").toString();
     }
+
+    private static String toString(JsonRecord jsonRecord)
+    {
+        return new StringWriter().append(CODEC.toJson(jsonRecord)).append("\n").toString();
```
martint, line 87: Format as:

```
return new StringWriter()
    .append(CODEC.toJson(jsonRecord))
    .append("\n")
    .toString();
```

log-manager/src/main/java/io/airlift/log/StaticFormatter.java
```diff
@@ -21,9 +25,10 @@
 import static java.time.temporal.ChronoField.YEAR;
 
 class StaticFormatter
-        extends Formatter
+        extends java.util.logging.Formatter
```
martint, line 28: Why?

log-manager/src/main/java/io/airlift/log/StaticFormatter.java
```diff
@@ -66,6 +81,13 @@ public String format(LogRecord record)
                 .append('\t')
                 .append(record.getMessage());
 
+        if (hasLogAnnotations()) {
+            stringWriter.append('\t')
```
martint, line 85: The annotations should probably go before the message, which may contain new lines in some cases.

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
mattstep: oops, I think because the order I refactored things in, didn't notice it.

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
mattstep: I'd need to special case the JsonRecord getter to return null then or else all the json log lines would contain an empty map.

martint: Yeah, but it's better to special-case for JsonRecord when it gets constructed than special case everywhere

log-manager/src/main/java/io/airlift/log/StaticFormatter.java
mattstep: My only hesitation with that is if a parser is splitting on tabs, it would mean the ordering of fields would change if the annotations are present or not.

docs/add_logging.md
```diff
@@ -38,3 +38,44 @@ public class MyClass
...
+
+  For example:
+  ```
+  #config.properties
+  log.annotation-file=annotations.properties
+  log.format=json
+  ```
+  ```
+  #annotations.properties
+  hostIp: ${ENV:HOST_IP}
```
electrum, line 66: I think the `=` syntax is better here, since it's not obvious why these two properties files use a different syntax, and we always use the `=` syntax so most people wouldn't even know this is valid.

docs/add_logging.md
```diff
@@ -38,3 +38,44 @@ public class MyClass
...
+
+* `log.output-file` can be set to the desired output for logging and also provides a mechanism
+  to stream logs over tcp. Setting this to the format `tcp://<host>:<port>` will enable the socket
+  logger for tcp streaming.
+
+* `log.format` can be set to either `text` or `json`. When set to `json` all fields are encapsulated
+  in potentially multi-line `json` fields. This allows for capturing and indexing exceptions as singular fields in
+  a logging search system.
+
+* `log.annontation-file` allows a file containing fields to be set at startup that can be
```
electrum, line 55: Typo `annontation`

docs/add_logging.md
```diff
@@ -38,3 +38,44 @@ public class MyClass
...
+
+The logging system has several configuration options that are provided to make it easier
+to use in cloud native environments.
+
+* `log.output-file` can be set to the desired output for logging and also provides a mechanism
+  to stream logs over tcp. Setting this to the format `tcp://<host>:<port>` will enable the socket
+  logger for tcp streaming.
+
+* `log.format` can be set to either `text` or `json`. When set to `json` all fields are encapsulated
+  in potentially multi-line `json` fields. This allows for capturing and indexing exceptions as singular fields in
```
electrum, line 52: Is the multi-line part correct? We don't enable indentation, so the JSON itself shouldn't have newlines, and the string values should have newlines escaped. Maybe rephrase as

> When set to `json`, the log record is formatted as a JSON object, one per line. Any newlines in the field values, such as exception stack traces, will be escaped as normal in JSON object. This allows ...

electrum: Nit: write the second `json` instance as JSON without code quotes.

log-manager/src/main/java/io/airlift/log/JsonFormatter.java
```diff
@@ -72,4 +84,12 @@ String minimalJsonErrorLogLine(JsonRecord jsonRecord, Exception exception)
         }
         return stringWriter.append("\n").toString();
     }
+
+    private static String toString(JsonRecord jsonRecord)
+    {
+        return new StringWriter()
```
electrum, line 90: No need for `StringWriter`
```java
return CODEC.toJson(jsonRecord) + "\n";

File: log-manager/src/main/java/io/airlift/log/JsonRecord.java [hunk 14]
```diff
@@ -26,14 +28,16 @@ public JsonRecord(
...
+            Throwable throwable,
+            Map<String, String> logAnnotations)
     {
         this.timestamp = requireNonNull(timestamp);
         this.level = requireNonNull(level);
         this.thread = thread;
         this.loggerName = loggerName;
         this.message = message;
         this.throwable = throwable;
+        this.logAnnotations = logAnnotations == null || logAnnotations.isEmpty() ? null : logAnnotations;
```
electrum, line 40: Nit: add parens around the boolean part

log-manager/src/main/java/io/airlift/log/LoggingConfiguration.java
```diff
@@ -155,4 +156,16 @@ public LoggingConfiguration setLegacyLoggerImplementationEnabled(boolean enabled
...
         return this;
     }
+
+    public String getLogAnnotationFile()
+    {
+        return logAnnotationFile;
+    }
+
+    @Config("log.annotation-file")
+    public LoggingConfiguration setLogAnnotationFile(String logAnnotationFile)
```
electrum, line 166: We could annotate with `@FileExists`. The config system will validate it exists if not null.

log-manager/src/main/java/io/airlift/log/StaticFormatter.java
```diff
@@ -43,6 +50,16 @@
             .appendOffset("+HHMM", "Z")
             .toFormatter(Locale.US);
 
+    public StaticFormatter(Map<String, String> logAnnotations)
+    {
+        this.logAnnotations = ImmutableMap.copyOf(requireNonNull(logAnnotations, "logAnnotations is null"));
+    }
+
+    public StaticFormatter()
```
electrum, line 58: Move this above and chain to other constructor
```java
this(ImmutableMap.of());
```
It's preferable to have everything go through a single constructor if possible

log-manager/src/main/java/io/airlift/log/StaticFormatter.java
```diff
@@ -62,8 +79,16 @@ public String format(LogRecord record)
                 .append('\t')
                 .append(Thread.currentThread().getName())
                 .append('\t')
-                .append(record.getLoggerName())
-                .append('\t')
+                .append(record.getLoggerName());
+
+        if (!logAnnotations.isEmpty()) {
+            stringWriter.append('\t')
```
electrum, line 85: This can be
```java
stringWriter.append('\t');
Joiner.on(",").withKeyValueSeparator("=").appendTo(stringWriter, logAnnotations);
```

log-manager/src/test/java/io/airlift/log/TestJsonFormatter.java
```diff
@@ -66,4 +68,19 @@ public void testLogFormatting()
         assertEquals(jsonMap.get("throwableMessage"), testException.getMessage());
         assertEquals(jsonMap.get("stackTrace"), getStackTraceAsString(testException));
     }
+
+    @Test
+    public void testLogAnnotations()
+    {
+        LogRecord record = new LogRecord(Level.DEBUG.toJulLevel(), "Test Log Message");
+        ImmutableMap<String, String> logAnnotations = ImmutableMap.of("foo", "apple", "bar", "banana");
```
electrum, line 76: Declare as `Map`

electrum (approved):
Nit: don't end commit title in period

log-manager/src/main/java/io/airlift/log/StaticFormatter.java
mattstep: .appendTo throws IOException, so I just did .join and appended the output

mattstep:
> Nit: don't end commit title in period

fixed

PR #964, opened 2021-12-01
Preserve stdErr and pass it into the ErrorManager in BufferedHandler

mattstep:
… errors

### DIFF
--- a/log-manager/src/main/java/io/airlift/log/BufferedHandler.java
+++ b/log-manager/src/main/java/io/airlift/log/BufferedHandler.java
@@ -10,6 +10,7 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.logging.ErrorManager;
 import java.util.logging.Formatter;
 import java.util.logging.Handler;
 import java.util.logging.LogRecord;
@@ -33,9 +34,10 @@ class BufferedHandler
     private final AtomicBoolean closed = new AtomicBoolean();
     private final AtomicLong droppedMessages = new AtomicLong(0);
 
-    public BufferedHandler(MessageOutput messageOutput, Formatter formatter)
+    public BufferedHandler(MessageOutput messageOutput, Formatter formatter, ErrorManager errorManager)
     {
         this.messageOutput = requireNonNull(messageOutput, "messageOutput is null");
+        setErrorManager(requireNonNull(errorManager, "errorManager is null"));
         setFormatter(requireNonNull(formatter, "formatter is null"));
 
         thread = new Thread(this::logging);
--- a/log-manager/src/main/java/io/airlift/log/Logging.java
+++ b/log-manager/src/main/java/io/airlift/log/Logging.java
@@ -53,6 +53,7 @@ public class Logging
     private static final String ROOT_LOGGER_NAME = "";
     private static final java.util.logging.Logger ROOT = java.util.logging.Logger.getLogger("");
     private static Logging instance;
+    private static final PrintStream stdErr = System.err;
 
     // hard reference to loggers for which we set the level
     @GuardedBy("this")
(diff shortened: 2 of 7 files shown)


log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -24,25 +26,53 @@
...
+        private boolean reported;
+
+        public synchronized void error(String msg, Exception ex, int code)
+        {
+            if (reported) {
+                // We only report the first error, to avoid clogging
+                // the screen.
+                return;
+            }
+            reported = true;
```
erichwang, line 64: use

private final AtomicBoolean reported = new AtomicBoolean();

if (!reported.compareAndSet(false, true)) {
  return;
}

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -53,6 +53,7 @@
     private static final String ROOT_LOGGER_NAME = "";
     private static final java.util.logging.Logger ROOT = java.util.logging.Logger.getLogger("");
     private static Logging instance;
+    private static PrintStream stdErr;
```
erichwang, line 56: just make this a private static final constant

erichwang (approved):
just a couple comments, but seems reasonable

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -24,25 +26,50 @@
 class BufferedHandler
         extends Handler
 {
-    private static final int MAX_BATCH_COUNT = 1024;
+    private static final int MAX_BATCH_COUNT = 10240;
```
martint, line 29: Why did this change?

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -24,25 +26,50 @@
...
     }
 
+    private class BufferedHandlerErrorManager
+            extends ErrorManager
+    {
+        private final AtomicBoolean reported = new AtomicBoolean();
+
+        public synchronized void error(String msg, Exception ex, int code)
+        {
+            if (!reported.compareAndSet(false, true)) {
```
martint, line 59: No need for an AtomicBoolean. This runs in a syncrhonized block, and no one other than this method accesses `reported`.

martint: Also, I'm not sure I understand the logic here. Once this method gets called, it will forever after short-circuit and exit early. What's the purpose of this `BufferedHandlerErrorManager`?

electrum:
Nit: commit title is too long

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -24,25 +26,51 @@
...
+    {
+        private boolean reported;
+
+        public synchronized void error(String msg, Exception ex, int code)
+        {
+            if (!reported) {
+                return;
+            }
+            reported = true;
+            String text = "java.util.logging.ErrorManager: " + code;
```
electrum, line 63: I would do
```java
String text = ErrorManager.class.getName() + ": " + code;

File: log-manager/src/main/java/io/airlift/log/RollingFileMessageOutput.java [hunk 6]
```diff
@@ -101,10 +102,11 @@ public static BufferedHandler createRollingFileHandler(
             DataSize maxFileSize,
             DataSize maxTotalSize,
             CompressionType compressionType,
-            Formatter formatter)
+            Formatter formatter,
+            PrintStream stdErr)
```
dain, line 106: I'd pass in ErrorManager here instead of the PrintStream

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -24,25 +26,51 @@
...
+            if (msg != null) {
+                text = text + ": " + msg;
+            }
+            stdErr.println(text);
+            if (ex != null) {
+                ex.printStackTrace(stdErr);
```
electrum, line 69: Using `Throwables.getStackTraceAsString()` would be better, so that stderr is written to only once, potentially behaving better if there are other concurrent writes to stderr.

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -24,25 +26,51 @@
...
+
         thread = new Thread(this::logging);
         thread.setName("log-writer");
         thread.setDaemon(true);
     }
 
+    private class BufferedHandlerErrorManager
+            extends ErrorManager
+    {
+        private boolean reported;
```
dain, line 55: `@GuardedBy("this")`

dain (approved):

Issue #968, opened 2021-12-27
Allow matching environment variable names with dashes for secret replacements


oyilmaz:
In our Trino setup, we would like to use oauth2 configuration that reads oauth2 client id and client secret from default k8s platform injected variable names that we do not control, i.e. `trino-client-id`, `trino-client-secret`. 

https://github.com/airlift/airlift/blob/master/configuration/src/main/java/io/airlift/configuration/ConfigurationUtils.java#L16 pattern only considers replacing environment variables with underscores.

PR #969, opened 2021-12-27
Replace environment variables that contain dashes

oyilmaz:
closes https://github.com/airlift/airlift/issues/968


electrum:
Thanks!

PR #971, opened 2022-01-01
Fix log file rename on startup

dain:
Log file was not being detected as a symlink as was being renamed


log-manager/src/main/java/io/airlift/log/RollingFileMessageOutput.java
```diff
@@ -104,25 +109,33 @@ public static BufferedHandler createRollingFileHandler(
...
     {
-        RollingFileMessageOutput output = new RollingFileMessageOutput(filename, maxFileSize, maxTotalSize, compressionType, formatter);
+        RollingFileMessageOutput output = new RollingFileMessageOutput(filename, maxFileSize, maxTotalSize, compressionType, formatter, fileTimeRoller);
         BufferedHandler handler = new BufferedHandler(output, formatter, errorManager);
         handler.start();
         return handler;
     }
 
-    private RollingFileMessageOutput(String filename, DataSize maxFileSize, DataSize maxTotalSize, CompressionType compressionType, Formatter formatter)
+    private RollingFileMessageOutput(String filename,
```
martint, line 121: Argument goes on next line

martint (commented):
First commit looks good. Still reviewing the other one.

log-manager/src/main/java/io/airlift/log/RollingFileMessageOutput.java
```diff
@@ -444,4 +459,44 @@ else if (fileName.startsWith(DELETED_PREFIX)) {
...
+        @VisibleForTesting
+        synchronized long getNextRollNanos()
+        {
+            return nextRollNanos;
+        }
+
+        public synchronized void updateNextRollTime(LocalDateTime logFileCreateTime)
+        {
+            long epochMilli = logFileCreateTime.toLocalDate().plusDays(1).atStartOfDay(ZoneId.systemDefault()).toInstant().toEpochMilli();
+            nextRollNanos = ticker.read() + TimeUnit.MILLISECONDS.toNanos(epochMilli - systemTimeMillis.getAsLong());
```
martint, line 499: This can overflow, and subsequent comparison with `nextRollNanos` will be broken. Specifically, from the System.nanoTime() javadocs:


> To compare elapsed time against a timeout, use
>  `if (System.nanoTime() - startTime >= timeoutNanos) ...`
> instead of
>  `if (System.nanoTime() >= startTime + timeoutNanos) ...`
> because of the possibility of numerical overflow.
```

martint (changes requested):

martint (approved):

Issue #976, opened 2022-01-14
NPE when the keystore password is `NULL` in ReloadableSslContextFactoryProvider


findinpath:
While trying to start Trino with the following `config.properties`

```
#single node install config
coordinator=true
node-scheduler.include-coordinator=true
discovery.uri=http://localhost:8080

http-server.https.enabled=true
http-server.https.port=8443
http-server.https.keystore.path=/tmp/presto-master.jks
#http-server.https.keystore.key=123456
http-server.http.port=8080
http-server.authentication.allow-insecure-over-http=true
```

the following exception occurs:

```
trino-coordinator_1  | 1) [Guice/ErrorInCustomProvider]: NullPointerException
trino-coordinator_1  |   while locating HttpServerProvider
trino-coordinator_1  |   at GeneratedMethodAccessor26.invoke(Unknown Source)
trino-coordinator_1  |   while locating HttpServer
trino-coordinator_1  | 
trino-coordinator_1  | Learn more:
trino-coordinator_1  |   https://github.com/google/guice/wiki/ERROR_IN_CUSTOM_PROVIDER
trino-coordinator_1  | Caused by: NullPointerException
trino-coordinator_1  | 	at ReloadableSslContextFactoryProvider.loadKeyStore(ReloadableSslContextFactoryProvider.java:212)
trino-coordinator_1  | 	at ReloadableSslContextFactoryProvider.loadContextFactory(ReloadableSslContextFactoryProvider.java:107)
trino-coordinator_1  | 	at ReloadableSslContextFactoryProvider.<init>(ReloadableSslContextFactoryProvider.java:99)
trino-coordinator_1  | 	at HttpServer.createReloadingSslContextFactory(HttpServer.java:534)
trino-coordinator_1  | 	at HttpServer.lambda$new$0(HttpServer.java:231)
trino-coordinator_1  | 	at java.base/Optional.orElseGet(Optional.java:369)
trino-coordinator_1  | 	at HttpServer.<init>(HttpServer.java:231)
trino-coordinator_1  | 	at HttpServerProvider.get(HttpServerProvider.java:149)
trino-coordinator_1  | 	at HttpServerProvider.get(HttpServerProvider.java:46)
trino-coordinator_1  | 	at ProviderInternalFactory.provision(ProviderInternalFactory.java:86)
trino-coordinator_1  | 	at BoundProviderFactory.provision(BoundProviderFactory.java:72)
[...]

PR #977, opened 2022-01-14
Avoid NPE for unspecified keystore key

findinpath:
Fixes https://github.com/airlift/airlift/issues/976

### DIFF
--- a/http-server/src/main/java/io/airlift/http/server/ReloadableSslContextFactoryProvider.java
+++ b/http-server/src/main/java/io/airlift/http/server/ReloadableSslContextFactoryProvider.java
@@ -209,7 +209,7 @@ final class ReloadableSslContextFactoryProvider
 
         try (InputStream in = new FileInputStream(file)) {
             KeyStore keyStore = KeyStore.getInstance("JKS");
-            keyStore.load(in, keystorePassword.toCharArray());
+            keyStore.load(in, Optional.ofNullable(keystorePassword).map(String::toCharArray).orElse(null));
             return keyStore;
         }
         catch (IOException | GeneralSecurityException e) {


findinpath:
After testing this change with Trino, the error occuring is way more misleading.
```
Caused by: RuntimeException: UnrecoverableKeyException: Get Key failed: Given final block not properly padded. Such issues can arise if a bad key is used during decryption.
	at HttpServerProvider.get(HttpServerProvider.java:173)
	at HttpServerProvider.get(HttpServerProvider.java:46)
	at ProviderInternalFactory.provision(ProviderInternalFactory.java:86)
	at BoundProviderFactory.provision(BoundProviderFactory.java:72)
	at ProviderInternalFactory$1.call(ProviderInternalFactory.java:67)
```


Loading a keystore with a `NULL` password actually works.

```
        String keystorePassword = null;
        try (InputStream in = new FileInputStream(new File("presto-master.jks"))) {
            KeyStore keystore = KeyStore.getInstance("JKS");
            keystore.load(in, Optional.ofNullable(keystorePassword).map(String::toCharArray).orElse(null));
        }
```

Detailed reasoning behind this choice:

https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/sun/security/pkcs12/PKCS12KeyStore.java#L2044-L2051


https://github.com/openjdk/jdk17/blob/master/src/java.base/share/classes/sun/security/pkcs12/PKCS12KeyStore.java#L2065-L2087


NOTE that in case of providing an invalid password we'll receive here
```
Exception in thread "main" java.io.IOException: keystore password was incorrect
	at java.base/sun.security.pkcs12.PKCS12KeyStore.engineLoad(PKCS12KeyStore.java:2159)
```


A possibility to go around this is to have a special handling in case of  `NULL` keystore password by going over the aliases and check whether the certificates can be retrieved


```
            Enumeration<String> enumeration = keystore.aliases();
            while(enumeration.hasMoreElements()) {
                String alias = enumeration.nextElement();
                Certificate certificate = keystore.getCertificate(alias);
                 if (certificate == null){
                       // we're dealing with an invalid password
                 }
            }
```

electrum:
Your suggested code sounds better. In the case where we need a password but one isn't provided, it seems better to fail up front with a good error message.

findinpath:
I'll refresh the PR and come up with a better version of this PR.

PR #980, opened 2022-01-20
More rolling log file manager improvements

dain:


log-manager/src/main/java/io/airlift/log/RollingFileMessageOutput.java
```diff
@@ -241,7 +241,7 @@ public synchronized void close()
     public synchronized void writeMessage(byte[] message)
             throws IOException
     {
-        if (currentFileSize + message.length > maxFileSize) {
+        if (currentFileSize > 0 && currentFileSize + message.length > maxFileSize) {
```
electrum, line 244: Maybe add parentheses here

electrum (approved):

PR #981, opened 2022-01-21
Add "bootstrap.quiet" configuration property

electrum:


bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -233,8 +235,9 @@ public Injector initialize()
             throw new ApplicationConfigurationException(errors, warnings);
         }
 
-        // Log effective configuration
-        if (!quiet) {
+        // Log effective configuration. The config property can force either true/false,
+        // irrespective of if quiet() was set for this Bootstrap instance.
+        if ((!quiet && !Boolean.TRUE.equals(quietConfig)) || Boolean.FALSE.equals(quietConfig)) {
```
dain, line 240: I think this would be simpler `if (!(quietConfig == null ? quiet : quietConfig)) {`

dain (approved):

CHANGES
```diff
@@ -1,6 +1,7 @@
 214
 - Fix log file rename in launcher script.
 - Remove Executed class from bootstrap module.
+- Add "bootstrap.quiet" configuration property.
```
findepi, line 4: What does it do?

Also, will it help with @nineinchnick's https://github.com/trinodb/trino/pull/10689?

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -234,7 +236,7 @@ public Injector initialize()
         }
 
         // Log effective configuration
-        if (!quiet) {
+        if (!((quietConfig == null) ? quiet : quietConfig)) {
```
findepi, line 239: `Bootstrap.quiet()` (`quiet`) is local to invocation, so should take precedence over config file.

nineinchnick: Can we log a single message saying that this is effective? It'll help to figure out why the configuration is not being logged.

CHANGES
electrum: It would help with that. The intent is to set it in `DistributedQueryRunner` without needing to touch any connectors.

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
electrum: No, the `quiet()` call is intended to act as a default for the configuration, so the configuration can override in either direction.

electrum: @nineinchnick The intent of this flag is to not log anything, so logging a message saying we're not logging anything would be counterproductive.

PR #990, opened 2022-05-16
Ignore trailing whitespace in log levels configuration

findepi:


log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -159,7 +159,7 @@ public void setLevels(File file)
         }
 
         fromProperties(properties).forEach((loggerName, value) ->
-                setLevel(loggerName, Level.valueOf(value.toUpperCase(Locale.US))));
+                setLevel(loggerName, Level.valueOf(value.toUpperCase(Locale.US).trim())));
```
electrum, line 162: Why not switch this method to use `loadPropertiesFrom` which you already fixed?

electrum (approved):

log-manager/src/main/java/io/airlift/log/Logging.java
findepi: trim is used in loadPropertiesFrom.

i think the whitespace definition here is fine (doesn't matter), but you may want to change `loadPropertiesFrom` to use strip instead

findepi: > Why not switch this method to use `loadPropertiesFrom` which you already fixed?

Good point, done

PR #991, opened 2022-06-02
Add AsyncSemaphore#processAll utility method

arhimondr:
Allows limiting the number of async tasks running concurrently.
Useful for limiting the number of async requests sent concurrently to a remote
server.

### DIFF
--- a/concurrent/src/main/java/io/airlift/concurrent/AsyncSemaphore.java
+++ b/concurrent/src/main/java/io/airlift/concurrent/AsyncSemaphore.java
@@ -13,13 +13,12 @@
  */
 package io.airlift.concurrent;
 
-import com.google.common.util.concurrent.FutureCallback;
-import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
 
 import javax.annotation.concurrent.ThreadSafe;
 
+import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
@@ -27,7 +26,12 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Verify.verify;
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.util.concurrent.Futures.immediateCancelledFuture;
+import static com.google.common.util.concurrent.Futures.immediateFailedFuture;
 import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
+import static io.airlift.concurrent.MoreFutures.allAsListWithCancellationOnFailure;
 import static java.util.Objects.requireNonNull;
 
 /**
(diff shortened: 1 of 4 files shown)


linzebing (approved):

erichwang:
@arhimondr and @linzebing, take a look at AsyncSemaphore in airlift, this behavior is very similar I think.

If you want to do bounded execution as an executor, check out BoundedExecutor

arhimondr:
@erichwang We don't need to submit to an executor, we are using a client that provides a native non blocking asynchronous API

electrum:
It seems like `AsyncSemaphore` does exactly what you need here.

arhimondr:
@electrum We are currently discussing this with @erichwang . It looks like `AsyncSemaphore` is designed to limit the number of independent concurrent tasks running in parallel. We need a primitive that runs a number of related tasks as an atomic unit with a concurrency limit. In theory it should be possible to implement what we need via `AsyncSemaphore` internally, however It's not clear whether it will reduce complexity significantly as the most of the complexity comes from making sure all the tasks running as a single unit (cancellation handling, cancellation on failure, etc.)

concurrent/src/main/java/io/airlift/concurrent/AsyncSemaphore.java
```diff
@@ -37,26 +42,70 @@
...
+     * If any of the submitted tasks fails or is canceled, the remaining tasks are cancelled.
+     */
+    public static <T, R> ListenableFuture<List<R>> processAll(List<T> tasks, Function<T, ListenableFuture<R>> submitter, int maxPermits, Executor executor)
+    {
+        AtomicReference<ListenableFuture<List<R>>> resultFutureReference = new AtomicReference<>();
+        AsyncSemaphore<T, R> semaphore = new AsyncSemaphore<>(maxPermits, executor, task -> {
+            ListenableFuture<List<R>> resultFuture = resultFutureReference.get();
+            if (resultFuture != null && resultFuture.isDone()) {
+                return immediateCancelledFuture();
+            }
```
erichwang, line 86: Since you are already doing `if (!queuedTask.getCompletionFuture().isDone()) {` in `runNext()`, you actually don't need to do this here with resultFutureReference. If allAsList is explicitly cancelled, or if any of the sub futures fail, everything that hasn't run yet will be cancelled (and will thus be done). There will be no additional value to return a `immediateCancelledFuture()` here.

concurrent/src/main/java/io/airlift/concurrent/AsyncSemaphore.java
```diff
@@ -37,26 +42,70 @@
...
+     * For example:
+     * <pre>{@code
+     * List<Integer> userIds = Lists.of(1, 2, 3);
+     * ListenableFuture<List<UserInfo>> future = processAll(ids, client::getUserInfoById, 2, executor);
+     * List<UserInfo> userInfos = future.get(...);
+     * }</pre>
+     *
+     * @param tasks tasks to process
+     * @param submitter task submitter
+     * @param maxPermits maximum number of tasks permitted to run in parallel
```
erichwang, line 71: maxPermits => maxConcurrency (makes more sense when you are looking at processing in general, and not looking at an explicit semaphore)

concurrent/src/main/java/io/airlift/concurrent/AsyncSemaphore.java
```diff
@@ -37,26 +42,70 @@
...
+     * @param executor task submission executor
+     * @return {@link ListenableFuture} containing a list of values returned by the {@code tasks}.
+     * The order of elements in the list matches the order of {@code tasks}.
+     * If the result future is cancelled all the remaining tasks are cancelled (submitted tasks will be cancelled, pending tasks will not be submitted).
+     * If any of the submitted tasks fails or is canceled, the result future is too.
+     * If any of the submitted tasks fails or is canceled, the remaining tasks are cancelled.
```
erichwang, line 77: Use a consistent spelling for `cancelled`

erichwang, line 77: "submitted tasks fails or is canceled" => "submitted tasks fail or are cancelled"

erichwang (commented):
I like this a lot more. Mostly minor style things, but note some of the comments on doing extra logic I don't think we need.

concurrent/src/main/java/io/airlift/concurrent/AsyncSemaphore.java
arhimondr: This is to cover a corner case I discovered when testing. When the executor provided is a `directExecutor()` cancelling the result future will attempt to synchronously cancel underlying futures triggering the `releasePermit` callback on every future resulting in unnecessary submissions.

erichwang: @arhimondr, ok yea, I don't think there is a good way around this other than what you suggested. But I would suggest this minor refactoring of this function:

```java
    public static <T, R> ListenableFuture<List<R>> processAll(List<T> tasks, Function<T, ListenableFuture<R>> submitter, int maxConcurrency, Executor submitExecutor)
    {
        SettableFuture<List<R>> resultFuture = SettableFuture.create();
        AsyncSemaphore<T, R> semaphore = new AsyncSemaphore<>(maxConcurrency, submitExecutor, task -> {
            if (resultFuture.isCancelled()) {
                // Task cancellation tends to happen in task submission order, which can race with subsequent task submissions after previous cancellations.
                // This check prevents this race ever from occurring.
                return immediateCancelledFuture();
            }
            return submitter.apply(task);
        });
        resultFuture.setFuture(allAsListWithFailureCancellation(tasks.stream()
                .map(semaphore::submit)
                .collect(toImmutableList())));
        return resultFuture;
    }

    /**
     * Same as {@link com.google.common.util.concurrent.Futures#allAsList(ListenableFuture[])}, but additionally cancels any remaining component Futures
     * if the result has already failed.
     *
     * @param futures futures to combine
     * @return a future that provides a list of the results of the component futures
     */
    private static <V> ListenableFuture<List<V>> allAsListWithFailureCancellation(Iterable<? extends ListenableFuture<? extends V>> futures)
    {
        List<ListenableFuture<? extends V>> futuresSnapshot = ImmutableList.copyOf(futures);
        ListenableFuture<List<V>> listFuture = allAsList(futuresSnapshot);
        addExceptionCallback(listFuture, () -> futuresSnapshot.forEach(future -> future.cancel(true)));
        return listFuture;
    }
```

I would also greatly appreciate if you put allAsListWithFailureCancellation into MoreFutures, since this is a behavior that I would very much like to reuse in the future.

erichwang (approved):

PR #992, opened 2022-06-07
Regenerate testing certificates using secure crypto algorithms

arhimondr:
SHA1withRSA is deprecated on certain operation systems


erichwang:
@dain, might want to take a look at this. @arhimondr was hitting some issues where it seems like we are using deprecated crypto algos for testing.

dain (approved):

PR #993, opened 2022-06-10
Auto detect records for Jackson serialization

Randgalt:
Currently, record serialization requires annotating records with
`@JsonAutoDetect(getterVisibility = PUBLIC_ONLY)`. This is a nuisance
and can be hard to debug when it's forgotten (Jackson serializes
to an empty object when it's missing).

Add a Jackson module that auto-detects Records and sets
the visibility to match the annotation. Note: Airlift is
currently at Java 11 so record detection is not ideal (though it
mimics what the JDK does). Once Airlift is at Java 17 it can
be updated and a test can be added.

Note: I tested this in Java 17 project independently to verify that it works.

### DIFF
--- a/json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
+++ b/json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
@@ -1,0 +1,81 @@
+package io.airlift.json;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.core.Version;
+import com.fasterxml.jackson.databind.AnnotationIntrospector;
+import com.fasterxml.jackson.databind.introspect.AnnotatedClass;
+import com.fasterxml.jackson.databind.introspect.VisibilityChecker;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.util.Optional;
+
+import static java.lang.invoke.MethodType.methodType;
+
+public class RecordAutoDetectModule
+        extends SimpleModule
+{
+    private static final Optional<MethodHandle> IS_RECORD;
+
+    static {
+        MethodHandle methodHandle;
+        try {
+            //noinspection JavaLangInvokeHandleSignature
+            methodHandle = MethodHandles.lookup()
+                    .findVirtual(Class.class, "isRecord", methodType(boolean.class));
+        }
+        catch (Exception e) {
+            // ignore - assume we're not on Java 14+
+            methodHandle = null;
+        }
+        IS_RECORD = Optional.ofNullable(methodHandle);
+    }
+
+    @Override
+    public void setupModule(SetupContext context)
+    {
+        super.setupModule(context);
+        context.insertAnnotationIntrospector(new Introspector());
+    }
+
+    private static class Introspector
+            extends AnnotationIntrospector
+    {
+        private static final VisibilityChecker.Std RECORD_VISIBILITY_CHECKER = VisibilityChecker.Std.defaultInstance()
+                .withGetterVisibility(JsonAutoDetect.Visibility.PUBLIC_ONLY)
+                .withCreatorVisibility(JsonAutoDetect.Visibility.DEFAULT)
+                .withFieldVisibility(JsonAutoDetect.Visibility.DEFAULT)
+                .withIsGetterVisibility(JsonAutoDetect.Visibility.DEFAULT);
+
+        @Override
+        public VisibilityChecker<?> findAutoDetectVisibility(AnnotatedClass ac, VisibilityChecker<?> checker)
+        {
+            if (IS_RECORD.map(methodHandle -> isRecord(ac, methodHandle)).orElse(false)) {
+                JsonAutoDetect overrideAnnotation = ac.getRawType().getAnnotation(JsonAutoDetect.class);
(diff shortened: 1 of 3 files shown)


Randgalt:
cc @electrum, @dain and/or @martint

findepi:
> Note: I tested this in Java 17 project independently to verify that it works.

Would it make sense to have a Java 17-based test module to verify Java 17-related Airlift behavior?

Randgalt:
> > Note: I tested this in Java 17 project independently to verify that it works.
> 
> Would it make sense to have a Java 17-based test module to verify Java 17-related Airlift behavior?

Yes - I think it does. However, I would do it as a separate PR. It will make the CI more complicated.

json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
```diff
@@ -0,0 +1,63 @@
...
+        {
+            return Version.unknownVersion();
+        }
+    }
+
+    // TODO replace with clazz.isRecord() once Airlift is at Java 17+
+    private static boolean isRecord(Class<?> clazz)
+    {
+        Class<?> superclass = (clazz != null) ? clazz.getSuperclass() : null;
+        return (superclass != null) && superclass.getName().equals("java.lang.Record") && ((clazz.getModifiers() & Modifier.FINAL) != 0);
```
electrum, line 61: I think `Modifier.isFinal(clazz.getModifiers())` is the preferred way

json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
```diff
@@ -0,0 +1,63 @@
...
+
+        @Override
+        public Version version()
+        {
+            return Version.unknownVersion();
+        }
+    }
+
+    // TODO replace with clazz.isRecord() once Airlift is at Java 17+
+    private static boolean isRecord(Class<?> clazz)
```
electrum, line 58: Maybe we should do this by calling `Class.isRecord()` reflectively if it's available?

electrum (approved):

json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
Randgalt: @electrum I just pushed a change that uses a MethodHandle instead. wdyt?

electrum (approved):

json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
```diff
@@ -0,0 +1,87 @@
...
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.util.Optional;
+
+import static java.lang.invoke.MethodType.methodType;
+
+public class RecordAutoDetectModule
+        extends SimpleModule
+{
+    private static final Optional<MethodHandle> isRecord;
```
electrum, line 19: Make uppercase since this is a constant

json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private static class Introspector
+            extends AnnotationIntrospector
+    {
+        private final VisibilityChecker.Std recordVisibilityChecker;
+
+        public Introspector()
+        {
+            recordVisibilityChecker = VisibilityChecker.Std.defaultInstance()
+                    .withGetterVisibility(JsonAutoDetect.Visibility.PUBLIC_ONLY)
```
electrum, line 50: Is it intended that this is duplicated below?

json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private static class Introspector
+            extends AnnotationIntrospector
+    {
+        private final VisibilityChecker.Std recordVisibilityChecker;
+
+        public Introspector()
+        {
+            recordVisibilityChecker = VisibilityChecker.Std.defaultInstance()
```
electrum, line 49: Make this a constant?

json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
```diff
@@ -0,0 +1,87 @@
...
+        {
+            return Version.unknownVersion();
+        }
+    }
+
+    private static boolean isRecord(AnnotatedClass ac, MethodHandle methodHandle)
+    {
+        try {
+            // TODO replace with real call to isRecord() when Airlift is on Java 14+
+            return (Boolean) methodHandle.invoke(ac.getRawType());
```
electrum, line 81: You should be able to use `invokeExact()` here and cast to `(boolean)` to avoid boxing.

json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
```diff
@@ -0,0 +1,87 @@
...
+        extends SimpleModule
+{
+    private static final Optional<MethodHandle> isRecord;
+
+    static {
+        MethodHandle methodHandle;
+        try {
+            //noinspection JavaLangInvokeHandleSignature
+            methodHandle = MethodHandles.lookup()
+                    .findVirtual(Class.class, "isRecord", methodType(Boolean.TYPE));
```
electrum, line 26: One more nit: use `boolean.class` here for the primitive class reference

electrum (approved):

PR #996, opened 2022-07-14
[log-manager] add retries to TCP log handler

vinay-kosaraju:
This PR adds retry feature to the TCP log handler. It retries transmission of failed log records to the configured TCP listener

### DIFF
--- a/log-manager/src/main/java/io/airlift/log/BufferedHandler.java
+++ b/log-manager/src/main/java/io/airlift/log/BufferedHandler.java
@@ -1,13 +1,18 @@
 package io.airlift.log;
 
 import com.google.common.annotations.VisibleForTesting;
+import io.airlift.units.Duration;
 import org.weakref.jmx.Managed;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.ErrorManager;
@@ -17,6 +22,7 @@ import java.util.logging.LogRecord;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.logging.ErrorManager.CLOSE_FAILURE;
 import static java.util.logging.ErrorManager.FLUSH_FAILURE;
 import static java.util.logging.ErrorManager.FORMAT_FAILURE;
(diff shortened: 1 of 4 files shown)


log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -165,6 +182,34 @@ private void processQueue()
...
+                Thread.currentThread().interrupt();
+            }
+            else {
+                Duration delay = errorBackOff.failed();
+                log.warn("Processing log records to TCP listener failed. Will retry in " + delay);
+                try {
+                    Thread.sleep(delay.toMillis());
+                }
+                catch (InterruptedException ex) {
+                    Thread.currentThread().interrupt();
```
mattstep, line 210: You should exit the loop at this point, so everything can shut down

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -174,12 +219,33 @@ private synchronized void logMessageBatch(List<byte[]> batch)
...
+            }
+        }
+        // always flush at the end of a batch, so logs aren't delayed
+        flush();
+    }
+
+    private synchronized boolean logRetryMessageBatch()
+    {
+        boolean success = true;
+        for (Iterator<byte[]> message = retryBatch.iterator(); message.hasNext(); ) {
```
mattstep, line 235: Why not use foreach syntax or Iterables in guava?

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -25,20 +29,33 @@
...
 {
+    private static final Logger log = Logger.get(BufferedHandler.class);
     private static final int MAX_BATCH_COUNT = 1024;
     private static final byte[] POISON_MESSAGE = new byte[0];
+    private static final int MIN_WAIT_MILLIS = 100;
+    private static final int MAX_WAIT_MILLIS = 360000;
 
     private final MessageOutput messageOutput;
     private final Thread thread;
+    private Thread retryThread;
```
erichwang, line 40: Generally nobody uses raw Threads in Java anymore. Instead, you should always use an ExecutorService instead which properly decouples the concerns of task execution, from thread pool management -- and is strictly always better.

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -165,6 +182,34 @@ private void processQueue()
...
+            if (isSuccess) {
+                retryBatch.trimToSize();
+                errorBackOff.success();
+                Thread.currentThread().interrupt();
+            }
+            else {
+                Duration delay = errorBackOff.failed();
+                log.warn("Processing log records to TCP listener failed. Will retry in " + delay);
+                try {
+                    Thread.sleep(delay.toMillis());
```
erichwang, line 207: TimeUnit.MILLISECONDS.sleep(delay.toMillis())

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -165,6 +182,34 @@ private void processQueue()
 
             logMessageBatch(batch);
             batch.clear();
+
+            if (!retryBatch.isEmpty() && (retryThread == null || !retryThread.isAlive())) {
```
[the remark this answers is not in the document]
erichwang: Yep, nobody should manage raw threads anymore. That's almost always a bad idea.

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -25,20 +29,33 @@
...
     private static final int MAX_BATCH_COUNT = 1024;
     private static final byte[] POISON_MESSAGE = new byte[0];
+    private static final int MIN_WAIT_MILLIS = 100;
+    private static final int MAX_WAIT_MILLIS = 360000;
 
     private final MessageOutput messageOutput;
     private final Thread thread;
+    private Thread retryThread;
     private final BlockingQueue<byte[]> queue = new ArrayBlockingQueue<>(MAX_BATCH_COUNT);
+    private final ArrayList<byte[]> retryBatch = new ArrayList<>(MAX_BATCH_COUNT);
```
erichwang, line 42: Seeing a non-threadsafe object like this in an otherwise threadsafe class is highly suspicious for developers. Usually it's better to define it in the context of the loop that needs it where it is obviously only used by a single thread. If it's very very hard to pass the object around in the single threaded context, then you probably want to document this as thread confined.

erichwang, line 42: We generally use List instead of ArrayList as the type

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -165,6 +182,34 @@ private void processQueue()
...
+
+    private void processRetryQueue()
+    {
+        boolean isSuccess = false;
+        while (!isSuccess) {
+            isSuccess = logRetryMessageBatch();
+            if (isSuccess) {
+                retryBatch.trimToSize();
+                errorBackOff.success();
+                Thread.currentThread().interrupt();
```
erichwang, line 201: We are interrupting to finish on success? Why don't we just return from the task normally?

mattstep:
This is looking good to me. I think @erichwang and @dain should take a pass while I'm out next week. Thanks for getting the changes in so far from feedback from me and @erichwang.

PR #997, opened 2022-07-18
Delay announcement error logging

djsstarburst:
Before this change, as Trino starts up, a flurry of exceptions are
logged at the ERROR level as workers try to contact the coordinator.
These errors typically end after a few milliseconds, but they
pollute the logs.  This commit logs these exceptions at the
DEBUG level for the first 500ms after the first request, thereafter
logging at the ERROR level as before.

### DIFF
--- a/discovery/src/main/java/io/airlift/discovery/client/Announcer.java
+++ b/discovery/src/main/java/io/airlift/discovery/client/Announcer.java
@@ -156,8 +156,6 @@ public final class Announcer
 
                 // wait 80% of the suggested delay
                 expectedDelay = new Duration(expectedDelay.toMillis() * 0.8, MILLISECONDS);
-                log.debug("Service announcement succeeded after %s. Next request will happen within %s", Duration.nanosSince(requestStart), expectedDelay);
-
                 scheduleNextAnnouncement(expectedDelay);
             }
 
@@ -165,8 +163,6 @@ public final class Announcer
             public void onFailure(Throwable t)
             {
                 Duration duration = errorBackOff.failed(t);
-                // todo this is a duplicate log message and should be remove after root cause of announcement delay is determined
-                log.error("Service announcement failed after %s. Next request will happen within %s", Duration.nanosSince(requestStart), expectedDelay);
                 scheduleNextAnnouncement(duration);
             }
         }, executor);
--- a/discovery/src/main/java/io/airlift/discovery/client/ExponentialBackOff.java
+++ b/discovery/src/main/java/io/airlift/discovery/client/ExponentialBackOff.java
@@ -11,11 +11,15 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 class ExponentialBackOff
 {
+    private static final long ERROR_LOGGING_DELAY_NANOS = MILLISECONDS.toNanos(500);
+
     private final long initialWait;
     private final long maxWait;
     private final String serverUpMessage;
     private final String serverDownMessage;
     private final Logger log;
+
+    private final long requestStart = System.nanoTime();
 
     @GuardedBy("this")
     private boolean serverUp = true;
@@ -46,8 +50,11 @@ class ExponentialBackOff
     public synchronized Duration failed(Throwable t)
     {
         if (serverUp) {
-            serverUp = false;
-            log.error(t, serverDownMessage);
+            // skip logging until 500ms has passed
+            if ((System.nanoTime() - requestStart) >= ERROR_LOGGING_DELAY_NANOS) {
+                serverUp = false;
+                log.error(t, serverDownMessage);
+            }
         }
 
         if (currentWaitInMillis <= 0) {


djsstarburst:
I would also like to suppress error logging of this trino-worker exception we see during trino cluster suspend.  Would it be ok to _always_ log the exception at the debug level?

```
io.airlift.discovery.client.DiscoveryException: Unannouncement failed for http://trino-coordinator-service
	at io.airlift.discovery.client.HttpDiscoveryAnnouncementClient$DiscoveryResponseHandler.handleException(HttpDiscoveryAnnouncementClient.java:195)
	at io.airlift.http.client.jetty.JettyResponseFuture.failed(JettyResponseFuture.java:120)
	at io.airlift.http.client.jetty.BufferingResponseListener.onComplete(BufferingResponseListener.java:85)
	at org.eclipse.jetty.client.ResponseNotifier.notifyComplete(ResponseNotifier.java:218)
	at org.eclipse.jetty.client.ResponseNotifier.notifyComplete(ResponseNotifier.java:210)
	at org.eclipse.jetty.client.HttpExchange.notifyFailureComplete(HttpExchange.java:276)
	at org.eclipse.jetty.client.HttpExchange.abort(HttpExchange.java:247)
	at org.eclipse.jetty.client.HttpConversation.abort(HttpConversation.java:164)
	at org.eclipse.jetty.client.HttpRequest.abort(HttpRequest.java:819)
	at org.eclipse.jetty.client.HttpDestination.abort(HttpDestination.java:502)
	at org.eclipse.jetty.client.HttpDestination.failed(HttpDestination.java:252)
	at org.eclipse.jetty.client.AbstractConnectionPool$FutureConnection.failed(AbstractConnectionPool.java:557)
	at org.eclipse.jetty.client.HttpClient$1.failed(HttpClient.java:621)
	at org.eclipse.jetty.util.SocketAddressResolver$Async.lambda$resolve$1(SocketAddressResolver.java:192)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:883)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1034)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.net.UnknownHostException: trino-coordinator-service
	at java.base/java.net.InetAddress$CachedAddresses.get(InetAddress.java:948)
	at java.base/java.net.InetAddress.getAllByName0(InetAddress.java:1628)
	at java.base/java.net.InetAddress.getAllByName(InetAddress.java:1494)
	at org.eclipse.jetty.util.SocketAddressResolver$Async.lambda$resolve$1(SocketAddressResolver.java:170)
	... 3 more
```

discovery/src/main/java/io/airlift/discovery/client/Announcer.java
```diff
@@ -166,7 +167,13 @@ public void onFailure(Throwable t)
             {
                 Duration duration = errorBackOff.failed(t);
                 // todo this is a duplicate log message and should be remove after root cause of announcement delay is determined
```
electrum, line 169: This logging was added in 633e4d5e85f0bfc49d722c6d024c870bf25ebc6c

@dain should we remove it, along with the one in `onSuccess`?

discovery/src/main/java/io/airlift/discovery/client/ExponentialBackOff.java
```diff
@@ -8,15 +8,20 @@
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
 
 class ExponentialBackOff
 {
+    public static final long ERROR_LOGGING_DELAY_NANOS = Duration.succinctDuration(500, MILLISECONDS).roundTo(NANOSECONDS);
```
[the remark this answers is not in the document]
djsstarburst: LOL, much simpler.  Changed.

discovery/src/main/java/io/airlift/discovery/client/Announcer.java
djsstarburst: Removed.

djsstarburst:
@electrum - - That leaves the last announcement-related log pollution, logging of by workers that the cannot reach the coordinator to unannounce.  I'd like to get rid of this error logging.

```
io.airlift.discovery.client.DiscoveryException: Unannouncement failed for http://trino-coordinator-service
	at io.airlift.discovery.client.HttpDiscoveryAnnouncementClient$DiscoveryResponseHandler.handleException(HttpDiscoveryAnnouncementClient.java:195)
	at io.airlift.http.client.jetty.JettyResponseFuture.failed(JettyResponseFuture.java:120)
        ...
Caused by: java.net.UnknownHostException: trino-coordinator-service
	at java.base/java.net.InetAddress$CachedAddresses.get(InetAddress.java:948)
        ...
```

Here is the response handler's `handleException` method, which throws under four conditions.  How about if we just don't throw in the last two conditions when we are unannouncing?

```
        @Override
        public final T handleException(Request request, Exception exception)
        {
            if (exception instanceof InterruptedException) {
                throw new DiscoveryException(name + " was interrupted for " + uri);
            }
            if (exception instanceof CancellationException) {
                throw new DiscoveryException(name + " was canceled for " + uri);
            }
            if (exception instanceof DiscoveryException) {
                throw (DiscoveryException) exception;
            }

            throw new DiscoveryException(name + " failed for " + uri, exception);
        }
```

electrum:
How about we make it at WARN level? This only happens when the cluster is shutting down, so it shouldn't be ERROR.

djsstarburst:
> How about we make it at WARN level? This only happens when the cluster is shutting down, so it shouldn't be ERROR.

I'm not sure what's a good place to trap the error - - right now it's propagating all the way down to the thread launch.

But maybe shouldn't worry about this one.  Right now, we're getting ~700/day in prod, so there are bigger fish to fry.

electrum (approved):

PR #1002, opened 2022-07-29
Add method add(value, count) to the DistributionStat

radek-starburst:
There is no method to add a collected metric with a parameter `count` as there is for `Distribution`. This pull request adds that.


findepi (approved):
Please add an entry in CHANGES

findepi (approved):

PR #989, opened 2022-05-11
Bump byte-buddy from 1.10.22 to 1.12.6

findinpath:

PR #1004, opened 2022-08-09
Update byte-buddy from 1.12.6 to 1.12.13

nineinchnick:
The latest byte-buddy version is using the latest version of JNA, 5.12.1. The current one uses 5.8.0. I want to align them in Trino so that I can add a plugin that doesn't use Trino as a parent (so it doesn't inherit dependencies) but uses `trino-main` as a dependency. See https://github.com/trinodb/trino/pull/12929

I ran tests from `trino-main` with this version and all passed.

I also saw that it was quite recently updated in #989 but not to the latest available version. Maybe we could bump it in Tempto too.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 218
 
 - Add method add(value, count) to the DistributionStat.
+- Update to Byte Buddy 1.12.13.
 
 217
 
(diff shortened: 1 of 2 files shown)


CHANGES
```diff
@@ -1,6 +1,7 @@
 218
 
 - Add method add(value, count) to the DistributionStat.
+- Update to byte-buddy 1.12.13.
```
findepi, line 4: "Update to Byte Buddy 1.12.13."

PR #1014, opened 2022-09-30
Add openmetrics module to airlift to expose jmx metrics and stats via openmetrics

mattstep:
For reference to the spec see https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md

I started off with just adding explicit types to 2 stats objects, but we can certainly do more in the PR if there are critical ones for Trino.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ test-output/
 .pmdruleset.xml
 .pmd
 .flattened-pom.xml
+.attach_pid*
--- a/openmetrics/src/main/java/io/airlift/openmetrics/JmxOpenMetricsModule.java
+++ b/openmetrics/src/main/java/io/airlift/openmetrics/JmxOpenMetricsModule.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.openmetrics;
+
+import com.google.inject.Binder;
+import com.google.inject.Module;
+
+import static io.airlift.configuration.ConfigBinder.configBinder;
+import static io.airlift.jaxrs.JaxrsBinder.jaxrsBinder;
+
+public class JmxOpenMetricsModule
+        implements Module
+{
+    @Override
+    public void configure(Binder binder)
+    {
+        binder.disableCircularProxies();
+
+        configBinder(binder).bindConfig(MetricsConfig.class);
+        jaxrsBinder(binder).bind(MetricsResource.class);
+    }
+}
--- a/openmetrics/src/main/java/io/airlift/openmetrics/MetricsConfig.java
+++ b/openmetrics/src/main/java/io/airlift/openmetrics/MetricsConfig.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.openmetrics;
+
+import com.google.common.collect.ImmutableList;
+import io.airlift.configuration.Config;
+import io.airlift.configuration.ConfigDescription;
+
+import javax.management.MalformedObjectNameException;
+import javax.management.ObjectName;
+
+import java.util.List;
+
+public class MetricsConfig
+{
+    private List<ObjectName> jmxObjectNames = ImmutableList.of();
+
+    public List<ObjectName> getJmxObjectNames()
+    {
+        return jmxObjectNames;
+    }
+
+    @Config("openmetrics.jmx-object-names")
+    @ConfigDescription("JMX object names to include when retrieving all metrics.")
+    public MetricsConfig setJmxObjectNames(List<String> jmxObjectNames)
+    {
+        ImmutableList.Builder objectNames = ImmutableList.builder();
+
+        for (String jmxObjectName : jmxObjectNames) {
+            try {
+                objectNames.add(new ObjectName(jmxObjectName));
+            }
+            catch (MalformedObjectNameException e) {
+                throw new RuntimeException(e);
+            }
+        }
+
+        this.jmxObjectNames = objectNames.build();
+        return this;
+    }
+}
--- a/openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
+++ b/openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
@@ -1,0 +1,325 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.openmetrics;
+
+import com.google.common.collect.ImmutableList;
+import io.airlift.log.Logger;
+import io.airlift.openmetrics.types.Counter;
+import io.airlift.openmetrics.types.Gauge;
+import io.airlift.openmetrics.types.Metric;
+import io.airlift.openmetrics.types.Summary;
+import io.airlift.stats.CounterStat;
+import io.airlift.stats.TimeDistribution;
+import org.weakref.jmx.MBeanExporter;
+import org.weakref.jmx.ManagedClass;
+
+import javax.inject.Inject;
+import javax.management.InstanceNotFoundException;
+import javax.management.IntrospectionException;
+import javax.management.JMException;
+import javax.management.MBeanAttributeInfo;
+import javax.management.MBeanInfo;
+import javax.management.MBeanServer;
+import javax.management.MalformedObjectNameException;
+import javax.management.ObjectName;
+import javax.management.ReflectionException;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+import static java.util.Objects.requireNonNull;
+
+@Path("/metrics")
+public class MetricsResource
+{
+    private static final Logger log = Logger.get(MetricsResource.class);
+    private static final String OPENMETRICS_CONTENT_TYPE = "application/openmetrics-text; version=1.0.0; charset=utf-8";
+    private static final String ATTRIBUTE_SEPARATOR = "_ATTRIBUTE_";
+    private static final String TYPE_SEPARATOR = "_TYPE_";
+    private static final String NAME_SEPARATOR = "_NAME_";
+    private static final Pattern METRIC_NAME_PATTERN = Pattern.compile("[[a-zA-Z]][\\w_]*");
+
+    private final MBeanServer mbeanServer;
+    private final MBeanExporter mbeanExporter;
+
+    private final List<ObjectName> allMetricsObjectNames;
+
+    @Inject
+    public MetricsResource(MBeanServer mbeanServer, MBeanExporter mbeanExporter, MetricsConfig metricsConfig)
+    {
+        this.mbeanServer = requireNonNull(mbeanServer, "mbeanServer is null");
+        this.mbeanExporter = requireNonNull(mbeanExporter, "mbeanExporter is null");
+        this.allMetricsObjectNames = metricsConfig.getJmxObjectNames();
+    }
+
+    @GET
+    @Produces(OPENMETRICS_CONTENT_TYPE)
+    public String getMetrics(@QueryParam("name[]") List<String> filter)
+    {
+        StringBuilder body = new StringBuilder();
+        if (filter != null && !filter.isEmpty()) {
+            for (String metricName : filter) {
+                toMetricExposition(metricName).ifPresent(exposition -> body.append(exposition));
+            }
+        }
+        else {
+            body.append(managedMetricExpositions());
+            for (ObjectName metricObjectNames : allMetricsObjectNames) {
+                body.append(jmxMetricExpositions(metricObjectNames));
+            }
+        }
+        body.append("# EOF\n");
+        return body.toString();
+    }
+
+    private Set<ObjectName> objectNamesFromMetricName(String metricName)
+            throws MalformedObjectNameException
+    {
+        int nameStart = metricName.indexOf(NAME_SEPARATOR);
+        int typeStart = metricName.indexOf(TYPE_SEPARATOR);
+        int attributeStart = metricName.indexOf(ATTRIBUTE_SEPARATOR);
+
+        String domain;
+
+        if (nameStart != -1) {
+            domain = metricName.substring(0, nameStart).replace("_", ".");
+        }
+        else if (typeStart != -1) {
+            domain = metricName.substring(0, typeStart).replace("_", ".");
+        }
+        else {
+            domain = metricName.substring(0, attributeStart).replace("_", ".");
+        }
+
+        StringBuilder objectNameBuilder = new StringBuilder(domain).append(":");
+
+        if (nameStart != -1) {
+            objectNameBuilder.append("name=")
+                    .append(metricName, nameStart + NAME_SEPARATOR.length(), typeStart == -1 ? attributeStart : typeStart)
+                    .append(",");
+        }
+        if (typeStart != -1) {
+            objectNameBuilder.append("type=")
+                    .append(metricName.substring(typeStart + TYPE_SEPARATOR.length(), attributeStart).replace("_", "$"))
+                    .append(",");
+        }
+
+        return mbeanServer.queryNames(ObjectName.getInstance(objectNameBuilder.append("*").toString()), null);
+    }
+
+    private String attributeNameFromMetricName(String metricName)
+    {
+        int attributeNameStart = metricName.indexOf(ATTRIBUTE_SEPARATOR);
+        if (attributeNameStart == -1) {
+            throw new RuntimeException("Metric name invalid, no attribute separator %s".formatted(metricName));
+        }
+        return metricName.substring(attributeNameStart + ATTRIBUTE_SEPARATOR.length()).replace("_", ".");
+    }
+
+    private String mBeanNameToMetricName(ObjectName objectName, String attributeName)
+    {
+        if (objectName.getDomain().contains("_")) {
+            log.warn("Unable to expose JMX metric with domain name %s, package names with underscores are unsupported.", objectName.getDomain());
+            throw new RuntimeException("Bad domain name %s".formatted(objectName.getDomain()));
+        }
+
+        StringBuilder metricNameBuilder = new StringBuilder("JMX_")
+                .append(objectName.getDomain().replace(".", "_"));
+
+        if (objectName.getKeyProperty("name") != null) {
+            metricNameBuilder.append(NAME_SEPARATOR)
+                    .append(objectName.getKeyProperty("name"));
+        }
+
+        if (objectName.getKeyProperty("type") != null) {
+            metricNameBuilder.append(TYPE_SEPARATOR)
+                    .append(objectName.getKeyProperty("type").replace("$", "_"));
+        }
+
+        metricNameBuilder.append(ATTRIBUTE_SEPARATOR)
+                .append(attributeName.replace(".", "_"));
+
+        String metricName = metricNameBuilder.toString();
+
+        if (!METRIC_NAME_PATTERN.matcher(metricName).matches()) {
+            log.warn("Calculated metric name has invalid characters %s skipping", metricName);
+        }
+        return metricName;
+    }
+
+    private Optional<String> toMetricExposition(String metricName)
+    {
+        if (metricName.startsWith("JMX_")) {
+            final String jmxMetricName = metricName.substring(4);
+            try {
+                String attributeName = attributeNameFromMetricName(jmxMetricName);
+                return objectNamesFromMetricName(jmxMetricName).stream()
+                        .map(objectName -> getMetric(objectName, attributeName, jmxMetricName, ""))
+                        .flatMap(Optional::stream)
+                        .map(Metric::getMetricExposition)
+                        .findFirst();
+            }
+            catch (MalformedObjectNameException e) {
+                log.warn(e, "Unable to retrieve metric %s.", metricName);
+                return Optional.empty();
+            }
+        }
+        else {
+            Stream<Metric> metricStream = getManagedMetricsStream();
+            return metricStream
+                    .filter(metric -> metric.metricName().equals(metricName))
+                    .findFirst()
+                    .map(Metric::getMetricExposition);
+        }
+    }
+
+    private Optional<Metric> getMetric(ObjectName objectName, String attributeName, String metricName, String description)
+    {
+        try {
+            Object attributeValue = mbeanServer.getAttribute(objectName, attributeName);
+
+            if (attributeValue == null) {
+                return Optional.empty();
+            }
+
+            if (attributeValue instanceof Number) {
+                return Optional.of(Gauge.from(metricName, (Number) attributeValue, description));
+            }
+
+            return Optional.empty();
+        }
+        catch (JMException ex) {
+            log.debug(ex, "Unable to get metric for ObjectName %s and Attribute %s.", objectName.getCanonicalName(), attributeName);
+            return Optional.empty();
+        }
+    }
+
+    private String inferAttributesForObjectName(ObjectName objectName)
+    {
+        StringBuilder expositions = new StringBuilder();
+        try {
+            MBeanInfo mbeanInfo = mbeanServer.getMBeanInfo(objectName);
+            for (MBeanAttributeInfo mBeanAttributeInfo : mbeanInfo.getAttributes()) {
+                String attributeName = mBeanAttributeInfo.getName();
+                String description = mBeanAttributeInfo.getDescription();
+                try {
+                    getMetric(objectName, attributeName, mBeanNameToMetricName(objectName, attributeName), description)
+                            .ifPresent(value -> expositions.append(value.getMetricExposition()));
+                }
+                catch (RuntimeException e) {
+                    log.debug(e, "Unable to get Metric for ObjectName %s and Attribute %s, skipping", objectName.getCanonicalName(), attributeName);
+                }
+            }
+        }
+        catch (InstanceNotFoundException | IntrospectionException | ReflectionException e) {
+            log.debug(e, "Unable to get MBeanInfo for object %s, skipping", objectName.getCanonicalName());
+        }
+        return expositions.toString();
+    }
+
+    private String sanitizeMetricName(String name)
+    {
+        return name.replace(".", "_")
+                .replace("$", "_")
+                .replace(":", "_")
+                .replace("=", "_")
+                .replace(",", "_");
+    }
+
+    private List<Metric> getMetricsRecursively(String prefix, ManagedClass managedClass)
+    {
+        String metricName = sanitizeMetricName(prefix);
+
+        ImmutableList.Builder<Metric> metrics = ImmutableList.builder();
+
+        for (String attributeName : managedClass.getAttributeNames()) {
+            try {
+                String metricAndAttribute = managedClass.isAttributeFlatten(attributeName) ? metricName : metricName + "_" + attributeName;
+                String attributeDescription = managedClass.getAttributeDescription(attributeName);
+
+                ManagedClass child = managedClass.getChildren().get(attributeName);
+                if (child != null) {
+                    // The managed class is directly translatable to an openmetrics type, don't recurse any further
+                    Optional<Metric> metricFromTarget = getMetricFromTarget(child, metricAndAttribute, attributeDescription);
+                    if (metricFromTarget.isPresent()) {
+                        metrics.add(metricFromTarget.get());
+                    }
+                    else {
+                        // Recurse this nested child
+                        metrics.addAll(getMetricsRecursively(metricAndAttribute, child));
+                    }
+                }
+                else {
+                    // Attempt to infer a numeric gauge
+                    Object attributeValue = managedClass.invokeAttribute(attributeName);
+                    if (attributeValue instanceof Number) {
+                        metrics.add(Gauge.from(metricAndAttribute, (Number) attributeValue, attributeDescription));
+                    }
+                }
+            }
+            catch (ReflectiveOperationException e) {
+                log.debug("Unable to invoke getter for managed attribute : " + attributeName);
+            }
+        }
+
+        return metrics.build();
+    }
+
+    private Optional<Metric> getMetricFromTarget(ManagedClass managedClass, String metricName, String description)
+    {
+        if (managedClass.getTargetClass() == CounterStat.class) {
+            return Optional.of(Counter.from(metricName, (CounterStat) managedClass.getTarget(), description));
+        }
+
+        if (managedClass.getTargetClass() == TimeDistribution.class) {
+            return Optional.of(Summary.from(metricName, (TimeDistribution) managedClass.getTarget(), description));
+        }
+        return Optional.empty();
+    }
+
+    private String jmxMetricExpositions(ObjectName initialObjectName)
+    {
+        StringBuilder stringBuilder = new StringBuilder();
+
+        mbeanServer.queryNames(initialObjectName, null).forEach(objectName -> stringBuilder.append(inferAttributesForObjectName(objectName)));
+
+        return stringBuilder.toString();
+    }
+
+    private Stream<Metric> getManagedMetricsStream()
+    {
+        Map<String, ManagedClass> managedClasses = this.mbeanExporter.getManagedClasses();
+
+        return managedClasses.keySet().stream()
+                .map(objectName -> getMetricsRecursively(objectName, managedClasses.get(objectName)))
+                .flatMap(List::stream);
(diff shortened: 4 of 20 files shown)


openmetrics/src/main/java/io/airlift/openmetrics/MetricsConfig.java
```diff
@@ -0,0 +1,81 @@
...
+    private boolean inferAll;
+    private List<String> jmxDomains = ImmutableList.of();
+
+    public boolean isSanitizeMetricNames()
+    {
+        return sanitizeMetricNames;
+    }
+
+    @Config("openmetrics.sanitize-metric-names")
+    @ConfigDescription("Prevent metric names that don't match the openmetrics format to be returned from the metrics endpoint. This can be costly in CPU time if metrics are frequently scraped.")
```
dain, line 23: Why would we allow unsanitary metric names at all

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,265 @@
...
+            metricNameBuilder.append(TYPE_SEPARATOR)
+                    .append(objectName.getKeyProperty("type").replace("$", "_"));
+        }
+
+        metricNameBuilder.append(ATTRIBUTE_SEPARATOR)
+                .append(attributeName.replace(".", "_"));
+
+        String metricName = metricNameBuilder.toString();
+
+        if (sanitizeMetricNames && !metricName.matches("[[a-zA-Z]][\\w_]*")) {
```
dain, line 175: Precompile the regular expression and store it in a constant.  Also if this is expensive we can memoize

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,265 @@
...
+            log.warn("Calculated metric name has invalid characters %s skipping", metricName);
+        }
+        return metricName;
+    }
+
+    private ImmutableList<String> toMetricExposition(String metricName)
```
dain, line 181: Don't use Immutable* as a return, parameter, field, etc type

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,265 @@
...
+@Path("/v1/metrics")
+public class MetricsResource
+{
+    public static final String OPENMETRICS_CONTENT_TYPE = "application/openmetrics-text; version=1.0.0; charset=utf-8";
+    public static final String ATTRIBUTE_SEPARATOR = "_ATTRIBUTE_";
+    public static final String TYPE_SEPARATOR = "_TYPE_";
+    public static final String NAME_SEPARATOR = "_NAME_";
+
+    private final MBeanServer mbeanServer;
+    private static final Logger log = Logger.get(MetricsResource.class);
```
dain, line 62: move the constants up, so the instance fields are together

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,265 @@
...
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
+
+@Path("/v1/metrics")
+public class MetricsResource
+{
+    public static final String OPENMETRICS_CONTENT_TYPE = "application/openmetrics-text; version=1.0.0; charset=utf-8";
+    public static final String ATTRIBUTE_SEPARATOR = "_ATTRIBUTE_";
+    public static final String TYPE_SEPARATOR = "_TYPE_";
+    public static final String NAME_SEPARATOR = "_NAME_";
```
dain, line 59: Do these need to be public?

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,265 @@
...
+        try {
+            Object attributeValue = mbeanServer.getAttribute(objectName, attributeName);
+
+            if (attributeValue == null) {
+                return Optional.empty();
+            }
+
+            return inferMetricType(attributeValue);
+        }
+        catch (MBeanException | AttributeNotFoundException | InstanceNotFoundException | ReflectionException ex) {
```
dain, line 208: You can likely use JMException here

dain (commented):
Some initial comments

openmetrics/src/main/java/io/airlift/openmetrics/MetricsConfig.java
mattstep: There shouldn't be any unsanitary names, but this enables an extra check that will add some additional logging if the fetching agent (datadog, prometheus, etc.) doesn't like the names we generate.

openmetrics-api/src/main/java/io/airlift/openmetrics/types/BigCounter.java
```diff
@@ -0,0 +1,19 @@
...
+
+public record BigCounter(BigInteger value) implements Metric
+{
+    @Override
+    public String getMetricExposition(String metricName)
+    {
+        StringBuilder stringBuilder = new StringBuilder()
+                .append(ExpositionHelper.typeLine(metricName, "counter"))
+                .append("\n")
+                .append(ExpositionHelper.valueLine(metricName, value.toString()))
```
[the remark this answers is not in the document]
mattstep: I made the helper take CharSequence, so a builder or string could be passed in, if it's a builder it looks like it does the right thing digging in the StringBuilder class... I also have them return builders now which also eliminates another potential copy of the string. I'll push the new code today, but it seems CharSequence is the interface that makes this all work wonderfully well.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsConfig.java
mattstep: Removed with the precompiled regex and the memoization, it'll be fast enough now..

dain:
Remove the `discovery/.attach_pid1876507` file

openmetrics-api/src/main/java/io/airlift/openmetrics/ExpositionHelper.java
```diff
@@ -0,0 +1,32 @@
+package io.airlift.openmetrics;
```
dain, line 1: missing license headers

openmetrics/src/main/java/io/airlift/openmetrics/JmxOpenMetricsModule.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2010 Proofpoint, Inc.
```
dain, line 2: Don't put copyright on new files

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,284 @@
...
+    {
+        StringBuilder stringBuilder = new StringBuilder();
+
+        for (ObjectName objectName : mbeanServer.queryNames(initialObjectName, null)) {
+            try {
+                MBeanInfo mbeanInfo = mbeanServer.getMBeanInfo(objectName);
+                for (String attributeName : Arrays.stream(mbeanInfo.getAttributes()).map(MBeanAttributeInfo::getName).collect(toImmutableList())) {
+                    try {
+                        Optional<Metric> metric = getMetric(objectName, attributeName);
+                        if (metric.isPresent()) {
```
dain, line 268: you can use function style here:
```
getMetric(objectName, attributeName)
        .ifPresent(value -> stringBuilder.append(value.getMetricExposition(mBeanNameToMetricName(objectName, attributeName))));

File: openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java [hunk 11]
```diff
@@ -0,0 +1,284 @@
...
+@Path("/v1/metrics")
+public class MetricsResource
+{
+    private static final Logger log = Logger.get(MetricsResource.class);
+    private static final String OPENMETRICS_CONTENT_TYPE = "application/openmetrics-text; version=1.0.0; charset=utf-8";
+    private static final String ATTRIBUTE_SEPARATOR = "_ATTRIBUTE_";
+    private static final String TYPE_SEPARATOR = "_TYPE_";
+    private static final String NAME_SEPARATOR = "_NAME_";
+
+    private final Pattern metricNamePattern = Pattern.compile("[[a-zA-Z]][\\w_]*");
```
dain, line 61: IntelliJ says this should be:
```
private final Pattern metricNamePattern = Pattern.compile("[a-zA-Z][\\w_]*");

File: openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java [hunk 12]
```diff
@@ -0,0 +1,284 @@
...
+            case Byte value -> inferNumeric ? Optional.of(new Counter(value)) : Optional.empty();
+
+            // Interpret floating point values as gauges
+            case Double value -> inferNumeric ? Optional.of(new Gauge(value)) : Optional.empty();
+            case Float value -> inferNumeric ? Optional.of(new Gauge(value)) : Optional.empty();
+
+            // Strings should be info
+            case String value -> inferString ? Optional.of(new Info(value)) : Optional.empty();
+
+            // If we don't know what it is, we produce an info with it's string representation
```
dain, line 253: IntelliJ says this should be `its string representation`

sample-server/src/main/java/io/airlift/sample/Main.java
```diff
@@ -53,7 +54,8 @@ public static void main(String[] args)
                 new LogJmxModule(),
                 new HttpEventModule(),
                 new TraceTokenModule(),
-                new MainModule());
+                new MainModule(),
+                new JmxOpenMetricsModule());
```
dain, line 58: move this up to be after the JMX stuff

skeleton-server/src/main/java/io/airlift/skeleton/Main.java
```diff
@@ -53,6 +54,7 @@ public static void main(String[] args)
                 new LogJmxModule(),
                 new HttpEventModule(),
                 new TraceTokenModule(),
+                new JmxOpenMetricsModule(),
```
dain, line 57: same

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
dain: This can be a constant

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,284 @@
...
+public class MetricsResource
+{
+    private static final Logger log = Logger.get(MetricsResource.class);
+    private static final String OPENMETRICS_CONTENT_TYPE = "application/openmetrics-text; version=1.0.0; charset=utf-8";
+    private static final String ATTRIBUTE_SEPARATOR = "_ATTRIBUTE_";
+    private static final String TYPE_SEPARATOR = "_TYPE_";
+    private static final String NAME_SEPARATOR = "_NAME_";
+
+    private final Pattern metricNamePattern = Pattern.compile("[[a-zA-Z]][\\w_]*");
+    private final Map<String, Boolean> memoizedValidNames = new HashMap<>();
```
dain, line 62: This is not thread safe... use a concurrent map

dain: Actually use a Guava cache

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,284 @@
...
+    private final boolean inferNumeric;
+    private final boolean inferString;
+    private final boolean inferAll;
+
+    private final List<ObjectName> allMetricsObjectNames;
+
+    @Inject
+    public MetricsResource(MBeanServer mbeanServer, MetricsConfig metricsConfig)
+    {
+        this.mbeanServer = mbeanServer;
```
dain, line 74: check null

mattstep:
> Remove the `discovery/.attach_pid1876507` file

done, and added a .gitignore, I think jmc drops these randomly...

openmetrics-api/src/test/java/io/airlift/openmetrics/TestMetricExpositions.java
```diff
@@ -0,0 +1,90 @@
...
+import io.airlift.openmetrics.types.Gauge;
+import io.airlift.openmetrics.types.Info;
+import io.airlift.openmetrics.types.Summary;
+import org.testng.annotations.Test;
+
+import java.math.BigInteger;
+
+import static org.testng.Assert.assertEquals;
+
+public class TestMetricExpositions
```
losipiuk, line 28: nit: I think test would me more readable if you use multiline literals instead concatenating.

Also use of production class `ExpositionHelper` to generate expected result is antipattern.

openmetrics-api/src/main/java/io/airlift/openmetrics/types/Summary.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Map;
+
+import static io.airlift.openmetrics.ExpositionHelper.typeLine;
+import static io.airlift.openmetrics.ExpositionHelper.valueLine;
+
+public record Summary(Long count, Double sum, Double created, Map<Double, Double> quantiles)
+        implements Metric
+{
+    @Override
+    public CharSequence getMetricExposition(String metricName)
```
losipiuk, line 25: cruious: why `CharSequence` and not `String`? Performance? I am not a big fan of returning mutable object, even if hidden behind interface. Maybe this is fine.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,281 @@
...
+@Path("/v1/metrics")
+public class MetricsResource
+{
+    private static final Logger log = Logger.get(MetricsResource.class);
+    private static final String OPENMETRICS_CONTENT_TYPE = "application/openmetrics-text; version=1.0.0; charset=utf-8";
+    private static final String ATTRIBUTE_SEPARATOR = "_ATTRIBUTE_";
+    private static final String TYPE_SEPARATOR = "_TYPE_";
+    private static final String NAME_SEPARATOR = "_NAME_";
+    private static final Pattern METRIC_NAME_PATTERN = Pattern.compile("[[a-zA-Z]][\\w_]*");
+    private static final LoadingCache<String, Boolean> VALID_NAMES_CACHE = CacheBuilder.newBuilder().build(new CacheLoader<String, Boolean>()
```
losipiuk, line 61: should keys expire eventually.

also nit: `<String, Boolean>` -> `<>`

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,281 @@
...
+import static java.util.Objects.requireNonNull;
+
+@Path("/v1/metrics")
+public class MetricsResource
+{
+    private static final Logger log = Logger.get(MetricsResource.class);
+    private static final String OPENMETRICS_CONTENT_TYPE = "application/openmetrics-text; version=1.0.0; charset=utf-8";
+    private static final String ATTRIBUTE_SEPARATOR = "_ATTRIBUTE_";
+    private static final String TYPE_SEPARATOR = "_TYPE_";
+    private static final String NAME_SEPARATOR = "_NAME_";
```
losipiuk, line 59: It feels like those can appear not that rarely in object names. Maybe use double underscores at beginning/end?

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,281 @@
...
+        else {
+            for (ObjectName metricObjectNames : allMetricsObjectNames) {
+                body.append(allMetricExpositions(metricObjectNames));
+            }
+        }
+        body.append("# EOF\n");
+        return body.toString();
+    }
+
+    private Set<ObjectName> objectNamesFromMetricName(String metricName)
```
losipiuk, line 119: those domain conversion methods are tricky. 
I strongly suggest extracting those as a helper and cover with unit tests.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,281 @@
...
+
+    private final List<ObjectName> allMetricsObjectNames;
+
+    @Inject
+    public MetricsResource(MBeanServer mbeanServer, MetricsConfig metricsConfig)
+    {
+        this.mbeanServer = requireNonNull(mbeanServer, "mbeanServer is null");
+        this.inferNumeric = metricsConfig.isInferNumeric();
+        this.inferString = metricsConfig.isInferStrings();
+        this.inferAll = metricsConfig.isInferAll();
```
losipiuk, line 83: When I read `@ConfigDescription` i though that `inferAll` implies `inferNumeric` and `inferString`.
If that is not the case maybe rename the property to `inferOther`.

Or change above lines to:
```
        this.inferNumeric = metricsConfig.isInferAll() || metricsConfig.isInferNumeric();
        this.inferString = metricsConfig.isInferAll() || metricsConfig.isInferStrings();
```

losipiuk (commented):
I skimmed. This is cool and highly needed :) Is there an ETA. Any help required?

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,340 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.util.Objects.requireNonNull;
+
+@Path("/metrics")
+public class MetricsResource
```
losipiuk, line 56: There is mostly no test coverage for this class. Especially when it comes to its interactions with 
`MBeanExporter`. Can this be improved? I am worried about the maintainability of this code.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,340 @@
...
+    private String sanitizeMetricName(String name)
+    {
+        return name.replace(".", "_")
+                .replace("$", "_")
+                .replace(":", "_")
+                .replace("=", "_")
+                .replace(",", "_");
+    }
+
+    private List<Metric> getMetricsRecusively(String prefix, ManagedClass managedClass)
```
losipiuk, line 268: nit: typo in method name

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,340 @@
...
+            return Optional.of(Summary.from(metricName, (TimeDistribution) managedClass.getTarget(), description));
+        }
+        return Optional.empty();
+    }
+
+    private String jmxMetricExpositions(ObjectName initialObjectName)
+    {
+        StringBuilder stringBuilder = new StringBuilder();
+
+        mbeanServer.queryNames(initialObjectName, null).stream().forEach(objectName -> stringBuilder.append(inferAttributesForObjectName(objectName)));
```
losipiuk, line 323: nit: drop `.stream()`

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,340 @@
...
+        }
+        catch (JMException ex) {
+            log.debug(ex, "Unable to get metric for ObjectName %s and Attribute %s.", objectName.getCanonicalName(), attributeName);
+            return Optional.empty();
+        }
+    }
+
+    private String inferAttributesForObjectName(ObjectName objectName)
+    {
+        StringBuilder expositions = new StringBuilder();
```
losipiuk, line 238: There are `Metric` to `String` conversions and then `String` concatenations at multiple layers in this metrics traversal logic. 
E.g we concatenate here in `inferAttributesForObjectName` and the method which calls it (`jmxMetricExpositions`).

I think the code would be easier to read if we obtained all the `Metric` objects we care about and we had just one place when we call `getMetricExposition` and concatenate those to build the response (in `getMetrics`).

Obtaining list of `Metric` objects can be done with an accumulator (E.g sth based on `ImmutableList.Builder`) you pass around and add stuff to it. The changed method signatures would look sth like:
* `void managedMetricExpositions(Consumer<Metric>)`
* `void getMetricsRecusively(String prefix, ManagedClass managedClass, Consumer<Metric> accumulator)`
* `String jmxMetricExpositions(ObjectName initialObjectName)` -> `void jmxMetrics(ObjectName initialObjectName, Consumer<Metric> accumulator)`
* ...

I believe code-wise it would very good chage. But maybe there is a reason why we cannot do so? E.g. maybe is keeping `Metric` objects in memory for request duration significantly more costly than theirs `String` based expositions?

openmetrics/src/main/java/io/airlift/openmetrics/JmxOpenMetricsModule.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class JmxOpenMetricsModule
+        implements Module
+{
+    @Override
+    public void configure(Binder binder)
+    {
+        binder.disableCircularProxies();
+
+        binder.bind(MBeanServer.class).toInstance(ManagementFactory.getPlatformMBeanServer());
```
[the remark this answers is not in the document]
mattstep: This doesn't depend on JmxModule, it can be used with or without it.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,340 @@
...
+    @Inject
+    public MetricsResource(MBeanServer mbeanServer, MBeanExporter mbeanExporter, MetricsConfig metricsConfig)
+    {
+        this.mbeanServer = requireNonNull(mbeanServer, "mbeanServer is null");
+        this.mbeanExporter = requireNonNull(mbeanExporter, "mbeanExporter is null");
+
+        if (metricsConfig.getJmxDomains().isEmpty()) {
+            this.allMetricsObjectNames = ImmutableList.of();
+        }
+        else {
```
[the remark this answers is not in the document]
mattstep: I don't follow

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,340 @@
...
+@Path("/metrics")
+public class MetricsResource
+{
+    private static final Logger log = Logger.get(MetricsResource.class);
+    private static final String OPENMETRICS_CONTENT_TYPE = "application/openmetrics-text; version=1.0.0; charset=utf-8";
+    private static final String ATTRIBUTE_SEPARATOR = "_ATTRIBUTE_";
+    private static final String TYPE_SEPARATOR = "_TYPE_";
+    private static final String NAME_SEPARATOR = "_NAME_";
+    private static final Pattern METRIC_NAME_PATTERN = Pattern.compile("[[a-zA-Z]][\\w_]*");
+    private static final LoadingCache<String, Boolean> VALID_NAMES_CACHE = CacheBuilder.newBuilder().build(new CacheLoader<String, Boolean>()
```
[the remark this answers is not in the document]
mattstep: I don't see how this could grow to a large size considering it's logically bound by the amount of jmx mbeans. If a user is attempting to pull so many metrics that this grows to a large size it would likely never actually return the metrics list.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,340 @@
...
+        if (filter != null && !filter.isEmpty()) {
+            for (String metricName : filter) {
+                body.append(Joiner.on("").join(toMetricExposition(metricName)));
+            }
+        }
+        else {
+            body.append(managedMetricExpositions());
+            for (ObjectName metricObjectNames : allMetricsObjectNames) {
+                body.append(jmxMetricExpositions(metricObjectNames));
+            }
```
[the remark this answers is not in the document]
mattstep: It'll return both and be a configuration issue that's very difficult to test for. In general this is very difficult to do, you'd have to define a custom object namer in jmxutils that returns the prefix I'm using to namespace the jmx metrics apart from the managed beans.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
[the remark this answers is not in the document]
losipiuk: I meant you can just leave what is in `else` clause

openmetrics/src/main/java/io/airlift/openmetrics/JmxOpenMetricsModule.java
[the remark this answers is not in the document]
martint: It might be detecting that the object returned by `ManagementFactory.getPlatformMBeanServer()` is the same instance, but that's quite brittle. 

Indeed, this module should not bind the MBeanServer -- that's the job of JmxModule. This module should just be in charge of wiring the REST resource, config, etc.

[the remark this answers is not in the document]
mattstep: It will always be the same instance, it's even documented in the jdk to be the same instance. https://docs.oracle.com/en/java/javase/18/docs/api/java.management/java/lang/management/ManagementFactory.html#getPlatformMBeanServer() I'll look further next week if you feel strongly about this.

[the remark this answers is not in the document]
mattstep: This has been a part of this PR for 4 months, so if it hasn't been a priority in that timeframe, then I don't see it as a priority to resolve. I will get to this and the other feedback in the coming weeks. Please take time to review this fully and I'll resolve as I have time.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
martint, line 64: What's the purpose of this cache? Seems like a premature optimization?

openmetrics/src/main/java/io/airlift/openmetrics/MetricsConfig.java
```diff
@@ -0,0 +1,38 @@
...
+public class MetricsConfig
+{
+    private List<String> jmxDomains = ImmutableList.of();
+
+    public List<String> getJmxDomains()
+    {
+        return jmxDomains;
+    }
+
+    @Config("openmetrics.jmx-domains")
```
martint, line 31: It'd be more powerful and generic for these to be objectname patterns. It would allow for more complex patterns that filter out certain object names that are not interesting, or to name object names explicitly.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,340 @@
...
+                    if (metricFromTarget.isPresent()) {
+                        metrics.add(metricFromTarget.get());
+                    }
+                    else {
+                        // Recurse this nested child
+                        metrics.addAll(getMetricsRecusively(metricAndAttribute, child));
+                    }
+                }
+                else {
+                    // Attempt to infer a numeric guage
```
martint, line 292: Typo: guage

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,340 @@
...
+
+    private String managedMetricExpositions()
+    {
+        StringBuilder stringBuilder = new StringBuilder();
+
+        final Map<String, ManagedClass> managedClasses = this.mbeanExporter.getManagedClasses();
+
+        managedClasses.keySet().stream().map(
+                objectName -> getMetricsRecusively(objectName, managedClasses.get(objectName))
+        ).flatMap(List::stream).forEach(metric -> stringBuilder.append(metric.getMetricExposition()));
```
martint, line 336: Format as:

```java
 managedClasses.keySet().stream()
    .map(objectName -> getMetricsRecusively(objectName, managedClasses.get(objectName)))
    .flatMap(List::stream)
    .forEach(metric -> stringBuilder.append(metric.getMetricExposition()));
```

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,340 @@
...
+        StringBuilder stringBuilder = new StringBuilder();
+
+        mbeanServer.queryNames(initialObjectName, null).stream().forEach(objectName -> stringBuilder.append(inferAttributesForObjectName(objectName)));
+
+        return stringBuilder.toString();
+    }
+
+    private String managedMetricExpositions()
+    {
+        StringBuilder stringBuilder = new StringBuilder();
```
martint, line 330: This could be named `builder` for conciseness and without loss of clarity.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,340 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.util.Objects.requireNonNull;
+
+@Path("/metrics")
```
martint, line 55: This name is too generic and it might conflict with other usages or other forms of metrics providers. Let's rename it to "/openmetrics"

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,340 @@
...
+        mbeanServer.queryNames(initialObjectName, null).stream().forEach(objectName -> stringBuilder.append(inferAttributesForObjectName(objectName)));
+
+        return stringBuilder.toString();
+    }
+
+    private String managedMetricExpositions()
+    {
+        StringBuilder stringBuilder = new StringBuilder();
+
+        final Map<String, ManagedClass> managedClasses = this.mbeanExporter.getManagedClasses();
```
martint, line 332: Don't use final for local variables.

openmetrics/src/main/java/io/airlift/openmetrics/types/BigCounter.java
```diff
@@ -0,0 +1,37 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+public record BigCounter(String metricName, BigInteger value, String help)
+        implements Metric
+{
+    public BigCounter(String metricName, BigInteger value, String help)
+    {
+        this.metricName = requireNonNull(metricName, "metricName is null");
+        this.value = requireNonNull(value, "Value is null");
```
martint, line 26: `Value` -> `value`

openmetrics/src/main/java/io/airlift/openmetrics/types/BigCounter.java
```diff
@@ -0,0 +1,37 @@
...
+        this.help = help;
+    }
+
+    @Override
+    public String getMetricExposition()
+    {
+        return TYPE_LINE_FORMAT.formatted(metricName, "counter") +
```
martint, line 33: This could be generalized into a common method for formatting all metrics:

```java
public static String formatMetric(String name, String type, String help, String value)
{
    return "# TYPE %s %s\\n".formatted(name, type) +
           (Strings.isNullOrEmpty(help) ? "" : "# HELP %s %s\\n".formatted(name, help)) +
           "%s %s\\n".formatted(name, value);
}
```

And then used in each implementation:

```java
return formatMetric(metricName, "counter", help, value.toString());
```

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
mattstep: As we chatted about this on slack, not all collection daemons support configuring custom paths. Changing this would make scraping more difficult.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
mattstep: I'm happy to chat about improving this out of band from this PR review.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
mattstep: It's just that running that regex hundreds of times per invocation of the /metrics endpoint seems wasteful. It shouldn't grow substantially and just reduces the potential cpu load of a misconfigured scraping client (prometheus can be configured to scrape as frequently as 1ms).

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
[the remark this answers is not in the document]
mattstep: This would change with one of Martin's suggestions around objectnames, so I'm going to look into that refactoring instead.

.gitignore
```diff
@@ -15,3 +15,4 @@ test-output/
 .pmdruleset.xml
 .pmd
 .flattened-pom.xml
+.attach_pid*
```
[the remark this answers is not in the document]
mattstep: Broken into a separate commit, I can also break it into a separate PR if need be, but this is just a file that gets dropped by java mission control, which I was using to explore the mbeans.

losipiuk:
I put some comments here but, please treat those just as suggestions. I do not deeply care about those. I do care about the feature though :) 

@martint (you are the Airlift maintainer so I ask you), are there any substantial issues to be addressed still or is it more or less ready to go?
If we are afraid of pushing out something which is not 100% baked, and care about future incompatible changes, maybe it is worth to mark feature as `experimental` to set proper expectations on API stability and merge as is?

openmetrics/src/main/java/io/airlift/openmetrics/types/BigCounter.java
mattstep: This works for single valued metrics, I've updated my PR to use it in those places, but it doesn't work for histograms/summaries.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -0,0 +1,351 @@
...
+@Path("/metrics")
+public class MetricsResource
+{
+    private static final Logger log = Logger.get(MetricsResource.class);
+    private static final String OPENMETRICS_CONTENT_TYPE = "application/openmetrics-text; version=1.0.0; charset=utf-8";
+    private static final String ATTRIBUTE_SEPARATOR = "_ATTRIBUTE_";
+    private static final String TYPE_SEPARATOR = "_TYPE_";
+    private static final String NAME_SEPARATOR = "_NAME_";
+    private static final Pattern METRIC_NAME_PATTERN = Pattern.compile("[[a-zA-Z]][\\w_]*");
+    public static final int MAXIMUM_METRICS = 1000000;
```
martint, line 63: Any reason for this to be public?

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
martint: I would get rid of it until we prove it's a problem. If someone misconfigures prometheus to scrape every 1ms, they have bigger problems, and they will notice the error very quickly.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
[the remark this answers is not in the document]
mattstep: It's hard to do that considering the jmx metrics are prefixed with JMX and the others are in a different naming format entirely.

martint (approved):

PR #1015, opened 2022-10-13
Fix duplicate logging appender when re-configuring logging

elecnix:
Logging is a singleton that's initialized once, but it may be configured more than once. Doing so used to add a duplicate appender if a `log.path` configuration is provided. 

Calling `LogManager.reset()` and resetting the root logger's appenders during Logging configuration fixes this duplicate appender problem.

### DIFF
--- a/log-manager/src/main/java/io/airlift/log/Logging.java
+++ b/log-manager/src/main/java/io/airlift/log/Logging.java
@@ -81,15 +81,22 @@ public class Logging
 
     private Logging()
     {
-        ROOT.setLevel(Level.INFO.toJulLevel());
+        resetConfiguration();
+        log.info("Logging to stderr");
+
+        redirectStdStreams();
+    }
+
+    private void resetConfiguration()
+    {
+        // Reset named loggers, remove their handlers, and reset root to INFO
+        LogManager.getLogManager().reset();
+
         for (Handler handler : ROOT.getHandlers()) {
             ROOT.removeHandler(handler);
         }
 
         enableConsole();
-        log.info("Logging to stderr");
-
-        redirectStdStreams();
     }
 
     private static void redirectStdStreams()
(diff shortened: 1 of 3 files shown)


log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -81,15 +81,22 @@ public static synchronized Logging initialize()
 
     private Logging()
     {
-        ROOT.setLevel(Level.INFO.toJulLevel());
```
mattstep, line 84: Why remove this?

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -81,15 +81,22 @@ public static synchronized Logging initialize()
...
+        resetConfiguration();
+        log.info("Logging to stderr");
+
+        redirectStdStreams();
+    }
+
+    private void resetConfiguration()
+    {
+        // Reset named loggers, remove their handlers, and reset root to INFO
+        LogManager.getLogManager().reset();
```
mattstep, line 93: It looks like this will drop logs when it does the reset if the appenders haven't flushed yet. You might be safer doing a no-op after the initial configuration if this is called a second time.

log-manager/src/main/java/io/airlift/log/Logging.java
elecnix: It became redundant with LogManager.reset() which does the same. See its resetLogger: (defaultLevel is Level.INFO)
```
        if (name != null && name.isEmpty()) {
            // This is the root logger.
            logger.setLevel(defaultLevel);
```

log-manager/src/main/java/io/airlift/log/Logging.java
elecnix: Hmm, good point. There are two issues:
* Appenders may not implement close() correctly by flushing their buffers.
* The app may log something after the appenders were removed, and before the new appenders are added.
I'll close this PR.

PR #1016, opened 2022-10-14
Bootstrap logging once

elecnix:
Logging is a singleton that's initialized once, but it may be configured more than once. Doing so used to add a duplicate appender if a log.path configuration is provided. An initial attempt at avoiding duplicate appenders was submitted in #1015.

When initializing subsequent Bootstrap instances, do not re-configure logging.

### DIFF
--- a/log-manager/src/main/java/io/airlift/log/Logging.java
+++ b/log-manager/src/main/java/io/airlift/log/Logging.java
@@ -62,6 +62,9 @@ public class Logging
     @GuardedBy("this")
     private OutputStreamHandler consoleHandler;
 
+    @GuardedBy("this")
+    private boolean configured;
+
     /**
      * Sets up default logging:
      * <p>
@@ -198,8 +201,14 @@ public class Logging
         return levels.build();
     }
 
-    public void configure(LoggingConfiguration config)
+    public synchronized void configure(LoggingConfiguration config)
     {
+        if (configured) {
+            log.warn("Logging already configured; ignoring new configuration.");
+            return;
+        }
+        configured = true;
+
         Map<String, String> logAnnotations = ImmutableMap.of();
         if (config.getLogAnnotationFile() != null) {
             try {
--- a/bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrap.java
+++ b/bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrap.java
@@ -25,6 +25,7 @@ import javax.inject.Inject;
 import static io.airlift.testing.Assertions.assertContains;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.fail;
 
 public class TestBootstrap
@@ -114,6 +115,16 @@ public class TestBootstrap
                     assertThat(e).hasStackTraceContaining("Configuration property 'test-required' was not used");
                     assertThat(e).hasStackTraceContaining("An exception was caught and reported. Message: happy user error");
                 });
+    }
+
+    @Test
+    public void testLoggingConfiguredOnce()
+    {
+        java.util.logging.Logger root = java.util.logging.Logger.getLogger("");
+        new Bootstrap().setOptionalConfigurationProperty("log.path", "tcp://0.0.0.0:0").initialize();
+        int configuredHandlerCount = root.getHandlers().length;
+        new Bootstrap().setOptionalConfigurationProperty("log.path", "tcp://0.0.0.0:0").initialize();
+        assertEquals(root.getHandlers().length, configuredHandlerCount);
     }
 
     public static class Instance {}


bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -209,9 +210,15 @@ public Injector initialize()
 
         // initialize logging
         if (logging != null) {
```
mattstep, line 212: We should probably also check `initializeLogging` here, it seems like we don't want to configure logging if that's set.

elecnix: If `initializeLogging` is set, then `logging` will be too.

djsstarburst (approved):
This certainly isn't my area of expertise, but it looked ok to me, and apparently Matt looked at an earlier version, so I'm approving it.

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
electrum: More specifically, `logging` is only set if `initializeLogging` is set.

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -72,6 +72,7 @@
     private boolean strictConfig = true;
 
     private boolean initialized;
+    private static boolean loggingConfigured;
```
electrum, line 75: Make this an `AtomicBoolean` since this could be used from multiple threads.

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -209,9 +210,15 @@ public Injector initialize()
 
         // initialize logging
         if (logging != null) {
-            log.info("Initializing logging");
-            LoggingConfiguration configuration = configurationFactory.build(LoggingConfiguration.class);
-            logging.configure(configuration);
+            if (loggingConfigured) {
+                log.warn("Logging already configured; ignoring new configuration.");
```
electrum, line 214: I don't think we want to log a warning when the configuration is the same. It would be better to silently ignore it. Otherwise, this is effectively making it a mistake to initialize logging twice, with the result being that we should change any such application code to not initialize the second time. But if we were to do that, we wouldn't need this change at all (except to catch mistakes).

electrum (commented):
I assume you intend to squash these commits?

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
mattstep: Ah, yeah, it might read better if we just check the same boolean here...

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
elecnix: I would then change to info, then.

Comparing configurations is something I'm trying to avoid, as it's not trivial, and doesn't bring much value. We're just trying to avoid adding the appender twice.

elecnix:
> I assume you intend to squash these commits?

Yes, but I usually do this when merging. However, I do not have merge rights, so would you prefer that I squash them all now?

Also, I've addressed all the comments in the PR. Please have another look. Thanks!

erichwang:
Looking at this code, it almost feels like the stateful check to configure logging should be in the logging class itself. For example, the Logging.initialize() controls itself to make sure it never initializes more than once. It feels cleaner to me to add a new method to Logging such as `configureOnce()` or `initialConfiguration()` which does the same thing and guarantees that it doesn't apply more than once.

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -72,6 +73,7 @@
     private boolean strictConfig = true;
 
     private boolean initialized;
+    private static AtomicBoolean loggingConfigured = new AtomicBoolean(true);
```
elecnix, line 76: Should be `false`

martint:
Agreed with @erichwang. Since Logging is a stateful singleton, if we need to guarantee it only initializes once, it should be done inside that class itself, independently of Bootstrap. At the end of the day, anyone can attempt to initialize Logging from anywhere, and callers may not even be relying on Bootstrap.

elecnix:
@erichwang and @martint Thanks for the comments. I've updated the PR and it is now ready to review again.

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -62,6 +62,8 @@
     @GuardedBy("this")
     private OutputStreamHandler consoleHandler;
 
+    private boolean configured;
```
electrum, line 65: Use an atomic boolean or add the `@GuardedBy` annotation

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -198,8 +201,12 @@ public Map<String, Level> getAllLevels()
         return levels.build();
     }
 
-    public void configure(LoggingConfiguration config)
+    public synchronized void configure(LoggingConfiguration config)
     {
+        if (configured) {
+            log.info("Logging already configured; ignoring new configuration.");
```
martint, line 207: We shouldn't be logging this at info level. It's likely a programming error, so it should be a warning if anything.

elecnix: Multiple Bootstrap apps can be composed to be re-used together, and used independently as well. It seems like a valid use-case. The problem is that each app may want its logging to be configured differently, and that Logging configuration is shared with a singleton. It could be a programming error, or not. I'm seeing many opinions, and I've opted for `info` because we don't know, at that level.

elecnix: @electrum was [arguing](https://github.com/airlift/airlift/pull/1016#discussion_r1001185166) for silently ignoring it. I feel like this is something worth logging, or someone could be very surprised down the road and waste time investigating.

martint: Yes, that's why I suggested it should be logged at warning level.

electrum: You're misrepresenting what I said.

The original problem that you were trying to solve was an application where `Bootstrap` was initialized twice. I said we could solve this trivially by adding `doNotInitializeLogging()` to the second invocation. You didn't like this solution, IIRC because the fact that the first invocation initialized logging would not be obvious.

By adding this logging at warning level, you're now forced to change the application to do what I originally suggested and add `doNotInitializeLogging()` to the second invocation to avoid the warning. At that point, this change is no longer necessary, although it's still good to have as a safety check.

electrum: I'll copy the linked comment here:

> I don't think we want to log a warning when the configuration is the same. It would be better to silently ignore it. Otherwise, this is effectively making it a mistake to initialize logging twice, with the result being that we should change any such application code to not initialize the second time. But if we were to do that, we wouldn't need this change at all (except to catch mistakes).

elecnix: Sorry, not trying to misrepresent anything you said, only what I understood of it. Thanks for clarifying!

I've updated the PR to log a warning.

Using `doNotInitializeLogging()` to suppress the warning and work around this "leaky abstraction" seems to be the best option users have to acknowledge that they known what they are doing.

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -238,5 +245,6 @@ public void configure(LoggingConfiguration config)
                 throw new UncheckedIOException(e);
             }
         }
+        configured = true;
```
martint, line 248: Move this to right after the if block at the beginning of the method. If something fails with an exception, we still need to flag that the configure() method has been called.

electrum (approved):

mattstep (approved):

PR #1020, opened 2022-11-23
Add exponential retries to SocketMessageOutput

vinay-kosaraju:
SocketMessageOutput currently retries messages 5 times before dropping records. This adds exponential backoff between retries and helps in minimizing the dropped log records when the TCP listener is down

### DIFF
--- a/log-manager/src/main/java/io/airlift/log/BufferedHandler.java
+++ b/log-manager/src/main/java/io/airlift/log/BufferedHandler.java
@@ -4,10 +4,9 @@ import com.google.common.annotations.VisibleForTesting;
 import org.weakref.jmx.Managed;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.ErrorManager;
@@ -69,6 +68,9 @@ class BufferedHandler
         }
 
         try {
+            if (queue.remainingCapacity() == 0) {
+                queue.remove();
+            }
             putUninterruptibly(queue, message);
         }
         catch (Exception e) {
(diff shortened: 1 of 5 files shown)


log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -143,43 +146,25 @@ private void logging()
...
+        while (!closed.get()) {
+            try {
+                message = queue.poll(100, TimeUnit.MILLISECONDS);
             }
-
-            int poisonMessageIndex = getPoisonMessageIndex(batch);
-            if (poisonMessageIndex >= 0) {
-                poisonMessageSeen = true;
-                batch = batch.subList(0, poisonMessageIndex);
+            catch (InterruptedException ignored) {
```
mattstep, line 154: You'll want to call `Thread.currentThread().interrupt()` at least here.

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -143,43 +146,25 @@ private void logging()
...
-                }
-                catch (InterruptedException ignored) {
-                }
-            }
-            else {
-                queue.drainTo(batch, MAX_BATCH_COUNT);
+        byte[] message = new byte[0];
+        while (!closed.get()) {
+            try {
+                message = queue.poll(100, TimeUnit.MILLISECONDS);
```
mattstep, line 152: This is all in-memory, so we could probably just do 1ms

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -143,43 +146,25 @@ private void logging()
...
-
-    private synchronized void logMessageBatch(List<byte[]> batch)
-    {
-        for (byte[] message : batch) {
-            try {
-                messageOutput.writeMessage(message);
-            }
-            catch (Exception e) {
-                reportError(null, e, WRITE_FAILURE);
+            if (message != POISON_MESSAGE && message != null) {
```
mattstep, line 157: It seems like there is a difference here where before we were stopping when we saw the poison message, this doesn't seem to stop, just skip when it encounters the poison message..

PR #1021, opened 2022-11-27
Util to bind polymorphic subtypes

Randgalt:
Proposal for a JSON subtype/polymorphic binder. wdyt?

So-called polymorphic subtypes are cumbersome in Jackson. This binding utility makes it easier by:

- Moving all meta-code into the binder so that no class/record decorations/annotations are needed - de-uglifies these classes
- Automatically adds the type property to the generated JSON
- Automatically generate bindings for sealed classes/hierarchies
- Eliminates potential programmer errors by automating naming and mapping of the type field in Jackson/JSON sub-types

### DIFF
--- a/json/src/main/java/io/airlift/json/JsonSubType.java
+++ b/json/src/main/java/io/airlift/json/JsonSubType.java
@@ -1,0 +1,246 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.json;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.TreeNode;
+import com.fasterxml.jackson.core.Version;
+import com.fasterxml.jackson.databind.AnnotationIntrospector;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.Module;
+import com.fasterxml.jackson.databind.PropertyName;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.cfg.MapperConfig;
+import com.fasterxml.jackson.databind.introspect.AnnotatedClass;
+import com.fasterxml.jackson.databind.introspect.AnnotatedMember;
+import com.fasterxml.jackson.databind.introspect.VirtualAnnotatedMember;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.databind.node.ValueNode;
+import com.fasterxml.jackson.databind.ser.BeanPropertyWriter;
+import com.fasterxml.jackson.databind.ser.impl.AttributePropertyWriter;
+import com.fasterxml.jackson.databind.util.SimpleBeanPropertyDefinition;
+import com.google.common.base.Suppliers;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static java.util.Objects.requireNonNull;
+
+public class JsonSubType
+{
+    private final Set<Module> modules;
+
+    public static JsonSubType.Builder builder()
+    {
+        return new Builder();
+    }
+
+    public interface SubTypeSubBuilder<B>
+    {
+        default <T extends B> SubTypeSubBuilder<B> add(Class<T> subClass)
+        {
+            return add(subClass, subClass.getSimpleName());
+        }
+
+        <T extends B> SubTypeSubBuilder<B> add(Class<T> subClass, String propertyValue);
+
+        default SubTypeSubBuilder<B> addPermittedSubClasses()
+        {
+            return addPermittedSubClasses(Class::getSimpleName);
+        }
+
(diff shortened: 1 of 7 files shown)


martint:
One downside is that this forces you to use guice for constructing the json codecs. How would you do this when guice is not available?

Randgalt:
> One downside is that this forces you to use guice for constructing the json codecs. How would you do this when guice is not available?

I could make Jackson classes public and provide a way to use them directly

jklamer (commented):
Super impressive as always, especially with all the options. I would love to use this. I was wondering what you think about creating some interesting language interoperability with this?
These sealed interface types I believe are a way of creating a tagged union type (aka enum in Rust ). The defacto standard I've encountered in Rust for their JSON representation is to have the property key signify the type. Such that Programmer looks like 
```
{
    "Programmer":{
        "name":"Jack"
    }
}
```
and manager:
```
{
    "Manager":{
        "name":"Jack",
        "reports":[]
    }
}
```
The advantages of this representation being there is no need to worry about weird property conflict edge cases with the metadata field embedded into the class and we'd gain a way to map 1-1 with models in other languages. A downside being we'd lose  the ability to encode something as Employee and read as Manager (while using "Ignore unknown field property" ofc). So this wouldn't necessarily be "polymorphic" as I understand the word.

Randgalt:
> I would love to use this. I was wondering what you think about creating some interesting language interoperability with this? These sealed interface types I believe are a way of creating a tagged union type (aka enum in Rust, and others ). The defacto standard I've encountered in Rust for their JSON representation is to have the property key signify the type. Such that Programmer looks like
> ... snip ...

Sorry it took so long to respond to this...

I don't think Jackson would support the extra level - at least I don't know how to do that in Jackson.

Randgalt:
cc @martint and @jklamer 

I reworked this so that there is a `JsonSubType` builder that builds a new entity, `JsonSubType` that is independent of Guice. There is a new `JsonSubTypeBinder` that can bind a `JsonSubType` for Guice. Otherwise, `JsonSubType` can be used with `ObjectMapperProvider` without requiring Guice.

json/src/main/java/io/airlift/json/JsonSubType.java
```diff
@@ -0,0 +1,231 @@
...
+        @Override
+        public B deserialize(JsonParser p, DeserializationContext ctxt)
+                throws IOException
+        {
+            TreeNode treeNode = p.readValueAsTree();
+            TreeNode propertyNode = treeNode.get(propertyName);
+
+            String propertyValue = ((propertyNode != null) && propertyNode.isValueNode()) ? ((ValueNode) propertyNode).textValue() : null;
+            if (propertyValue == null) {
+                throw new IllegalArgumentException("JSON expected to have a property named %s. Double check the addBinding() or addPermittedSubClassBindings()".formatted(propertyName));
```
electrum, line 219: Add quotes around `%s`

json/src/main/java/io/airlift/json/JsonSubType.java
```diff
@@ -0,0 +1,231 @@
...
+            }
+
+            Class<?> subClass = subClassPropertyValues.get().get(propertyValue);
+            if (subClass == null) {
+                throw new IllegalArgumentException("No binding was made for property name %s and value %s. Double check the addBinding() or addPermittedSubClassBindings().".formatted(propertyName, propertyValue));
```
electrum, line 224: Same, add quotes around values

json/src/main/java/io/airlift/json/JsonSubType.java
```diff
@@ -0,0 +1,231 @@
...
+            }
+
+            //noinspection unchecked
```
electrum, line 227: I don't think this suppression works for `javac`. Probably need a helper method for this.

json/src/main/java/io/airlift/json/JsonSubTypeBinder.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2010 Proofpoint, Inc.
```
electrum, line 2: Remove, this is new code

json/src/main/java/io/airlift/json/JsonSubTypeBinder.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public void bindJsonSubType(JsonSubType jsonSubType)
+    {
+        subTypeBinder.addBinding().toInstance(jsonSubType);
+    }
+
+    private JsonSubTypeBinder(Binder binder)
+    {
+        subTypeBinder = Multibinder.newSetBinder(binder, JsonSubType.class);
```
electrum, line 37: Nit: static import `newSetBinder`

json/src/test/java/io/airlift/json/Employee.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2010 Proofpoint, Inc.
```
electrum, line 2: Remove all of these

json/src/test/java/io/airlift/json/TestJsonSubType.java
```diff
@@ -0,0 +1,196 @@
...
+        String item1Json = objectMapper.writeValueAsString(item1);
+        String item2Json = objectMapper.writeValueAsString(item2);
+        String containerJson = objectMapper.writeValueAsString(container);
+
+        for (int i = 0; i < 2; ++i) {
+            Part deserializedItem1 = (i == 0) ? objectMapper.readValue(item1Json, Part.class) : jsonCodec.fromJson(item1Json);
+            Part deserializedItem2 = (i == 0) ? objectMapper.readValue(item2Json, Part.class) : jsonCodec.fromJson(item2Json);
+            Part deserializedContainer = (i == 0) ? objectMapper.readValue(containerJson, Part.class) : jsonCodec.fromJson(containerJson);
+
+            assertTrue(deserializedItem1 instanceof Item);
```
electrum, line 154: Don't use `assertTrue` with assertion expressions. Instead, use AssertJ:
```java
assertThat(deserializedItem1).isInstanceOf(Item.class);

File: json/src/main/java/io/airlift/json/JsonSubType.java [hunk 8]
```diff
@@ -0,0 +1,231 @@
...
+        {
+            this.propertyName = requireNonNull(propertyName, "propertyName is null");
+            this.subClassPropertyValues = ImmutableMap.copyOf(subClassPropertyValues);
+        }
+
+        @Override
+        public void findAndAddVirtualProperties(MapperConfig<?> config, AnnotatedClass ac, List<BeanPropertyWriter> properties)
+        {
+            String propertyValue = subClassPropertyValues.get(ac.getRawType());
+            if (propertyValue != null) {
```
electrum, line 171: I'd return early here

electrum (commented):
I like this.

The test coverage is good, but since it's fully automatic, we're not verifying what the JSON actually looks like (only that it round trips). We should add the expected JSON output.

Also, it would be good to have a standalone test that doesn't use Guice, to show how the APIs work and make sure Guice isn't required.

Randgalt:
@electrum all comments addressed

electrum (approved):

PR #1024, opened 2022-12-16
Make CpuTimer user time collection optional

pettyjamesm:
Adds a constructor overload to `CpuTimer` to disable CPU user time collection. On Linux, user CPU time reporting [requires reading and parsing `/proc/self/task/<thread id>/stat`](https://github.com/openjdk/jdk/blob/df6cf1e41d0fc2dd5f5c094f66c7c8969cf5548d/src/hotspot/os/linux/os_linux.cpp#L5032,L5074) on each call, whereas collecting system and user time combined [has a fast-path](https://github.com/openjdk/jdk/blob/df6cf1e41d0fc2dd5f5c094f66c7c8969cf5548d/src/hotspot/os/linux/os_linux.cpp#L5023,L5029). Collecting user time therefor adds significant overhead if the caller does not care about the breakout of system vs user time.

This change adds handling to detect when `ThreadMXBean` does not support collecting user CPU time and returns `CpuDuration` values that allow the caller to detect when there is no user cpu time value present via `CpuDuration#hasUser()`, although `getUser()` will still return zero nanoseconds to preserve compatibility in that situation.

Also adds support for taking a custom `Ticker` argument to `CpuTimer` if the caller would like to be able to control the wall time collection source.

### DIFF
--- a/stats/src/main/java/io/airlift/stats/CpuTimer.java
+++ b/stats/src/main/java/io/airlift/stats/CpuTimer.java
@@ -13,6 +13,7 @@
  */
 package io.airlift.stats;
 
+import com.google.common.base.Ticker;
 import io.airlift.units.Duration;
 
 import javax.annotation.Nullable;
@@ -22,6 +23,7 @@ import java.lang.management.ThreadMXBean;
 import java.util.Optional;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
+import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 
 public class CpuTimer
@@ -29,6 +31,7 @@ public class CpuTimer
     private static final ThreadMXBean THREAD_MX_BEAN = ManagementFactory.getThreadMXBean();
     private static final Duration ZERO_NANOS = new Duration(0, NANOSECONDS);
 
+    private final Ticker ticker;
     private final long wallStartTime;
     private final long cpuStartTime;
     private final long userStartTime;
@@ -39,12 +42,18 @@ public class CpuTimer
 
     public CpuTimer()
     {
-        this(true);
+        this(Ticker.systemTicker(), true);
     }
 
     public CpuTimer(boolean collectUserTime)
     {
-        wallStartTime = System.nanoTime();
+        this(Ticker.systemTicker(), collectUserTime);
+    }
+
+    public CpuTimer(Ticker ticker, boolean collectUserTime)
+    {
+        this.ticker = requireNonNull(ticker, "ticker is null");
+        wallStartTime = ticker.read();
         cpuStartTime = THREAD_MX_BEAN.getCurrentThreadCpuTime();
         // ThreadMXBean will return -1 if user CPU time collection is not supported
         userStartTime = collectUserTime ? THREAD_MX_BEAN.getCurrentThreadUserTime() : -1;
@@ -56,7 +65,7 @@ public class CpuTimer
 
     public CpuDuration startNewInterval()
     {
-        long currentWallTime = System.nanoTime();
+        long currentWallTime = ticker.read();
         long currentCpuTime = THREAD_MX_BEAN.getCurrentThreadCpuTime();
         // ThreadMXBean will return -1 if user CPU time collection is not supported
         long currentUserTime = intervalUserStart == -1 ? -1 : THREAD_MX_BEAN.getCurrentThreadUserTime();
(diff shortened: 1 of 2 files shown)


pettyjamesm:
Related finding: https://github.com/prestodb/presto/pull/17649

stats/src/test/java/io/airlift/stats/TestCpuTimer.java
```diff
@@ -0,0 +1,50 @@
...
+    {
+        CpuTimer timer = new CpuTimer(false);
+        assertFalse(timer.elapsedTime().hasUser());
+        assertFalse(timer.startNewInterval().hasUser());
+        assertFalse(timer.elapsedIntervalTime().hasUser());
+        assertFalse(timer.elapsedTime().add(new CpuTimer.CpuDuration()).hasUser());
+        assertFalse(timer.elapsedTime().subtract(new CpuTimer.CpuDuration()).hasUser());
+    }
+
+    @Test(expectedExceptions = NullPointerException.class, expectedExceptionsMessageRegExp = "ticker is null")
```
martint, line 36: User Assertj's `assertThatThrownBy()`

stats/src/main/java/io/airlift/stats/CpuTimer.java
```diff
@@ -66,24 +81,28 @@ public CpuDuration elapsedIntervalTime()
...
+                currentUserTime == -1 ? null : nanosBetween(intervalUserStart, currentUserTime));
     }
 
     public CpuDuration elapsedTime()
     {
         long currentWallTime = System.nanoTime();
         long currentCpuTime = THREAD_MX_BEAN.getCurrentThreadCpuTime();
-        long currentUserTime = THREAD_MX_BEAN.getCurrentThreadUserTime();
+        // ThreadMXBean will return -1 if user CPU time collection is not supported
+        long currentUserTime = (collectUserTime && userStartTime != -1) ? THREAD_MX_BEAN.getCurrentThreadUserTime() : -1;
```
martint, line 108: No need to check `userStartTime != -1` ?

stats/src/main/java/io/airlift/stats/CpuTimer.java
```diff
@@ -66,24 +81,28 @@ public CpuDuration elapsedIntervalTime()
     {
         long currentWallTime = System.nanoTime();
         long currentCpuTime = THREAD_MX_BEAN.getCurrentThreadCpuTime();
-        long currentUserTime = THREAD_MX_BEAN.getCurrentThreadUserTime();
+        // ThreadMXBean will return -1 if user CPU time collection is not supported
+        long currentUserTime = (collectUserTime && intervalUserStart != -1) ? THREAD_MX_BEAN.getCurrentThreadUserTime() : -1;
```
martint, line 94: No need to check `intervalUserStart != -1` ?

stats/src/main/java/io/airlift/stats/CpuTimer.java
```diff
@@ -48,12 +61,14 @@ public CpuDuration startNewInterval()
...
```
martint, line 74: No need to check `intervalUserStart != -1` ?

stats/src/main/java/io/airlift/stats/CpuTimer.java
pettyjamesm: When `collectUserTime == true` but `ThreadMXBean.getCurrentThreadUserTime()` has already returned `-1` (not supported) on the initial attempt, we know that we can't produce a user time metric anyway and will skip the call. If for whatever reason the return value is not stable, we don't want to attempt to subtract an error code from whatever this call returns to produce a user time metric anyway.

stats/src/main/java/io/airlift/stats/CpuTimer.java
pettyjamesm: See explanation in https://github.com/airlift/airlift/pull/1024#discussion_r1051025454

stats/src/main/java/io/airlift/stats/CpuTimer.java
pettyjamesm: See explanation in https://github.com/airlift/airlift/pull/1024#discussion_r1051025454

stats/src/test/java/io/airlift/stats/TestCpuTimer.java
pettyjamesm: Done

stats/src/main/java/io/airlift/stats/CpuTimer.java
pettyjamesm: I've decided this can be simplified to remove the `collectUserTime` boolean and to consolidate the logic to only need to test whether the user start time is `-1`.

PR #1025, opened 2022-12-19
Add public getter for CpuTimer wallStartTime

pettyjamesm:
Allows the caller to access `wallStartTime` and `intervalWallStart` values read from the Ticker which may allow the caller to avoid an additional redundant call to `Ticker#read()` at the point of construction or the initiation of a new interval.

This is an additional follow up item after https://github.com/airlift/airlift/pull/1024 to reduce overheads in Trino `PrioritizedSplitRunner` usage.


pettyjamesm:
Thanks!

PR #1032, opened 2023-01-05
Update to Jetty 10

electrum:


event-http/pom.xml
```diff
@@ -90,6 +90,12 @@
             <groupId>org.eclipse.jetty</groupId>
             <artifactId>jetty-server</artifactId>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.eclipse.jetty.toolchain</groupId>
+                    <artifactId>jetty-servlet-api</artifactId>
```
joakime, line 96: This artifact contains the servlet schemas, the official artifact does not.

electrum: Is this a problem for embedded use with Jersey? I was trying to minimize the changes needed for downstream projects. We'll update to Jetty 11 with the new Jakarta namespace after updating to Jetty 10.

joakime: Jersey can use our artifact without issue.

But we cannot use the standard artifact (`jakarta.servlet:jakarta.servlet.api`), without lots of XML loading/parsing/validation issues on the server level. (it's a valid artifact for web applications, not servers)
This was fixed in Servlet 6.0 (ee10, Jetty 12) the official artifact now has the schemas (and a valid JPMS module definition too).

joakime: Also, the use of the standard artifact can (and does) result in your XML library hitting various external servers to obtain the XML dtds/xsds when the webapps are deployed. (even on embedded use).
This is because the XML schemas are not present in that jar and will result in the XML library needing to go fetch them from remote servers (if the public ids are uris, which not all references are).

electrum: Is it a problem if we don't use webapps? I assume this is a new thing in Jetty 10+?

joakime: If you use Jersey, you are technically using the Servlet API.

The "webapp" in this sense is merely the use of the Servlet API.
While you don't have to worry about all of the machinery of a war file deployment using Jetty embedded, you are not immune from the effects of the Servlet API and running a webapp.

This is the list of schemas we are talking about ...

```
javax/servlet/resources/
javax/servlet/resources/datatypes.dtd
javax/servlet/resources/j2ee_1_4.xsd
javax/servlet/resources/j2ee_web_services_1_1.xsd
javax/servlet/resources/j2ee_web_services_client_1_1.xsd
javax/servlet/resources/javaee_5.xsd
javax/servlet/resources/javaee_6.xsd
javax/servlet/resources/javaee_7.xsd
javax/servlet/resources/javaee_8.xsd
javax/servlet/resources/javaee_web_services_1_2.xsd
javax/servlet/resources/javaee_web_services_1_3.xsd
javax/servlet/resources/javaee_web_services_1_4.xsd
javax/servlet/resources/javaee_web_services_client_1_2.xsd
javax/servlet/resources/javaee_web_services_client_1_3.xsd
javax/servlet/resources/javaee_web_services_client_1_4.xsd
javax/servlet/resources/permissions_7.xsd
javax/servlet/resources/web-app_2_2.dtd
javax/servlet/resources/web-app_2_3.dtd
javax/servlet/resources/web-app_2_4.xsd
javax/servlet/resources/web-app_2_5.xsd
javax/servlet/resources/web-app_3_0.xsd
javax/servlet/resources/web-app_3_1.xsd
javax/servlet/resources/web-app_4_0.xsd
javax/servlet/resources/web-common_3_0.xsd
javax/servlet/resources/web-common_3_1.xsd
javax/servlet/resources/web-common_4_0.xsd
javax/servlet/resources/web-fragment_3_0.xsd
javax/servlet/resources/web-fragment_3_1.xsd
javax/servlet/resources/web-fragment_4_0.xsd
javax/servlet/resources/XMLSchema.dtd
javax/servlet/resources/xml.xsd
```

This contains not only war deployment schemas, but also schemas for standard XML stuff (`xml.xsd`, `XMLSchema.dtd`, `datatypes.dtd`) and things related to web services (which has nothing to do with war deployment)

electrum: Are you saying that Jetty 10 will "result in your XML library hitting various external servers"? Is this a new behavior that is not present in Jetty 9?

I can run `io.airlift.sample.TestServer` which starts Jetty with Jersey and exercises the various JAX-RS endpoints. It works while offline, so I'm wondering how this is a problem in practice.

Do we absolutely need to update downstream projects to the new servlet API JAR when we update to Jetty 10, or is that something that can wait?

joakime: If you are building a server, the server dependencies should be evaluated for their version is your new "minimum version" to use.

In other words ...

Scenario 1: 
Jetty 10.0.13, uses jetty-servlet-api 4.0.6.
Jersey uses Jakarta Servlet API 4.0.2.

You MUST use jetty-servlet-api 4.0.6.

Scenario 2:
Jetty 10.0.14 uses jetty-servlet-api 4.0.6
Jersey uses Jakarta Servlet API 4.0.8

Then you are free to use 4.0.6 and have no demand to upgrade to 4.0.8.

Another thing that will make you upgrade is bugs in the API (yep, this happens).
Example: Jakarta Servlet API 4.0.2 has [a bad ServletOutputStream bug](https://github.com/jakartaee/servlet/issues/228) which was fixed in Jakarta Servlet API 4.0.3.

electrum: Thanks for explaining. I'll update everything: https://github.com/airlift/airbase/pull/334

joakime:
Please fix the GzipHandler usage in your tests as well.

``` patch
Index: http-client/src/test/java/io/airlift/http/client/TestingHttpServer.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/http-client/src/test/java/io/airlift/http/client/TestingHttpServer.java b/http-client/src/test/java/io/airlift/http/client/TestingHttpServer.java
--- a/http-client/src/test/java/io/airlift/http/client/TestingHttpServer.java	(revision 43a92d75aa0e880897f2fe161dff75a43ae36763)
+++ b/http-client/src/test/java/io/airlift/http/client/TestingHttpServer.java	(date 1674055431174)
@@ -85,14 +85,15 @@
 
         ServletHolder servletHolder = new ServletHolder(servlet);
         ServletContextHandler context = new ServletContextHandler(ServletContextHandler.NO_SESSIONS);
-        context.setGzipHandler(new GzipHandler());
         context.addServlet(servletHolder, "/*");
 
         HandlerCollection handlers = new HandlerCollection();
         additionalHandle.ifPresent(handlers::addHandler);
         handlers.addHandler(context);
 
-        server.setHandler(handlers);
+        GzipHandler gzipHandler = new GzipHandler();
+        gzipHandler.setHandler(handlers);
+        server.setHandler(gzipHandler);
 
         this.server = server;
         this.server.start();
```

http-client/src/test/java/io/airlift/http/client/TestingHttpServer.java
```diff
@@ -85,7 +85,7 @@ public TestingHttpServer(Optional<String> keystore, Servlet servlet, Consumer<Ht
 
         ServletHolder servletHolder = new ServletHolder(servlet);
         ServletContextHandler context = new ServletContextHandler(ServletContextHandler.NO_SESSIONS);
-        context.setGzipHandler(new GzipHandler());
+        context.insertHandler(new GzipHandler());
```
joakime, line 88: It would be better to put `GzipHandler` outside of the `ServletContextHandler`.

In other words, the `GzipHandler` wraps `ContextHandlerCollection`.

joakime:
The Request body content layer has changed in Jetty 10.

The `ContentProvider` API is deprecated.

which means things like this in `JettyHttpClient` are deprecated now.

``` java
jettyRequest.content(new BodyGeneratorContentProvider(bodyGenerator, httpClient.getExecutor()));
```

Use the new `Request.Content` API, and reimplement these content providers to use this new API.
This new API supports the subscribe model that the Jetty Client is built on now, as well as supporting a flag for "reproducible", meaning, can the content be used again for a followup request (like when following a redirect).

Instead of `BodyGenerator` consider using `org.eclipse.jetty.client.util.OutputStreamRequestContent` and `org.eclipse.jetty.client.util.PathRequestContent` instead.

joakime:
Be careful of the mutability of the HttpFields object as well.
It's not always mutable.

Here's a fix ...

``` patch
Index: http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java b/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java	(revision 43a92d75aa0e880897f2fe161dff75a43ae36763)
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java	(date 1674056767095)
@@ -700,9 +700,11 @@
 
         jettyRequest.method(finalRequest.getMethod());
 
-        for (Entry<String, String> entry : finalRequest.getHeaders().entries()) {
-            jettyRequest.header(entry.getKey(), entry.getValue());
-        }
+        jettyRequest.headers(headers -> {
+            for (Entry<String, String> entry : finalRequest.getHeaders().entries()) {
+                headers.add(entry.getKey(), entry.getValue());
+            }
+        });
 
         BodyGenerator bodyGenerator = finalRequest.getBodyGenerator();
         if (bodyGenerator != null) {
Index: http-client/src/main/java/io/airlift/http/client/jetty/AuthorizationPreservingHttpClient.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/http-client/src/main/java/io/airlift/http/client/jetty/AuthorizationPreservingHttpClient.java b/http-client/src/main/java/io/airlift/http/client/jetty/AuthorizationPreservingHttpClient.java
--- a/http-client/src/main/java/io/airlift/http/client/jetty/AuthorizationPreservingHttpClient.java	(revision 43a92d75aa0e880897f2fe161dff75a43ae36763)
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/AuthorizationPreservingHttpClient.java	(date 1674056177158)
@@ -28,10 +28,9 @@
 
         if (isPreserveAuthorization(oldRequest)) {
             setPreserveAuthorization(newRequest, true);
-            for (HttpField field : oldRequest.getHeaders()) {
-                if (field.getHeader() == AUTHORIZATION) {
-                    newRequest.header(field.getName(), field.getValue());
-                }
+            HttpField httpField = oldRequest.getHeaders().getField(AUTHORIZATION);
+            if (httpField != null) {
+                newRequest.headers(headers -> headers.add(httpField));
             }
         }
```

joakime:
There's a javadoc comment referencing a deprecated/removed constructor as well.

``` patch
Index: http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java b/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java	(revision 43a92d75aa0e880897f2fe161dff75a43ae36763)
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java	(date 1674055919269)
@@ -86,7 +86,7 @@
     /**
      * This property is initialized with Jetty's default excluded ciphers list.
      *
-     * @see org.eclipse.jetty.util.ssl.SslContextFactory#SslContextFactory(boolean, String)
+     * @see org.eclipse.jetty.util.ssl.SslContextFactory#SslContextFactory(boolean)
      */
     private List<String> excludedCipherSuites = ImmutableList.of("^.*_(MD5|SHA|SHA1)$", "^TLS_RSA_.*$", "^SSL_.*$", "^.*_NULL_.*$", "^.*_anon_.*$");
 
```

http-client/src/test/java/io/airlift/http/client/TestingHttpServer.java
electrum: Thanks, updated. Should we make a similar change in `HttpServer`?

joakime: I didn't notice the behavior in HttpServer, but yes, it should change to match this pattern.

electrum:
Thanks for the review.

Note that the change to `AuthorizationPreservingHttpClient` needs to use `getFields()` since we can have multiple authorization header values.

For `BodyGenerator`, I think we need to deprecate that, since using `OutputStream` is problematic and not actually needed. If we ever need to produce large content in a streaming way, we'll come up with a new API that fits with Jetty's model.

joakime:
> Note that the change to `AuthorizationPreservingHttpClient` needs to use `getFields()` since we can have multiple authorization header values.

OK, But don't add/update/change an `HttpFields` directly though via the (deprecated) `request.header(String,String)` method, do it through the `request.headers(Consumer<HttpFields>)` method.
This ensures you'll have the right headers in all scenarios.

electrum:
We [deprecated custom `BodyGenerator` implementations](https://github.com/airlift/airlift/pull/1036) and switched to eagerly generate the full body, so now we only use the standard `StaticBodyGenerator` and `FileBodyGenerator` implementations.

dain (approved):
looks good

PR #1035, opened 2023-01-11
New non-blocking buffered handler

erichwang:
In the previous implementation of BufferedHandler, the log buffer would
cause calling threads to hang if the buffer ever became full. Instead of
hanging the calling threads, we can instead retain a summary of dropped
messages that couldn't fit in the buffer. This dropped summary can then
be emitted once the underlying logging system catches up again.

Note: we need to modify SocketMessageOutput here since previously it
would throw an IOException even if it was successful in it's retries.
Now, it only throws if it fails.

### DIFF
--- a/log-manager/src/main/java/io/airlift/log/BufferedHandler.java
+++ b/log-manager/src/main/java/io/airlift/log/BufferedHandler.java
@@ -1,114 +1,407 @@
 package io.airlift.log;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.HashMultiset;
+import com.google.common.collect.ImmutableMultiset;
+import com.google.common.collect.Multiset;
+import com.google.common.util.concurrent.RateLimiter;
 import org.weakref.jmx.Managed;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.BlockingQueue;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
+import javax.annotation.concurrent.ThreadSafe;
+
+import java.time.Duration;
+import java.util.Deque;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.ErrorManager;
 import java.util.logging.Formatter;
 import java.util.logging.Handler;
+import java.util.logging.Level;
 import java.util.logging.LogRecord;
 
+import static com.google.common.base.MoreObjects.firstNonNull;
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+import static io.airlift.concurrent.Threads.daemonThreadsNamed;
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Comparator.comparing;
 import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.Executors.newSingleThreadExecutor;
 import static java.util.logging.ErrorManager.CLOSE_FAILURE;
 import static java.util.logging.ErrorManager.FLUSH_FAILURE;
 import static java.util.logging.ErrorManager.FORMAT_FAILURE;
+import static java.util.logging.ErrorManager.GENERIC_FAILURE;
 import static java.util.logging.ErrorManager.WRITE_FAILURE;
+import static java.util.stream.Collectors.joining;
 
+@ThreadSafe
 class BufferedHandler
         extends Handler
 {
-    private static final int MAX_BATCH_COUNT = 1024;
-    private static final byte[] POISON_MESSAGE = new byte[0];
+    public interface DropSummaryFormatter
+    {
+        String formatDropSummary(Multiset<String> dropCountBySource);
+    }
+
+    private static final MessageAndSource TERMINAL_MESSAGE = new MessageAndSource(new byte[] {}, "");
+
+    private final ExecutorService bufferDrainExecutor = newSingleThreadExecutor(daemonThreadsNamed("log-buffer-drainer"));
+    private final AtomicBoolean initialized = new AtomicBoolean();
+    private final AtomicBoolean inputClosed = new AtomicBoolean();
+    private final AtomicLong droppedMessages = new AtomicLong();
 
     private final MessageOutput messageOutput;
-    private final Thread thread;
-    private final BlockingQueue<byte[]> queue = new ArrayBlockingQueue<>(MAX_BATCH_COUNT);
-    private final AtomicBoolean closed = new AtomicBoolean();
-    private final AtomicLong droppedMessages = new AtomicLong(0);
+    private final DropSummaryFormatter dropSummaryFormatter;
+    private final RateLimiter errorRetryLimiter;
+    private final Duration maxCloseTime;
+    private final int messageFlushCount;
+
+    // queueDrainLock ensures that queue and dropCountBySource are kept in sync as data is extracted or moved between these structures
+    private final ReentrantLock queueDrainLock = new ReentrantLock();
+    private final Condition recordEnqueued = queueDrainLock.newCondition();
+    private final Deque<MessageAndSource> queue;
+    @GuardedBy("queueDrainLock")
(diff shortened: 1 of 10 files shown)


erichwang:
Updated to do the string formatting on the publishing thread

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -1,219 +1,399 @@
...
-        thread.start();
+        while (!queue.offerLast(message)) {
+            synchronized (queueDrainLock) {
+                MessageAndSource dropCandidate = queue.peekFirst();
+                if (dropCandidate == null) {
+                    continue;
+                }
+                if (dropCandidate == TERMINAL_MESSAGE) {
+                    // Queue already terminated
+                    return;
```
dain, line 293: shouldn't this increment `droppedMessages` like the old code did?  If it too annoying, to count messages after close perfectly, just add a comment to the field

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -1,219 +1,399 @@
...
+                }
+                verify(dropCandidate == queue.pollFirst());
+                droppedMessagesBySource.add(dropCandidate.sourceName());
+                droppedMessages.incrementAndGet();
+            }
+        }
+
+        // Notify drainer about new message
+        synchronized (queueDrainLock) {
```
dain, line 302: Is the contention between writers a problem?

log-manager/src/test/java/io/airlift/log/TestBufferedHandler.java
```diff
@@ -0,0 +1,449 @@
...
+            }
+        });
+        executor.execute(() -> {
+            for (int i = 0; i < 1000; i++) {
+                bufferedHandler.publish(logRecord(INFO, "B-TestLogger", String.valueOf(i)));
+            }
+        });
+
+        // Short period where writes and flushes will fail, before allowing the system to recover
+        sleepUninterruptibly(100, TimeUnit.MILLISECONDS);
```
dain, line 204: maybe use a countdown latch

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -237,5 +263,38 @@ public synchronized void configure(LoggingConfiguration config)
                 throw new UncheckedIOException(e);
             }
         }
+
+        exportJmxOnConfiguration(jmxExporterFuture, jmxExportCollector);
+    }
+
+    private static void exportJmxOnConfiguration(ListenableFuture<MBeanExporter> jmxExporterFuture, List<JmxExport> jmxExports)
```
dain, line 270: I'm pretty sure we need a way to unexport

log-manager/src/main/java/io/airlift/log/LogJmxModule.java
```diff
@@ -34,12 +36,28 @@ public void configure(Binder binder)
...
+        public LoggingProvider(MBeanExporter exporter)
+        {
+            this.exporter = requireNonNull(exporter, "exporter is null");
+        }
+
+        @Override
+        public Logging get()
+        {
+            Logging logging = Logging.initialize();
+            logging.registerJmxExports(exporter);
```
dain, line 59: I'm pretty sure we need to unexport when the system is shutdown.  Todo this, I suggest reverting this code back to a signleton provider method and adding this:
```java
@Override
{
    binder.disableCircularProxies();

    binder.bind(LoggingMBean.class).in(Scopes.SINGLETON);
    newExporter(binder).export(LoggingMBean.class).as("io.airlift.log:name=Logging");
    binder.bind(LogExporter.class).asEagerSingleton();
}

@Provides
@Singleton
public Logging getLogging()
{
    return Logging.initialize();
}

public static class LogExporter
{

    private final Logging logging;
    private final MBeanExporter exporter;

    @Inject
    public LogExporter(Logging logging, MBeanExporter exporter)
    {
        this.logging = requireNonNull(logging, "logging is null");
        this.exporter = requireNonNull(exporter, "exporter is null");
        logging.export(exporter);
    }

    @PreDestroy
    public void destroy()
    {
        logging.unexport(exporter);
    }
}
```

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -95,6 +104,11 @@ private Logging()
         redirectStdStreams();
     }
 
+    public void registerJmxExports(MBeanExporter exporter)
```
dain, line 107: add a comment stating that logging typically is initialized before JMX so export must be lazy

log-manager/src/main/java/io/airlift/log/LogJmxModule.java
```diff
@@ -34,12 +36,28 @@ public void configure(Binder binder)
...
         newExporter(binder).export(LoggingMBean.class).as("io.airlift.log:name=Logging");
+
+        // Eager to ensure JMX registration
+        binder.bind(Logging.class).toProvider(LoggingProvider.class).asEagerSingleton();
     }
 
-    @Provides
-    @Singleton
-    public Logging getLogging()
+    public static class LoggingProvider
```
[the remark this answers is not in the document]
erichwang: The `@Singleton` here is not eager. I couldn't find a way to force an eager singleton using a provider method.

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -1,219 +1,399 @@
...
+
+                            queueDrainLock.wait();
+                        }
+                    }
+
+                    // Dropped messages occurred before the current, and so drop snapshot need to be written first
+                    if (!droppedSnapshot.isEmpty()) {
+                        if (!writeMessageOutputSafe(formatMessageBytes(createDropSummaryRecord(droppedSnapshot)))) {
+                            errorRetryLimiter.acquire();
+                            continue;
```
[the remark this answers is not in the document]
erichwang: Yea anything not explicitly unreferenced will be requeued, so that keeps us robust to things like errors as well.

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -1,219 +1,399 @@
...
+                    }
+                }
+                finally {
+                    // Return messages that failed
+                    synchronized (queueDrainLock) {
+                        // Return a failed message to the queue, or drop it if at capacity.
+                        if (message != null) {
+                            // If droppedMessagesBySource is nonempty, then need to drop the current message because messages after it have already been capacity dropped.
+                            // If the queue is full, then also need to drop the current message.
+                            if (!droppedMessagesBySource.isEmpty() || !queue.offerFirst(message)) {
```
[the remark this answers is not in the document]
erichwang: OH, GOOD CATCH. Yea that could potentially drop the terminal message then.

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
erichwang: Probably not a problem for contention. If you look at the implementation inside java's blocking queue, it has a lock for notification just like this as well.

log-manager/src/main/java/io/airlift/log/LogJmxModule.java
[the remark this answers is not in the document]
dain: Trino always initializes singleton... eager is for some kind of development mode that Trino doesn't use.  cc @electrum

log-manager/src/main/java/io/airlift/log/Logging.java
erichwang: Added

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
erichwang: Based on our discussion, it sounds like we won't be counting the dropped messages after close, since this is typically not what the caller cares about if they are closing the stream explicitly already.

log-manager/src/main/java/io/airlift/log/LogJmxModule.java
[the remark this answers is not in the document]
erichwang: Used the sample you provided below.

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -1,219 +1,438 @@
...
     {
-        thread.start();
+        while (!queue.offerLast(message)) {
+            synchronized (queueDrainLock) {
+                MessageAndSource toDrop = queuePollFirst();
+                if (toDrop == null) {
+                    if (terminalMessageDequeued) {
+                        return; // Queue already terminated
+                    }
+                    else {
```
dain, line 323: remove redundant else

dain (commented):
Looks good.  Only a few minor suggestions

dain (approved):

PR #1038, opened 2023-01-31
Add ByteBufferGenerator

linzebing:
Currently `generateBody` will materialize everything and wrap it with `BytesRequestContent`. This is not ideal since it will require extra copy and additional memory. In many cases we can pass an array of ByteBuffers directly. This is useful in the context of buffer service.

Tested with https://github.com/starburstdata/trino-buffer-service/pull/188 and result looks good

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/ByteBufferBodyGenerator.java
+++ b/http-client/src/main/java/io/airlift/http/client/ByteBufferBodyGenerator.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.http.client;
+
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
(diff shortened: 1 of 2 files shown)


linzebing:
cc @arhimondr @electrum

http-client/src/main/java/io/airlift/http/client/ByteBufferBodyGenerator.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public ByteBuffer[] getByteBuffers()
+    {
+        return byteBuffers;
+    }
+
+    @Override
+    public void write(OutputStream out)
+            throws Exception
```
arhimondr, line 41: nit: Drop the `Exception` from the interface if possible

arhimondr (commented):
LGTM % comments

arhimondr (approved):

electrum (approved):

http-client/src/main/java/io/airlift/http/client/ByteBufferBodyGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+
+import static java.util.Objects.requireNonNull;
+
+public class ByteBufferBodyGenerator
+        implements BodyGenerator
+{
+    public static ByteBufferBodyGenerator createByteBuffersGenerator(ByteBuffer... byteBuffers)
```
electrum, line 24: Sorry for the nit, but since this is the public API, can you rename to `createByteBufferGenerator`?

electrum: Actually, I'm thinking we should remove this and just have the constructor. I don't see any value in having this factory method. We don't have one for `FileBodyGenerator`.

electrum (approved):

PR #1039, opened 2023-02-08
Guard against collected weak references for openmetrics

mattstep:


openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -286,12 +286,17 @@ private List<Metric> getMetricsRecursively(String prefix, ManagedClass managedCl
 
     private Optional<Metric> getMetricFromTarget(ManagedClass managedClass, String metricName, String description)
     {
```
dain, line 288: I'd write this as:
```
Class<?> targetClass;
try {
    targetClass = managedClass.getTargetClass();
}
catch (IllegalStateException ignored) {
    return Optional.empty();
}

dain (approved):

Issue #1049, opened 2023-03-16
HTTP client's UnexpectedResponseException can log sensitive info 


Laonel:
`io.airlift.http.client.UnexpectedResponseException` overrides `toString()` and adds to it`io.airlift.http.client.Request` object. This logs the entire request object, including headers, when `UnexpectedResponseException` is thrown, and can potentially leak sensitive information, like authorization header value.

Is it worth limiting the data returned by the exception's `toString()` method, maybe hiding the request headers?

electrum:
Yeah, that seems like a good idea. I think the request method and URI is sufficient.

Laonel:
Thanks @electrum, I've prepared a PR with a fix https://github.com/airlift/airlift/pull/1050.

PR #1050, opened 2023-03-16
Keep only request URI and method in UnexpectedResponseException

Laonel:
It is enough for UnexpectedResponseException to mention the HTTP method and URI of the request for which it was thrown. Adding the entire request object could clutter the logs and potentially expose sensitive data contained in the request's headers.

Fixes https://github.com/airlift/airlift/issues/1049

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/UnexpectedResponseException.java
+++ b/http-client/src/main/java/io/airlift/http/client/UnexpectedResponseException.java
@@ -20,6 +20,7 @@ import com.google.common.collect.ListMultimap;
 
 import javax.annotation.Nullable;
 
+import java.net.URI;
 import java.util.List;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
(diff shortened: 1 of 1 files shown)


http-client/src/main/java/io/airlift/http/client/UnexpectedResponseException.java
```diff
@@ -81,7 +84,8 @@ public ListMultimap<HeaderName, String> getHeaders()
     public String toString()
     {
         return toStringHelper(this)
-                .add("request", request)
```
electrum, line 84: How about
```java
.add("request", request.getMethod() + " " + request.getUri())

Laonel: Yeah, that's definitely better, thanks!

Laonel:
@electrum Have you had a chance to take a look at this?

PR #1051, opened 2023-03-23
Avoid empty lines when stdout/stderr redirect to logging

findepi:
After `io.airlift.log.Logging#redirectStdStreams`, whenever application printed a line to `System.out` or `System.err`, console handler output two lines. This was because the `System.out.println` prints a newline to the output stream and `LoggingOutputStream` passed it back to logging which added another new line when printing to console.

This commit strips newline in `LoggingOutputStream` so that for a line printed via `System.out.println`, one line gets logged via console handler.

### DIFF
--- a/log-manager/src/main/java/io/airlift/log/LoggingOutputStream.java
+++ b/log-manager/src/main/java/io/airlift/log/LoggingOutputStream.java
@@ -24,13 +24,11 @@ import java.io.IOException;
 class LoggingOutputStream
         extends ByteArrayOutputStream
 {
-    private final String lineSeparator;
     private final Logger logger;
 
     public LoggingOutputStream(Logger logger)
     {
         this.logger = logger;
-        lineSeparator = System.getProperty("line.separator");
     }
 
     /**
@@ -44,7 +42,10 @@ class LoggingOutputStream
         String record = this.toString();
         reset();
 
-        if (record.isEmpty() || record.equals(lineSeparator)) {
+        // Strip trailing new line typically added when this class is used via System.out.println
+        record = record.stripTrailing();
+
+        if (record.isEmpty()) {
             // avoid empty records
             return;
         }
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,7 @@
+229
+
+- Strip trailing whitespace when redirecting standard output streams to log
+
 228
 
 - Exclude Hamcrest from jaxrs-testing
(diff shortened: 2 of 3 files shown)


log-manager/src/main/java/io/airlift/log/LoggingOutputStream.java
```diff
@@ -44,7 +44,13 @@ public synchronized void flush()
         String record = this.toString();
         reset();
 
-        if (record.isEmpty() || record.equals(lineSeparator)) {
+        // Strip trailing new line typically added when this class is used via System.out.println
+        if (record.length() >= lineSeparator.length() &&
```
electrum, line 48: I think we can eliminate `lineSeparator` and simplify this code to
```java
record = record.stripTrailing();
```

electrum (approved):
My first thought about this is the question "is this safe?" from the perspective of munging lines when we write multiple lines at once. It took me a bit to realize the simple logic here: strip the trailing newline when logging a message. The log record itself is a boundary to separate multiple lines, so there's no danger here, and no point in logging a trailing blank line.

Please update CHANGES

log-manager/src/main/java/io/airlift/log/LoggingOutputStream.java
findepi: pushed a change to strip all trailing new lines

the existing logic treats lineSeparator specially.
if we don't care about line endings, then yes, we can simplify this further

electrum: I don’t think we care about platform line endings. Stripping all leading and trailing whitespace seems fine to me. Do you see any issues with that?

electrum (approved):

mattstep (approved):

log-manager/src/main/java/io/airlift/log/LoggingOutputStream.java
findepi: i don't see a reason to strip leading whitespace and have no evidence yet that it would be useful.  it wasn't this code's behavior before my changes, so i didn't want to introduce it.

do you think that my changes require stripping leading whitespace as well?

mattstep: It can make stack traces look different to strip leading whitespace, but that doesn't bother me.

PR #1053, opened 2023-03-29
Add Scale method to QuantileDigest

skrzypo987:


martint:
What’s the use case for this?

skrzypo987:
> What’s the use case for this?

Adding a scale method to Trino. We can do it in Trino, but I thought this is the proper place

martint:
Understood, but what's the use case for having a scale method?

Praveen2112:
The purpose of the scale method is to scale up or down `QDigest` summary for a given set of data. Say if we have a summary of an order delivered from point A to point B stored in `QDigest`, the scale function allows us to extrapolate if the point A or point B diverges or converges.

Praveen2112:
@martint Do we need any additional information here ?

PR #1054, opened 2023-04-01
Remove non-primitive version of TDigest#valuesAt

raunaqmorarka:
Using this API incurrs cost of converting input list to array and output array to list.
There is no reason to use this over the more efficient primitive arrays version.

### DIFF
--- a/stats/src/main/java/io/airlift/stats/DecayTDigest.java
+++ b/stats/src/main/java/io/airlift/stats/DecayTDigest.java
@@ -16,7 +16,6 @@ package io.airlift.stats;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Ticker;
 
-import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 import static io.airlift.stats.ExponentialDecay.weight;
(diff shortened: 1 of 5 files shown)


stats/src/main/java/io/airlift/stats/DecayTDigest.java
```diff
@@ -127,11 +126,6 @@ public double valueAt(double quantile)
         return digest.valueAt(quantile);
     }
 
-    public List<Double> valuesAt(List<Double> quantiles)
```
martint, line 130: These methods serve two different purposes. Not always the caller has a fixed list of values, so a list is necessary.

martint (changes requested):

stats/src/main/java/io/airlift/stats/DecayTDigest.java
raunaqmorarka: Is there a real use case where the caller has a variable list of quantiles ? I couldn't find any actual usage of this particular method.
For the same method on TDigest, I found all the existing usage of it be unnecessary and am proposing to removing it in https://github.com/trinodb/trino/pull/16811 
The implementation of this method is just converting the list to an array and then using the primitive version of the implementation. That can be done explicitly by the caller anyway.
If there is a real need for variable input, then there should an implementation that doesn't incur the overheads of List<Double> to double[] conversion. Otherwise it's just hiding the cost of conversions from the caller and can lead to callers not thinking about this unnecessary overhead.

PR #1055, opened 2023-04-09
Add OpenTelemetry tracing

electrum:


http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -1017,9 +1023,26 @@ static void recordRequestComplete(RequestStats requestStats, Request request, lo
...
+                requestBytes,
                 response.getBytesRead(),
                 requestProcessingTime,
                 responseProcessingTime);
     }
+
+    private static class RequestSizeListener
+            implements org.eclipse.jetty.client.api.Request.ContentListener
+    {
+        private long bytes;
```
dain, line 1121: Is this the correct thread safety for a content listener callback?

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -686,6 +749,29 @@ private Request applyRequestFilters(Request request)
...
+    {
+        String method = request.getMethod().toUpperCase(ENGLISH);
+        return request.getSpanBuilder()
+                .orElseGet(() -> tracer.spanBuilder(name + " " + method))
+                .setSpanKind(SpanKind.CLIENT)
+                .setAttribute(CLIENT_NAME, name)
+                .setAttribute(SemanticAttributes.HTTP_URL, request.getUri().toString())
+                .setAttribute(SemanticAttributes.HTTP_METHOD, method)
+                .setAttribute(SemanticAttributes.NET_PEER_NAME, request.getUri().getHost())
+                .setAttribute(SemanticAttributes.NET_PEER_PORT, (long) request.getUri().getPort())
```
dain, line 762: `getPort` will return -1 if the port is not set

jaxrs/src/main/java/io/airlift/jaxrs/tracing/TracingServletFilter.java
```diff
@@ -0,0 +1,33 @@
...
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws IOException, ServletException
+    {
+        request.setAttribute(TracingFilter.REMOTE_ADDRESS, request.getRemoteAddr());
+
+        try {
+            chain.doFilter(request, response);
+        }
+        catch (Throwable t) {
+            try (var ignored = (Scope) request.getAttribute(TracingFilter.REQUEST_SCOPE)) {
```
dain, line 26: Add a comment explaining why this is not just a finally block (avoid exception getting overwritten)

jaxrs/src/main/java/io/airlift/jaxrs/tracing/TracingFilter.java
```diff
@@ -0,0 +1,149 @@
...
+        SpanBuilder spanBuilder = tracer.spanBuilder(method + " " + route)
+                .setParent(parent)
+                .setSpanKind(SpanKind.SERVER)
+                .setAttribute(SemanticAttributes.HTTP_METHOD, method)
+                .setAttribute(SemanticAttributes.HTTP_ROUTE, route)
+                .setAttribute(SemanticAttributes.HTTP_SCHEME, uri.getScheme())
+                .setAttribute(SemanticAttributes.HTTP_TARGET, target)
+                .setAttribute(SemanticAttributes.HTTP_CLIENT_IP, remoteAddress)
+                .setAttribute(SemanticAttributes.NET_HOST_NAME, uri.getHost())
+                .setAttribute(SemanticAttributes.NET_HOST_PORT, (long) uri.getPort())
```
dain, line 80: This can be -1

jaxrs/src/main/java/io/airlift/jaxrs/tracing/TracingFilter.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Override
+    public void filter(ContainerRequestContext requestContext)
+    {
+        Context parent = propagator.extract(Context.root(), requestContext, JaxrsTextMapGetter.INSTANCE);
+
+        ContainerRequest request = (ContainerRequest) requestContext.getRequest();
+
+        String method = request.getMethod().toUpperCase(ENGLISH);
+        String remoteAddress = (String) request.getProperty(REMOTE_ADDRESS);
```
dain, line 58: consider moving this down to the section with `userAgent` and making it conditional.  The remote address could be there, but might not

dain (approved):

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
```diff
@@ -0,0 +1,113 @@
...
+import static io.airlift.tracing.Tracing.attribute;
+import static java.util.Objects.requireNonNull;
+
+public class OpenTelemetryModule
+        implements Module
+{
+    private final String serviceName;
+    private final String serviceVersion;
+
+    public OpenTelemetryModule(String serviceName, String serviceVersion)
```
mattstep, line 38: Could we infer this from environment variables via launcher.py or via the jar manifest? It looks like we set some properties at build in the jar manifest like `artifactId` and `version`

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
```diff
@@ -0,0 +1,113 @@
...
+        };
+    }
+
+    private static String hostArch()
+    {
+        return switch (nullToEmpty(OS_ARCH.value())) {
+            case "amd64", "x86_64" -> "amd64";
+            case "aarch64" -> "arm64";
+            case "ppc64le" -> "ppc64";
+            default -> null;
```
mattstep, line 110: Is it safe to let this fall through and just pass the denormalized string through if it's a new arch type? I think there are some cloud vendors looking at risc-v machines.

mattstep (approved):

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
electrum: This matches the other listeners, so it should be ok.

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
electrum: We decided it's better to have people set this directly, since the name and version comes from the application.

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
electrum: According to https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/host/

> host.arch has the following list of well-known values. If one of them applies, then the respective value MUST be used, otherwise a custom value MAY be used.

In order to know if it's not a well known value like `x86`, you'd have to map all the values like `x86_32`, `i686`, etc., so I only passed the known values that we currently support in Trino. Surprisingly, they don't have a value for RISC-V yet.

PR #1057, opened 2023-04-18
Remove broken uncaught exception handler from Bootstrap

electrum:
It was only being set for the current thread. An application that wants this behavior can use `Thread.setDefaultUncaughtExceptionHandler()` which will set it for all threads.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,7 @@
+230
+
+- Remove broken uncaught exception handler from Bootstrap
+
 229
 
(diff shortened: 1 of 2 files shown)


dain (commented):
Maybe describe this as remove broken uncaught exception handler... this is only setting on thread.  If soneone wants to do they should `setDefaultUncaughtExceptionHandler`

dain (approved):

PR #1060, opened 2023-04-25
Use labels in OpenMetrics

nineinchnick:
Currently, the whole object name of an MBean is converted to a metric name, including key properties. This results in very long metric names and can make it hard to find similar metrics with different keys. This PR converts key properties to labels.

Before:
```
# TYPE io_airlift_http_client_type_HttpClient_name_ForSecondService_WrittenBytes_AllTime_Total gauge
io_airlift_http_client_type_HttpClient_name_ForSecondService_WrittenBytes_AllTime_Total 2.0
```

After:
```
# TYPE io_airlift_http_client_WrittenBytes_AllTime_Total gauge
io_airlift_http_client_WrittenBytes_AllTime_Total{name="ForSecondService",type="HttpClient",domain="io.airlift.http_client"} 2.0
```

This is similar to how the DataDog agent converts JMX metrics, described at the end of this section: https://docs.datadoghq.com/integrations/java/?tab=host#configuration-options


openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -202,7 +204,7 @@ private Optional<Metric> getMetric(ObjectName objectName, String attributeName,
             }
 
             if (attributeValue instanceof Number) {
-                return Optional.of(Gauge.from(metricName, (Number) attributeValue, description));
+                return Optional.of(Gauge.from(metricName, List.of(), (Number) attributeValue, description));
```
nineinchnick, line 207: This is clearly wrong, but I'm not sure what's the purpose of constructing metrics from advanced filters with these separators like `_ATTRIBUTE_`.

openmetrics/src/test/resources/expected-metrics.txt
```diff
@@ -0,0 +1,513 @@
+# TYPE io_airlift_http_client_WrittenBytes_AllTime_Total gauge
```
nineinchnick, line 1: Metrics for both `ForFirstService` and `ForSecondService` should be grouped under a single `# TYPE`, but this is not implemented yet. I wanted to get an initial approval of this approach first.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -319,7 +321,21 @@ private Stream<Metric> getManagedMetricsStream()
         Map<String, ManagedClass> managedClasses = this.mbeanExporter.getManagedClasses();
 
         return managedClasses.keySet().stream()
-                .map(objectName -> getMetricsRecursively(objectName, managedClasses.get(objectName)))
+                .map(name -> {
+                    ObjectName objectName;
```
nineinchnick, line 325: Something similar should be done in `jmxMetricExpositions()`.

mattstep:
An alternative would be to add labels during scraping of the metrics that match k8s labels for the pod (pod name, app name, etc.) This can be done with datadog or the opentelemetry collector iirc.

nineinchnick:
> An alternative would be to add labels during scraping of the metrics that match k8s labels for the pod (pod name, app name, etc.)

This PR is about changing the generated metric names to something more manageable. For example, tracking bytes written in different HTTP clients should be done using a single metric with different labels, not different metrics.

mattstep:
> > An alternative would be to add labels during scraping of the metrics that match k8s labels for the pod (pod name, app name, etc.)
> 
> This PR is about changing the generated metric names to something more manageable. For example, tracking bytes written in different HTTP clients should be done using a single metric with different labels, not different metrics.

Got it, hmm, we should chat with @martint about this as well then. We should think through how this interoperates with the jmxutils ObjectNameGenerator.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -145,6 +147,8 @@ private String mBeanNameToMetricName(ObjectName objectName, String attributeName
         StringBuilder metricNameBuilder = new StringBuilder("JMX_")
                 .append(objectName.getDomain().replace(".", "_"));
 
+        // TODO name and type are not necessary here since they're included in labels
```
nineinchnick, line 150: Need to solve this. Also, I did not add any tests for non-Airlift MBeans.

nineinchnick: I'm not sure what's the purpose of the filter. It allows fetching any JMX metric, even those not present in `openmetrics.jmx-object-names`. Parsing the names is complicated and has no test coverage. I couldn't find any referenced about this in the OpenMetrics spec or Prometheus handler.

nineinchnick:
> We should think through how this interoperates with the jmxutils ObjectNameGenerator.

I don't see any issues with converting MBeans to OpenMetrics, but going the other way would require parsing labels. I'm not sure if we need it, what's the value in looking up a single metric vs filtering the output on the client side.

nineinchnick:
Another reference would be https://github.com/prometheus/jmx_exporter#default-format, which includes composite (tabular) values. Maybe we could use it as a library?

mattstep:
> > We should think through how this interoperates with the jmxutils ObjectNameGenerator.
> 
> I don't see any issues with converting MBeans to OpenMetrics, but going the other way would require parsing labels. I'm not sure if we need it, what's the value in looking up a single metric vs filtering the output on the client side.

The filter scrape API is used by Prometheus and datadog iirc.

mattstep:
> Another reference would be https://github.com/prometheus/jmx_exporter#default-format, which includes composite (tabular) values. Maybe we could use it as a library?

I explored this, but it's dependency scope is pretty big iirc.

openmetrics/src/test/java/io/airlift/openmetrics/TestMetricExpositions.java
```diff
@@ -36,8 +37,8 @@ public void testCounterExposition()
                 metric_name 0
                 """;
 
-        Counter counter = new Counter("metric_name", 0, "metric_help");
-        BigCounter bigCounter = new BigCounter("metric_name", BigInteger.ZERO, "metric_help");
+        Counter counter = new Counter("metric_name", List.of(), 0, "metric_help");
+        BigCounter bigCounter = new BigCounter("metric_name", List.of(), BigInteger.ZERO, "metric_help");
```
mattstep, line 41: We should test this with labels and without

openmetrics/src/test/java/io/airlift/openmetrics/TestMetricsResource.java
```diff
@@ -0,0 +1,56 @@
...
+import java.lang.management.ManagementFactory;
+import java.util.List;
+
+import static java.util.Objects.requireNonNull;
+import static org.testng.Assert.assertEquals;
+
+public class TestMetricsResource
+{
+    @Test
+    public void testMetrics()
```
mattstep, line 33: We should do some testing of very large numbers of metrics and labels and make sure it's not too slow.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
mattstep: It's exactly for grabbing metrics from jmx without having to specify the objectname config, it's basically the most efficient route for us to capture the jvm garbage collection metrics and such for trino, if we allow all the jmx metrics from the jvm we'll be scraping a very large amount of metrics and only using a very small subset.

openmetrics/src/test/resources/expected-metrics.txt
mattstep: In the openmetrics spec they should be, they should also share the same help section.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -245,7 +247,7 @@ private String sanitizeMetricName(String name)
                 .replace(",", "_");
     }
 
-    private List<Metric> getMetricsRecursively(String prefix, ManagedClass managedClass)
+    private List<Metric> getMetricsRecursively(String prefix, List<Map.Entry<String, String>> labels, ManagedClass managedClass)
```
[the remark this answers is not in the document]
mattstep: Why not just use a map?

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -319,7 +333,17 @@ private Stream<Metric> getManagedMetricsStream()
         Map<String, ManagedClass> managedClasses = this.mbeanExporter.getManagedClasses();
 
         return managedClasses.keySet().stream()
-                .map(objectName -> getMetricsRecursively(objectName, managedClasses.get(objectName)))
+                .map(name -> {
+                    ObjectName objectName;
+                    try {
+                        objectName = new ObjectName(name);
```
mattstep, line 339: This won't work if you use a custom ObjectNameGenerator, the ManagedClasses in jmx utils do not have object names that are guaranteed to match the jmx object name naming pattern, so this will fail for anything that implements it's own ObjectNameGenerator.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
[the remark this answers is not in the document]
nineinchnick: Do maps in Java always preserve order?

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -319,7 +333,17 @@ private Stream<Metric> getManagedMetricsStream()
         Map<String, ManagedClass> managedClasses = this.mbeanExporter.getManagedClasses();
```
mattstep, line 333: These are not guaranteed to be jmx mbeans

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
nineinchnick: I only do this because `mbeanExporter.getManagedClasses()` returns a map indexed by a string representation of the object name. Do you think we could change it to use objects?

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
nineinchnick: TIL that are can be non-jmx mbeans :D Can we filter them somehow?

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
mattstep: We want to also support non-jmx classes in the managedclasses from jmxutils in the future.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -319,7 +333,17 @@ private Stream<Metric> getManagedMetricsStream()
...
 
         return managedClasses.keySet().stream()
-                .map(objectName -> getMetricsRecursively(objectName, managedClasses.get(objectName)))
+                .map(name -> {
+                    ObjectName objectName;
+                    try {
+                        objectName = new ObjectName(name);
+                    }
+                    catch (MalformedObjectNameException e) {
+                        // objectName is a string representation of an existing ObjectName
```
mattstep, line 342: Not always, this is not strictly guaranteed.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -236,6 +241,15 @@ private String inferAttributesForObjectName(ObjectName objectName)
         return expositions.toString();
     }
 
+    private static List<Map.Entry<String, String>> getLabels(ObjectName objectName)
+    {
+        List<Map.Entry<String, String>> labels = Stream.concat(
```
mattstep, line 246: It might be simpler and more efficient to just generate an immutablemap with an immutablemap.builder in a loop here. That said, again, we don't want to tie ourselves to jmx here.

mattstep:
One thing that is missing here is that the entire point of using the managed classes directly from jmxutils is to reduce the dependence we have on jmx for metrics, this change ties us back to objectnames that conform to the jmx objectname syntax, something that the ObjectNameGenerator allows us to change in jmxutils.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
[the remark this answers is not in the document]
electrum: Depends on the implementation. LinkedHashMap and Guava InmutableMap do. It will be added directly here: https://openjdk.org/jeps/431

electrum:
We’ve talked about going with more generic JMX names in Trino. Using class names and packages is leaking an implementation detail. We need to improve the ObjectNameGenerator API since it requires a class name (with a qualified package name) today.

nineinchnick:
I'm closing this PR, since the goals are not clear here. I need to reevaluate any naming scheme changes.

martint:
> We need to improve the ObjectNameGenerator API since it requires a class name (with a qualified package name) today.

You can bind JMX objects with any name by using:

```
 exporter.export(ManagedObject.class).as("<some name>");
```

It doesn't need to be based on the package/class name. ObjectNameGenerator is there for convenience.

PR #1062, opened 2023-04-27
Use global OpenTelemetry if tracing is not explicitly enabled

nineinchnick:
This allows running with the following JVM flags:
```
-javaagent:/otel/opentelemetry-javaagent.jar
-Dotel.exporter.otlp.endpoint=http://host.docker.internal:4317
-Dotel.service.name=load-tests
-Dotel.resource.attributes=env=local
```

instead of these:
```
-Dtracing.enabled=true
-Dtracing.exporter.endpoint=http://host.docker.internal:4317
```

The advantages are:
* enabling more built-in integrations, like getting additional spans from Jetty and JAX-RS
* being able to set the service name and additional attributes

One disadvantage is having to provide the agent jar.

Refs: https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md#opentelemetry-resource

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
@@ -27,6 +27,7 @@ import com.google.inject.TypeLiteral;
 import io.airlift.configuration.ConfigDefaults;
 import io.airlift.http.client.jetty.JettyHttpClient;
 import io.airlift.node.NodeInfo;
+import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.api.trace.TracerProvider;
@@ -98,7 +99,7 @@ public class HttpClientModule
         private final Class<? extends Annotation> annotation;
         private Injector injector;
         private NodeInfo nodeInfo;
-        private OpenTelemetry openTelemetry = OpenTelemetry.noop();
+        private OpenTelemetry openTelemetry = GlobalOpenTelemetry.get();
         private Tracer tracer = TracerProvider.noop().get("noop");
 
         private HttpClientProvider(String name, Class<? extends Annotation> annotation)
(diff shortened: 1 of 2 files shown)


nineinchnick:
@electrum PTAL

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
```diff
@@ -27,6 +27,7 @@
 import io.airlift.configuration.ConfigDefaults;
 import io.airlift.http.client.jetty.JettyHttpClient;
 import io.airlift.node.NodeInfo;
+import io.opentelemetry.api.GlobalOpenTelemetry;
```
mattstep, line 30: Why would we want to default to this over noop?

mattstep:
> This allows running with the following JVM flags:
> 
> ```
> -javaagent:/otel/opentelemetry-javaagent.jar
> -Dotel.exporter.otlp.endpoint=http://host.docker.internal:4317
> -Dotel.service.name=load-tests
> -Dotel.resource.attributes=env=local
> ```
> 
> instead of these:
> 
> ```
> -Dtracing.enabled=true
> -Dtracing.exporter.endpoint=http://host.docker.internal:4317
> ```
> 
> The advantages are:
> 
> * enabling more built-in integrations, like getting additional spans from Jetty and JAX-RS
> * being able to set the service name and additional attributes
> 
> One disadvantage is having to provide the agent jar.
> 
> Refs: https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md#opentelemetry-resource

I don't think this would work well with trino connectors, and I don't really like having to run with an agent all the time, it would be best if we could move away from using an agent.

electrum:
We already have integration with Jetty and JAX-RS. What is lacking?

I don’t have a philosophical problem with using the agent, but it shouldn’t be necessary, since we already did the work to integrate manually.

mattstep:
> We already have integration with Jetty and JAX-RS. What is lacking?
> 
> I don’t have a philosophical problem with using the agent, but it shouldn’t be necessary, since we already did the work to integrate manually.

In my experience each obsevability provider has a custom built agent, but also supports sending spans via grpc, so we should probably pick one or the other and so far we've gone down the path of manually integrating and using grpc to forward traces.

nineinchnick:
Closing in favor of #1065 - the main reason for this was to be able to configure some OpenTelemetry defaults.

PR #1063, opened 2023-04-28
Add Scale method to QuantileDigest

huberty89:
Cherry picked from https://github.com/prestodb/airlift/pull/12


huberty89:
I did not see there is already PR for that https://github.com/airlift/airlift/pull/1053

PR #1065, opened 2023-05-05
Allow setting additional OpenTelemetry resource attributes

nineinchnick:
Alternative to #1062


nineinchnick:
@mattstep @electrum PTAL

tracing/src/main/java/io/airlift/tracing/OpenTelemetryConfig.java
```diff
@@ -22,4 +28,22 @@ public OpenTelemetryConfig setEndpoint(String endpoint)
...
     }
+
+    @NotNull
+    public Map<String, String> getResourceAttributes()
+    {
+        return resourceAttributes;
+    }
+
+    @Config("tracing.resource.attributes")
+    public OpenTelemetryConfig setResourceAttributes(String resourceAttributes)
```
mattstep, line 39: You should take a look at how we allow setting additional annotations in logging in the logmanager. We allow for setting a separate property file with key-value pairs. It's a little cleaner.

mattstep: Actually, it would be great if we refactored that for this so we just set it up once.

nineinchnick: Done! We'll probably be able to reuse the same properties file here, this is better, thanks!

I'm not sure what refactoring you mean, this is already called from a singleton provider.

tracing/src/main/java/io/airlift/tracing/OpenTelemetryConfig.java
```diff
@@ -22,4 +24,17 @@ public OpenTelemetryConfig setEndpoint(String endpoint)
...
         return this;
     }
+
+    @FileExists
+    public String getResourceAttributesFile()
+    {
+        return resourceAttributesFile;
+    }
+
+    @Config("tracing.resource-attributes-file")
```
mattstep, line 34: We should just have a common name, we are calling it different things in the separate modules name with the same intention, and we could use a common property name so we just set it once in config. This looks good otherwise.

tracing/src/main/java/io/airlift/tracing/OpenTelemetryConfig.java
mattstep: Yeah, if you want to rename that one to be something more generic.

log-manager/src/main/java/io/airlift/log/LoggingConfiguration.java
```diff
@@ -140,7 +140,11 @@ public String getLogAnnotationFile()
         return logAnnotationFile;
     }
 
-    @Config("log.annotation-file")
+    /* Reuse the same property as defined in io.airlift.node.NodeConfig because NodeInfo objects
+       are not yet available when the Logging is being configured.
+     */
+    @Config("node.annotation-file")
+    @LegacyConfig("log.annotation-file")
```
nineinchnick, line 147: I'm not sure if I should use `@Deprecated` here instead, would this raise an exception if the old name is used?

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -249,6 +273,15 @@ public long getStartTime()
         return startTime;
     }
 
+    /**
+     * Annotations describing this server and/or the environment in which it is running.
+     */
+    @Managed
```
nineinchnick, line 279: Is it safe to tag a map as managed?

nineinchnick:
@mattstep @electrum PTAL, I refactored this after talking with Matt.

node/src/main/java/io/airlift/node/NodeInfo.java
mattstep: I don't think it'll export properly, you want to use this as labels I think, not as a time series value

mattstep (approved):

node/src/main/java/io/airlift/node/NodeInfo.java
nineinchnick: Thanks, removed.

log-manager/src/main/java/io/airlift/log/LoggingConfiguration.java
nineinchnick: The docs say that "diagnostics will be logged": https://github.com/airlift/airlift/blob/master/docs/ref_configuration.md#legacyconfig

log-manager/src/main/java/io/airlift/log/LoggingConfiguration.java
```diff
@@ -140,7 +140,11 @@ public String getLogAnnotationFile()
         return logAnnotationFile;
     }
 
-    @Config("log.annotation-file")
+    /* Reuse the same property as defined in io.airlift.node.NodeConfig because NodeInfo objects
```
electrum, line 143: Nit: we would use `//` comments for this. Using a block comment here makes it look like a Javadoc.

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -83,7 +91,8 @@ public NodeInfo(String environment,
             String location,
             String binarySpec,
             String configSpec,
-            AddressSource internalAddressSource)
+            AddressSource internalAddressSource,
+            String annotationsFile)
```
electrum, line 95: Name this `annotationFile` to be consistent

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -132,6 +141,21 @@ public NodeInfo(String environment,
         else {
             this.externalAddress = this.internalAddress;
         }
+
+        if (annotationsFile != null) {
+            try {
+                this.annotations = replaceEnvironmentVariables(loadPropertiesFrom(annotationsFile));
+            }
+            catch (FileNotFoundException e) {
```
electrum, line 149: No need to handle this specially. Wrapping with `UncheckedIOException` is fine.

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -132,6 +141,21 @@ public NodeInfo(String environment,
         else {
             this.externalAddress = this.internalAddress;
         }
+
+        if (annotationsFile != null) {
+            try {
+                this.annotations = replaceEnvironmentVariables(loadPropertiesFrom(annotationsFile));
```
electrum, line 147: Add `ImmutableMap.copyOf()` since `replaceEnvironmentVariables()` returns a mutable map and we expose it from this class.

node/src/test/java/io/airlift/node/TestNodeConfig.java
```diff
@@ -46,12 +49,20 @@ public void testDefaults()
...
                 .setConfigSpec(null)
-                .setInternalAddressSource(IP));
+                .setInternalAddressSource(IP)
+                .setAnnotationFile(null));
     }
 
     @Test
     public void testExplicitPropertyMappings()
     {
+        File annotationFile;
```
electrum, line 59: You can use `TempFile` from the `testing` module which supports try-with-resources

electrum: Actually, we could reuse the `annotations.properties` testing resource file here

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
```diff
@@ -63,6 +63,7 @@ public OpenTelemetry createOpenTelemetry(NodeInfo nodeInfo, OpenTelemetryConfig
                 .putAll(attribute(ResourceAttributes.OS_NAME, OS_NAME.value()))
                 .putAll(attribute(ResourceAttributes.OS_VERSION, OS_VERSION.value()))
                 .putAll(attribute(ResourceAttributes.HOST_ARCH, hostArch()));
+        nodeInfo.getAnnotations().forEach(attributes::put);
```
electrum, line 66: @mattstep Should we namespace these somehow? Are the values typically used for log annotations appropriate to use in this way?

electrum (approved):

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
mattstep: Good call, I'd say yes, we basically do that in logging.

mattstep: They are in a nested field "annotations" : https://github.com/airlift/airlift/blob/master/log-manager/src/main/java/io/airlift/log/JsonRecord.java#L129

nineinchnick: Any ideas for the namespace? I think `annotations` would be too generic in this context. Also, as Matt mentioned, we already put this in a nested `annotations` field for logging, so this would stutter. Maybe `node-info`?

On the other hand, should we leave this for users to come up with meaningful prefix? This would also allow overriding some of the default attributes, like `service.name`, which is configurable in every OpenTelemetry instrumentation.

log-manager/src/main/java/io/airlift/log/LoggingConfiguration.java
nineinchnick: Thanks, it felt weird but I wasn't sure what's the convention.

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
mattstep: It's a very generic feature, just about anything can be added in. Idk, we used annotations for logging as a name because it's just really generic. We probably don't want to support overriding the fields we set already.

mattstep: Looking through https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md#recommendations-for-application-developers I believe we should namespace the attributes something like `io.airlift.custom` for anything passed in via the annotations file, and the naming there might be better as `custom-attributes` or so... For any attributes we're adding in that are specific to airlift and not based on something in ResourceAttributes we should use the airlift packaging from the module it comes from as it's namespace, so anything derived from NodeInfo that isn't already a standard attribute would be `io.airlift.node`. How does that sound?

nineinchnick: Sounds good, I added `io.airlift.node`, PTAL and thanks!

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
```diff
@@ -27,11 +27,13 @@
 import static com.google.common.base.Strings.nullToEmpty;
 import static io.airlift.configuration.ConfigBinder.configBinder;
 import static io.airlift.tracing.Tracing.attribute;
+import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 
 public class OpenTelemetryModule
         implements Module
 {
+    private static final String NODE_ANNOTATION_PREFIX = "io.airlift.node";
```
findepi, line 36: just `airlift`?
or do we need any prefix at all?
since we pass-through user-provided annotations, we could as well just add them without any prefix.

tracing/src/main/java/io/airlift/tracing/OpenTelemetryConfig.java
```diff
@@ -7,6 +7,7 @@
 public class OpenTelemetryConfig
 {
     private String endpoint = "http://localhost:4317";
+    private String resourceAttributesFile;
```
findepi, line 10: is this used?

node/src/test/java/io/airlift/node/TestNodeModule.java
```diff
@@ -78,6 +82,7 @@ public void testFullConfig()
         String binarySpec = "binary";
         String configSpec = "config";
         String publicAddress = "public";
+        File annotationFile = new File(requireNonNull(getClass().getClassLoader().getResource("annotations.properties")).getFile());
```
findepi, line 85: requireNonNull -> use guava `Resources`
`URL.getFile()` -> this will pass for resource URLs like `jar:...`, but you want it to fail

use `new File(url.toURI())`

node/src/test/java/io/airlift/node/TestNodeConfig.java
```diff
@@ -45,12 +47,14 @@ public void testDefaults()
...
                 .setConfigSpec(null)
-                .setInternalAddressSource(IP));
+                .setInternalAddressSource(IP)
+                .setAnnotationFile(null));
     }
 
     @Test
     public void testExplicitPropertyMappings()
     {
+        File annotationFile = new File(requireNonNull(getClass().getClassLoader().getResource("annotations.properties")).getFile());
```
findepi, line 57: same here

> requireNonNull -> use guava `Resources`
`URL.getFile()` -> this will pass for resource URLs like `jar:...`, but you want it to fail

> use `new File(url.toURI())`

node/src/test/java/io/airlift/node/TestNodeModule.java
nineinchnick: This is a test and the resource name is hardcoded, not provided by users. I'm not sure if I understand the requested change, can you add a code suggestion?

tracing/src/main/java/io/airlift/tracing/OpenTelemetryConfig.java
nineinchnick: No, it's a leftover, thanks!

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
findepi: the link doesn't work for me

node/src/test/java/io/airlift/node/TestNodeModule.java
findepi:
```
File annotationFile = new File([Resources.]getResource("annotations.properties").toURI());
```

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
nineinchnick: Sorry, this one's correct: https://github.com/airlift/airlift/pull/1065#discussion_r1197221348

PR #1067, opened 2023-05-12
Switch from javax.* to jakarta.*

wendigo:
This PR moves airlift to Jakarta namespace with updated dependencies like Jetty 11, Jersey 3.1, Servlet 6.0, Guice 7.0.

All tests are passing


wendigo:
cc @electrum

electrum:
Can you remove the inject annotations, since we'll use #1068 instead?

wendigo:
I can merge those two PRs together @electrum

wendigo:
Nevermind, I'm rebasing this PR after #1068 was merged

wendigo:
Rebased. PTAL @electrum

electrum:
Commit message doesn’t mention the Jersey update

wendigo:
@electrum do you mean jakarta.rs-ws.api?

wendigo:
@electrum please ammend commit message as you please

wendigo:
ping @electrum

pom.xml
```diff
@@ -284,37 +285,45 @@
...
                 <groupId>org.eclipse.jetty</groupId>
                 <artifactId>jetty-bom</artifactId>
-                <version>10.0.14</version>
+                <version>11.0.15</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
 
             <!-- jersey -->
             <dependency>
```
electrum, line 301: We should add these three dependencies to Airbase instead, since they are needed in consumers

pom.xml
```diff
@@ -32,8 +32,9 @@
 
         <dep.airlift.version>232-SNAPSHOT</dep.airlift.version>
         <dep.packaging.version>${dep.airlift.version}</dep.packaging.version>
-        <dep.jersey.version>2.39</dep.jersey.version>
+        <dep.jersey.version>3.1.2</dep.jersey.version>
         <dep.jjwt.version>0.11.2</dep.jjwt.version>
+        <dep.guice.version>7.0.0</dep.guice.version>
```
electrum, line 37: We need to update this in Airbase

electrum:
I'm going to prepare a PR to Trino to verify that the new version of everything works.

event-http/pom.xml
```diff
@@ -71,7 +71,7 @@
 
         <dependency>
```
electrum, line 72: I think we want to remove this and use `jakarta.servlet:jakarta.servlet-api` instead

electrum: I'm asking the Jetty folks via support about this

electrum:
Here's the draft Trino PR: https://github.com/trinodb/trino/pull/17764

electrum:
@wendigo can you put up an Airbase PR with those dependency additions plus the Guice version update?

electrum:
Draft PR for discovery-server: https://github.com/airlift/discovery/pull/50

electrum:
Async processing is broken. Unfortunately, we don't have any tests for this in Airlift.

electrum:
I updated your branch with some fixes.

event-http/pom.xml
electrum: Confirmed that this is the right dependency

electrum:
We also need to add `jakarta.annotation:jakarta.annotation-api:2.1.1` to Airbase as that's needed in consumers for specifying JAX-RS filter priority.

wendigo:
@electrum I've migrated annotation-api to jakarta as well

pom.xml
wendigo: Done

pom.xml
wendigo: Done

wendigo:
This now requires: https://github.com/airlift/airbase/pull/361

jaxrs/pom.xml
```diff
@@ -138,6 +114,30 @@
             <scope>runtime</scope>
         </dependency>
 
+        <dependency>
```
electrum, line 117: Let's not move these dependencies. The comment above "used by tests but also needed transitively" is only true for `json` but not for these. We could revisit this later if we apply the POM order enforcement.

wendigo: Sure. Restored previous order. I'll follow up with the pedantic pom enforcer

wendigo:
Updated with airbase 141

electrum (approved):

PR #1068, opened 2023-05-15
Use google.inject annotations instead of javax

wendigo:
Follow up to the https://github.com/airlift/airlift/pull/1067 as discussed with @electrum

### DIFF
--- a/bootstrap/pom.xml
+++ b/bootstrap/pom.xml
@@ -46,13 +46,6 @@
             <artifactId>javax.annotation-api</artifactId>
         </dependency>
 
-        <!-- used by tests but also needed transitively -->
-        <dependency>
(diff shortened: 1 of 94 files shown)


wendigo:
cc @electrum @findepi

wendigo:
I'd like to switch Trino to google.inject as well. We've discussed it with @electrum that it'd make switch to guice 7.0 easier (it is jakarta based).

findepi:
does this mean that Guice 7 no longer honors `javax.inject.Inject` annotations?

wendigo:
@findepi yes, it has moved to jakarta.* namespace. Hence this change.

wendigo:
ping @electrum @martint

PR #1069, opened 2023-05-16
Restore previous string limit for JSON 

wendigo:
Prior to the Jackson 2.15 there was no upper bound on the length of the string fields.

In order to increase compatibility with the existing use cases, previous behavior is restored.

### DIFF
--- a/jmx-http/src/main/java/io/airlift/jmx/JmxHttpModule.java
+++ b/jmx-http/src/main/java/io/airlift/jmx/JmxHttpModule.java
@@ -87,6 +87,7 @@ public class JmxHttpModule
         }
 
         @Override
+        @SuppressWarnings("deprecation")
         public JsonNode getSchema(SerializerProvider provider, Type typeHint)
                 throws JsonMappingException
         {
@@ -148,6 +149,7 @@ public class JmxHttpModule
         }
 
         @Override
+        @SuppressWarnings("deprecation")
         public JsonNode getSchema(SerializerProvider provider, Type typeHint)
         {
             return createSchemaNode("object", true);
--- a/json/src/main/java/io/airlift/json/ObjectMapperProvider.java
+++ b/json/src/main/java/io/airlift/json/ObjectMapperProvider.java
@@ -17,6 +17,7 @@ package io.airlift.json;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.StreamReadConstraints;
 import com.fasterxml.jackson.core.Version;
 import com.fasterxml.jackson.databind.DeserializationFeature;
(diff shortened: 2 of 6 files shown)


wendigo:
cc @findepi @electrum

jmx-http/src/main/java/io/airlift/jmx/JmxHttpModule.java
```diff
@@ -148,6 +149,7 @@ public void serialize(CompositeData data, JsonGenerator jsonGenerator, Serialize
         }
 
         @Override
+        @SuppressWarnings("deprecation")
```
findepi, line 152: why added now?

json/src/main/java/io/airlift/json/ObjectMapperProvider.java
```diff
@@ -215,7 +216,15 @@ public ObjectMapper get()
 
         objectMapper.addModules(modules);
 
-        return objectMapper.build();
+        ObjectMapper mapper = objectMapper.build();
+
+        // Disable the length limit, caller will be responsible for validating the input length
+        mapper.getFactory().setStreamReadConstraints(StreamReadConstraints
```
findepi, line 222: is `mapper.getFactory()` same field as our `jsonFactory` ?

jmx-http/src/main/java/io/airlift/jmx/JmxHttpModule.java
wendigo: Jackson was updated with the airbase bump and these methoda are now deprecated

json/src/main/java/io/airlift/json/ObjectMapperProvider.java
wendigo: No. Internal jackson thingy

jmx-http/src/main/java/io/airlift/jmx/JmxHttpModule.java
findepi: why deprecated? should we stop using them?

json/src/main/java/io/airlift/json/ObjectMapperProvider.java
```diff
@@ -69,6 +70,12 @@ public ObjectMapperProvider(JsonFactory jsonFactory)
     {
         this.jsonFactory = requireNonNull(jsonFactory, "jsonFactory is null");
 
+        // Disable the length limit, caller will be responsible for validating the input length
+        jsonFactory.setStreamReadConstraints(StreamReadConstraints
+                .builder()
+                .maxStringLength(Integer.MAX_VALUE)
```
findepi, line 76: add a note to CHANGES

wendigo: Done!

PR #1070, opened 2023-05-24
Allow custom configuration interpolation logic

wendigo:


wendigo:
cc @martint @electrum

configuration/src/main/java/io/airlift/configuration/ConfigurationInterpolator.java
```diff
@@ -0,0 +1,114 @@
...
+                String interpolatorName = matcher.group("name");
+                String keyName = matcher.group("key");
+                ValueInterpolator valueInterpolator = valueProviders.get(interpolatorName);
+                if (valueInterpolator == null) {
+                    onError.accept(propertyKey, "Configuration property '%s' references value interpolator '%s' that does not exist".formatted(propertyKey, interpolatorName));
+                    return;
+                }
+
+                Optional<String> replacedValue = valueInterpolator.interpolate(keyName);
+                if (replacedValue.orElse(null) == null) {
```
nineinchnick, line 57: Would `isEmpty()` be more idiomatic? Should we assert that `replacedValue` is not null? That would signal some other issues with the interpolator, right?
```suggestion
                if (replacedValue.isEmpty()) {
```

configuration/src/main/java/io/airlift/configuration/ConfigurationInterpolator.java
```diff
@@ -0,0 +1,114 @@
...
+            }
+            matcher.appendTail(replacedPropertyValue);
+            replaced.put(propertyKey, replacedPropertyValue.toString());
+        });
+        return replaced;
+    }
+
+    public interface ValueInterpolator
+    {
+        Optional<String> interpolate(String key);
```
nineinchnick, line 71: If this interface would also have a `name()` method, we could avoid passing `valueProviders` as a map and maybe use a set. I was wondering if the `ENV` key should be defined as a const, but then figured out that if we always associate it with `EnvValueInterpolator` it should be embedded in it.

electrum:
What is the use case for this?

wendigo:
@electrum interpolating secrets from other sources than env (i.e. AWS secretmanager or Hashicorp vault)

PR #1071, opened 2023-05-24
Add HttpStatusListener to http client

Randgalt:
An HTTP status listener hooks into Jetty's response listening so that users of this client can listen for the beginning of the response and examine the status before anything else from the request is processed. This type of listener has many uses cases one of which is to short circuit a response that has a status code that indicates the server is busy, no longer available, etc. It also gives users of this client an opportunity to throw an exception so that Jetty removes the connection from its connection pool. There are cases where Jetty's connection pool can get poisoned with bad connections.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientBinder.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientBinder.java
@@ -25,16 +25,19 @@ import java.lang.annotation.Annotation;
 import java.util.Collection;
 
 import static com.google.inject.multibindings.Multibinder.newSetBinder;
+import static java.util.Objects.requireNonNull;
 
 public class HttpClientBinder
 {
     private final Binder binder;
     private final Multibinder<HttpRequestFilter> globalFilterBinder;
+    private final Multibinder<HttpStatusListener> globalStatusListenerBinder;
 
     private HttpClientBinder(Binder binder)
     {
         this.binder = binder.skipSources(getClass());
         this.globalFilterBinder = newSetBinder(binder, HttpRequestFilter.class, GlobalFilter.class);
+        this.globalStatusListenerBinder = newSetBinder(binder, HttpStatusListener.class, GlobalFilter.class);
     }
 
     public static HttpClientBinder httpClientBinder(Binder binder)
@@ -46,12 +49,17 @@ public class HttpClientBinder
     {
         HttpClientModule module = new HttpClientModule(name, annotation);
         binder.install(module);
-        return new HttpClientBindingBuilder(module, newSetBinder(binder, HttpRequestFilter.class, annotation));
+        return new HttpClientBindingBuilder(module, newSetBinder(binder, HttpRequestFilter.class, annotation), newSetBinder(binder, HttpStatusListener.class, annotation));
     }
 
     public LinkedBindingBuilder<HttpRequestFilter> addGlobalFilterBinding()
     {
         return globalFilterBinder.addBinding();
+    }
+
+    public LinkedBindingBuilder<HttpStatusListener> addGlobalStatusListenerBinding()
+    {
+        return globalStatusListenerBinder.addBinding();
     }
 
     public HttpClientBinder bindGlobalFilter(Class<? extends HttpRequestFilter> filterClass)
@@ -69,12 +77,14 @@ public class HttpClientBinder
     public static class HttpClientBindingBuilder
     {
         private final HttpClientModule module;
-        private final Multibinder<HttpRequestFilter> multibinder;
+        private final Multibinder<HttpRequestFilter> filterBinder;
+        private final Multibinder<HttpStatusListener> statusListenerBinder;
 
-        public HttpClientBindingBuilder(HttpClientModule module, Multibinder<HttpRequestFilter> multibinder)
+        public HttpClientBindingBuilder(HttpClientModule module, Multibinder<HttpRequestFilter> filterBinder, Multibinder<HttpStatusListener> statusListenerBinder)
         {
-            this.module = module;
-            this.multibinder = multibinder;
+            this.module = requireNonNull(module, "module is null");
+            this.filterBinder = requireNonNull(filterBinder, "multibinder is null");
+            this.statusListenerBinder = requireNonNull(statusListenerBinder, "statusListenerBinder is null");
         }
 
         public HttpClientBindingBuilder withAlias(Class<? extends Annotation> alias)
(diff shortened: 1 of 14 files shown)


http-client/src/main/java/io/airlift/http/client/HttpClientBinder.java
```diff
@@ -46,7 +49,12 @@ public HttpClientBindingBuilder bindHttpClient(String name, Class<? extends Anno
     {
         HttpClientModule module = new HttpClientModule(name, annotation);
         binder.install(module);
-        return new HttpClientBindingBuilder(module, newSetBinder(binder, HttpRequestFilter.class, annotation));
+        return new HttpClientBindingBuilder(module, newSetBinder(binder, HttpRequestFilter.class, annotation), newSetBinder(binder, HttpStatusListener.class, annotation));
+    }
+
+    public LinkedBindingBuilder<HttpStatusListener> addGlobalStatusListenerBinding()
```
electrum, line 55: Nit: move this below `addGlobalFilterBinding()` so that the ordering matches the field ordering

http-client/src/main/java/io/airlift/http/client/HttpClientBinder.java
```diff
@@ -70,11 +78,13 @@ public HttpClientBinder bindGlobalFilter(HttpRequestFilter filter)
...
         private final Multibinder<HttpRequestFilter> multibinder;
+        private final Multibinder<HttpStatusListener> statusListenerBinder;
 
-        public HttpClientBindingBuilder(HttpClientModule module, Multibinder<HttpRequestFilter> multibinder)
+        public HttpClientBindingBuilder(HttpClientModule module, Multibinder<HttpRequestFilter> multibinder, Multibinder<HttpStatusListener> statusListenerBinder)
         {
-            this.module = module;
-            this.multibinder = multibinder;
+            this.module = requireNonNull(module, "module is null");
+            this.multibinder = requireNonNull(multibinder, "multibinder is null");
```
electrum, line 86: Let's rename the field and parameter to `filterBinder` now that we have multiple multibinders (the old name was bad anyway)

http-client/src/main/java/io/airlift/http/client/HttpStatusListener.java
```diff
@@ -0,0 +1,21 @@
+/*
```
electrum, line 1: Remove this license header, since this is a new class and the copyright is not correct

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -200,13 +211,27 @@ public JettyHttpClient(
             Tracer tracer,
             Optional<String> environment,
             Optional<SslContextFactory.Client> maybeSslContextFactory)
+    {
+        this(name, config, requestFilters, openTelemetry, tracer, environment, maybeSslContextFactory, ImmutableList.of());
+    }
+
+    public JettyHttpClient(
```
electrum, line 218: As a follow up, we should probably deprecate most/all of these constructors and add a builder

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -619,6 +645,12 @@ public void onContent(Response response, ByteBuffer content)
                 }
                 super.onContent(response, content);
             }
+
+            @Override
+            public void onBegin(Response response)
```
electrum, line 650: Should we move this above `onContent()`, since this will be called first?

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -85,6 +87,7 @@
     private String scheme = "http";
     private String host = "127.0.0.1";
     protected String keystore;
+    protected final Map<Integer, Integer> statusCounts = new ConcurrentHashMap<>();
```
electrum, line 90: `Multiset<Integer>` seems better, since the intent is to have a count of the items

http-client/src/test/java/io/airlift/http/client/TestHttpClientBinder.java
```diff
@@ -220,6 +270,13 @@ private static void assertFilterCount(HttpClient httpClient, int filterCount)
         assertEquals(((JettyHttpClient) httpClient).getRequestFilters().size(), filterCount);
     }
 
+    private static void assertStatusListenerCount(HttpClient httpClient, int statusListenerCount)
+    {
+        assertNotNull(httpClient);
```
electrum, line 275: You can do
```java
assertThat(httpClient).isInstanceOfSatisfying(JettyHttpClient.class, jettyClient ->
        assertThat(jettyClient.getStatusListeners().size()).isEqualTo(statusListenerCount));
```

http-client/src/test/java/io/airlift/http/client/TestingStatusListener.java
```diff
@@ -0,0 +1,38 @@
...
+    @Inject
+    public TestingStatusListener(Map<Integer, Integer> statusCounter)
+    {
+        this.statusCounter = statusCounter;
+    }
+
+    @Override
+    public void statusReceived(int statusCode)
+    {
+        statusCounter.compute(statusCode, (ignore, count) -> (count == null) ? 1 : (count + 1));
```
electrum, line 36: With `Multiset` this would simply be `statusCounter.add(statusCode)`. Though you can simplify this for `Map`
```java
statusCounter.merge(statusCode, 1, Math::addExact);

File: http-client/src/test/java/io/airlift/http/client/jetty/AbstractHttpClientTestHttpProxy.java [hunk 9]
```diff
@@ -51,7 +53,7 @@ public void setUpHttpClient()
             throws Exception
     {
         testingHttpProxy = new TestingHttpProxy(Optional.ofNullable(keystore));
-        httpClient = new JettyHttpClient("test-shared", createClientConfig(), ImmutableList.of(new TestingRequestFilter()));
+        httpClient = new JettyHttpClient("test-shared", createClientConfig(), ImmutableList.of(new TestingRequestFilter()), ImmutableSet.of(new TestingStatusListener(statusCounts)));
```
electrum, line 56: Maybe add a method `getStatusListeners()` in the parent class.

electrum (approved):
Please update CHANGES

http-client/src/test/java/io/airlift/http/client/jetty/AbstractHttpClientTestHttpProxy.java
Randgalt: We don't have it for filters and it's not needed anywhere

Randgalt:
@electrum all comments applied. Please merge when you can.

PR #1073, opened 2023-06-07
Switch validation to Jakarta-based

wendigo:
Requires https://github.com/airlift/airbase/pull/363

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
@@ -40,13 +40,14 @@ import com.google.inject.spi.InstanceBinding;
 import com.google.inject.spi.Message;
 import com.google.inject.spi.ProviderInstanceBinding;
 import io.airlift.configuration.ConfigurationMetadata.AttributeMetadata;
-import org.apache.bval.jsr.ApacheValidationProvider;
+import jakarta.validation.ConstraintViolation;
+import jakarta.validation.Validation;
+import jakarta.validation.Validator;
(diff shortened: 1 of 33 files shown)


configuration/src/test/java/io/airlift/configuration/validation/TestFileExistsValidator.java
```diff
@@ -20,7 +20,13 @@
 
 public class TestFileExistsValidator
 {
-    private static final Validator VALIDATOR = Validation.byProvider(ApacheValidationProvider.class).configure().buildValidatorFactory().getValidator();
+    private static final Validator VALIDATOR = Validation.byProvider(HibernateValidator.class)
```
electrum, line 23: Can we switch this to use `ValidationAssertions`?

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -735,4 +746,122 @@ public void accept(ConfigurationProvider<?> configurationProvider)
             listener.configurationBound(configurationProvider.getConfigurationBinding(), configBinder);
         }
     }
+
+    private static class ThrowingClassLoader
```
electrum, line 750: Let's make this a package-private top level class

electrum (approved):

pom.xml
```diff
@@ -13,7 +13,7 @@
     <parent>
         <groupId>io.airlift</groupId>
         <artifactId>airbase</artifactId>
-        <version>141</version>
+        <version>142-SNAPSHOT</version>
```
electrum, line 16: `142` is released now

PR #1075, opened 2023-06-10
HttpStatusListener exceptions need to fail the response

Randgalt:
The most recent version of Jetty ignores exceptions thrown by listeners. This subverts the intent of the recently added HttpStatusListener. Modify the handling of HttpStatusListeners to fail the response if the HttpStatusListener throws an exception.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 234
 
 - Move to Jakarta validation API
+- Fail response if HttpStatusListener throws an exception
 
 233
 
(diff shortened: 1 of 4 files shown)


Randgalt:
@wendigo and/or @electrum is there any way to publish a 232.1 with this commit cherry picked?

wendigo (approved):

wendigo:
@Randgalt David will know.

As for the change itself - can we test it?

wendigo (approved):

electrum (approved):

PR #1078, opened 2023-06-22
HttpStatusListener exceptions need to fail the response

Randgalt:
The most recent version of Jetty ignores exceptions thrown
by listeners. This subverts the intent of the recently added
HttpStatusListener. Modify the handling of HttpStatusListeners
to fail the response if the HttpStatusListener throws an
exception.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,7 @@
+232.1
+
+- Fail response if HttpStatusListener throws an exception
+
 232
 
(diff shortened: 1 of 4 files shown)


electrum:
Looks like this was already merged?

Randgalt:
> Looks like this was already merged?

Yeah sorry. This was a mistake.

PR #1079, opened 2023-06-27
Sort POMs with sort-pom plugin

wendigo:
The plugin will be applied from the airbase 144: airlift/airbase#370


wendigo:
cc @electrum

electrum (approved):
Please remove the "for testing" section headers, since they're now in the wrong place and not needed with the automatic sorting.

PR #1080, opened 2023-07-07
Allow registering custom span processors

wendigo:


wendigo:
cc @electrum

wendigo:
The use case is to allow adding a custom span processor that will capture and process spans internally (instead of exporting them to the external sink like Jeager)

wendigo:
cc @electrum @nineinchnick

tracing/src/main/java/io/airlift/tracing/NoopSpanProcessor.java
```diff
@@ -0,0 +1,32 @@
+package io.airlift.tracing;
+
+import io.opentelemetry.context.Context;
+import io.opentelemetry.sdk.trace.ReadWriteSpan;
+import io.opentelemetry.sdk.trace.ReadableSpan;
+import io.opentelemetry.sdk.trace.SpanProcessor;
+
+public class NoopSpanProcessor
```
electrum, line 21: No need for this class. You can do `SpanProcessor.composite()`

tracing/src/main/java/io/airlift/tracing/NoopSpanProcessor.java
electrum, line 1: We haven't bothered with license headers for new files in this project. Everything is covered under the root `LICENSE` file. There are lots of other files without license headers, so let's not change this now.

tracing/src/main/java/io/airlift/tracing/OpenTelemetryConfig.java
```diff
@@ -21,4 +22,16 @@ public OpenTelemetryConfig setEndpoint(String endpoint)
         this.endpoint = endpoint;
         return this;
     }
+
+    public boolean isExportEnabled()
+    {
+        return exportEnabled;
+    }
+
+    @Config("tracing.exporter.enabled")
```
nineinchnick, line 31: It's not obvious what this does. Looks like it ignores the default span processor, which has the default exporter. If this is only used for tests, it shouldn't be a configuration option. Why users would enable tracing without any exporter?

wendigo: it allows disabling default exporter, in case if you want to add your own spanprocessor and don't use the default one

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
```diff
@@ -45,11 +51,12 @@ public OpenTelemetryModule(String serviceName, String serviceVersion)
     public void configure(Binder binder)
     {
         configBinder(binder).bindConfig(OpenTelemetryConfig.class);
+        newSetBinder(binder, SpanProcessor.class);
     }
 
     @Provides
     @Singleton
-    public OpenTelemetry createOpenTelemetry(NodeInfo nodeInfo, OpenTelemetryConfig config)
+    public OpenTelemetry createOpenTelemetry(NodeInfo nodeInfo, Set<SpanProcessor> extraSpanProcessors, OpenTelemetryConfig config)
```
nineinchnick, line 59: Maybe the span processor could be a dependency with a default, so it's possible to override it in tests?

tracing/src/main/java/io/airlift/tracing/OpenTelemetryConfig.java
nineinchnick: I don't think it's a good idea to mix configuration options with module arguments/dependencies.

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
wendigo: Yeah, makes sense.

nineinchnick (approved):
Looks much better!

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
```diff
@@ -110,4 +111,25 @@ private static String hostArch()
             default -> null;
         };
     }
+
+    private static class BatchSpanExporterProvider
```
electrum, line 115: I would call this the `OltpGrpcSpanExporterProvider` since that is more important than the "batch" aspect (which is some local buffering and is effectively an implementation detail)

electrum (commented):
I like this approach better, since it's not mixing a user configuration with something an application would configure. However, we shouldn't use an optional binder, since that still requires the config to be bound. We don't want to allow setting `tracing.exporter.endpoint` if the exporter is not used.

I'm thinking we can make a new `OltpGrpcSpanExporterModule` and add an additional constructor to `OpenTelemetryModule` that takes a `Module`. The existing constructor will pass `OltpGrpcSpanExporterModule` to the new constructor, preserving the existing behavior. Applications that want a different exporter can provide their own as a module.

What do you think?

findepi:
> The use case is to allow adding a custom span processor that will capture and process spans internally (instead of exporting them to the external sink like Jeager)

Is the use-cases for production or test purposes?
For test purposes i created https://github.com/airlift/airlift/pull/1090. That PR, however, is definitely not suitable for production uses (naming; not configuration-based -- API designed for tests' consumption).

PR #1081, opened 2023-07-07
Test on JDK 21-ea

wendigo:
To ensure forward-compatibility with the new Java LTS.

This adds a new matrix configuration that tests airlift with:

- JDK 17 and 17 as a target
- JDK 21-ea and 17 as a target
- JDK 21-ea and 21 as a target (not required to pass a build)


wendigo:
cc @electrum

electrum (approved):
Please drop last commit

PR #1082, opened 2023-07-11
Allow overriding default span processor

wendigo:
Alternative to https://github.com/airlift/airlift/pull/1080

### DIFF
--- a/tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
+++ b/tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
@@ -1,21 +1,19 @@
 package io.airlift.tracing;
 
 import com.google.inject.Binder;
-import com.google.inject.Module;
 import com.google.inject.Provides;
 import com.google.inject.Singleton;
+import io.airlift.configuration.AbstractConfigurationAwareModule;
 import io.airlift.node.NodeInfo;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.common.AttributesBuilder;
 import io.opentelemetry.api.trace.propagation.W3CTraceContextPropagator;
 import io.opentelemetry.context.propagation.ContextPropagators;
-import io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporter;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
-import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
-import io.opentelemetry.sdk.trace.export.SpanExporter;
+import io.opentelemetry.sdk.trace.SpanProcessor;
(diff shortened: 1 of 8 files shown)


wendigo:
ptal @electrum

wendigo:
ping @electrum

tracing/src/main/java/io/airlift/tracing/TracingModule.java
```diff
@@ -15,19 +16,27 @@
 public class TracingModule
         extends AbstractConfigurationAwareModule
 {
+    private final Module module;
     private final String serviceName;
     private final String serviceVersion;
 
-    public TracingModule(String serviceName, String serviceVersion)
+    public TracingModule(Module module, String serviceName, String serviceVersion)
```
findepi, line 23: Typically module dependencies aren't that explicit.
For example you need to install JsonModule  and NodeModule modules when you want to use TracingModule.

If this `module` parameter is something special, give it a name.

findepi:
Alternative, presumably simpler, approach purely for testing purposes -- https://github.com/airlift/airlift/pull/1090

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
```diff
@@ -25,13 +23,12 @@
...
 import static com.google.common.base.StandardSystemProperty.OS_VERSION;
 import static com.google.common.base.Strings.nullToEmpty;
-import static io.airlift.configuration.ConfigBinder.configBinder;
 import static io.airlift.tracing.Tracing.attribute;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 
 public class OpenTelemetryModule
-        implements Module
+        extends AbstractConfigurationAwareModule
```
findepi, line 31: in `Allow overriding default span processor` commit, why OpenTelemetryModule becomes AbstractConfigurationAwareModule? especially given the fact it no longer binds any configs

tracing/src/test/java/io/airlift/tracing/TestTracingModule.java
```diff
@@ -0,0 +1,82 @@
...
+        assertThat(instance.toString())
+                .contains("BatchSpanProcessor{spanExporter=io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporter");
+    }
+
+    @Test
+    public void testCreateTracingWithCustomSpanProcessor()
+    {
+        Bootstrap bootstrap = new Bootstrap(
+                new TracingModule(binder -> {
+                    binder.bind(SpanProcessor.class).toInstance(SimpleSpanProcessor.create(OtlpHttpSpanExporter.getDefault()));
```
findepi, line 52: you can use `InMemorySpanExporter.create()`
and then verify the span processor is being used to export some spans

i think it would be cleaner than the instanceof and toString-based assertions below.

findepi (commented):
Some comments for `Allow overriding default span processors`
Would it make sense to separate the changes from `Add tracing max-batch and max-queue size config` commit into independent PRs? I think the two commits address independent concerns.

electrum:
Repalced by https://github.com/airlift/airlift/pull/1107. You can open a new PR for the batch configs if you need.

PR #1084, opened 2023-07-23
Support streaming execution in the HttpClient

Randgalt:
Support end-to-end streaming. This is useful for endpoints that return
a large list of entities. Instead of needing to queue that entire list
in memory in both the server and client the entities are streamed.
This has even greater impact when there is a client in the middle:
client -> server A -> server B -> back to client.

The flow is:

- Client sends request to server
- Server begins to stream entities
- Client reads entities one by one

This necessitates giving the client ownership of the internal
input stream.

Additionally, adds JsonStreamingResponse to make it simpler/convenient to stream lists
of JSON entities.

### DIFF
--- a/http-client/pom.xml
+++ b/http-client/pom.xml
@@ -96,6 +96,16 @@
         </dependency>
 
         <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+        </dependency>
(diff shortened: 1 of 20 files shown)


Randgalt:
cc @electrum, @findepi

Randgalt:
I thought of something better... Stay tuned.

PR #1085, opened 2023-07-23
Support streaming execution in the HttpClient

Randgalt:
Support end-to-end streaming. This is useful for endpoints that return
a large list of entities. Instead of needing to queue that entire list
in memory in both the server and client the entities are streamed.
This has even greater impact when there is a client in the middle:
client -> server A -> server B -> back to client.

The flow is:

- Client sends request to server
- Server begins to stream entities
- Client reads entities one by one

This necessitates giving the client ownership of the internal
input stream.

Additionally, adds JsonStreamingResponse to make it simpler/convenient to stream lists
of JSON entities.

### DIFF
--- a/http-client/pom.xml
+++ b/http-client/pom.xml
@@ -96,6 +96,11 @@
         </dependency>
 
         <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>jakarta.annotation</groupId>
             <artifactId>jakarta.annotation-api</artifactId>
         </dependency>
@@ -145,6 +150,12 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
+            <scope>runtime</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
             <scope>runtime</scope>
         </dependency>
 
--- a/http-client/src/main/java/io/airlift/http/client/JsonStreamingResponseHandler.java
+++ b/http-client/src/main/java/io/airlift/http/client/JsonStreamingResponseHandler.java
@@ -1,0 +1,173 @@
+package io.airlift.http.client;
+
+import com.fasterxml.jackson.core.JsonToken;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ListMultimap;
+import com.google.common.primitives.Ints;
+import io.airlift.json.JsonCodec;
+import io.airlift.json.JsonCodecParser;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.util.NoSuchElementException;
+import java.util.Set;
+
+import static io.airlift.http.client.JsonResponseHandler.validateResponse;
+import static io.airlift.http.client.ResponseHandlerUtils.propagate;
+import static java.util.Objects.requireNonNull;
+
+public class JsonStreamingResponseHandler<T>
+        implements StreamingResponseHandler<T>
+{
+    private final Set<Integer> successfulResponseCodes;
+    private final JsonCodec<T> codec;
+
+    public static <T> JsonStreamingResponseHandler<T> createJsonStreamingResponse(JsonCodec<T> codec)
+    {
+        return new JsonStreamingResponseHandler<>(codec, 200, 201, 202, 203, 204, 205, 206);
+    }
+
+    public static <T> JsonStreamingResponseHandler<T> createJsonStreamingResponse(JsonCodec<T> codec, int firstSuccessfulResponseCode, int... otherSuccessfulResponseCodes)
+    {
+        return new JsonStreamingResponseHandler<>(codec, firstSuccessfulResponseCode, otherSuccessfulResponseCodes);
+    }
+
+    private JsonStreamingResponseHandler(JsonCodec<T> codec, int firstSuccessfulResponseCode, int... otherSuccessfulResponseCodes)
+    {
+        this.codec = requireNonNull(codec, "codec is null");
+        successfulResponseCodes = ImmutableSet.<Integer>builder().add(firstSuccessfulResponseCode).addAll(Ints.asList(otherSuccessfulResponseCodes)).build();
+    }
+
+    @Override
+    public StreamingResponseIterator<T> handle(Request request, StreamingResponse response)
+    {
+        try {
+            validateResponse(request, response, successfulResponseCodes);
+        }
+        catch (UnexpectedResponseException e) {
+            throw cleanup(request, null, response, e);
+        }
+
+        JsonCodecParser<T> jsonCodecParser = null;
+        try {
+            jsonCodecParser = codec.newParser(response.getInputStream());
+            if (jsonCodecParser.nextToken() != JsonToken.START_ARRAY) {
+                throw new UnexpectedResponseException("Expected JSON array", request, response);
+            }
+            jsonCodecParser.nextToken();
+        }
+        catch (IOException e) {
+            throw cleanup(request, jsonCodecParser, response, new UncheckedIOException(e));
+        }
+
+        return new StreamingIterator<>(request, response, jsonCodecParser);
+    }
+
+    private static RuntimeException cleanup(Request request, JsonCodecParser<?> jsonCodecParser, StreamingResponse response, RuntimeException cause)
+    {
+        if ((jsonCodecParser != null) && !jsonCodecParser.isClosed()) {
+            try {
+                jsonCodecParser.close();
+            }
+            catch (Throwable t) {
+                if (cause == null) {
+                    cause = propagate(request, t);
+                }
+                else {
+                    cause.addSuppressed(propagate(request, t));
+                }
+            }
+        }
+
+        try {
+            response.close();
+        }
+        catch (Throwable t) {
+            if (cause == null) {
+                cause = propagate(request, t);
+            }
+            else {
+                cause.addSuppressed(propagate(request, t));
+            }
+        }
+
+        return cause;
+    }
+
+    private static class StreamingIterator<T>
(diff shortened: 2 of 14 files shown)


Randgalt:
cc @electrum  @findepi  @losipiuk

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -703,26 +727,39 @@ public void onContent(Response response, ByteBuffer content)
...
-        JettyResponse jettyResponse = null;
         T value;
-        try {
-            jettyResponse = new JettyResponse(response, listener.getInputStream());
-            value = responseHandler.handle(request, jettyResponse);
+
+        if (isStreaming) {
+            JettyResponse jettyResponse;
+            try {
+                jettyResponse = new JettyResponse(response, listener.getInputStream());
```
Praveen2112, line 735: Can we extract this above the `if` statement ? Do we need to wrap between `try`

http-client/src/main/java/io/airlift/http/client/JsonStreamingResponseHandler.java
```diff
@@ -55,26 +67,45 @@ public StreamingResponseIterator<T> handle(Request request, StreamingResponse re
...
+        parser.nextToken();
+    }
+
+    @SuppressWarnings("unused")
+    protected JsonParser buildParser(Request request, StreamingResponse response)
+            throws IOException
+    {
+        JsonParser parser;
+        parser = objectMapper.createParser(new InputStreamReader(response.getInputStream(), StandardCharsets.UTF_8));
+        return parser;
```
Praveen2112, line 100: We can inline them right ?

http-client/src/main/java/io/airlift/http/client/JsonStreamingResponseHandler.java
```diff
@@ -41,11 +42,22 @@ public static <T> JsonStreamingResponseHandler<T> createJsonStreamingResponse(Js
...
+        this.closeStreamOnCompletion = closeStreamOnCompletion;
+
         objectMapper = codec.mapper();
-        JavaType javaType = objectMapper.getTypeFactory().constructType(codec.getType());
-        objectReader = objectMapper.readerFor(javaType);
+        objectReader = codec.mapper().readerFor(objectMapper.getTypeFactory().constructType(codec.getType()));
+    }
 
-        successfulResponseCodes = ImmutableSet.<Integer>builder().add(firstSuccessfulResponseCode).addAll(Ints.asList(otherSuccessfulResponseCodes)).build();
+    public JsonStreamingResponseHandler<T> withoutClosingStream()
```
Praveen2112, line 58: Do we use this withoutClosingStream.

Praveen2112 (commented):
LGTM. License header is missing

http-client/src/main/java/io/airlift/http/client/JsonStreamingResponseHandler.java
```diff
@@ -0,0 +1,184 @@
+package io.airlift.http.client;
```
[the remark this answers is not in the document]
Randgalt: In my recent airlift PRs @electrum has said we don't want license headers. The old header was Proofpoint and that doesn't apply anymore.

http-client/src/main/java/io/airlift/http/client/JsonStreamingResponseHandler.java
```diff
@@ -0,0 +1,184 @@
...
+        try {
+            validateResponse(request, response, successfulResponseCodes);
+        }
+        catch (UnexpectedResponseException e) {
+            throw cleanup(request, null, response, e);
+        }
+
+        JsonParser parser;
+        try {
+            parser = objectMapper.createParser(new InputStreamReader(response.getInputStream(), StandardCharsets.UTF_8));
```
oskar-szwajkowski: shouldn't
```
new InputStreamReader(response.getInputStream(), StandardCharsets.UTF_8)
```
be moved to previous line using try-with-resources?

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -807,6 +815,103 @@ public void testHttpStatusListenerException()
...
+            throws Exception
+    {
+        int qty = 25;
+
+        List<Thing> things = IntStream.range(0, qty).mapToObj(i -> new Thing(Integer.toString(i), i)).collect(toImmutableList());
+        String json = new ObjectMapperProvider().get().writeValueAsString(things);
+        JsonCodec<Thing> codec = jsonCodec(Thing.class);
+
+        servlet.setResponseBody(json);
+        servlet.addResponseHeader("Content-Type", "application/json");
```
oskar-szwajkowski, line 831: I wonder if support for `application/x-ndjson` is something to consider

http-client/src/main/java/io/airlift/http/client/JsonStreamingResponseHandler.java
Randgalt: re-written

http-client/src/main/java/io/airlift/http/client/JsonStreamingResponseHandler.java
Randgalt: removed

http-client/src/main/java/io/airlift/http/client/JsonStreamingResponseHandler.java
Randgalt: No - the need here is that the parser is not closed. The caller has the responsibility of closing.

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
Randgalt: We can have a new `StreamingResponseHandler` for that, adding it when Airlift's other JSON hander gets support.

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
Randgalt: I don't think that would help. In the non-streaming case it gets the entire value after allocating the `JettyResponse`. The streaming case doesn't do that.

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -215,6 +215,11 @@ public byte[] toJsonBytes(T instance)
         }
     }
 
+    public ObjectMapper mapper()
+    {
+        return mapper;
```
findepi, line 220: Mapper is mutable (and even potentially shared `io.airlift.json.JsonCodec#OBJECT_MAPPER_SUPPLIER`). This breaks encapsulation. This is a no-go from my perspective.

Randgalt: I'll find another solution.

Randgalt: @findepi I've come up with an alternate. PTAL.

oskar-szwajkowski: also `return mapper.copy();` would do, right?

json/src/main/java/io/airlift/json/JsonCodecParser.java
```diff
@@ -0,0 +1,96 @@
...
+        return parser.nextIntValue(defaultValue);
+    }
+
+    public long nextLongValue(long defaultValue)
+            throws IOException
+    {
+        return parser.nextLongValue(defaultValue);
+    }
+
+    public Boolean nextBooleanValue()
```
losipiuk, line 91: why `Boolean` hear (not `bool`) and `int` and `long` above?

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -703,26 +727,39 @@ public void onContent(Response response, ByteBuffer content)
...
+            catch (Throwable e) {
+                throw propagate(request, e);
+            }
+
+            Runnable finalizer = () -> finalizeExecution(request, span, requestStart, requestSize, responseStart, jettyResponse);
+            try {
+                value = responseHandler.handle(request, new JettyStreamingResponse(jettyResponse, span, finalizer));
+            }
+            catch (RuntimeException e) {
+                throw new StreamingException(e, finalizer);
```
losipiuk, line 746: why pass `finalizer` through `StreamingException` instead calling it immediately here before exception is thrown?

http-client/src/main/java/io/airlift/http/client/jetty/JettyStreamingResponse.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public InputStream getInputStream()
+            throws IOException
+    {
+        return response.getInputStream();
+    }
+
+    @Override
+    public void close()
```
losipiuk, line 54: nit: guard from calling more than once.

http-client/pom.xml
```diff
@@ -158,6 +153,12 @@
             <scope>runtime</scope>
         </dependency>
 
+        <dependency>
```
losipiuk, line 156: Merge `Tests for streaming execution` into commit which adds support for streaming responses.

losipiuk (approved):

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
Randgalt: So that `span.end()` is the very last thing called. As I read the code this needs to be the last operation. For example, `finalizeExecution()` sets an attribute on the span. It could be that I don't understand the span library well enough.

Randgalt: So that `span.end()` is the very last thing called. As I read the code this needs to be the last operation. For example, the catch in `execute` calls `span.setStatus()` and that should happen before `span.end()`.

wendigo:
@Randgalt please rebase

Randgalt:
Closing in favor of https://github.com/airlift/airlift/pull/1174

PR #1090, opened 2023-09-04
Add testing tracing module for in-memory tracing

findepi:
Add `TestingTracingModule` with enhances `TracingModule` with ability to override `SpanProcessor`. This can be used e.g. for in-memory tracing.

### DIFF
--- a/tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
+++ b/tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
@@ -14,9 +14,12 @@ import io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporter;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
+import io.opentelemetry.sdk.trace.SpanProcessor;
 import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
+
+import java.util.Optional;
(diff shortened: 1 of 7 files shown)


nineinchnick:
Here's an earlier approach, there's some discussion there: https://github.com/airlift/airlift/pull/1082

findepi:
Thanks for linking the PRs!

I definitely biased when comparing my code with some else's code, but I think the approach taken here is simpler.
Not for me to decide though.

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
```diff
@@ -69,12 +74,13 @@ public OpenTelemetry createOpenTelemetry(NodeInfo nodeInfo, OpenTelemetryConfig
 
         Resource resource = Resource.getDefault().merge(Resource.create(attributes.build()));
 
-        SpanExporter spanExporter = OtlpGrpcSpanExporter.builder()
-                .setEndpoint(config.getEndpoint())
-                .build();
-
         SdkTracerProvider tracerProvider = SdkTracerProvider.builder()
-                .addSpanProcessor(BatchSpanProcessor.builder(spanExporter).build())
+                .addSpanProcessor(spanProcessor.orElseGet(() -> {
```
nineinchnick, line 78: If there is a custom `spanProcessor`, the exporter endpoint configuration property is silently ignored.

wendigo: @findepi I've consulted my change with David and he raised similiar concern for my initial attempt

findepi: You're right. We shouldn't bind the endpoint config when `spanProcessor` is provided. I can address this once we have clarity which way we want to go.

electrum:
Replaced by https://github.com/airlift/airlift/pull/1107

PR #1094, opened 2023-10-05
Config option to enable setting SO_KEEPALIVE on the http client socket

Randgalt:


http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -661,6 +663,18 @@ public HttpClientConfig setLogCompressionEnabled(boolean logCompressionEnabled)
         return this;
     }
 
+    @NotNull
+    public Optional<Duration> getTcpKeepAliveIdleTime()
+    {
+        return tcpKeepAliveIdleTime;
+    }
+
+    public HttpClientConfig setTcpKeepAliveIdleTime(Duration tcpKeepAliveIdleTime)
```
electrum, line 672: This is missing `@Config`

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -423,6 +436,22 @@ public JettyHttpClient(
...
     }
 
+    private static void setKeepAlive(SelectableChannel selectable, Duration tcpKeepAliveIdleTime)
+    {
+        if (selectable instanceof NetworkChannel channel) {
+            try {
+                channel.setOption(StandardSocketOptions.SO_KEEPALIVE, true);
+                channel.setOption(ExtendedSocketOptions.TCP_KEEPIDLE, toIntExact(tcpKeepAliveIdleTime.roundTo(SECONDS)));
+            }
+            catch (Throwable e) {
```
electrum, line 446: Don’t carch Throwable as it can hide internal errors. Instead, catch the expected error.

PR #1101, opened 2023-10-25
Add detection of invalid security sensitive config

wendigo:
cc @findepi 

to prevent situations like https://github.com/trinodb/trino/pull/19519

### DIFF
--- a/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
@@ -67,6 +67,8 @@ public class Bootstrap
 
     private Map<String, String> requiredConfigurationProperties;
     private Map<String, String> optionalConfigurationProperties;
+    private List<String> securitySensitivePatterns;
+
     private boolean initializeLogging = true;
     private boolean quiet;
     private boolean strictConfig = true;
@@ -116,6 +118,15 @@ public class Bootstrap
             this.optionalConfigurationProperties = new TreeMap<>();
         }
         this.optionalConfigurationProperties.putAll(optionalConfigurationProperties);
+        return this;
+    }
+
+    public Bootstrap setSensitiveConfigPatterns(List<String> patterns)
+    {
+        if (this.securitySensitivePatterns == null) {
+            this.securitySensitivePatterns = new ArrayList<>();
+        }
+        this.securitySensitivePatterns.addAll(patterns);
         return this;
     }
 
--- a/configuration/src/test/java/io/airlift/configuration/TestConfig.java
+++ b/configuration/src/test/java/io/airlift/configuration/TestConfig.java
@@ -45,6 +45,7 @@ import static java.lang.annotation.ElementType.FIELD;
 import static java.lang.annotation.ElementType.METHOD;
 import static java.lang.annotation.ElementType.PARAMETER;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.testng.Assert.assertEquals;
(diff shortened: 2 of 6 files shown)


bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -67,6 +67,8 @@ public class Bootstrap
 
     private Map<String, String> requiredConfigurationProperties;
     private Map<String, String> optionalConfigurationProperties;
+    private List<String> securitySensitivePatterns;
```
kokosing, line 70: Is pattern a regex? Below it looks like you are using `contains`. Can we have a good default list? So one does not need to configure whenever `Boostrap` is used?

kokosing (commented):
Nice!

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
wendigo: @kokosing it's as substring, not a pattern, I could make it a regex

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -119,6 +121,15 @@ public Bootstrap setOptionalConfigurationProperties(Map<String, String> optional
         return this;
     }
 
+    public Bootstrap setSensitiveConfigPatterns(List<String> patterns)
+    {
+        if (this.securitySensitivePatterns == null) {
+            this.securitySensitivePatterns = new ArrayList<>();
+        }
+        this.securitySensitivePatterns.addAll(patterns);
```
findepi, line 129: setSensitiveConfigPatterns -> add....

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -119,6 +121,15 @@ public Bootstrap setOptionalConfigurationProperties(Map<String, String> optional
         return this;
     }
 
+    public Bootstrap setSensitiveConfigPatterns(List<String> patterns)
+    {
+        if (this.securitySensitivePatterns == null) {
+            this.securitySensitivePatterns = new ArrayList<>();
```
findepi, line 127: why not initialize the list directly on the field?

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -119,6 +121,15 @@ public Bootstrap setOptionalConfigurationProperties(Map<String, String> optional
         return this;
     }
 
+    public Bootstrap setSensitiveConfigPatterns(List<String> patterns)
```
findepi, line 124: unused?

configuration/src/test/java/io/airlift/configuration/TestConfig.java
```diff
@@ -204,6 +205,33 @@ public void testDetectsNoConfigAnnotations()
         }
     }
 
+    @Test
+    public void testDetectsSecuritySensitiveConfig()
+    {
+        try {
+            Injector injector = createInjector(Collections.<String, String>emptyMap(), ImmutableList.of("access-key"), createModule(Key.get(SecuritySensitiveConfig.class), SecuritySensitiveConfig.class, null));
```
findepi, line 212: I understand that this provides a form of a unit test that the annotation is there.
However, that means i either need to remember to add annotation or to add a test (and then add annotation), which relies on remembering.

so the add value of this is that, once i have a test, one cannot remove the annotation without updating the test

findepi:
if i am reading this correctly, then I don't understand how it helps find configs missing security sensitive annotation.
how does it work?

wendigo:
@findepi this will fail on runtime while constructing config with keys that are not annotated but match patterns

PR #1106, opened 2023-11-14
Deprecate attribute utility methods in Tracing

electrum:
OpenTelemetry explicitly ignores null or empty values.


tracing/src/main/java/io/airlift/tracing/Tracing.java
```diff
@@ -17,11 +17,13 @@ public static Tracer noopTracer()
         return TracerProvider.noop().get("noop");
     }
 
+    @Deprecated
```
dain, line 20: for removal?

dain (approved):

PR #1110, opened 2023-11-24
Update dependencies & improve code style

wendigo:


wendigo:
@martint @electrum PTAL

wendigo:
ping @martint @electrum

PR #1111, opened 2023-12-12
Update Jackson to 2.16.0

wendigo:
Disable write constraints that were introduced in that version


wendigo:
cc @electrum @martint

wendigo:
Ping @electrum @martint

wendigo:
Ping @martint

PR #1113, opened 2024-01-03
Match lifecycle annotations coming from different classloader

wendigo:
If there are multiple jakarta.annotations.PreDestroy on a class-path, it's safer to match annotations by name, rather than by class equivalence.


bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleMethods.java
```diff
@@ -28,6 +28,7 @@
 import java.util.Set;
```
kokosing, line 28: > If there are multiple jakarta.annotations.PreDestroy on a class-path, it's safer to match annotations

Should this class part of the SPI and use single classloader?

PR #1119, opened 2024-02-06
NodeConfig errors display regex that's supposed to match

augurpl:
Currently node config errors shown when environment, pool or id have quite cryptic message: is malformed. I'd like to make this more readable, so I've added regex pattern that validates field to the message.

### DIFF
--- a/node/src/test/java/io/airlift/node/TestNodeConfig.java
+++ b/node/src/test/java/io/airlift/node/TestNodeConfig.java
@@ -94,11 +94,11 @@ public class TestNodeConfig
                 .setEnvironment("test")
                 .setNodeId(UUID.randomUUID().toString()));
 
-        assertFailsValidation(new NodeConfig().setNodeId("abc/123"), "nodeId", "is malformed", Pattern.class);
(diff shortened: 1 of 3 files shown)


augurpl:
@electrum this is fix for one of SEP's issues

node/src/test/java/io/airlift/node/TestNodeConfig.java
```diff
@@ -94,11 +94,11 @@ public void testValidations()
                 .setEnvironment("test")
                 .setNodeId(UUID.randomUUID().toString()));
 
-        assertFailsValidation(new NodeConfig().setNodeId("abc/123"), "nodeId", "is malformed", Pattern.class);
+        assertFailsValidation(new NodeConfig().setNodeId("abc/123"), "nodeId", NodeConfig.ID_REGEXP_ERROR, Pattern.class);
```
electrum, line 97: Let's put the literal string here, so we can tell that the error message is what we expect.

augurpl: Good point. Updated.

electrum (approved):

PR #1120, opened 2024-02-14
Add support for escaping commas when passing list of strings

vinay-kosaraju:
When passing jmx object names with multiple properties, object names were incorrectly parsed and throws an `MalformedObjectNameException`

```
Caused by: javax.management.MalformedObjectNameException: Domain part must be specified
	at java.management/javax.management.ObjectName.construct(ObjectName.java:467)
	at java.management/javax.management.ObjectName.<init>(ObjectName.java:1405)
	at io.airlift.openmetrics.MetricsConfig.setJmxObjectNames(MetricsConfig.java:42)
	... 31 more
```

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
@@ -61,6 +61,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.regex.Pattern;
 
 import static com.google.common.base.CaseFormat.LOWER_CAMEL;
 import static com.google.common.base.CaseFormat.UPPER_CAMEL;
@@ -78,7 +79,7 @@ public class ConfigurationFactory
     private static final Validator VALIDATOR;
 
     private static final TypeToken<List<String>> LIST_OF_STRINGS_TYPE_TOKEN = new TypeToken<List<String>>() {};
-    private static final Splitter VALUE_SPLITTER = Splitter.on(",").omitEmptyStrings().trimResults();
+    private static final Splitter VALUE_SPLITTER = Splitter.on(Pattern.compile("(?<!\\\\),")).omitEmptyStrings().trimResults();
 
     static {
         ClassLoader currentClassLoader = Thread.currentThread().getContextClassLoader();
@@ -667,6 +668,7 @@ public class ConfigurationFactory
 
         if (LIST_OF_STRINGS_TYPE_TOKEN.isSubtypeOf(TypeToken.of(type))) {
             return VALUE_SPLITTER.splitToStream(value)
+                    .map((v) -> v.replaceAll("\\\\,", ","))
                     .collect(toImmutableList());
         }
 
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
@@ -17,9 +17,9 @@ package io.airlift.configuration;
 
 import com.google.common.collect.ImmutableSortedMap;
 
-import java.io.FileInputStream;
 import java.io.IOException;
-import java.io.InputStream;
+import java.io.StringReader;
+import java.nio.file.Path;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
@@ -27,6 +27,9 @@ import java.util.TreeMap;
 
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.Maps.fromProperties;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.nio.file.Files.readAllLines;
+import static java.util.stream.Collectors.joining;
 
 public final class ConfigurationLoader
 {
@@ -57,9 +60,10 @@ public final class ConfigurationLoader
             throws IOException
     {
         Properties properties = new Properties();
-        try (InputStream inputStream = new FileInputStream(path)) {
-            properties.load(inputStream);
-        }
+        String escapedProperties = readAllLines(Path.of(path), UTF_8).stream()
+                .map(line -> line.replaceAll("\\\\,", "\\\\\\\\,"))
+                .collect(joining("\n"));
+        properties.load(new StringReader(escapedProperties));
 
         return fromProperties(properties).entrySet().stream()
                 .collect(toImmutableMap(Entry::getKey, entry -> entry.getValue().trim()));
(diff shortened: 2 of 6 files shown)


openmetrics/src/main/java/io/airlift/openmetrics/MetricsConfig.java
```diff
@@ -39,7 +39,7 @@ public MetricsConfig setJmxObjectNames(List<String> jmxObjectNames)
 
         for (String jmxObjectName : jmxObjectNames) {
             try {
-                objectNames.add(new ObjectName(jmxObjectName));
+                objectNames.add(new ObjectName(jmxObjectName.replace("\\", "")));
```
mattstep, line 42: I wonder if we could be more generic and support removing the escaping characters in the configuration module.

mattstep (approved):

lozbrown:
@vinay-kosaraju  any chance the @ConfigDescription can be updated with details of the delimiter and escaping?

Eg :"comma delimeted, commas in object names just be escaped with backlash"

Or something to that affect?

lozbrown:
> @vinay-kosaraju  any chance the @ConfigDescription can be updated with details of the delimiter and escaping?
> 
> Eg :"comma delimeted, commas in object names just be escaped with backlash"
> 
> Or something to that affect?

@vinay-kosaraju thanks for the update.

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -78,7 +79,7 @@ public class ConfigurationFactory
     private static final Validator VALIDATOR;
 
     private static final TypeToken<List<String>> LIST_OF_STRINGS_TYPE_TOKEN = new TypeToken<List<String>>() {};
-    private static final Splitter VALUE_SPLITTER = Splitter.on(",").omitEmptyStrings().trimResults();
+    private static final Splitter VALUE_SPLITTER = Splitter.on(Pattern.compile("(?<!\\\\),")).omitEmptyStrings().trimResults();
```
elonazoulay, line 82: lgtm: object names cannot have `\`'s in them (suggested to use identifier chars), i.e. so no need to worry about multiple escapes, etc.

elonazoulay (approved):

configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
```diff
@@ -57,8 +60,14 @@ public static Map<String, String> loadPropertiesFrom(String path)
             throws IOException
     {
         Properties properties = new Properties();
-        try (InputStream inputStream = new FileInputStream(path)) {
-            properties.load(inputStream);
+        try (Reader fileReader = new FileReader(path, UTF_8);
+                BufferedReader reader = new BufferedReader(fileReader)) {
+            StringBuilder escapedPropertiesBuilder = new StringBuilder();
+            String currentLine;
```
elonazoulay, line 66: can also use ```
String escapedProperties = Files.readAllLines(path, UTF_8).stream()
  .map(line -> escapedPropertiesBuilder.append(line.replaceAll(...)))
  .collector(joining("\n"));
```

File: openmetrics/src/main/java/io/airlift/openmetrics/MetricsConfig.java [hunk 4]
```diff
@@ -32,7 +32,7 @@ public List<ObjectName> getJmxObjectNames()
     }
 
     @Config("openmetrics.jmx-object-names")
-    @ConfigDescription("JMX object names to include when retrieving all metrics.")
+    @ConfigDescription("Comma delimited JMX object names to include when retrieving all metrics. Escape commas in object names with backslash")
```
elonazoulay, line 35: nit: `backslash.`

lozbrown:
@martint any chance we can get this merged in the hope of seeing it get into trino 440?

configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
vinay-kosaraju: Done, updated

configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
```diff
@@ -57,9 +60,10 @@ public static Map<String, String> loadPropertiesFrom(String path)
             throws IOException
     {
         Properties properties = new Properties();
-        try (InputStream inputStream = new FileInputStream(path)) {
-            properties.load(inputStream);
-        }
+        String escapedProperties = readAllLines(Path.of(path), UTF_8).stream()
+                .map(line -> line.replaceAll("\\\\,", "\\\\\\\\,"))
+                .collect(joining("\n"));
+        properties.load(new StringReader(escapedProperties));
```
martint, line 66: What's the purpose of this?

As an aside, this is a behavioral change. By default, properties files are expected to be encoded using ISO 8859-1. I think it makes sense to change to UTF-8 nowadays, but this should be done separately and intentionally.

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -667,6 +668,7 @@ private static Object coerce(Class<?> type, String value)
 
         if (LIST_OF_STRINGS_TYPE_TOKEN.isSubtypeOf(TypeToken.of(type))) {
             return VALUE_SPLITTER.splitToStream(value)
+                    .map((v) -> v.replaceAll("\\\\,", ","))
```
electrum, line 671: Nit: no need for parenthesis around a single lambda variable

configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
```diff
@@ -57,9 +60,10 @@ public static Map<String, String> loadPropertiesFrom(String path)
             throws IOException
     {
         Properties properties = new Properties();
-        try (InputStream inputStream = new FileInputStream(path)) {
-            properties.load(inputStream);
-        }
+        String escapedProperties = readAllLines(Path.of(path), UTF_8).stream()
+                .map(line -> line.replaceAll("\\\\,", "\\\\\\\\,"))
```
electrum, line 64: Why are we changing the way that properties are loaded? The intent of this PR is to change how `ConfigurationFactory` parses a value into a list for `List<String>` config properties. The above is effectively changing how properties files are parsed -- they no longer adhere to the Java properties specification.

electrum (changes requested):

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
electrum: Note that this is changing how parsing is done for all config values of type `List<String>`. It's not limited to just object names.

electrum (changes requested):
This escape handling seems incomplete. It allows backslashes as literals, but doesn't allow a backslash to precede a comma, and changes the behavior for existing configs.

I think that if we want to support escaping, it should be consistent and complete: a backslash escapes any character, including itself. A backslash at the end of the string is illegal.

Alternatively, we could change `MetricsConfig` to accept a `String` and do the splitting there. Maybe use a pipe rather than a comma, since that can't appear in an object name, and thus users won't need to escape commas. The downside is that this would be a silent breaking change.

electrum:
Alternative: https://github.com/airlift/airlift/pull/1129

PR #1124, opened 2024-03-02
Disable blocking in local http log

elonazoulay:
Related to https://jira.qos.ch/browse/LOGBACK-1424


elonazoulay:
Can repro the failure locally, and it succeeds w/o `setNeverBlock`. Investigating.

wendigo:
@elonazoulay do we still need this?

PR #1127, opened 2024-03-13
Enable legacy URI mode in Jetty and Servlets

Randgalt:
Latest versions of Servlet/Jetty specs disallow encoded slashes in paths. However, there is a legacy mode currently that turns off the compliance validation for this and, thus, allows encoded slashes to work as they have in the past. Enable this legacy mode and add a JAX-RS test to make sure it's passed through.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 243
 
 - Use pipe as separator for openmetrics object names config
+- Add support for Jetty legacy URI compliance
 
 242
 - Update airbase to 151
--- a/http-server/src/main/java/io/airlift/http/server/HttpServer.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpServer.java
@@ -102,7 +102,6 @@ public class HttpServer
     private final Server server;
     private final MonitoredQueuedThreadPoolMBean monitoredQueuedThreadPoolMBean;
     private final MonitoredQueuedThreadPoolMBean monitoredAdminQueuedThreadPoolMBean;
-    private final boolean showStackTrace;
     private final DelimitedRequestLog requestLog;
     private ConnectionStats httpConnectionStats;
     private ConnectionStats httpsConnectionStats;
@@ -154,7 +153,8 @@ public class HttpServer
         }
         server = new Server(threadPool);
         this.monitoredQueuedThreadPoolMBean = new MonitoredQueuedThreadPoolMBean(threadPool);
-        showStackTrace = config.isShowStackTrace();
+
+        boolean showStackTrace = config.isShowStackTrace();
 
         this.sslContextFactory = maybeSslContextFactory;
 
--- a/http-server/src/main/java/io/airlift/http/server/HttpServer.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpServer.java
@@ -34,6 +34,7 @@ import org.eclipse.jetty.ee10.servlet.ServletContextHandler;
 import org.eclipse.jetty.ee10.servlet.ServletHolder;
 import org.eclipse.jetty.ee10.servlet.security.ConstraintMapping;
 import org.eclipse.jetty.ee10.servlet.security.ConstraintSecurityHandler;
+import org.eclipse.jetty.http.UriCompliance;
 import org.eclipse.jetty.http2.server.HTTP2CServerConnectionFactory;
 import org.eclipse.jetty.io.ConnectionStatistics;
 import org.eclipse.jetty.jmx.MBeanContainer;
@@ -67,6 +68,7 @@ import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -87,6 +89,8 @@ import static java.util.Collections.list;
 import static java.util.Comparator.naturalOrder;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.Executors.newSingleThreadScheduledExecutor;
+import static org.eclipse.jetty.http.UriCompliance.Violation.AMBIGUOUS_PATH_ENCODING;
+import static org.eclipse.jetty.http.UriCompliance.Violation.AMBIGUOUS_PATH_SEPARATOR;
 import static org.eclipse.jetty.security.Constraint.ALLOWED;
 import static org.eclipse.jetty.util.VirtualThreads.getNamedVirtualThreadsExecutor;
 
@@ -121,6 +125,7 @@ public class HttpServer
             Map<String, String> adminParameters,
             Set<Filter> adminFilters,
             boolean enableVirtualThreads,
+            boolean enableLegacyUriCompliance,
             ClientCertificate clientCertificate,
             MBeanServer mbeanServer,
             LoginService loginService,
@@ -175,6 +180,12 @@ public class HttpServer
         }
         if (config.getMaxResponseHeaderSize() != null) {
             baseHttpConfiguration.setResponseHeaderSize(toIntExact(config.getMaxResponseHeaderSize().toBytes()));
+        }
+
+        if (enableLegacyUriCompliance) {
+            // allow encoded slashes to occur in URI paths
+            UriCompliance uriCompliance = UriCompliance.from(EnumSet.of(AMBIGUOUS_PATH_SEPARATOR, AMBIGUOUS_PATH_ENCODING));
+            baseHttpConfiguration.setUriCompliance(uriCompliance);
         }
 
         // set up HTTP connector
@@ -330,11 +341,11 @@ public class HttpServer
 
         // add handlers to Jetty
         StatisticsHandler statsHandler = new StatisticsHandler();
-        statsHandler.setHandler(createServletContext(theServlet, resources, parameters, filters, tokenManager, loginService, Set.of("http", "https"), showStackTrace));
+        statsHandler.setHandler(createServletContext(theServlet, resources, parameters, filters, tokenManager, loginService, Set.of("http", "https"), showStackTrace, enableLegacyUriCompliance));
 
         ContextHandlerCollection rootHandlers = new ContextHandlerCollection();
         if (theAdminServlet != null && config.isAdminEnabled()) {
-            rootHandlers.addHandler(createServletContext(theAdminServlet, resources, adminParameters, adminFilters, tokenManager, loginService, Set.of("admin"), showStackTrace));
+            rootHandlers.addHandler(createServletContext(theAdminServlet, resources, adminParameters, adminFilters, tokenManager, loginService, Set.of("admin"), showStackTrace, enableLegacyUriCompliance));
         }
         rootHandlers.addHandler(statsHandler);
         StatsRecordingHandler statsRecordingHandler = new StatsRecordingHandler(stats);
@@ -375,13 +386,19 @@ public class HttpServer
             TraceTokenManager tokenManager,
             LoginService loginService,
             Set<String> connectorNames,
-            boolean showStackTrace)
+            boolean showStackTrace,
+            boolean enableLegacyUriCompliance)
     {
         ServletContextHandler context = new ServletContextHandler(ServletContextHandler.NO_SESSIONS);
         if (!showStackTrace) {
             ErrorHandler handler = new ErrorHandler();
             handler.setShowStacks(false);
             context.setErrorHandler(handler);
+        }
+
+        if (enableLegacyUriCompliance) {
+            // allow encoded slashes to occur in URI paths
+            context.getServletHandler().setDecodeAmbiguousURIs(true);
         }
 
         context.addFilter(new FilterHolder(new TimingFilter()), "/*", null);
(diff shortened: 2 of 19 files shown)


wendigo (approved):

wendigo (approved):

wendigo:
Thanks @Randgalt. @electrum can you review and merge this change and then release new Airlift?

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -177,6 +178,9 @@ public HttpServer(
             baseHttpConfiguration.setResponseHeaderSize(toIntExact(config.getMaxResponseHeaderSize().toBytes()));
         }
 
+        // allow encoded slashes to occur in URI paths
+        baseHttpConfiguration.setUriCompliance(UriCompliance.LEGACY);
```
electrum, line 182: I think we should use `UriCompliance.AMBIGUOUS_PATH_SEPARATOR`

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -384,6 +388,9 @@ private static ServletContextHandler createServletContext(Servlet theServlet,
             context.setErrorHandler(handler);
         }
 
+        // allow encoded slashes to occur in URI paths
+        context.getServletHandler().setDecodeAmbiguousURIs(true);
```
electrum, line 392: Can this cause any security issues?

jaxrs/src/test/java/io/airlift/jaxrs/TestLegacyUriMode.java
```diff
@@ -0,0 +1,74 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+import static org.testng.Assert.assertEquals;
+
+public class TestLegacyUriMode
+{
+    @Path("/legacy")
+    public static class MyResource
+    {
+        @GET
+        @Path("{a:.*}/{b:.*}/{c:.*}")
```
electrum, line 38: Isn't this normally done as
```java
@Path("{a}/{b}/{c}")

File: jaxrs/src/test/java/io/airlift/jaxrs/TestLegacyUriMode.java [hunk 4]
```diff
@@ -0,0 +1,74 @@
...
+        public List<String> test(@PathParam("a") String a, @PathParam("b") String b, @PathParam("c") String c)
+        {
+            return ImmutableList.of(a, b, c);
+        }
+    }
+
+    @Retention(RUNTIME)
+    @BindingAnnotation
+    public @interface ForTest
```
electrum, line 47: Nit:
```java
public @interface ForTest {}

File: jaxrs/src/test/java/io/airlift/jaxrs/TestLegacyUriMode.java [hunk 5]
```diff
@@ -0,0 +1,74 @@
...
+
+    @Test
+    public void testLegacyUriMode()
+    {
+        Module module = binder -> {
+            jaxrsBinder(binder).bind(MyResource.class);
+            httpClientBinder(binder).bindHttpClient("test", ForTest.class);
+        };
+        Injector injector = new Bootstrap(
+                module,
```
electrum, line 59: You could inline this

http-server/src/main/java/io/airlift/http/server/HttpServer.java
Randgalt: `UriCompliance.AMBIGUOUS_PATH_SEPARATOR` is a `Violation`. Only `UriCompliance` instances are accepted which is a `final` class. The only one that worked was `LEGACY`.

http-server/src/main/java/io/airlift/http/server/HttpServer.java
Randgalt: The only thing it skips are `UriCompliance.AMBIGUOUS_VIOLATIONS`. So, it seems safe.

jaxrs/src/test/java/io/airlift/jaxrs/TestLegacyUriMode.java
Randgalt: I wasn't sure how it was used. I can try both.

wendigo:
@electrum @Randgalt does it make sense to enable this behaviour per http instance? Make it configurable?

Randgalt:
> @electrum @Randgalt does it make sense to enable this behaviour per http instance? Make it configurable?

Unfortunately it's a server behavior, not a client behavior. We could still make it configurable though but it's for the entire server always.

wendigo:
@Randgalt correct. And not every server requires legacy behaviour

Randgalt:
@wendigo done - the default is new behavior (legacy mode not enabled)

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -495,4 +496,16 @@ public HttpServerConfig setHttp2StreamIdleTimeout(Duration http2StreamIdleTimeou
         this.http2StreamIdleTimeout = http2StreamIdleTimeout;
         return this;
     }
+
+    public boolean isLegacyUriComplianceEnabled()
```
electrum, line 500: This shouldn't be a config for end users, as the usage of this is specific to the application. If the application requires it, disabling the behavior via config would break the application. This should be configurable at the application level via `HttpServerBinder`, similar to virtual threads.

CHANGES
```diff
@@ -1,3 +1,7 @@
+243
+
+- Add support for Jetty/Eclipse legacy URI compliance
```
electrum, line 3: Remove Eclipse

electrum (commented):
Looks good other than the config. I sent a link to the Jetty team and asked them about the implications of enabling this.

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -177,6 +179,11 @@ public HttpServer(
             baseHttpConfiguration.setResponseHeaderSize(toIntExact(config.getMaxResponseHeaderSize().toBytes()));
         }
 
+        if (legacyUriComplianceEnabled) {
+            // allow encoded slashes to occur in URI paths
+            baseHttpConfiguration.setUriCompliance(UriCompliance.LEGACY);
```
joakime, line 184: Don't use LEGACY.

You DEFINITELY do not want to enable all of the other things that LEGACY also brings to the table.  Namely: AMBIGUOUS_EMPTY_SEGMENT, AMBIGUOUS_PATH_ENCODING, AMBIGUOUS_PATH_PARAMETER, and AMBIGUOUS_PATH_SEGMENT.

Be surgical, only turn on what you need/want to allow, not everything.

Eg:

``` java
UriCompliance uriCompliance = UriCompliance.from("RFC3986,AMBIGUOUS_PATH_SEPARATOR");
baseHttpConfiguration.setUriCompliance(uriCompliance);
```

or 

``` java
UriCompliance uriCompliance = UriCompliance.from(EnumSet.of(UriCompliance.Violation.AMBIGUOUS_PATH_SEPARATOR));
baseHttpConfiguration.setUriCompliance(uriCompliance);
```

joakime (changes requested):
Enabling ambiguous and suspicious characters in the path portion of the URL/URI is going to break the following things in the Servlet API.

* Cookies
* Sessions
* RequestDispatcher
* Error Dispatching
* Location response header
* HttpServletRequest .getRequestURL()
* HttpServletRequest .getRequestURI()
* HttpServletResponse sendRedirect()
* Servlet Security Constraints
* url-pattern matching logic in Servlets and Filters

Basically anything that does, or needs, reliable URI Canonicalization.

The Servlet Spec says we should throw a 400 Bad Request if a request contains the `%2F` character (along with other suspicious sequences)

See:
* https://github.com/jakartaee/servlet/blob/6.0.0-RELEASE/spec/src/main/asciidoc/servlet-spec-body.adoc#352-uri-path-canonicalization (Look for the list titled "Rejecting Suspicious Sequences.")

The Servlet spec does not recommend using `%2F` in the path portion of the URI/URL, it is safe to use in the query section.

Also, these problems (the bullet list above) are not new to the Servlet 6 Spec, and has been known about for a number of years.
It's just that Servlet 6 finally stopped trying to fix / band-aid the problems and acknowledged that there is no solution.

Note that you also do not want to use the UriCompliance.LEGACY default, as that's too broad for your narrow use case.
And that UriCompliance is will have more violations in the next release.
See: https://github.com/jetty/jetty.project/pull/11516

joakime:
Also note that JAX-RS itself is no longer supporting suspicious sequences (eg: `%2F`, `/%2e%2e`, etc) in it's own TCK.

https://github.com/jakartaee/rest/issues/1138

Those TCK tests are now disabled and no jax-rs implementation based on ee10 (or newer) is expected to support it.

http-server/src/main/java/io/airlift/http/server/HttpServer.java
Randgalt: Thanks for this. It wasn't obvious how to construct that.

Randgalt:
@electrum, @wendigo, @joakime - comments addressed:

- Changed to use `UriCompliance.from(EnumSet.of(UriCompliance.Violation.AMBIGUOUS_PATH_SEPARATOR));`
- Changed config method to match how virtual threads are enabled

This works but we may want to pass given comments from @joakime. At least this is available should we need some time or decide to manage future changes as they arise.

wendigo:
@Randgalt lgtm % checkstyle

http-server/src/main/java/io/airlift/http/server/HttpServerBinder.java
```diff
@@ -41,6 +41,12 @@ public HttpServerBinder enableVirtualThreads()
         return this;
     }
```
wendigo, line 43: Let's mark it as `@Deprecated`

wendigo (approved):

http-server/src/main/java/io/airlift/http/server/HttpServerBinder.java
Randgalt: done

Randgalt:
@joakime the solution in this PR is:

```
            UriCompliance uriCompliance = UriCompliance.from(EnumSet.of(AMBIGUOUS_PATH_SEPARATOR, AMBIGUOUS_PATH_ENCODING));
            baseHttpConfiguration.setUriCompliance(uriCompliance);
```

and

```
            context.getServletHandler().setDecodeAmbiguousURIs(true);
```

Can these be considered a supported feature of Jetty or is it merely a temporary solution? If temporary, is the direction from Servlet/Jetty that URI paths cannot express path segments with reserved characters in it - ever? My reading of the W3 spec on URIs implies this is far more restrictive than the W3 spec. Regardless, if that's the case we'll need to handle it. If Jetty/Servlet will support loosened validation per this PR into the future then we won't need to rework our APIs. Your advice would be appreciated. Thank you.

wendigo:
PTAL @electrum

http-server/src/test/java/io/airlift/http/server/testing/TestTestingHttpServerWithAllEnabled.java
```diff
@@ -0,0 +1,25 @@
+/*
```
electrum, line 1: Remove copyright headers on new files

electrum (approved):

http-server/src/test/java/io/airlift/http/server/testing/TestTestingHttpServerWithAllEnabled.java
Randgalt: done

joakime:
> Can these be considered a supported feature of Jetty or is it merely a temporary solution? If temporary, is the direction from Servlet/Jetty that URI paths cannot express path segments with reserved characters in it - ever? My reading of the W3 spec on URIs implies this is far more restrictive than the W3 spec. Regardless, if that's the case we'll need to handle it. If Jetty/Servlet will support loosened validation per this PR into the future then we won't need to rework our APIs. Your advice would be appreciated. Thank you.

These APIs are not going anywhere in Jetty.

The W3 URI spec is at odds with the HTTP and URI specs at the IETF in many ways.

Examples:
* W3 URL/URI spec allows `+` in the query section to represent and encoded SPACE, the HTTP and URI specs do not allow that. (Jetty still supports this, and has no plans to stop supporting it)
* The W3 URL/URI spec does not attempt to specify path canonicalization, path normalization, path parameters, path navigation, etc.  Concepts that the IETF does address and is the source of some of our UriCompliance.Violation entries on ambiguous paths..
* The W3 URL/URI spec focuses on how URL should behave in a browser and in javascript, ignoring the rest of the user-agent use cases, something that IETF addresses.

Since Jetty has to support a server side, with a mix of user-agents (browsers, rest clients, spiders, etc) we choose to follow what the IETF and Servlet specs bring to the table, not the W3 spec.  (eg: if we have to take a request path submitted from a user-agent, and apply a resolved request-path for a redirect `Location` line, we follow the IETF spec, as W3 doesn't even cover this)

wendigo:
Since it's temporary workaround for some of the APIs that we control, I think that we should just avoid the problem altogether and adopt more strict specs @Randgalt

PR #1128, opened 2024-03-14
Skip Object defined methods in configuration classes

wendigo:
This skips checking if methods from Object are annotated which causes NoSuchMethodException being thrown multiple times for every object, which forces JVM to construct a stacktrace, which is costly.


electrum (approved):

electrum:
Does this have an actual performance impact somewhere?

PR #1129, opened 2024-03-14
Use pipe as separator for openmetrics object names config

electrum:


mattstep (approved):

lozbrown:
@electrum Can you confirm whether this made it onto Trino 442 or not, so we know to change the configuration?

PR #1131, opened 2024-03-18
Update Jackson to 2.17.0

wendigo:
Add a test to ensure that setting limits for JSON is respected when setReadContraints method is used.

https://github.com/FasterXML/jackson-core/issues/1207

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 243
 
 - Use pipe as separator for openmetrics object names config
+- Update jackson to 2.17.0
 
 242
 - Update airbase to 151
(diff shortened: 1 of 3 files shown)


wendigo:
Alternative to https://github.com/airlift/airlift/pull/1130

CHANGES
```diff
@@ -1,6 +1,7 @@
 243
 
 - Use pipe as separator for openmetrics object names config
+- Update jackson to 2.17.0
```
findepi, line 4: This also fixes a bug (from  ObjectMapperProvider's user perspective). Worth calling it out in CHANGES?

wendigo: this comes with the version upgrade, isn't it? There are other bugs solved between 2.16.1 and 2.17.0 (3 releases in between)

PR #1136, opened 2024-04-02
Test Jetty 12.0.8

wendigo:


wendigo:
@martint @electrum PTAL

http-server/pom.xml
```diff
@@ -13,6 +13,8 @@
 
     <properties>
         <air.main.basedir>${project.parent.basedir}</air.main.basedir>
+        <!-- allow mocking on JDK 23 -->
+        <air.test.jvm.additional-arguments>-Dnet.bytebuddy.experimental=true</air.test.jvm.additional-arguments>
```
martint, line 17: What's the issue if this is not present?

http-server/pom.xml
```diff
@@ -184,7 +184,7 @@
         <dependency>
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcpkix-jdk15on</artifactId>
-            <version>1.68</version>
+            <version>1.70</version>
```
martint, line 187: Why is this different from the other bouncycastle version in the next commit?

wendigo: Separate artifact with separate versioning. I can migrate to jdk18

wendigo: Done. 1.77 will be used only

PR #1137, opened 2024-04-03
Modernize to JDK 21 APIs

wendigo:
Overall make the codebase smaller.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,17 @@ on:
   - pull_request
 
 jobs:
+  error-prone:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: 21
+      - name: Compile with error-prone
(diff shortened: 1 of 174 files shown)


wendigo:
PTAL @electrum @martint

wendigo:
I've checked necessary Trino changes: https://github.com/wendigo/trino/pull/new/serafin/airlift-jdk-21

event/src/main/java/io/airlift/event/client/AbstractEventClient.java
```diff
@@ -50,6 +50,5 @@ public final <T> ListenableFuture<Void> post(EventGenerator<T> eventGenerator)
         return immediateFuture(null);
     }
 
-    protected abstract <T> void postEvent(T event)
-            throws IOException;
```
dain, line 54: The removal of exceptions on "production" classes that can be extended, needs to be thought about.  In this case, and code extending this that needs to throw an IOException would break.

wendigo: I've dropped this commit. Anything else that could break compatibility?

wendigo:
Closing this PR because of the lack of interest from project maintainers.

PR #1139, opened 2024-04-04
Allow providing a custom ErrorHandler

wendigo:
This will allow providing a custom Jetty error handler in Trino that always returns a JSON-encoded error message that can be interpreted correctly by internal http clients.

Improves error reporting in cases like: https://github.com/trinodb/trino/issues/18843

### DIFF
--- a/http-server/src/main/java/io/airlift/http/server/ErrorWriter.java
+++ b/http-server/src/main/java/io/airlift/http/server/ErrorWriter.java
@@ -1,0 +1,27 @@
+package io.airlift.http.server;
+
+import java.io.PrintWriter;
+import java.nio.charset.Charset;
+import java.util.Optional;
+
+import static java.util.Objects.requireNonNull;
+
+public interface ErrorWriter
+{
+    void writeError(PrintWriter writer, ContentType contentType, Charset charset, ErrorDetails error);
+
+    enum ContentType
+    {
+        JSON,
+        HTML,
+        PLAIN;
+    }
+
+    record ErrorDetails(int responseCode, String message, Optional<Throwable> cause)
+    {
+        public ErrorDetails {
+            requireNonNull(message, "message is null");
+            requireNonNull(cause, "cause is null");
+        }
+    }
+}
--- a/http-server/src/main/java/io/airlift/http/server/ErrorWriterHandler.java
+++ b/http-server/src/main/java/io/airlift/http/server/ErrorWriterHandler.java
@@ -1,0 +1,46 @@
+package io.airlift.http.server;
+
+import io.airlift.http.server.ErrorWriter.ErrorDetails;
+import org.eclipse.jetty.server.Request;
+import org.eclipse.jetty.server.handler.ErrorHandler;
+
+import java.io.PrintWriter;
+import java.io.Writer;
+import java.nio.charset.Charset;
+import java.util.Optional;
+
+import static io.airlift.http.server.ErrorWriter.ContentType.HTML;
+import static io.airlift.http.server.ErrorWriter.ContentType.JSON;
+import static io.airlift.http.server.ErrorWriter.ContentType.PLAIN;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Objects.requireNonNull;
+
+public class ErrorWriterHandler
+        extends ErrorHandler
+{
+    private final ErrorWriter errorWriter;
+
+    public ErrorWriterHandler(ErrorWriter errorWriter)
+    {
+        this.errorWriter = requireNonNull(errorWriter, "errorWriter is null");
+    }
+
+    @Override
+    protected void writeErrorHtml(Request request, Writer writer, Charset charset, int code, String message, Throwable cause, boolean showStacks)
+    {
+        PrintWriter printWriter = (writer instanceof PrintWriter value) ? value : new PrintWriter(writer);
+        errorWriter.writeError(printWriter, HTML, charset, new ErrorDetails(code, message, Optional.ofNullable(cause)));
+    }
+
+    @Override
+    protected void writeErrorPlain(Request request, PrintWriter writer, int code, String message, Throwable cause, boolean showStacks)
+    {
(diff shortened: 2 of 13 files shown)


wendigo:
Ptal @martint @electrum

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -428,14 +431,41 @@ public boolean isShowStackTrace()
...
 
+    @Config("http-server.show-error-in-title")
+    @ConfigDescription("Show the error message in a title when generating an error page")
+    public HttpServerConfig setShowErrorInTitle(boolean showErrorInTitle)
+    {
+        this.showErrorInTitle = showErrorInTitle;
+        return this;
+    }
+
+    public boolean isShowErrorInTitle()
```
electrum, line 451: Getter should be above the setter

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -428,14 +431,41 @@ public boolean isShowStackTrace()
...
     }
 
-    @Config("http-server.show-stack-trace")
-    @ConfigDescription("Show the stack trace when generating an error response")
+    @LegacyConfig("http-server.show-stack-trace")
+    @Config("http-server.show-error-stack-trace")
+    @ConfigDescription("Show the stack trace when generating an error page")
     public HttpServerConfig setShowStackTrace(boolean showStackTrace)
     {
         this.showStackTrace = showStackTrace;
```
electrum, line 439: Rename to `showErrorStackTrace`

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -428,14 +431,41 @@ public boolean isShowStackTrace()
...
+        this.showErrorInTitle = showErrorInTitle;
+        return this;
+    }
+
+    public boolean isShowErrorInTitle()
+    {
+        return showErrorInTitle;
+    }
+
+    public boolean isShowCauses()
```
electrum, line 456: Rename field and methods to `showErrorCauses`

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -89,6 +90,8 @@ public class HttpServerConfig
     private int adminMaxThreads = 200;
 
     private boolean showStackTrace = true;
+    private boolean showErrorInTitle = true;
```
electrum, line 93: Do we need these new configs? Having a single config seems fine -- show stack traces or not. Especially considering that we want to replace the error handler in Trino, I don't see a need for extra configuration.

http-server/src/main/java/io/airlift/http/server/HttpServerBinder.java
```diff
@@ -51,6 +53,18 @@ public HttpServerBinder enableLegacyUriCompliance()
         return this;
     }
 
+    public HttpServerBinder bindErrorHandler(ErrorHandler errorHandler)
```
electrum, line 56: We don't want to expose Jetty specific classes. Instead, we should have on own, simpler interface here.

wendigo: What would the interface look like @electrum ?

http-server/src/main/java/io/airlift/http/server/ErrorWriterHandler.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Override
+    protected void writeErrorHtml(Request request, Writer writer, Charset charset, int code, String message, Throwable cause, boolean showStacks)
+    {
+        if (writer instanceof PrintWriter printWriter) {
+            errorWriter.writeError(printWriter, HTML, charset, new ErrorWriter.ErrorDetails(code, message, Optional.ofNullable(cause)));
+        }
+        else {
+            errorWriter.writeError(new PrintWriter(writer), HTML, charset, new ErrorWriter.ErrorDetails(code, message, Optional.ofNullable(cause)));
+        }
```
electrum, line 35:
```suggestion
        PrintWriter printWriter = (writer instanceof PrintWriter value) ? value : new PrintWriter(writer);
        errorWriter.writeError(printWriter, HTML, charset, new ErrorWriter.ErrorDetails(code, message, Optional.ofNullable(cause)));
        }
```

http-server/src/main/java/io/airlift/http/server/ErrorWriterHandler.java
```diff
@@ -0,0 +1,49 @@
...
+    public ErrorWriterHandler(ErrorWriter errorWriter)
+    {
+        this.errorWriter = requireNonNull(errorWriter, "errorWriter is null");
+    }
+
+    @Override
+    protected void writeErrorHtml(Request request, Writer writer, Charset charset, int code, String message, Throwable cause, boolean showStacks)
+    {
+        if (writer instanceof PrintWriter printWriter) {
+            errorWriter.writeError(printWriter, HTML, charset, new ErrorWriter.ErrorDetails(code, message, Optional.ofNullable(cause)));
```
electrum, line 31:
```java
import io.airlift.http.server.ErrorWriter.ErrorDetails;

File: http-server/src/main/java/io/airlift/http/server/ErrorWriter.java [hunk 8]
```diff
@@ -0,0 +1,26 @@
...
+    void writeError(PrintWriter writer, ContentType contentType, Charset charset, ErrorDetails error);
+
+    enum ContentType
+    {
+        JSON,
+        HTML,
+        PLAIN;
+    }
+
+    record ErrorDetails(int responseCode, String message, Optional<Throwable> cause)
```
electrum, line 20: It looks like `message` can't be null, so check for null.

electrum (commented):
The new interface looks good. Let's get rid of the config changes for the default error handler.

wendigo:
Closing this PR because of the lack of interest from project maintainers.

PR #1140, opened 2024-04-05
Update Jersey to 3.1.6

wendigo:
This is first Jersey release that improves compatibility with virtual threads avoiding thread pinning in synchronized blocks


wendigo:
cc @electrum @martint

PR #1141, opened 2024-04-11
Exclude non-fields from record serialization

findepi:
Before the change, a record's getter-like methods would be considered
getters and would be parsed of serialized form. This is undesirable
because

- For record component `foo`, the `foo()` method should be considered
  the accessor. If `getFoo()` exists, it may return something else.
  Using it can lead to correctness problems.
- A method like `getFoo()` may not be accessor to a record component at
  all, it should not be part of serialization.
- The default behavior is different than one for regular classes, so a
  benign refactor from a class to a record changes what gets serialized
  as JSON -- the getter-like non-getters on a class are ignored (do not
  need `@JsonIgnore`) and are included on a record (unless annotated
  with `@JsonIgnore`). This changes serialized form in a hard to notice
  manner, because we ignore those additional properties during
  deserialization.

This commit fixes this. Now only the canonical accessors are considered
a getter in a record.

Alternative solution would be to rely on Jackson built-in record
support. However, it relies on private field access, and we disable
CAN_OVERRIDE_ACCESS_MODIFIERS mapper feature.

### DIFF
--- a/json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
+++ b/json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
@@ -11,6 +11,8 @@ import com.fasterxml.jackson.databind.introspect.AnnotatedMethod;
 import com.fasterxml.jackson.databind.introspect.VisibilityChecker;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
 import java.lang.reflect.Field;
 import java.lang.reflect.Member;
 import java.lang.reflect.Method;
@@ -20,10 +22,17 @@ import java.util.Set;
 import java.util.stream.Stream;
 
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 public class RecordAutoDetectModule
         extends SimpleModule
 {
+    @Deprecated(forRemoval = true)
+    @Retention(RUNTIME)
+    @Target(TYPE)
+    public @interface LegacyRecordIntrospection {}
+
     @Override
     public void setupModule(SetupContext context)
     {
@@ -34,6 +43,12 @@ public class RecordAutoDetectModule
     private static class Introspector
             extends AnnotationIntrospector
     {
+        private static final VisibilityChecker.Std RECORD_VISIBILITY_CHECKER = VisibilityChecker.Std.defaultInstance()
+                .withGetterVisibility(JsonAutoDetect.Visibility.PUBLIC_ONLY)
+                .withCreatorVisibility(JsonAutoDetect.Visibility.DEFAULT)
+                .withFieldVisibility(JsonAutoDetect.Visibility.DEFAULT)
+                .withIsGetterVisibility(JsonAutoDetect.Visibility.DEFAULT);
+
         @Override
         public VisibilityChecker<?> findAutoDetectVisibility(AnnotatedClass ac, VisibilityChecker<?> checker)
         {
@@ -41,6 +56,9 @@ public class RecordAutoDetectModule
                 JsonAutoDetect overrideAnnotation = ac.getRawType().getAnnotation(JsonAutoDetect.class);
                 if (overrideAnnotation != null) {
                     return VisibilityChecker.Std.construct(JsonAutoDetect.Value.from(overrideAnnotation));
+                }
+                if (ac.getRawType().isAnnotationPresent(LegacyRecordIntrospection.class)) {
+                    return RECORD_VISIBILITY_CHECKER;
                 }
                 return new RecordVisibilityChecker(ac.getRawType().asSubclass(Record.class));
             }
--- a/json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
+++ b/json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
@@ -1,11 +1,25 @@
 package io.airlift.json;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.PropertyAccessor;
 import com.fasterxml.jackson.core.Version;
 import com.fasterxml.jackson.databind.AnnotationIntrospector;
 import com.fasterxml.jackson.databind.introspect.AnnotatedClass;
+import com.fasterxml.jackson.databind.introspect.AnnotatedField;
+import com.fasterxml.jackson.databind.introspect.AnnotatedMember;
+import com.fasterxml.jackson.databind.introspect.AnnotatedMethod;
 import com.fasterxml.jackson.databind.introspect.VisibilityChecker;
 import com.fasterxml.jackson.databind.module.SimpleModule;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Member;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.RecordComponent;
+import java.util.Set;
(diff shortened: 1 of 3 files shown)


losipiuk (approved):

findepi:
cc @wendigo

wendigo:
LGTM. This will make Record usage across Trino codebase consistent without a need to annotate components and add @JsonIgnore to everywhere else

json/src/test/java/io/airlift/json/TestJsonCodec.java
```diff
@@ -257,4 +260,142 @@ public void testToJsonWithLengthLimitComplex()
...
+
+        // Looks like a getter for "bar" property, there is no such record component
+        public String getBar()
+        {
+            return "there is no bar field in the record";
+        }
+
+        // Not a record component, but explicitly requested to be included in serialization
+        @JsonProperty
+        @SuppressWarnings("unused")
```
martint, line 342: Remove these suppressions

martint (commented):
Just skimmed - direction looks good. I'm doing a deeper review.

json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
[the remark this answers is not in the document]
findepi: is it a negative change? sure, can remove it

json/src/test/java/io/airlift/json/TestJsonCodec.java
findepi: i don't find it beneficial to have warnings reported by my IDE, even if IDE is wrong. i prefer to suppress them once rather than have them annoy me every time i work with the class

what is the benefit of removing the suppression?

json/src/test/java/io/airlift/json/TestJsonCodec.java
```diff
@@ -257,4 +260,142 @@ public void testToJsonWithLengthLimitComplex()
...
+    {
+        assertSerializationRoundTrip(
+                jsonCodec(MyRecord.class),
+                new MyRecord("my value"),
+                """
+                        {
+                          "foo" : "my value"
+                        }""");
+
+        assertSerializationRoundTrip(
```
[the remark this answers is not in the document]
findepi: This changes the value of the literal. With such a change the tests do not pass anymore.
I don't want to add a trim inside assertSerializationRoundTrip, so will keep the code as-is. 
If you prefer trim inside assertSerializationRoundTrip, or something else, let me know

findepi:
AC & added entry in CHANGES.

json/src/test/java/io/airlift/json/TestJsonCodec.java
martint: Because suppressions are a code smell. They are a signal that something is not quite right with the code. Additionally, they remove the warning but don't tell you when the warning no longer applies, so over time you end up with "dead" suppressions.

For this specific one, I would suggest you make your IDE ignore the unused warning when a method has  the `@JsonProperty` annotation. Since that expected to be an over-the-wire interface, there may be uses of the interface that aren't captured by code references.

findepi: >  make your IDE ignore the unused warning when a method has the @JsonProperty annotation. Since that expected to be an over-the-wire interface

thanks for the tip!

that would make it harder to notice fields carried around (part of over-the-wire interface), but are otherwise unused

> Because suppressions are a code smell.

Agreed!
I use suppressions only when necessary to suppress warnings, so that i can still see warnings about real problems. Too many times i have seen my IDE warn me about a problem that's real and should be fixed, like lack of synchronization. 

I understand suppressions are even bigger code small then code warnings themselves, at least in this project. I will drop those annotations.

findepi:
thanks for awesome review. AC

json/src/test/java/io/airlift/json/TestJsonCodec.java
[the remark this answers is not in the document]
martint: You can do this, instead:
```
                """
                {
                  "foo" : "my value"
                }\
                """
```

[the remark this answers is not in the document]
findepi: done!

findepi:
thank you for your review. AC

json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
```diff
@@ -45,4 +53,134 @@ public Version version()
             return Version.unknownVersion();
         }
     }
+
+    private static class RecordVisibilityChecker
+            implements VisibilityChecker<RecordVisibilityChecker>
+    {
+        private final Set<String> componentNames;
+
+        RecordVisibilityChecker(Class<? extends Record> recordClass)
```
martint, line 80: `public`, since it's part of the public interface of this class

json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
```diff
@@ -45,4 +53,134 @@ public Version version()
...
+        {
+            componentNames = Stream.of(recordClass.getRecordComponents())
+                    .map(RecordComponent::getName)
+                    .collect(toImmutableSet());
+        }
+
+        @Override
+        public RecordVisibilityChecker with(JsonAutoDetect ann)
```
martint, line 88: Avoid abbreviations

json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
```diff
@@ -20,10 +22,17 @@
...
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 public class RecordAutoDetectModule
         extends SimpleModule
 {
+    @Deprecated(forRemoval = true)
+    @Retention(RUNTIME)
+    @Target(TYPE)
+    public @interface LegacyRecordIntrospection {}
```
martint, line 34: Why do we need legacy mode? Upgrading to the new version of airlift requires changing callers to use this annotation, anyway, so they might as well just add/remove the necessary annotations needed for the new serialization behavior.

json/src/test/java/io/airlift/json/TestJsonCodec.java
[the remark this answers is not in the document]
findepi: indeed. that's probably  a bug in intellij 's formatter  and i didn't notice this myself, sorry
i will reformat the code correctly without IDE's "Reformat Code" action

json/src/main/java/io/airlift/json/RecordAutoDetectModule.java
findepi: good catch! this matches parameter names in jackson interface we're implementing here, but i will change them nonetheless.

json/src/test/java/io/airlift/json/TestJsonCodec.java
```diff
@@ -257,4 +260,142 @@ public void testToJsonWithLengthLimitComplex()
...
+
+        // Might shadow actual getter for condition -- the condition() method
+        public boolean getCondition()
+        {
+            // TODO this method should not be called during serialization
+            return condition();
+        }
+    }
+
+    public enum FakeGetterBehavior
```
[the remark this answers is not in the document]
martint: Thanks for doing that extra work. I'd prefer to just fix the bug and add/fix the relevant tests. That should be self-explanatory and not require someone to try to understand code that's removed immediately. So, in short, let's squash those commits.

martint (approved):
See my last comment. Otherwise, looks good.

json/src/test/java/io/airlift/json/TestJsonCodec.java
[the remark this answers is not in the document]
findepi: squashed

PR #1142, opened 2024-04-13
Update bouncycastle to 1.78

wendigo:
https://bouncycastle.org/releasenotes.html#r1rv78


wendigo:
cc @martint @electrum

PR #1144, opened 2024-04-19
Enable Java Vector module for Java 22 in launcher.py

raunaqmorarka:


electrum:
This seems too specific. Also, this will break on older JVMs, or newer JVMs (after Vector is final). Let me think about this more. We have a seemingly related problem of other required options like `--add-opens`.

raunaqmorarka:
> This seems too specific. Also, this will break on older JVMs, or newer JVMs (after Vector is final). Let me think about this more. We have a seemingly related problem of other required options like `--add-opens`.

I have refactored my Trino PR so that the Vector module is no longer required. We will automatically fall back to existing code paths if Vector module isn't found. So now it's just a matter of finding a convenient way to add `--add-modules=jdk.incubator.vector`. We could just add it to all the jvm configs in the codebase and add it to our jvm configs documentation and leave it at that. The other option is to add it to `core/docker/bin/run-trino` which will cover our main mode of deployment.

findepi:
> We could just add it to all the jvm configs in the codebase and add it to our jvm configs documentation and leave it at that. The other option is to add it to `core/docker/bin/run-trino` which will cover our main mode of deployment.

I like this direction.

raunaqmorarka:
> > We could just add it to all the jvm configs in the codebase and add it to our jvm configs documentation and leave it at that. The other option is to add it to `core/docker/bin/run-trino` which will cover our main mode of deployment.
> 
> I like this direction.

We had an offline discussion about this, the consensus was to change launcher.py to add `--add-modules=jdk.incubator.vector` if we detect the JVM to be version 22 and to update this version check with each subsequent new Java version until it is no longer needed.

electrum (approved):

electrum:
Thanks!

PR #1148, opened 2024-04-27
Add support for probabilistic sampling in Tracing

vinay-kosaraju:


mattstep (approved):

elonazoulay (approved):

tracing/src/main/java/io/airlift/tracing/OpenTelemetryConfig.java
```diff
@@ -1,12 +1,15 @@
...
 import io.airlift.configuration.Config;
+import jakarta.validation.constraints.Max;
+import jakarta.validation.constraints.Min;
 import jakarta.validation.constraints.NotNull;
 import jakarta.validation.constraints.Pattern;
 
 public class OpenTelemetryConfig
 {
     private String endpoint = "http://localhost:4317";
+    private Double samplingRatio = 1d;
```
erichwang, line 12: unless this is nullable, you should make this a `double` instead of the object

vinay-kosaraju: updated

tracing/src/main/java/io/airlift/tracing/OpenTelemetryConfig.java
```diff
@@ -21,4 +24,18 @@ public OpenTelemetryConfig setEndpoint(String endpoint)
         this.endpoint = endpoint;
         return this;
     }
+
+    @DecimalMax("1")
```
electrum, line 28: We can use `@Max` here. `@DecimalMax` is only needed for non-integer constraint values.

Note that per the Bean Validation specification, neither validation annotations are required to support `double`, but the Hibernate Validator implementation that we use supports `double` for both annotations.

tracing/src/main/java/io/airlift/tracing/OpenTelemetryConfig.java
```diff
@@ -21,4 +24,18 @@ public OpenTelemetryConfig setEndpoint(String endpoint)
...
+    @DecimalMin("0")
+    public double getSamplingRatio()
+    {
+        return samplingRatio;
+    }
+
+    @Config("tracing.exporter.sampling-ratio")
```
electrum, line 35: Name this `tracing.sampling-ratio`, since it's not related to the exporter

tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
```diff
@@ -47,11 +50,12 @@ public OpenTelemetryModule(String serviceName, String serviceVersion)
     public void configure(Binder binder)
     {
         newSetBinder(binder, SpanProcessor.class);
+        configBinder(binder).bindConfig(OpenTelemetryConfig.class);
```
electrum, line 53: The existing `OpenTelemetryConfig` is only bound when tracing is enabled. We shouldn't bind it when tracing is disabled, as that would allow configuration of the exporter endpoint, which wouldn't be used.

Instead, let's rename `OpenTelemetryConfig` to `OpenTelemetryExporterConfig`, then create a new `OpenTelemetryConfig` with the sampling config.

electrum (changes requested):

electrum (approved):
Sorry, one more thing, I should have mentioned it: can you do the rename of `OpenTelemetryConfig` in a separate commit? That will make the commits easier to read. This should be easy since the changes are in separate files.

electrum:
Thanks!

PR #1152, opened 2024-04-30
Change ConfigurationFactory.instanceCache to a LoadingCache

wendigo:


configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -101,14 +102,24 @@ public class ConfigurationFactory
     private final Map<String, String> properties;
     private final WarningsMonitor warningsMonitor;
     private final Problems.Monitor monitor;
-    private final ConcurrentMap<ConfigurationProvider<?>, Object> instanceCache = new ConcurrentHashMap<>();
+    private final LoadingCache<ConfigurationProvider<?>, Object> instanceCache = CacheBuilder.newBuilder()
+            .maximumSize(1024)
```
Randgalt, line 106: Maybe softValues instead?

wendigo: @electrum WDYT?

martint:
What's the purpose of this change? Generally, the number of configuration objects in an application would be naturally bounded, already.

PR #1157, opened 2024-05-06
Support dynamic properties in HTTP announcement 

findepi:
Extend discovery binder's bindHttpAnnouncement's capabilities by
allowing bindings of non-constant properties that require Guice
injection to be calculated.

For https://github.com/trinodb/trino/pull/21744

### DIFF
--- a/discovery/src/main/java/io/airlift/discovery/client/DiscoveryBinder.java
+++ b/discovery/src/main/java/io/airlift/discovery/client/DiscoveryBinder.java
@@ -15,19 +15,28 @@
  */
 package io.airlift.discovery.client;
 
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import com.google.inject.Binder;
 import com.google.inject.Inject;
+import com.google.inject.Injector;
 import com.google.inject.Key;
 import com.google.inject.Provider;
 import com.google.inject.Scopes;
+import com.google.inject.TypeLiteral;
+import com.google.inject.multibindings.MapBinder;
 import com.google.inject.multibindings.Multibinder;
 import io.airlift.discovery.client.ServiceAnnouncement.ServiceAnnouncementBuilder;
 
+import java.lang.annotation.Annotation;
+import java.util.Map;
+
+import static com.google.inject.multibindings.MapBinder.newMapBinder;
 import static com.google.inject.multibindings.Multibinder.newSetBinder;
 import static io.airlift.configuration.ConfigBinder.configBinder;
 import static io.airlift.discovery.client.ServiceAnnouncement.serviceAnnouncement;
 import static io.airlift.discovery.client.ServiceTypes.serviceType;
 import static java.util.Objects.requireNonNull;
+import static java.util.UUID.randomUUID;
 
 public class DiscoveryBinder
 {
@@ -84,11 +93,12 @@ public class DiscoveryBinder
         serviceAnnouncementBinder.addBinding().toProvider(announcementProviderClass);
     }
 
-    public ServiceAnnouncementBuilder bindHttpAnnouncement(String type)
+    public HttpAnnouncementBindingBuilder bindHttpAnnouncement(String type)
     {
-        ServiceAnnouncementBuilder serviceAnnouncementBuilder = serviceAnnouncement(type);
-        bindServiceAnnouncement(new HttpAnnouncementProvider(serviceAnnouncementBuilder));
-        return serviceAnnouncementBuilder;
+        HttpAnnouncement annotation = new HttpAnnouncementImpl(type + "." + randomUUID());
+        MapBinder<String, String> propertiesBinder = newMapBinder(binder, String.class, String.class, annotation);
+        bindServiceAnnouncement(new HttpAnnouncementProvider(type, annotation));
+        return new HttpAnnouncementBindingBuilder(propertiesBinder);
     }
 
     public void bindHttpSelector(String type)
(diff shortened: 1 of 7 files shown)


discovery/src/main/java/io/airlift/discovery/client/DiscoveryBinder.java
```diff
@@ -84,11 +96,12 @@ public <T extends ServiceAnnouncement> void bindServiceAnnouncement(Class<? exte
         serviceAnnouncementBinder.addBinding().toProvider(announcementProviderClass);
     }
 
-    public ServiceAnnouncementBuilder bindHttpAnnouncement(String type)
+    public HttpServiceAnnouncementBuilder bindHttpAnnouncement(String type)
     {
-        ServiceAnnouncementBuilder serviceAnnouncementBuilder = serviceAnnouncement(type);
-        bindServiceAnnouncement(new HttpAnnouncementProvider(serviceAnnouncementBuilder));
-        return serviceAnnouncementBuilder;
+        Named annotation = named("http-announcement.%s.%s".formatted(type, randomUUID()));
```
wendigo, line 101: Do we need the generated annotation?

findepi: how do you suggest to avoid it?

wendigo: I don't understand this code at all, so I don't know what you are trying to achieve here :)

discovery/src/main/java/io/airlift/discovery/client/DiscoveryBinder.java
```diff
@@ -104,15 +117,90 @@ public void bindHttpSelector(ServiceType serviceType)
         binder.bind(HttpServiceSelector.class).annotatedWith(serviceType).toProvider(new HttpServiceSelectorProvider(serviceType.value())).in(Scopes.SINGLETON);
     }
 
+    public static class HttpServiceAnnouncementBuilder
```
electrum, line 120: End the name with `BindingBuilder` to match the convention. We could shorten it to `HttpAnnoucementBindingBuilder`, matching the name of `HttpAnnouncementProvider`, with the reasoning that `service` is generic and is replaced with `http`.

discovery/src/main/java/io/airlift/discovery/client/DiscoveryBinder.java
electrum: I think this should be idempotent rather than random, which would allow multiple modules to independently add bindings for the same announcement. Rather than using `Named`, we should have a `HttpAnnouncementType` annotation and implementation, like we have with `ServiceType` and `ServiceTypeImpl`.

discovery/src/main/java/io/airlift/discovery/client/DiscoveryBinder.java
```diff
@@ -104,15 +117,90 @@ public void bindHttpSelector(ServiceType serviceType)
...
             implements Provider<ServiceAnnouncement>
     {
-        private final ServiceAnnouncementBuilder builder;
+        private final String type;
+        private final Annotation annotation;
+        private Injector injector;
         private AnnouncementHttpServerInfo httpServerInfo;
 
-        public HttpAnnouncementProvider(ServiceAnnouncementBuilder serviceAnnouncementBuilder)
+        public HttpAnnouncementProvider(String type, Annotation annotation)
```
electrum, line 194: Add `equals` and `hashCode` based on `type` and `annotation` so that Guice can de-dupe if the provider is bound multiple times.

discovery/src/test/java/io/airlift/discovery/client/TestHttpAnnouncementBinder.java
```diff
@@ -54,17 +54,15 @@ public void testHttpAnnouncement()
                 .addProperty("http-external", httpServerInfo.getHttpExternalUri().toASCIIString())
                 .build();
 
-        Set<ServiceAnnouncement> announcements = injector.getInstance(Key.get(new TypeLiteral<Set<ServiceAnnouncement>>()
-        {
-        }));
+        Set<ServiceAnnouncement> announcements = injector.getInstance(Key.get(new TypeLiteral<Set<ServiceAnnouncement>>() {}));
```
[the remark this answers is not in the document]
findepi: nice. this commit is just formatting fix. i'd prefer the semantic improvement to be a followup, if you don't mind

(i can undo formatting fixes, if necessary)

discovery/src/main/java/io/airlift/discovery/client/DiscoveryBinder.java
```diff
@@ -104,15 +117,90 @@ public void bindHttpSelector(ServiceType serviceType)
...
+        public HttpServiceAnnouncementBuilder addProperties(Map<String, String> properties)
+        {
+            properties.forEach(this::addProperty);
+            return this;
+        }
+
+        public interface PropertyBinding
+        {
+            @CanIgnoreReturnValue
+            default HttpServiceAnnouncementBuilder toInstance(String value)
```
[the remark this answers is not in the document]
findepi: yes, it's redundant, but more consistent with how binders generally work

i think it could be unfriendly to type `.bindProperty("foo") ...` and see code completion for more sophisticated use, but not for the less sophisticated.

we can deprecate `addProperty` if needed.

or remove `bindProperty` replacing it with `addProperty` overloads that take a provider

discovery/src/main/java/io/airlift/discovery/client/DiscoveryBinder.java
```diff
@@ -104,15 +117,90 @@ public void bindHttpSelector(ServiceType serviceType)
...
+    public static class HttpServiceAnnouncementBuilder
+    {
+        private final MapBinder<String, String> propertiesBinder;
+
+        public HttpServiceAnnouncementBuilder(MapBinder<String, String> propertiesBinder)
+        {
+            this.propertiesBinder = propertiesBinder;
+        }
+
+        public PropertyBinding bindProperty(String key)
```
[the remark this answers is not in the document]
findepi: this would resolve https://github.com/airlift/airlift/pull/1157#discussion_r1591426356. sure, can do that

discovery/src/main/java/io/airlift/discovery/client/DiscoveryBinder.java
findepi: HttpAnnouncementProvider is pre-existing and didn't have equals/hashcode
it feels it would be separate commit. since i don't fully understand the rationale, can this be a follow-up?

discovery/src/test/java/io/airlift/discovery/client/TestHttpAnnouncementBinder.java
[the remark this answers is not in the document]
findepi: i will use this in the newly added code though

findepi:
AC, PTAL

discovery/src/main/java/io/airlift/discovery/client/DiscoveryBinder.java
findepi: Slight modification of `io.airlift.discovery.client.TestHttpAnnouncementBinder#testHttpAnnouncementWithCustomProperties`:

```java
@Test
public void testHttpAnnouncementWithCustomProperties()
{
    final StaticAnnouncementHttpServerInfoImpl httpServerInfo = new StaticAnnouncementHttpServerInfoImpl(
            URI.create("http://127.0.0.1:4444"),
            URI.create("http://example.com:4444"),
            URI.create("https://127.0.0.1:4444"),
            URI.create("https://example.com:4444"));

    Injector injector = Guice.createInjector(
            new TestingDiscoveryModule(),
            binder -> {
                binder.bind(AnnouncementHttpServerInfo.class).toInstance(httpServerInfo);
                discoveryBinder(binder).bindHttpAnnouncement("apple").addProperty("a", "apple");
            },
            // Bind again to add new property
            binder -> {
                binder.bind(AnnouncementHttpServerInfo.class).toInstance(httpServerInfo);
                discoveryBinder(binder).bindHttpAnnouncement("apple").addProperty("b", "banana");
            });

    ServiceAnnouncement announcement = serviceAnnouncement("apple")
            .addProperty("a", "apple")
            .addProperty("b", "banana")
            .addProperty("http", httpServerInfo.getHttpUri().toASCIIString())
            .addProperty("http-external", httpServerInfo.getHttpExternalUri().toASCIIString())
            .addProperty("https", httpServerInfo.getHttpsUri().toASCIIString())
            .addProperty("https-external", httpServerInfo.getHttpsExternalUri().toASCIIString())
            .build();

    Set<ServiceAnnouncement> announcements = injector.getInstance(Key.get(new TypeLiteral<Set<ServiceAnnouncement>>() {
    }));

    assertAnnouncement(announcements, announcement);
}
```

This doesn't pass in current master, as it creates two announcements, not one.
The behavior is retained by this PR.

discovery/src/main/java/io/airlift/discovery/client/DiscoveryBinder.java
electrum: Not having this will prevent properties from being provided from multiple locations. It's needed to allow
```java
discoveryBinder(binder).bindHttpAnnouncement("apple").addProperty("aaa", "aaa");
discoveryBinder(binder).bindHttpAnnouncement("apple").addProperty("xxx", "xxx");
```
But we could add this later if needed.

discovery/src/main/java/io/airlift/discovery/client/DiscoveryBinder.java
electrum: Hmm, I see how this works now. In the existing code, `bindHttpAnnouncement()` creates `ServiceAnnouncementBuilder` which has a UUID, so we get a unique announcement for each call.

electrum (approved):
The API looks good, but we should still create custom annotation class rather than using `Named`. We avoid using `Named` as it's single shared namespace. While this usage of using it with UUIDs is unlikely to actually be a problem, using a custom annotation is best practice (and keeps the simple rule of "don't use `Named`").

findepi:
Custom annotation is less encapsulated, as someone could instantiate it outside of the dedicated builder, so i think this usage of Named is actually better. We will need a custom annotation if we want idempotence (https://github.com/airlift/airlift/pull/1157#discussion_r1591481022). Although I don't agree, I will still switch to custom annotation.

discovery/src/test/java/io/airlift/discovery/client/TestHttpAnnouncementBinder.java
```diff
@@ -150,4 +185,22 @@ private void assertAnnouncement(Set<ServiceAnnouncement> actualAnnouncements, Se
...
+
+    public static class StringPropertyProvider
+            implements Provider<String>
+    {
+        private final AnnouncementHttpServerInfo httpServerInfo;
+
+        @Inject
+        public StringPropertyProvider(AnnouncementHttpServerInfo httpServerInfo)
+        {
+            this.httpServerInfo = httpServerInfo;
```
wendigo, line 197: rnn httpServerInfo

PR #1158, opened 2024-05-07
Add HTTP version to client traces

wendigo:


http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -729,6 +730,10 @@ public void onContent(Response response, ByteBuffer content)
         // record attributes
         span.setAttribute(HttpAttributes.HTTP_RESPONSE_STATUS_CODE, response.getStatus());
 
+        // negotiated http version
+        span.setAttribute(NetworkAttributes.NETWORK_PROTOCOL_NAME, "http");
+        span.setAttribute(NetworkAttributes.NETWORK_PROTOCOL_VERSION, Integer.toString(response.getVersion().getVersion()));
```
electrum, line 735: This seems incorrect, since the Jetty integer would be `20` or `30`. The [specification](https://opentelemetry.io/docs/specs/semconv/attributes-registry/network/) shows `1.1` and `2` as examples. We probably need to do a switch statement here.

electrum (dismissed):

wendigo:
@electrum PTAL

wendigo:
@findepi PTAL

findepi (approved):

PR #1160, opened 2024-05-09
Fix time to first byte for async servlet

wendigo:


wendigo:
Replaced with https://github.com/airlift/airlift/pull/1161

PR #1161, opened 2024-05-09
Reimplement request/response timing - improves accuracy and correctness 

wendigo:
- Increase metrics precision to nanoseconds (and use correct way of measuring elapsed times - using monotonic time source rather than a system time that could go backwards)
- Drop TimingFilter and capture metrics on the Jetty level
- Validate timings invariants

Fixes:
- Metrics capture for async responses


wendigo:
I know that this PR is pretty big but it drops a lot of unnecessary code and aligns the code for better readability. It's best to review it with the checked out code to see how it looks like after changes.

wendigo:
Alternative to https://github.com/airlift/airlift/pull/1160

wendigo:
@electrum PTAL. Right now this doesn't work for async servlet at all. I've tested it thoroughly and added invariants that check that corresponding processing stages happen monotonically. In HTTP/2 where requests are really fast (us/ms) this matters for accuracy:

Screenshot 2024-06-05 at 16 44 40

electrum (approved):

PR #1165, opened 2024-05-17
Fix HTTP request builder setHeader case sensitivity

findepi:
This fixes a bug in request builder. The `io.airlift.http.client.Request.Builder#setHeader` is supposed to replace existing headers of given name and save the new value. Before this PR, it works only when `setHeader` uses same header name casing as the previously set headers (eg `Cookie` and not `cookie`).

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/Request.java
+++ b/http-client/src/main/java/io/airlift/http/client/Request.java
@@ -15,7 +15,6 @@
  */
 package io.airlift.http.client;
 
-import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.MultimapBuilder;
@@ -39,10 +38,7 @@ public final class Request
     private final Optional<HttpVersion> httpVersion;
     private final URI uri;
     private final String method;
-    private final ListMultimap<String, String> headers = MultimapBuilder
-            .treeKeys(CASE_INSENSITIVE_ORDER)
-            .arrayListValues()
-            .build();
+    private final ListMultimap<String, String> headers = newHeaders();
     private final Optional<Duration> requestTimeout;
     private final Optional<Duration> idleTimeout;
     private final BodyGenerator bodyGenerator;
@@ -196,6 +192,14 @@ public final class Request
                 followRedirects);
     }
 
+    private static ListMultimap<String, String> newHeaders()
+    {
+        return MultimapBuilder
+                .treeKeys(CASE_INSENSITIVE_ORDER)
+                .arrayListValues()
+                .build();
+    }
+
     public static final class Builder
     {
         public static Builder prepareHead()
@@ -248,7 +252,7 @@ public final class Request
 
         private URI uri;
         private String method;
-        private final ListMultimap<String, String> headers = ArrayListMultimap.create();
+        private final ListMultimap<String, String> headers = newHeaders();
         private BodyGenerator bodyGenerator;
         private SpanBuilder spanBuilder;
         private Optional<HttpVersion> version = Optional.empty();
--- a/http-client/src/test/java/io/airlift/http/client/TestRequestBuilder.java
+++ b/http-client/src/test/java/io/airlift/http/client/TestRequestBuilder.java
@@ -23,6 +23,7 @@ import java.net.URI;
 import static io.airlift.http.client.Request.Builder.fromRequest;
 import static io.airlift.http.client.Request.Builder.prepareGet;
 import static io.airlift.http.client.StaticBodyGenerator.createStaticBodyGenerator;
+import static java.util.Map.entry;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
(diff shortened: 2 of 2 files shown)


http-client/src/main/java/io/airlift/http/client/Request.java
```diff
@@ -232,7 +232,7 @@ public Builder setMethod(String method)
 
         public Builder setHeader(String name, String value)
         {
-            this.headers.removeAll(name);
+            this.headers.keys().removeIf(key -> key.equalsIgnoreCase(name));
```
brybacki, line 235: good catch!
Does addHeader and addHeaders behave correctly?

brybacki (approved):

losipiuk (approved):

http-client/src/main/java/io/airlift/http/client/Request.java
findepi: add are supposed to accumulate

brybacki: Right, it accumulates so not a problem, but there is still a difference and results in names not normalized. But maybe deduplication and normalization is a responsibility of  a different class. 

So it depends if we want:
```        
ListMultimap<String, String> headers = ArrayListMultimap.create();
        headers.put("Key-One", "v1");
        headers.put("Key-One", "v1");
        headers.put("Key-One", "v2");

        headers.put("key-one", "v1");
        
        System.out.println(headers);
```
To result in:
```
{Key-One=[v1, v1, v2], key-one=[v1]}

```
or in:
```
{key-one=[v1, v1, v1, v2]}
```

findepi:
added lowercasing and a test

brybacki:
```
Error:    TestBasicAuthRequestFilter.testBasicAuthentication:39 expected [Basic QWxhZGRpbjpvcGVuIHNlc2FtZQ==] but found [null]
Error:    TestTraceTokenRequestFilter.testBasic:46 lists don't have the same size expected [1] but found [0]
```

findepi:
replaced lowercasing with case-preserving, per @losipiuk suggestion

findepi:
the code is not simplest possible, because it preserves the case of headers. the simplest would be to just lowercase.
preserving header case is not important from http spec perspective, but we've seen cases where not preservice headers breaks some buggy applications

cc @losipiuk @electrum @wendigo

wendigo:
Alternative https://github.com/airlift/airlift/pull/1166

http-client/src/main/java/io/airlift/http/client/Request.java
wendigo: Custom multimap does that:

```
ListMultimap<String, String> headers = Multimaps.newListMultimap(new TreeMap<>(CASE_INSENSITIVE_ORDER), ArrayList::new);
        headers.put("Key-One", "v1");
        headers.put("Key-One", "v1");
        headers.put("Key-One", "v2");
        headers.put("key-one", "v1");
        System.out.println(headers);
```

```
{Key-One=[v1, v1, v2, v1]}
```

http-client/src/test/java/io/airlift/http/client/TestTraceTokenRequestFilter.java
```diff
@@ -43,7 +44,7 @@ public void testBasic()
         assertEquals(filtered.getUri(), original.getUri());
         assertEquals(original.getHeaders().size(), 0);
         assertEquals(filtered.getHeaders().size(), 1);
-        assertEquals(filtered.getHeaders().get(TRACETOKEN_HEADER), ImmutableList.of("testBasic"));
+        assertEquals(filtered.getHeaders().get(TRACETOKEN_HEADER.toLowerCase(ENGLISH)), ImmutableList.of("testBasic"));
```
findepi, line 47: With lowercasing, apps using airlift may need to change too. For example https://github.com/trinodb/trino/blob/a27e5af2d7389a7bc0e906210eafb906f01e4365/core/trino-main/src/test/java/io/trino/operator/MockExchangeRequestProcessor.java#L103

findepi:
added lowercasing again.

ptal

findepi:
@electrum @wendigo @losipiuk can you please review?

losipiuk:
Non preserving hurts my aesthetic soul but I hope I can live with that.

losipiuk (approved):

martint:
We shouldn’t lower case header names. What problem are we trying to solve?

findepi:
> What problem are we trying to solve?

`setHeader` was incorrectly replacing headers only if they were given in exact same case. See tests.

Of course the simplest implementation would be to just change `setHeader`:
```diff
--- http-client/src/main/java/io/airlift/http/client/Request.java
+++ http-client/src/main/java/io/airlift/http/client/Request.java
@@ -232,7 +232,7 @@ public final class Request
 
         public Builder setHeader(String name, String value)
         {
-            this.headers.removeAll(name);
+            this.headers.keySet().removeIf(key -> key.equalsIgnoreCase(name));
             this.headers.put(name, value);
             return this;
         }
```

in fact this was the first version of this PR.
However i realized that keeping `io.airlift.http.client.Request.Builder#headers` as `ListMultimap` would no longer make sense, if we always search it linearly.

then the PR had case-preserving version (see "before" in this diff https://github.com/airlift/airlift/compare/257a0490937b4f1e23ddbd44e0bf47612c81704a..08e57c46f089d59d97cf29481e2b2d69edcce336)

Then @electrum suggested case-preserving isn't worth it.

BTW once headers gets thru airlift request builder, their case is later changed in jetty client. It looks like the case normalization in the jetty client happens for all headers registered with `org.eclipse.jetty.http.HttpHeader` enum.


> We shouldn’t lower case header names.

Why?

findepi:
@martint @electrum hope you can converge on what this should do (lowercase vs case-preserving vs something else).
Happy to implement whatever we want to do. If this helps, i can create separate PRs for every option considered, so that we don't need to use this PR push history.

wendigo:
Jersey is using follow implementation: https://github.com/eclipse-ee4j/jersey/blob/14c5c43ff227ba34bc27aa41c81f84d5f1f3080d/core-common/src/main/java/org/glassfish/jersey/internal/util/collection/StringKeyIgnoreCaseMultivaluedMap.java#L28

http-client/src/test/java/io/airlift/http/client/TestTraceTokenRequestFilter.java
electrum: We should probably make `Request.getHeaders()` return a case-insensitive map

electrum:
Given the HTTP/2 is always lowercase, it's not clear that there is value in preserving header case. This seems like a bad assumption in applications that can appear to work, but will break depending on the protocol version used at runtime. Lowercasing headers makes code simpler and avoids the surprise.

http-client/src/test/java/io/airlift/http/client/TestTraceTokenRequestFilter.java
findepi: happy to apply once we know for sure (without "probably"). there was quite a bit back and forth already on this PR. It doesn't cost much because it is small, but  it's not cool nonetheless



@electrum @martint can you explicitly decide on the behavior you want to see implemented? happy to apply whatever is needed

findepi:
Hopefully we have agreement that the current behavior of setHeader is wrong. It caused an incident internally at Starburst.

@electrum @martint how would you like the code to behave?

- preserve case vs lower-case in the builder?
- getHeader to stay as is OR match case-insensitively in request getHeader()
- case-sensitive OR case-insensitive map keys in request getHeaders(), OR something else (like removing this method altogether)
  - case-insensitive has the problem similar to identity-based collections -- taking defensive immutable copy changes semantics; so i'd actually prefer to have dedicated wrapper class to represent header name (either lowercasing or preserving case but comparing case insensitively; depending on the answer to the first question)

wendigo:
Is this still needed? I think that we've sorted it out

findepi:
> Is this still needed? I think that we've sorted it out

This PR solved the problem for `Request`, but not the `Request.Builder`

- https://github.com/airlift/airlift/pull/1257

therefore the bug is still there and is still worth fixing.

I updated this PR to use the logic we settled on in the above PR (case-insensitive map keys comparator in request headers).

PR #1166, opened 2024-05-17
Preserve HTTP header case but compare them case insensitive

wendigo:
altered copy of https://github.com/airlift/airlift/pull/1165

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/Request.java
+++ b/http-client/src/main/java/io/airlift/http/client/Request.java
@@ -15,26 +15,28 @@
  */
 package io.airlift.http.client;
 
-import com.google.common.collect.ArrayListMultimap;
-import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Multimap;
+import com.google.common.collect.Multimaps;
 import io.opentelemetry.api.trace.SpanBuilder;
 
 import java.net.URI;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.TreeMap;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
 import static com.google.common.base.Preconditions.checkArgument;
+import static java.lang.String.CASE_INSENSITIVE_ORDER;
 import static java.util.Objects.requireNonNull;
 
 public final class Request
 {
     private final URI uri;
     private final String method;
-    private final ListMultimap<String, String> headers;
+    private final ListMultimap<String, String> headers = createHeadersMultimap();
     private final BodyGenerator bodyGenerator;
     private final Optional<SpanBuilder> spanBuilder;
     private final boolean followRedirects;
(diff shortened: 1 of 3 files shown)


wendigo:
cc @losipiuk

http-client/src/main/java/io/airlift/http/client/Request.java
```diff
@@ -291,4 +293,9 @@ private static URI validateUri(URI uri)
         checkArgument(uri.getPort() != 0, "Cannot make requests to HTTP port 0");
         return uri;
     }
+
+    private static ListMultimap<String, String> createHeadersMultimap()
+    {
+        return Multimaps.newListMultimap(new TreeMap<>(comparing(String::toLowerCase)), ArrayList::new);
```
findepi, line 299: toLowerCase is typically used with a locale

but here you can use `java.lang.String#CASE_INSENSITIVE_ORDER`

http-client/src/test/java/io/airlift/http/client/TestRequestBuilder.java
```diff
@@ -56,6 +58,36 @@ public void testBuilderFromRequest()
...
+                        entry("my-header", "camel"));
+    }
+
+    @Test
+    public void testSetCaseInsensitive()
+    {
+        Request request = prepareGet()
+                .setUri(URI.create("http://example.com"))
+                .addHeader("my-header", "lower")
+                .addHeader("My-Header", "camel")
```
findepi, line 82: @losipiuk suggested better test cases for https://github.com/airlift/airlift/pull/1165, so please update

http-client/src/main/java/io/airlift/http/client/Request.java
```diff
@@ -15,26 +15,28 @@
...
 import static com.google.common.base.Preconditions.checkArgument;
+import static java.lang.String.CASE_INSENSITIVE_ORDER;
 import static java.util.Objects.requireNonNull;
 
 public final class Request
 {
     private final URI uri;
     private final String method;
-    private final ListMultimap<String, String> headers;
+    private final ListMultimap<String, String> headers = createHeadersMultimap();
```
findepi, line 39: keep this field immutable. 

the Request is built with builder, so the "case-insensitively duplicated"  keys are already collapsed to one. 
there is however a semantics difference in the `getHeader(String)` behavior. this should be test covered

wendigo: After reconsideration I think that we can actually lower-case headers everywhere in the airlift. According to the RFC 7540, Hypertext Transfer Protocol Version 2 (HTTP/2), Section 8.1.2:

```
Just as in HTTP/1.x, header field names are strings of ASCII characters that are compared in a case-insensitive fashion. However, header field names MUST be converted to lowercase prior to their encoding in HTTP/2. A request or response containing uppercase header field names MUST be treated as malformed
```

findepi: Of course. Nothing that I didn't know. The desire to preserve case is not resulted of a whim, or lack of understanding how http works. i have no hard data to support that desire, it's more like an intuitions that something else can break.

wendigo: @findepi so let's fix the broken places instead

findepi: see https://github.com/airlift/airlift/pull/1165#issuecomment-2117739408

PR #1169, opened 2024-05-20
Add request-scoped jaxrs binding support

wendigo:
The second commits matters for HTTP/2 support. As explained in the commit message:

```
Add AsyncResponseBinder that allows to interrupt a Future when client disconnects
from the server. This will not only interrupt work that is no longer necessary
but also makes sure that the AsyncResponse won't be completed with the Response
since it's a pointless thing to do anyway. This is important for HTTP/2 communication
because client explicitly sends a RESET frame that server handles and recycles request/response
objects immediately.
```

This can be used in the Trino following manner:

Instead of:

```
public void getTaskInfo(
            @PathParam("taskId") TaskId taskId,
            @HeaderParam(TRINO_CURRENT_VERSION) Long currentVersion,
            @HeaderParam(TRINO_MAX_WAIT) Duration maxWait,
            @Context UriInfo uriInfo,
            @Suspended AsyncResponse asyncResponse)
```

```
 bindAsyncResponse(asyncResponse, futureTaskInfo, responseExecutor)
      .withTimeout(timeout);
```

following pattern can be used:

```
public void getTaskInfo(
            @PathParam("taskId") TaskId taskId,
            @HeaderParam(TRINO_CURRENT_VERSION) Long currentVersion,
            @HeaderParam(TRINO_MAX_WAIT) Duration maxWait,
            @Context UriInfo uriInfo,
            @Context AsyncResponseBinder asyncResponseBinder,
            @Suspended AsyncResponse asyncResponse)
```

```
asyncResponseBinder
            .withTimeout(timeout)
            .withFuture(futureTaskInfo)
            .withExecutor(responseExecutor)
            .cancelOnClientDisconnection()
            .bindTo(asyncResponse);
```

### DIFF
--- a/jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseBinder.java
+++ b/jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseBinder.java
@@ -1,0 +1,110 @@
+package io.airlift.jaxrs;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import io.airlift.units.Duration;
+import jakarta.servlet.AsyncEvent;
+import jakarta.servlet.AsyncListener;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.ws.rs.container.AsyncResponse;
+import jakarta.ws.rs.core.Context;
+import jakarta.ws.rs.core.Response;
+
+import java.util.Optional;
+import java.util.concurrent.Executor;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import static io.airlift.jaxrs.AsyncResponseHandler.bindAsyncResponse;
+import static jakarta.ws.rs.core.Response.status;
+import static java.util.Objects.requireNonNull;
+
+public class AsyncResponseBinder
+{
+    private final HttpServletRequest servletRequest;
+    private boolean cancelOnClientDisconnect = true;
+    private Optional<Duration> timeout = Optional.empty();
+    private Function<Duration, Supplier<Response>> timeoutResponse = AsyncResponseBinder::defaultTimeoutResponse;
+
+    private AsyncResponseBinder(HttpServletRequest servletRequest)
+    {
+        this.servletRequest = requireNonNull(servletRequest, "servletRequest is null");
+    }
+
+    public AsyncResponseBinder cancelOnClientDisconnection(boolean cancelOnClientDisconnection)
+    {
+        this.cancelOnClientDisconnect = cancelOnClientDisconnection;
+        return this;
+    }
+
+    public AsyncResponseBinder withTimeout(Duration timeout)
+    {
+        this.timeout = Optional.of(timeout);
+        return this;
+    }
+
+    public AsyncResponseBinder withTimeoutResponse(Response response)
+    {
+        requireNonNull(response, "response is null");
+        this.timeoutResponse = timeout -> () -> response;
+        return this;
+    }
+
+    public AsyncResponseBinder withTimeoutResponse(Supplier<Response> responseSupplier)
+    {
+        requireNonNull(responseSupplier, "responseSupplier is null");
+        this.timeoutResponse = timeout -> responseSupplier;
+        return this;
+    }
+
+    public void bindTo(AsyncResponse response, ListenableFuture<?> future, Executor httpExecutor)
+    {
+        AsyncResponseHandler<?> asyncResponseHandler = bindAsyncResponse(response, future, httpExecutor);
+        if (cancelOnClientDisconnect) {
+            servletRequest.getAsyncContext().addListener(new AsyncListener()
+            {
+                @Override
+                public void onError(AsyncEvent event)
+                {
+                    // Jetty will signal the client disconnection with the EofException
+                    if (event.getThrowable().getMessage().contains("cancel_stream_error")) {
+                        asyncResponseHandler.clientDisconnected();
+                    }
+                }
+
+                @Override
+                public void onComplete(AsyncEvent event) {}
+
+                @Override
+                public void onTimeout(AsyncEvent event) {}
+
+                @Override
+                public void onStartAsync(AsyncEvent event) {}
+            });
+        }
+        timeout.ifPresent(timeout -> asyncResponseHandler.withTimeout(timeout, timeoutResponse.apply(timeout)));
+    }
+
+    private static Supplier<Response> defaultTimeoutResponse(Duration timeout)
+    {
+        return () -> status(Response.Status.SERVICE_UNAVAILABLE)
+                .entity("Timed out after waiting for " + timeout.convertToMostSuccinctTimeUnit())
+                .build();
+    }
+
+    public static class Factory
+            implements Supplier<AsyncResponseBinder>
+    {
+        private final HttpServletRequest servletRequest;
+
+        private Factory(@Context HttpServletRequest servletRequest)
+        {
+            this.servletRequest = requireNonNull(servletRequest, "servletRequest is null");
+        }
+
+        @Override
+        public AsyncResponseBinder get()
+        {
+            return new AsyncResponseBinder(servletRequest);
+        }
+    }
+}
--- a/jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseHandler.java
+++ b/jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseHandler.java
@@ -23,6 +23,7 @@ import jakarta.ws.rs.core.Response;
 import java.lang.ref.WeakReference;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 
 import static com.google.common.base.Preconditions.checkArgument;
@@ -30,27 +31,29 @@ import static jakarta.ws.rs.core.Response.status;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
-public class AsyncResponseHandler
+public class AsyncResponseHandler<T>
+        implements FutureCallback<T>
 {
+    private final AtomicBoolean clientDisconnected = new AtomicBoolean();
     private final AsyncResponse asyncResponse;
-    private final WeakReference<Future<?>> futureResponseReference;
+    private final WeakReference<Future<T>> futureResponseReference;
 
-    private AsyncResponseHandler(AsyncResponse asyncResponse, ListenableFuture<?> futureResponse)
+    private AsyncResponseHandler(AsyncResponse asyncResponse, ListenableFuture<T> futureResponse, Executor httpResponseExecutor)
     {
         this.asyncResponse = requireNonNull(asyncResponse, "asyncResponse is null");
         // the jaxrs implementation can hold on to the async timeout for a long time, and
         // the future can reference large expensive objects.  Since we are only interested
         // in canceling this future on a timeout, only hold a weak reference to the future
         this.futureResponseReference = new WeakReference<>(requireNonNull(futureResponse, "futureResponse is null"));
+        Futures.addCallback(futureResponse, this, httpResponseExecutor);
     }
 
-    public static AsyncResponseHandler bindAsyncResponse(AsyncResponse asyncResponse, ListenableFuture<?> futureResponse, Executor httpResponseExecutor)
+    public static <T> AsyncResponseHandler<T> bindAsyncResponse(AsyncResponse asyncResponse, ListenableFuture<T> futureResponse, Executor httpResponseExecutor)
     {
-        Futures.addCallback(futureResponse, toFutureCallback(asyncResponse), httpResponseExecutor);
-        return new AsyncResponseHandler(asyncResponse, futureResponse);
+        return new AsyncResponseHandler<>(asyncResponse, futureResponse, httpResponseExecutor);
     }
 
-    public AsyncResponseHandler withTimeout(Duration timeout)
+    public AsyncResponseHandler<T> withTimeout(Duration timeout)
     {
         return withTimeout(timeout,
                 status(Response.Status.SERVICE_UNAVAILABLE)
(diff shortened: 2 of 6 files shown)


findepi:
not sure i understand everything here. would it help to add PR description? with what it does and what it is for.

wendigo:
@findepi test says it all. Adds a support to define your own @Context request-scoped beans

findepi:
yeah, i saw the title. i should know better, thanks

martint:
It’s intentional that we don’t support request scoped bindings in airlift.

electrum:
@martint These are JAX-RS request-scoped bindings, not related to Guice scopes. We already use this functionality via `org.glassfish.jersey.server.spi.internal.ValueParamProvider` in applications, but this seems like a cleaner API and doesn't require applications to depend on Jersey directly.

wendigo:
@electrum yes, exactly, that's jax-rs only binding

wendigo:
This could i.e. allow us to inject `@Context Identity` in Trino instead of passing `HttpServletRequest` to access its' attributes

findepi:
is there any processing overhead involved for handling these request-scoped things?

wendigo:
injecting HttpHeaders or HttpServletRequest goes through the same mechanism

wendigo:
@electrum PTAL. The request-scoped factories can be also used to implement https://github.com/trinodb/trino/pull/22227

jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseBinder.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    public void bindTo(AsyncResponse response)
+    {
+        AsyncResponseHandler<?> asyncResponseHandler = bindAsyncResponse(response, future, httpExecutor);
+        if (cancelOnClientDisconnection) {
+            servletRequest.getAsyncContext().addListener(new AsyncListener()
+            {
+                @Override
+                public void onComplete(AsyncEvent event)
```
electrum, line 80: Nit: format empty methods like
```java
 public void onComplete(AsyncEvent event) {}

File: jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseBinder.java [hunk 2]
```diff
@@ -0,0 +1,130 @@
...
+                @Override
+                public void onStartAsync(AsyncEvent event)
+                {
+                }
+            });
+        }
+        timeout.ifPresent(timeout -> asyncResponseHandler.withTimeout(timeout, timeoutResponse.apply(timeout)));
+    }
+
+    private static Supplier<Response> defaultTimeoutResponse(Duration timeout)
```
electrum, line 107: Rather than duplicating the code from `AsyncResponseHandler`, just call the `withTimeout(Duration)` variant

jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseBinder.java
```diff
@@ -0,0 +1,130 @@
...
+    public AsyncResponseBinder withTimeout(Duration timeout)
+    {
+        this.timeout = Optional.of(timeout);
+        return this;
+    }
+
+    public AsyncResponseBinder withTimeoutResponse(Response response)
+    {
+        requireNonNull(response, "response is null");
+        this.timeoutResponse = (timeout) -> () -> response;
```
electrum, line 62: Nit: no need for parens on single argument lambdas

jaxrs/src/main/java/io/airlift/jaxrs/JaxrsBinder.java
```diff
@@ -5,24 +5,30 @@
...
 
 import static com.google.inject.Scopes.SINGLETON;
 import static com.google.inject.multibindings.Multibinder.newSetBinder;
+import static java.lang.reflect.Modifier.isFinal;
 import static java.util.Objects.requireNonNull;
 
 public class JaxrsBinder
 {
     private final Multibinder<Object> resourceBinder;
+    private final Multibinder<JaxrsModule.JerseyFactoryBinding> factoryBinder;
```
electrum, line 24: Import `JerseyFactoryBinding` directly

jaxrs/src/test/java/io/airlift/jaxrs/TestRequestScopedBinding.java
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    @Retention(RUNTIME)
+    @BindingAnnotation
+    public @interface ForTest {}
+
+    public record RemoteAddr(String value) {}
+
+    public static class HeaderValue
```
electrum, line 77: Add a comment
```java
// this is not a record in order to test that normal classes work

File: jaxrs/src/test/java/io/airlift/jaxrs/TestRequestScopedBinding.java [hunk 6]
```diff
@@ -0,0 +1,141 @@
...
+                .setHeader("X-Name", name)
+                .build();
+        return httpClient.execute(request, createStringResponseHandler()).getBody();
+    }
+
+    @Path("/hello")
+    public static class MyResource
+    {
+        @GET
+        public String returnUserScopedInfo(@Context RemoteAddr remoteAddr, @Context HeaderValue headerValue)
```
electrum, line 65: Nit: name this `returnRequestScopedInfo` since there is nothing "user" related here (it's confusing since `HttpServletRequest` does have some "user" related methods but those aren't used here)

jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseBinder.java
```diff
@@ -0,0 +1,130 @@
...
+import static io.airlift.jaxrs.AsyncResponseHandler.bindAsyncResponse;
+import static jakarta.ws.rs.core.Response.status;
+import static java.util.Objects.requireNonNull;
+
+public class AsyncResponseBinder
+{
+    private final HttpServletRequest servletRequest;
+    private boolean cancelOnClientDisconnect = true;
+    private ListenableFuture<?> future;
+    private Executor httpExecutor;
```
electrum, line 26: I don't like that these required fields need to be set using the builder methods. I think it would be better to make them required arguments to `bindTo()` similar to `bindAsyncResponse()`.

jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseHandler.java
```diff
@@ -23,47 +23,56 @@
...
+    public AsyncResponseHandler<T> withTimeout(Duration timeout)
     {
         return withTimeout(timeout,
                 status(Response.Status.SERVICE_UNAVAILABLE)
                         .entity("Timed out after waiting for " + timeout.convertToMostSuccinctTimeUnit())
                         .build());
     }
 
-    public AsyncResponseHandler withTimeout(Duration timeout, Response timeoutResponse)
+    public void clientDisconnected()
```
electrum, line 64: Move this method below the other `withTimeout()` methods so that the diff is cleaner

electrum (commented):
This looks good to me. @dain should review the APIs as he is more familiar with this

electrum:
Nit: commit title for second commit is too long

jaxrs/src/test/java/io/airlift/jaxrs/TestRequestScopedBinding.java
```diff
@@ -0,0 +1,141 @@
...
+            return new RemoteAddr(request.getRemoteAddr());
+        }
+    }
+
+    private static class HeaderValueFactory
+            implements Supplier<HeaderValue>
+    {
+        private final HttpHeaders headers;
+
+        @Inject
```
electrum, line 129: This shouldn't be needed as this object is not constructed with Guice

jaxrs/src/test/java/io/airlift/jaxrs/TestRequestScopedBinding.java
```diff
@@ -0,0 +1,141 @@
...
+                },
+                new TestingNodeModule(),
+                new TestingHttpServerModule(),
+                new JaxrsModule(),
+                new JsonModule())
+                .quiet()
+                .initialize();
+    }
+
+    private record RemoteAddrFactory(HttpServletRequest request)
```
electrum, line 109: How does the injection work if the classes and constructors are private? Does Jersey bypass permissions with reflection?

wendigo: As far as I know - yes

jaxrs/src/main/java/io/airlift/jaxrs/JaxrsBinder.java
```diff
@@ -49,6 +55,16 @@ public void bind(Key<?> targetKey)
         resourceBinder.addBinding().to(targetKey).in(SINGLETON);
     }
 
+    public <T> void bindRequestScoped(Class<T> requestScopeClass, Class<? extends Supplier<T>> scopedFactory)
+    {
+        factoryBinder.addBinding().toInstance(scopedFactoryBinding(requestScopeClass, scopedFactory));
+    }
+
+    public <T> void bindRequestScoped(Class<T> requestScopeClass, Supplier<T> scopedFactory)
```
electrum, line 63: This doesn't seem to be tested. How would it work? Is this binding a singleton?

jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseBinder.java
wendigo: Makes sense. Moved to method

jaxrs/src/main/java/io/airlift/jaxrs/JaxrsBinder.java
wendigo: Yeah, probably doesn't make sense. Dropped.

jaxrs/src/test/java/io/airlift/jaxrs/TestRequestScopedBinding.java
wendigo: Right, dropped.

jaxrs/src/test/java/io/airlift/jaxrs/TestRequestScopedBinding.java
```diff
@@ -0,0 +1,140 @@
...
+                .setHeader("X-Name", name)
+                .build();
+        return httpClient.execute(request, createStringResponseHandler()).getBody();
+    }
+
+    @Path("/hello")
+    public static class MyResource
+    {
+        @GET
+        public String returnRequestScopedInfo(@Context RemoteAddr remoteAddr, @Context HeaderValue headerValue)
```
dain, line 64: This can be accomplished with standard JaxRs.

First you use `@BeanParam` instead `@Context`:
```java
public String returnRequestScopedInfo(@BeanParam RemoteAddr remoteAddr, @BeanParam HeaderValue headerValue)
```
Then for the classes you can do:
```java
    public record RemoteAddr(String value)
    {
        public RemoteAddr(@Context HttpServletRequest request)
        {
            this(request.getRemoteAddr());
        }
    }

    public static class HeaderValue
    {
        private final String value;

        private HeaderValue(@Context HttpHeaders headers)
        {
            this.value = headers.getHeaderString("X-Name");
        }

        public String getValue()
        {
            return value;
        }
    }
```

jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseBinder.java
```diff
@@ -0,0 +1,110 @@
...
+import static java.util.Objects.requireNonNull;
+
+public class AsyncResponseBinder
+{
+    private final HttpServletRequest servletRequest;
+    private boolean cancelOnClientDisconnect = true;
+    private Optional<Duration> timeout = Optional.empty();
+    private Function<Duration, Supplier<Response>> timeoutResponse = AsyncResponseBinder::defaultTimeoutResponse;
+
+    private AsyncResponseBinder(HttpServletRequest servletRequest)
```
dain, line 28: If we go with the @BeanParam solution, then can be:
```java
private AsyncResponseBinder(@Context HttpServletRequest servletRequest, @Suspended AsyncResponse asyncResponse)
```

jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseBinder.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    public AsyncResponseBinder withTimeoutResponse(Supplier<Response> responseSupplier)
+    {
+        requireNonNull(responseSupplier, "responseSupplier is null");
+        this.timeoutResponse = timeout -> responseSupplier;
+        return this;
+    }
+
+    public void bindTo(AsyncResponse response, ListenableFuture<?> future, Executor httpExecutor)
```
dain, line 59: If we do the above the `AsyncResponse response` parameter here would not be necessary.

jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseBinder.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.Optional;
+import java.util.concurrent.Executor;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import static io.airlift.jaxrs.AsyncResponseHandler.bindAsyncResponse;
+import static jakarta.ws.rs.core.Response.status;
+import static java.util.Objects.requireNonNull;
+
+public class AsyncResponseBinder
```
dain, line 21: This should have a test... nothing complicated, just make sure it works at all

dain (commented):
I'd prefer we don't extend the `@Context` annotation usage. Jersey lets you extend the objects bound with context, but other implementations don't. Instead I think we can just use the built in `@BeanParam`.  If we don't like that name of that annotation, we could write a `ParamConverterProvider` to support a different annotation (chatGPT can provide a basic implementation)

jaxrs/src/main/java/io/airlift/jaxrs/AsyncResponseBinder.java
wendigo: That won't work. You can inject @Suspended but it's not handled properly.

wendigo: Actually, it does but you need to annotate an argument with both @BeanParam and @Suspended

PR #1171, opened 2024-05-25
Update Jersey to 3.1.7

wendigo:


wendigo:
Tested in https://github.com/trinodb/trino/pull/22133

PR #1172, opened 2024-05-26
Support request body streaming

Randgalt:
Supports streaming the response of one request as the body of another request. i.e. a very simple pipe between two requests/responses. For the Jetty implementation, overrides with a Jetty `InputStreamRequestContent`.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/StreamingBodyGenerator.java
+++ b/http-client/src/main/java/io/airlift/http/client/StreamingBodyGenerator.java
@@ -1,0 +1,35 @@
+package io.airlift.http.client;
+
+import java.io.InputStream;
+import java.io.OutputStream;
+
+import static java.util.Objects.requireNonNull;
+
+public class StreamingBodyGenerator
+        implements BodyGenerator
+{
+    private final InputStream source;
+
+    public static StreamingBodyGenerator streamingBodyGenerator(InputStream source)
+    {
+        return new StreamingBodyGenerator(source);
+    }
+
+    public InputStream source()
+    {
+        return source;
+    }
+
+    private StreamingBodyGenerator(InputStream source)
+    {
+        this.source = requireNonNull(source, "source is null");
+    }
+
+    @SuppressWarnings("deprecation")
+    @Override
+    public void write(OutputStream out)
+            throws Exception
+    {
+        throw new UnsupportedOperationException();
+    }
+}
--- a/http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
+++ b/http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
@@ -47,6 +47,8 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import static com.google.common.base.Throwables.getStackTraceAsString;
 import static com.google.common.base.Throwables.throwIfInstanceOf;
@@ -63,7 +65,9 @@ import static io.airlift.http.client.Request.Builder.prepareDelete;
 import static io.airlift.http.client.Request.Builder.prepareGet;
 import static io.airlift.http.client.Request.Builder.preparePost;
 import static io.airlift.http.client.Request.Builder.preparePut;
+import static io.airlift.http.client.ResponseHandlerUtils.propagate;
 import static io.airlift.http.client.StatusResponseHandler.createStatusResponseHandler;
+import static io.airlift.http.client.StreamingBodyGenerator.streamingBodyGenerator;
 import static io.airlift.http.client.StringResponseHandler.createStringResponseHandler;
 import static io.airlift.testing.Assertions.assertBetweenInclusive;
 import static io.airlift.testing.Assertions.assertGreaterThanOrEqual;
@@ -72,6 +76,7 @@ import static io.airlift.units.Duration.nanosSince;
 import static java.lang.String.format;
 import static java.lang.Thread.currentThread;
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.concurrent.Executors.newVirtualThreadPerTaskExecutor;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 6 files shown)


wendigo (approved):

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -511,6 +520,70 @@ public void testPutMethodWithFileBodyGenerator()
         assertTrue(testFile.delete());
     }
 
+    @Test
+    public void testPutMethodWithStreamingBodyGenerator()
```
wendigo, line 524: I don't fully understand this test tbh. Please explain

Randgalt: It tests the response stream can be outside after `httpClient.execute()` returns. Note that prior to this change, `httpClient.execute()` always closed the response stream and ended the request. This test proves that the response stream still works correctly, etc. after `httpClient.execute()`.

Randgalt: This test proves piping the result of one request as the body of another request, which is the main purpose of this change. For example, imagine a proxy that sits between a client and a server. The client, say, generates 32GB of data for the remote server. Without this change, the proxy would need to spool the 32GB somewhere before sending to the remote client. This test proves that this works. It executes a request and, in the response handler, executes _another_ request while piping the response of the first request as the body of the second request.

Randgalt: Update - I took a look again and see a big mistake. Let me correct it...

Randgalt: It's better now. Sorry about that.

wendigo: Maybe let's pipe from tmp file instead and on the servlet side assert on the received length?

Randgalt: Let's do both. I really want to test an existing, open request streaming into another request as that's the main use-case.

http-client/src/main/java/io/airlift/http/client/StreamingBodyGenerator.java
```diff
@@ -0,0 +1,36 @@
...
+        this.source = requireNonNull(source, "source is null");
+    }
+
+    @SuppressWarnings("deprecation")
+    @Override
+    public void write(OutputStream out)
+            throws Exception
+    {
+        source.transferTo(out);
+        out.flush();
```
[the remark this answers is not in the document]
Randgalt: In my testing, this was needed. It can't hurt anyway.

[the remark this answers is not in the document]
electrum: I'm confused. This method should never be called, right? We're transforming it into Jetty's `InputStreamRequestContent`. I'd prefer to make this method throw `UnsupportedOperationException`.

http-client/src/main/java/io/airlift/http/client/StreamingBodyGenerator.java
```diff
@@ -0,0 +1,36 @@
+package io.airlift.http.client;
+
+import java.io.InputStream;
+import java.io.OutputStream;
+
+import static java.util.Objects.requireNonNull;
+
+public class StreamingBodyGenerator
```
electrum, line 8: Update the Javadoc in `BodyGenerator` to mention this class.

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -511,6 +516,90 @@ public void testPutMethodWithFileBodyGenerator()
...
+        assertEquals(servlet.getRequestBytes(), contents);
+
+        assertTrue(testFile.delete());
+    }
+
+    @Test
+    public void testPutMethodWithStreamingBodyGenerator()
+            throws Exception
+    {
+        String responseBody = IntStream.range(0, 10000).mapToObj(ignore -> "hello").collect(Collectors.joining());
```
electrum, line 548: This is so small that I'm concerned we're not really testing the streaming behavior.

http-client/src/main/java/io/airlift/http/client/StreamingBodyGenerator.java
electrum: We should probably seal the interface, as a separate change.

Randgalt: https://github.com/airlift/airlift/issues/1181

http-client/src/main/java/io/airlift/http/client/StreamingBodyGenerator.java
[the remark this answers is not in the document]
Randgalt: I was just copying the others. I'll change to unsupported.

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
Randgalt: I'll change to `100_000` hellos. That should be enough.

http-client/src/test/java/io/airlift/http/client/jetty/TestAsyncJettyHttpClient.java
```diff
@@ -52,4 +53,14 @@ public <T, E extends Exception> T executeRequest(HttpClientConfig config, Reques
             return executeAsync(client, request, responseHandler);
         }
     }
+
+    @Test(enabled = false)
+    protected void testPutMethodWithStreamingBodyGenerator(boolean largeContent)
+            throws Exception
+    {
+        // don't test with async clients as they buffer responses and the LARGE content is too big
```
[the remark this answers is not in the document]
Randgalt: I need to think about it. It might be complex. Give me some time.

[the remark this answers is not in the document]
Randgalt: tbh - the easiest thing would be to wrap the sync exec call in a virtual thread. We could do that internally when streaming is used. wdyt?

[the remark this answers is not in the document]
electrum: I don't think there is a point to making the async client support streaming. You use async because you want to submit requests and get notified when they are complete, without blocking the current thread. To do streaming, you either need a completely different async API, or you block the thread to stream the results. If you're blocking the thread, then you might as well use the sync API. This is why the async API is for size limited request/response that can be fully buffered.

wendigo (approved):

PR #1174, opened 2024-05-26
Introduce StreamingResponseHandler for streaming responses

Randgalt:
Introduce `StreamingResponseHandler` which allows callers
to stream response content in whatever manner they want. The
HttpClient does not close response streams or connection. It
is the responsibility of the caller to close the `StreamingResponseHandler`
so that the request is completed.

`StreamingResponseHandler` is written in a way that maps well to
Jakarta's `StreamingOutput` class. Combined with an `AsyncResponse`
async streaming responses, piped request-to-responses, etc. are
possible.

`StreamingResponseHandler` implements `Closeable` to ensure/indicate
that the caller is responsible for finalizing/closing the response.


http-client/src/main/java/io/airlift/http/client/StreamingResponseHandler.java
```diff
@@ -0,0 +1,101 @@
...
+        {
+            InputStream inputStream = response.getInputStream();
+            inputStream.transferTo(output);
+            output.flush();
+        }
+
+        @Override
+        public void close()
+        {
+            requireNonNull(completer, "completer is null. The request did not end correctly.")
```
electrum, line 53: This should be `checkState()` so that we throw the proper exception type

http-client/src/main/java/io/airlift/http/client/StreamingResponseHandler.java
```diff
@@ -0,0 +1,101 @@
...
+        {
+            this.response = requireNonNull(response, "response is null");
+            this.statusCode = response.getStatusCode();
+            this.headers = ImmutableListMultimap.copyOf(response.getHeaders());
+        }
+
+        public void write(OutputStream output)
+                throws IOException
+        {
+            InputStream inputStream = response.getInputStream();
```
electrum, line 45: Nit: you can inline this variable

http-client/src/main/java/io/airlift/http/client/StreamingResponseHandler.java
```diff
@@ -0,0 +1,101 @@
...
+    public StreamingResponse handle(Request request, Response response)
+            throws RuntimeException
+    {
+        return new StreamingResponse(response);
+    }
+
+    @Override
+    public void completeRequest(Runnable completer)
+    {
+        if (this.completer != null) {
```
electrum, line 96: `checkState()`

http-client/src/main/java/io/airlift/http/client/StreamingResponseHandler.java
```diff
@@ -0,0 +1,101 @@
...
+            implements Closeable
+    {
+        private final int statusCode;
+        private final ListMultimap<HeaderName, String> headers;
+        private final Response response;
+
+        private StreamingResponse(Response response)
+        {
+            this.response = requireNonNull(response, "response is null");
+            this.statusCode = response.getStatusCode();
```
electrum, line 38: No need to copy these fields since we're holding onto `Response`. We can say that these are expected to be immutable in `Response` as we own the implementation.

Or, we could also fetch `InputStream` in the constructor and not hold `Response`, since there is no need for that call to be lazy. Either way is fine with me.

http-client/src/main/java/io/airlift/http/client/StreamingResponseHandler.java
```diff
@@ -0,0 +1,101 @@
...
+
+        public int statusCode()
+        {
+            return statusCode;
+        }
+
+        public Optional<String> header(String name)
+        {
+            List<String> values = headers().get(HeaderName.of(name));
+            return values.isEmpty() ? Optional.empty() : Optional.ofNullable(values.getFirst());
```
electrum, line 65: The value can't be null, as Guava immutable collections don't allow that

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -584,6 +589,25 @@ public void testResponseStatusCode()
         assertEquals(statusCode, 543);
     }
 
+    @Test
+    public void testStreamingResponseHandler()
+            throws Exception
+    {
+        String content = IntStream.range(0, 16_000).mapToObj(ignore -> "x").collect(Collectors.joining());
```
electrum, line 596: How long does the test take if we make this 100MB? The above seems too small to really verify that streaming large responses is working.

electrum (approved):

http-client/src/main/java/io/airlift/http/client/StreamingResponseHandler.java
Randgalt: I wasn't sure the consequences of accessing the `InputStream` early. Let's get rid of the fields and just use `response`.

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
Randgalt: Interesting result from that. 100MB allocates fast, let's keep that. But, it exposed HttpClient's executeAsync buffers the response. I've changed the code to throw an exception if `executeAsync` is used with `StreamingResponseHandler`. This seems reasonable to me. If you wanted async streaming you'd use a Virtual thread anyway.

Randgalt: Update: 100MB was breaking CI (OOM). I changed to 10MB.

Randgalt:
attn @electrum, @wendigo et al. If we want to support async response streaming we'd need to do something like this. I can't think of another way. Given we have Virtual threads now I believe we don't need to support async streaming. LMKWYT. https://github.com/Randgalt/airlift/compare/jordanz/http-client-response-streaming...Randgalt:airlift:jordanz/async-streaming?expand=1

Randgalt:
cc @wendigo

wendigo:
@Randgalt sorry, you need to rebase it once again. I've made changes to how test servers are created - these are no longer shared so that tests can now run in parallel.

Randgalt:
@wendigo PTAL

http-client/src/main/java/io/airlift/http/client/StreamingResponseHandler.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        public Optional<String> header(String name)
+        {
+            return hasHeader(name) ? Optional.of(headers(name).getFirst()) : Optional.empty();
+        }
+
+        public List<String> headers(String name)
+        {
+            return hasHeader(name) ? response.getHeaders(name) : ImmutableList.of();
```
wendigo, line 71: do we need these conditions?

```
Returns a view collection of the values associated with key in this multimap, if any. Note that when containsKey(key) is false, this returns an empty collection, not null.
```

wendigo, line 71: just `response.getHeaders(name)`

http-client/src/test/java/io/airlift/http/client/jetty/TestAsyncJettyHttpClient.java
```diff
@@ -40,6 +44,19 @@ public <T, E extends Exception> T executeRequest(CloseableTestHttpServer server,
         return executeAsync(client, request, responseHandler);
     }
 
+    @Test
+    public void testStreamingResponseHandler()
+            throws Exception
+    {
+        try {
```
wendigo, line 51: use: `assertThatThrownBy(super::testStreamingResponseHandler)` instead

http-client/src/main/java/io/airlift/http/client/StreamingResponseHandler.java
Randgalt: This is copy/paste from the other handler

http-client/src/main/java/io/airlift/http/client/ResponseHandler.java
```diff
@@ -15,11 +15,23 @@
...
 
     T handle(Request request, Response response)
             throws E;
+
+    default void completeRequest(Runnable completer)
+    {
+        completer.run();
+    }
+
+    default void endSpan(Span span)
```
wendigo, line 33: why do we need this one?

Randgalt: The span shouldn't end until all the bytes have been read

wendigo: Yeah, I figured

Randgalt:
Closing in favor of https://github.com/airlift/airlift/pull/1204

PR #1175, opened 2024-05-31
Properly catch errors thrown by request handlers

erichwang:
If the request handlers throw any exceptions, the call to JettyHttpClient would previously fail with an inline exception, rather than returning a failed future. This is fixed here.


wweiss-starburst (approved):

http-client/src/main/java/io/airlift/http/client/jetty/FailedHttpResponseFuture.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static com.google.common.util.concurrent.Futures.immediateFailedFuture;
+
+public class FailedHttpResponseFuture<T>
+        implements HttpClient.HttpResponseFuture<T>
```
erichwang, line 14: I hate that we have this custom future interface. But whatever.

electrum, line 14: I think this can be
```java
public class FailedHttpResponseFuture<T>
        extends SimpleForwardingListenableFuture<T>
        implements HttpClient.HttpResponseFuture<T>
{
    public FailedHttpResponseFuture(Throwable throwable)
    {
        super(immediateFailedFuture(throwable));
    }

    @Override
    public String getState()
    {
        return "FAILED";
    }
}

electrum (approved):

http-client/src/main/java/io/airlift/http/client/jetty/FailedHttpResponseFuture.java
erichwang: NICE

PR #1176, opened 2024-06-03
Support list and set injection into configuration

wendigo:


CHANGES
```diff
@@ -3,6 +3,7 @@
 - Fix failure when system properties modified when loading configuration
 - Update Jackson to 2.17.1
 - Update Jersey to 3.1.7
+- Add support for Enum array injection into configuration classes
```
findepi, line 6: The code looks like supporting not only enum arrays. Am i reading it wrong?

wendigo: You are right, you can also inject arrays of any other supported types

electrum:
The code change looks good, but I’d like to understand why we need this. What’s special about arrays of enums? We would typically prefer lists.

findepi:
> We would typically prefer lists.

that was supposed to be my next question.

findepi:
Lists support unlimited generics, while arrays are practically limited to non-generic types.

wendigo:
@electrum we don't support List of enums injection either.

findepi:
@wendigo you should be able to get full generic type for lists
Use TypeToken to resolve them

wendigo:
@findepi added support for list/set/array injection into configuration classes

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -665,15 +669,45 @@ private static Object coerce(Class<?> type, String value)
             return match;
         }
 
-        if (LIST_OF_STRINGS_TYPE_TOKEN.isSubtypeOf(TypeToken.of(type))) {
+        if (type.isArray()) {
```
findepi, line 672: let's not add array support. not needed

wendigo: Dropped

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -688,11 +687,6 @@ private static Object coerce(TypeToken<?> type, String value)
                     .collect(toImmutableList());
         }
 
-        if (LIST_OF_STRINGS_TYPE_TOKEN.isSubtypeOf(type)) {
```
findepi: This can be part of `Support list and set injection into configuration` commit

Also, reword the commit and CHANGES entry `Support generic list and set injection ...` -- this because we had _some_ list injection, but now we have better one.

findepi (approved):

PR #1178, opened 2024-06-04
Allow enabling Jetty's case sensitive header cache

Randgalt:
See: https://bugs.eclipse.org/bugs/show_bug.cgi?id=414449#c4

Jetty re-writes headers based on a cache. This can result in headers being sent to clients with a cached value that has a different case then what was sent from the remote client. In applications that must validate headers for hash signatures this is a severe problem as the value that was sent by the client is lost.

Add a binding that makes Jetty's header cache case sensitive. This is a documented feature of Jetty.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,7 @@
 - Support generic list and set injection into configuration classes
 - Update Jetty to 12.0.10
 - Update HdrHistogram to 2.2.2
+- Allow configuring Jetty's header cache case-sensitiveness
 
 248
 
--- a/http-server/src/test/java/io/airlift/http/server/testing/AbstractTestTestingHttpServer.java
+++ b/http-server/src/test/java/io/airlift/http/server/testing/AbstractTestTestingHttpServer.java
@@ -27,6 +27,7 @@ import io.airlift.http.client.HttpClient;
 import io.airlift.http.client.HttpClientConfig;
 import io.airlift.http.client.HttpStatus;
 import io.airlift.http.client.HttpUriBuilder;
+import io.airlift.http.client.Request;
 import io.airlift.http.client.StatusResponseHandler.StatusResponse;
 import io.airlift.http.client.StringResponseHandler;
 import io.airlift.http.client.jetty.JettyHttpClient;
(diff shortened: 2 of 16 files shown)


wendigo:
@Randgalt I've just updated jetty to version 12.0.10. Can you validate its' behaviour?

Randgalt:
I just tested. This fix is still good.

wendigo:
@Randgalt how we can test that?

Randgalt:
It's complex. I can set up a test if needed. It didn't seem worth it to me.

Randgalt:
@wendigo I figured out a test and added it

http-server/src/test/java/io/airlift/http/server/testing/AbstractTestTestingHttpServer.java
```diff
@@ -273,6 +276,48 @@ public void testGuiceInjectionWithResources()
         }
     }
 
+    @SuppressWarnings("SynchronizationOnLocalVariableOrMethodParameter")
+    @Test
+    public void testHeaderCaseSensitivity()
+            throws Exception
+    {
+        skipUnlessJdkHasVirtualThreads();
```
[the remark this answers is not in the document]
Randgalt: copy/paste

CHANGES
```diff
@@ -6,6 +6,7 @@
 - Support generic list and set injection into configuration classes
 - Update Jetty to 12.0.10
 - Update HdrHistogram to 2.2.2
+- Jetty re-writes many headers. Added config to ensure that case-sensitivity is retained.
```
wendigo, line 9: Allow configuring Jetty's header cache case-sensitiveness.

wendigo (approved):

Randgalt:
Thanks @wendigo !

electrum:
Why not always enable this?

Randgalt:
> Why not always enable this?

I’m not sure. It’s been as it is for a long time. But we could.

http-server/src/test/java/io/airlift/http/server/testing/AbstractTestTestingHttpServer.java
```diff
@@ -273,6 +276,47 @@ public void testGuiceInjectionWithResources()
         }
     }
 
+    @SuppressWarnings("SynchronizationOnLocalVariableOrMethodParameter")
```
findepi, line 279: is this avoidable?

Randgalt: We could move the method to the servlet and make it synchronized

PR #1180, opened 2024-06-04
HTTP/2 support for the client/server

wendigo:


http-client/src/main/java/io/airlift/http/client/jetty/JettyResponseFuture.java
```diff
@@ -62,6 +64,8 @@ private enum JettyAsyncHttpState
         this.span = requireNonNull(span, "span is null");
         this.stats = requireNonNull(stats, "stats is null");
         this.recordRequestComplete = recordRequestComplete;
+
+        jettyRequest.onRequestCommit(this);
```
electrum, line 68: Register this on the outside, so that we don't leak a partially constructed object

http-client/src/main/java/io/airlift/http/client/jetty/JettyResponseFuture.java
```diff
@@ -70,6 +74,19 @@ public String getState()
         return state.get().toString();
     }
 
+    @Override
+    public void onCommit(org.eclipse.jetty.client.Request request)
+    {
+        state.set(JettyAsyncHttpState.REQUEST_COMMITED);
+    }
+
+    public void idleTimeout()
```
electrum, line 83: Make this package-private

http-client/src/main/java/io/airlift/http/client/jetty/JettyResponseFuture.java
```diff
@@ -70,6 +74,19 @@ public String getState()
...
+    {
+        if (state.get() == JettyAsyncHttpState.WAITING_FOR_CONNECTION) {
```
electrum, line 85: Since this can be called from a different thread than the other Jetty state handling, we probably need to make all the methods synchronized. Otherwise, for example, we could be in the middle of executing `cancel()` and have set the status to cancelled but not updated the state yet. Also, this is just failing the future, but isn't touching the Jetty state, so `cancel()` or `completed()` could be called after this.

This idle timeout change seems to be more complicated than it appears. Why do we need to do this ourselves? Does the Jetty client already do this and maybe we're missing a callback or something?

How is it related to the HTTP/2 change?

If this change is required in the current form, then it will take some careful consideration of the response future state machine. Let's pull this out into a separate PR.

http-server/src/main/java/io/airlift/http/server/HttpServerBinder.java
```diff
@@ -57,6 +57,21 @@ public HttpServerBinder enableLegacyUriCompliance()
         return this;
     }
 
+    public void withFeatures(HttpServerFeatures serverFeatures)
```
electrum, line 60: Can you make this API change a separate commit or PR? It's not clear how this is related to HTTP/2.

PR #1182, opened 2024-06-05
Seal BodyGenerator - disallow custom implementations

### LINKED ISSUE #1181: Seal BodyGenerator
We should probably seal the interface, as a separate change.

_Originally posted by @electrum in https://github.com/airlift/airlift/pull/1172#discussion_r1626412524_

Randgalt:
Closes #1181

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/BodyGenerator.java
+++ b/http-client/src/main/java/io/airlift/http/client/BodyGenerator.java
@@ -15,14 +15,7 @@
  */
 package io.airlift.http.client;
 
-import java.io.OutputStream;
-
-public interface BodyGenerator
+public sealed interface BodyGenerator
+        permits ByteBufferBodyGenerator, FileBodyGenerator, StaticBodyGenerator, StreamingBodyGenerator
 {
-    /**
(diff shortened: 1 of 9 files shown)


electrum (approved):
We can also remove the deprecated `write()` method

Randgalt (commented):
@electrum I just added a commit that introduces a new `JettyBodyGenerator` to allow custom body generators. wdyt?

electrum:
I don't think we want that. The Airlift API is intended to be independent of Jetty. The Jetty content API doesn't seem like something you would want to implement directly, so if there is a use case, let's figure out an appropriate API.

electrum:
Why would we need that after #1172?

Randgalt:
> We can also remove the deprecated `write()` method

@electrum we need a solution for `JaxrsTestingHttpProcessor.getRequestBody()`. We could add a similar switch on `BodyGenerator` but that feels wrong. Maybe we remove `JaxrsTestingHttpProcessor`. What is it used for?

electrum:
Hmm. It's used to allow direct testing of a JAX-RS resource via `TestingHttpClient`. It's used in one place in Trino. Since this is part of Airlift and it's for testing, having knowledge of the body generators seems fine. Simply supporting `StaticBodyGenerator` is probably sufficient.

PR #1184, opened 2024-06-05
Small improvements

wendigo:
Extracted from HTTP/2 work (unrelated).

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyResponse.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyResponse.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ListMultimap;
 import com.google.common.io.CountingInputStream;
 import io.airlift.http.client.HeaderName;
+import io.airlift.http.client.HttpVersion;
 import org.eclipse.jetty.client.Response;
 import org.eclipse.jetty.http.HttpFields;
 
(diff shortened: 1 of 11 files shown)


http-client/src/main/java/io/airlift/http/client/jetty/JettyResponse.java
```diff
@@ -31,6 +32,16 @@ public int getStatusCode()
         return response.getStatus();
     }
 
+    @Override
+    public HttpVersion getHttpVersion()
+    {
+        return switch (response.getVersion()) {
+            case HTTP_1_1 -> HttpVersion.HTTP_1_1;
```
electrum, line 39: We could map `HTTP_0_9` and `HTTP_1_0` to `HTTP_1_1`. Or maybe name ours `HTTP_1` since we don't care about the distinction?

http-client/src/main/java/io/airlift/http/client/jetty/JettyResponse.java
```diff
@@ -31,6 +32,16 @@ public int getStatusCode()
...
+            case HTTP_2 -> HttpVersion.HTTP_2;
+            default -> throw new IllegalStateException("Unexpected HTTP version: " + response.getVersion());
```
electrum, line 41: This will break if the server uses `HTTP_3`. I think we need to add an `HTTP_3` enum, or make this return optional.

electrum (commented):
Looks good, but we need to make the response enum handling more robust.

http-client/src/main/java/io/airlift/http/client/jetty/JettyResponse.java
wendigo: Right, fixed, now the switch is exhaustive

wendigo:
@electrum improved

electrum (approved):

PR #1185, opened 2024-06-05
Update airbase to 157

wendigo:


wendigo:
Guava update requires stripping off Scope ID that is attached to InetAddress and previously was ignored when `.toAddrString` was called

node/src/main/java/io/airlift/node/AddressToHostname.java
```diff
@@ -28,16 +30,39 @@ private AddressToHostname() {}
...
+            default -> InetAddresses.toAddrString(inetAddress);
+        };
+
+        return hostname + IP_ENCODED_SUFFIX;
+    }
+
+    // Same as in the https://github.com/google/guava/commit/3f61870ac6e5b18dbb74ce6f6cb2930ad8750a43
+    private static String scopeWithDelimiter(Inet6Address ip)
+    {
+        // getHostAddress on android sometimes maps the scope id to an invalid interface name; if the
```
electrum, line 54: We don’t run on Android

wendigo: I wanted to keep this method the same as in Guava (since they are adding a scope id and we need to remove it)

wendigo: I've implemented it other way around (as suggested by the commit message in the Guava's change).

electrum (approved):

PR #1186, opened 2024-06-10
Notify AsyncContext on remote exception

wendigo:
https://github.com/jetty/jetty.project/pull/11897


findepi (commented):
cc @losipiuk 

is this testable?

wendigo:
@findepi hardly :(

wendigo:
Right now this is unguarded by the Jetty's config property (so on by default) but since this is going to be fixed we need to set this config explicitly.

findepi (approved):
code lgtm. it would be awesome to have a test.
i don't understand the limitation here

PR #1187, opened 2024-06-12
ALPN, HTTP/2 support for server and the client

wendigo:


wendigo:
Minimal change to get HTTP/2 support.

wendigo:
@dain @electrum PTAL. I've tested these changes and I think that we can land this.

dain (approved):

PR #1189, opened 2024-06-15
Add TOML configuration support

wendigo:


wendigo:
I wanted to keep it simple so instead of introducing new abstractions - TOML files are read and converted to Map<String, String> for consistency with .properties loading

martint:
Nice! I've been wanting to look at what it it would take to support TOML for configuration in Airlift.

I do think, however, that we need to explore how TOML types and general structure map into Airlift configs more carefully. The approach of interpreting everything as strings may make cause incompatibilities once we do proper and idiomatic mappings.

wendigo:
@martint we could replace coercion that we have right now in ConfigurationFactory with an explicit typed get

wendigo:
@martint i've added support for injecting lists and maps recently so we can leverage that. We could easily add configuration class nesting

PR #1190, opened 2024-06-18
Sanitize JMX metric names

vinay-kosaraju:
When scraping metrics from HikariCP, exported metrics are not compliant with openmetrics. With this change`JMX_com_zaxxer_hikari_TYPE_Pool (HikariPool-1)_ ATTRIBUTE_ActiveConnections` would become
`JMX_com_zaxxer_hikari_TYPE_Pool_HikariPool_1__ATTRIBUTE_ActiveConnections`

### DIFF
--- a/openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
+++ b/openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
@@ -13,6 +13,8 @@
  */
 package io.airlift.openmetrics;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.CharMatcher;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Inject;
 import io.airlift.log.Logger;
@@ -58,6 +60,13 @@ public class MetricsResource
     private static final String TYPE_SEPARATOR = "_TYPE_";
     private static final String NAME_SEPARATOR = "_NAME_";
     private static final Pattern METRIC_NAME_PATTERN = Pattern.compile("[[a-zA-Z]][\\w_]*");
+    private static final CharMatcher NON_ALLOWED_LABEL_CHARACTERS = CharMatcher
+            .inRange('a', 'z')
+            .or(CharMatcher.inRange('A', 'Z'))
+            .or(CharMatcher.inRange('0', '9'))
+            .or(CharMatcher.anyOf("_"))
+            .negate()
+            .precomputed();
 
     private final MBeanServer mbeanServer;
     private final MBeanExporter mbeanExporter;
@@ -146,7 +155,7 @@ public class MetricsResource
         }
 
         StringBuilder metricNameBuilder = new StringBuilder("JMX_")
-                .append(objectName.getDomain().replace(".", "_"));
+                .append(objectName.getDomain());
 
         if (objectName.getKeyProperty("name") != null) {
             metricNameBuilder.append(NAME_SEPARATOR)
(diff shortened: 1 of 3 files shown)


vinay-kosaraju:
cc: @mattstep @electrum

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -246,7 +246,10 @@ private String sanitizeMetricName(String name)
                 .replace(":", "_")
                 .replace("=", "_")
                 .replace(",", "_")
-                .replace("-", "_");
+                .replace("-", "_")
+                .replace(" ", "_")
+                .replace("(", "")
+                .replace(")", "");
```
findepi, line 252: what about having a list of allowed characters instead of a list of characters to replace?
for example, should `%` or `;` be allowed or not?

findepi (approved):

mattstep (approved):

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
wendigo: Yes, we should replace characters other than allowed. @vinay-kosaraju can you apply this logic?

vinay-kosaraju: Will update the PR

vinay-kosaraju: Updated the PR to replace any non-word character pattern match with `_`. Here's prometheus [metric name convention](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels)

```
\w | A word character: [a-zA-Z_0-9]
-- | --
\W | A non-word character: [^\w]
```

mattstep: A regex pattern might get costly here in cpu. I'd rather stick with a list or precompile a regex. We should only do this for non-jmxutils metrics at least, since those are restricted in the special characters that can come up.

vinay-kosaraju: Considering Matt's concerns regarding performance, I undid my change. @findepi @wendigo let me know what you think?

findepi: @vinay-kosaraju you can use `com.google.common.base.CharMatcher` for regex-like character classes without regex-like overhead.

wendigo:
```
    // https://github.com/OpenObservability/OpenMetrics/issues/53
    private static final CharMatcher NON_ALLOWED_LABEL_CHARACTERS = CharMatcher
            .inRange('a', 'z')
            .or(CharMatcher.inRange('A', 'Z'))
            .or(CharMatcher.inRange('0', '9'))
            .or(CharMatcher.anyOf("_:"))
            .negate()
            .precomputed();
```

and then:

```
private static String sanitizeMetricName(String name)
    {
        return NON_ALLOWED_LABEL_CHARACTERS.replaceFrom(name, '_');
    }
```

wendigo: Also please add test for that method

vinay-kosaraju: Thanks! Updated the PR

wendigo:
@vinay-kosaraju You need to rebase and use JUnit insted of testng as it was dropped. Sorry

wendigo:
I've rebased to use JUnit and AssertJ

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -239,14 +248,10 @@ private String inferAttributesForObjectName(ObjectName objectName)
...
+    @VisibleForTesting
+    static String sanitizeMetricName(String name)
     {
-        return name.replace(".", "_")
-                .replace("$", "_")
-                .replace(":", "_")
-                .replace("=", "_")
-                .replace(",", "_")
-                .replace("-", "_");
+        return NON_ALLOWED_LABEL_CHARACTERS.replaceFrom(name, "_");
```
findepi, line 254: `collapseFrom`

findepi (approved):

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
wendigo: neat!

PR #1197, opened 2024-06-19
Export cache size stats by fixing the method name

weijiii:
## Summary
* The cache size is not exported. E.g.
```
trino:default> describe jmx."current"."trino.sql.gen:name=joinfilterfunctioncompiler";
                   Column                    |  Type   | Extra | Comment 
---------------------------------------------+---------+-------+---------
 joinfilterfunctionfactorystats.hitrate      | double  |       |         
 joinfilterfunctionfactorystats.missrate     | double  |       |         
 joinfilterfunctionfactorystats.requestcount | bigint  |       |         
 node                                        | varchar |       |         
 object_name                                 | varchar |       |         
(5 rows)
```

### DIFF
--- a/jmx/src/main/java/io/airlift/jmx/CacheStatsMBean.java
+++ b/jmx/src/main/java/io/airlift/jmx/CacheStatsMBean.java
@@ -28,7 +28,7 @@ public class CacheStatsMBean
     }
 
     @Managed
-    public long size()
+    public long getSize()
     {
         return cache.size();
     }


findepi (approved):

findepi:
@losipiuk PTAL

jmx/src/main/java/io/airlift/jmx/CacheStatsMBean.java
```diff
@@ -28,7 +28,7 @@ public CacheStatsMBean(Cache<?, ?> cache)
     }
 
     @Managed
-    public long size()
+    public long getSize()
```
losipiuk, line 31: This not being exported looks more like a bug in code which processes @Managed properties.

cc: @martint @electrum

electrum:
This class is deprecated and will be removed. Please apply this to https://github.com/trinodb/trino/blob/master/lib/trino-cache/src/main/java/io/trino/cache/CacheStatsMBean.java

PR #1204, opened 2024-06-21
Add InputStreamResponseHandler

wendigo:


Randgalt:
The `span.end()` here needs to be handled: [`5d33449` (#1204)](https://github.com/airlift/airlift/pull/1204/commits/5d3344996ff052aa6282d62e46f9b78ecc80ec03#diff-a826b6dacbaf35356591584b9fd212d3f905a64c7e972a5222536928f09e2c85R673)

wendigo:
Yeah, right @Randgalt . Added

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -771,6 +774,7 @@ private JettyResponse.CompletionCallback onCompletion(Span span, Request request
                 recordRequestComplete(stats, request, requestSize.get(), requestStart, response, responseStart);
             }
             span.setAttribute(HttpIncubatingAttributes.HTTP_RESPONSE_BODY_SIZE, response.getBytesRead());
+            span.end();
```
Randgalt, line 777: The `finally` in `execute()` above needs to be handled:

```
finally {
            span.end();
        }
```

wendigo: not really, the span is completed by callback

Randgalt:
```
public <T, E extends Exception> T execute(Request request, ResponseHandler<T, E> responseHandler)
            throws E
    {
        request = applyRequestFilters(request);
        Span span = startSpan(request);
        request = injectTracing(request, span);
        try {
            return doExecute(request, responseHandler, span);
        }
        catch (Throwable t) {
            span.setStatus(StatusCode.ERROR, t.getMessage());
            span.recordException(t, Attributes.of(ExceptionAttributes.EXCEPTION_ESCAPED, true));
            throw t;
        }
        finally {
            span.end();
        }
```

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -876,6 +893,31 @@ public void testDowngradedHttpProtocolUsed()
...
+    {
+        try (CloseableTestHttpServer server = newServer()) {
+            server.servlet().setResponseBody(LARGE_CONTENT);
+
+            Request request = prepareGet()
+                    .setUri(server.baseURI())
+                    .build();
+
+            executeOnServer(server, createClientConfig(), client -> {
+                try (InputStreamResponseHandler.InputStreamResponse streamingResponse = client.execute(request, inputStreamResponseHandler())) {
```
Randgalt, line 908: nit: static import for `InputStreamResponseHandler.InputStreamResponse`

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -876,6 +893,31 @@ public void testDowngradedHttpProtocolUsed()
...
+                    ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+                    streamingResponse.write(outputStream);
+                    String responseString = outputStream.toString(UTF_8);
+                    assertThat(responseString).isEqualTo(LARGE_CONTENT);
+                    System.out.println(":: inputStream read completed");
```
Randgalt, line 913: Remove `System.out` and use logging if still desired

Randgalt (approved):

electrum:
Can you explain the motivation for this change? `InputStream` is already available in `Response`.

Randgalt:
> Can you explain the motivation for this change? `InputStream` is already available in `Response`.

@wendigo can explain this particular PR, which is a refactor of my original PR.

Here's the description from my [original PR](https://github.com/airlift/airlift/pull/1174):

Introduce `StreamingResponseHandler` which allows callers
to stream response content in whatever manner they want. The
HttpClient does not close response streams or connection. It
is the responsibility of the caller to close the `StreamingResponseHandler`
so that the request is completed.

`StreamingResponseHandler` is written in a way that maps well to
Jakarta's `StreamingOutput` class. Combined with an `AsyncResponse`
async streaming responses, piped request-to-responses, etc. are
possible.

`StreamingResponseHandler` implements `Closeable` to ensure/indicate
that the caller is responsible for finalizing/closing the response.

http-client/src/main/java/io/airlift/http/client/jetty/JettyResponseFuture.java
```diff
@@ -121,28 +123,30 @@ void completed(Response response, InputStream content)
...
-            if (jettyResponse != null) {
-                span.setAttribute(HttpIncubatingAttributes.HTTP_RESPONSE_BODY_SIZE, jettyResponse.getBytesRead());
-            }
+        catch (Exception e) {
+            throwIfUnchecked(e);
+            throw new RuntimeException(e);
+        }
+    }
+
+    private JettyResponse.CompletionCallback onCompletion(Span span, Request request, long requestStart, long responseStart, Supplier<Long> requestSize)
```
electrum, line 141: This should be `Response.CompletionCallback`

electrum (commented):
This code seems good, but it bothers me that we have to add these extra methods to `Response` that are only used by `InputStreamResponseHandler`. They are very specific to `InputStreamResponseHandler` and users of the HTTP client API shouldn't mess with them.

Thoughts on how we could keep these hidden?

Randgalt:
@wendigo and @electrum Here's my original version that does not touch `Response`. @wendigo can explain why he prefers this version. https://github.com/airlift/airlift/pull/1174

PR #1206, opened 2024-06-21
Add HTTP/2 client max concurrent streams setting

wendigo:


electrum (approved):
Code seems fine, but I'm not sure if we want to unconditionally enable virtual threads

PR #1207, opened 2024-06-23
Config option to enable setting IP_TOS on the http client socket

wenjianhn:


wendigo:
Can you describe impact of this change? How this could be leveraged?

wenjianhn:
@wendigo I will add the impact this change to https://github.com/trinodb/trino this week. 

The goal is to prioritize Trino internal traffic, e.g. heartbeats, over HDFS traffic.
Similar to this ceph issue: https://github.com/ceph/ceph/pull/3353

Thanks.

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -649,6 +650,19 @@ public HttpClientConfig setTcpKeepAliveIdleTime(Duration tcpKeepAliveIdleTime)
         return this;
     }
 
+    @NotNull
```
electrum, line 653: This isn't needed, as the code in this class ensures that the field is not null.

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -649,6 +650,19 @@ public HttpClientConfig setTcpKeepAliveIdleTime(Duration tcpKeepAliveIdleTime)
...
+    public Optional<Integer> getIpTos()
```
electrum, line 654: Add validation for the TOS range:
```java
@Min(0)
@Max(255)
```
This matches the validation in the JDK `sun.nio.ch.Net#setSocketOption`

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -267,6 +267,9 @@ protected void configure(SelectableChannel selectable)
                 if (config.getTcpKeepAliveIdleTime().isPresent()) {
                     setKeepAlive(selectable, config.getTcpKeepAliveIdleTime().get());
                 }
+                if (config.getIpTos().isPresent()) {
```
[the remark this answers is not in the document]
electrum: Unfortunately, `ifPresent()` won't work here as `setIpTos()` throws `IOException`

electrum (approved):

wendigo:
@wenjianhn can you rebase this change to resolve conflicts?

PR #1212, opened 2024-07-03
Prevent NoSuchMethodExceptions being thrown

oskar-szwajkowski:
Do not call getMethod directly, instead filter methods

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationMetadata.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationMetadata.java
@@ -26,6 +26,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -678,14 +679,11 @@ public class ConfigurationMetadata<T>
             return null;
         }
 
-        try {
-            Method method = configClass.getDeclaredMethod(methodName, paramTypes);
-            if (method != null && method.isAnnotationPresent(annotation)) {
+        Method[] configClassMethods = configClass.getDeclaredMethods();
+        for (Method method : configClassMethods) {
+            if (method.getName().equals(methodName) && Arrays.equals(method.getParameterTypes(), paramTypes) && method.isAnnotationPresent(annotation)) {
                 return method;
             }
-        }
-        catch (NoSuchMethodException e) {
-            // ignore
         }
 
         if (configClass.getSuperclass() != null) {
--- a/event/src/main/java/io/airlift/event/client/AnnotationUtils.java
+++ b/event/src/main/java/io/airlift/event/client/AnnotationUtils.java
@@ -19,6 +19,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 
 final class AnnotationUtils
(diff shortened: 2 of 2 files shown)


configuration/src/main/java/io/airlift/configuration/ConfigurationMetadata.java
```diff
@@ -678,15 +679,12 @@ public static Method findAnnotatedMethod(Class<?> configClass, Class<? extends j
             return null;
         }
 
-        try {
-            Method method = configClass.getDeclaredMethod(methodName, paramTypes);
-            if (method != null && method.isAnnotationPresent(annotation)) {
+        Method[] configClassMethods = configClass.getDeclaredMethods();
+        for (Method method : configClassMethods) {
+            if (method.getName().equals(methodName) && Arrays.equals(method.getParameterTypes(), paramTypes) && method.isAnnotationPresent(annotation)) {
```
findepi, line 684: is the linear search faster than JVM own lookup, or slower?

oskar-szwajkowski: Linear search is slower, when method that we search for exists within class
When method does not exists, and `NoSuchMethodException` is thrown, linear search is much faster

I created this sample benchmark:

```java
import org.openjdk.jmh.annotations.Benchmark;

import java.lang.reflect.Method;
import java.util.Arrays;

public class MyBenchmark
{

    @Benchmark
    public void testMethod()
    {
        // This is a demo/sample template for building your JMH benchmarks. Edit as needed.
        // Put your benchmark code here.
        for (int i = 0; i < 10_000; i++) {
            findAnnotatedMethod2(TestSubject.class, Override.class, "qwerty", String.class);
        }
    }

    public static Method findAnnotatedMethod1(Class<?> clazz, Class<? extends java.lang.annotation.Annotation> annotation, String methodName, Class<?>... paramTypes)
    {
        try {
            Method method = clazz.getMethod(methodName, paramTypes);
            if (method.isAnnotationPresent(annotation)) {
                return method;
            }
        }
        catch (NoSuchMethodException e) {
            //
        }
        return null;
    }

    public static Method findAnnotatedMethod2(Class<?> clazz, Class<? extends java.lang.annotation.Annotation> annotation, String methodName, Class<?>... paramTypes)
    {
        Method[] clazzDeclaredMethods = clazz.getDeclaredMethods();
        for (Method method : clazzDeclaredMethods) {
            if (method.getName().equals(methodName) && Arrays.equals(method.getParameterTypes(), paramTypes) && method.isAnnotationPresent(annotation)) {
                return method;
            }
        }
        return null;
    }

    static class TestSubject
        extends TestSubjectBase
    {
        @Override
        public String getName(String arg)
        {
            return "";
        }

        @Override
        public String getName2(String arg)
        {
            return "";
        }

        @Override
        public String getName3(String arg)
        {
            return "";
        }

        @Override
        public String getName4(String arg)
        {
            return "";
        }
    }

    static abstract class TestSubjectBase
    {
        public abstract String getName(String arg);
        public abstract String getName2(String arg);
        public abstract String getName3(String arg);
        public abstract String getName4(String arg);
    }
}
```

and here are results:

existing method with `getMethod`, no exception:

Iteration   1: 3939.791 ops/s
Iteration   2: 3818.048 ops/s
Iteration   3: 3933.231 ops/s
Iteration   4: 3927.964 ops/s
Iteration   5: 3928.367 ops/s

existing method with `getMethods` + search:

Iteration   1: 1504.927 ops/s
Iteration   2: 1498.824 ops/s
Iteration   3: 1460.544 ops/s
Iteration   4: 1503.224 ops/s
Iteration   5: 1473.149 ops/s

NOT existing method with `getMethod` + exception:

Iteration   1: 128.498 ops/s
Iteration   2: 128.420 ops/s
Iteration   3: 128.386 ops/s
Iteration   4: 125.401 ops/s
Iteration   5: 127.580 ops/s

NOT existing method with `getMethods` + search:

Iteration   1: 1708.281 ops/s
Iteration   2: 1710.228 ops/s
Iteration   3: 1612.066 ops/s
Iteration   4: 1697.531 ops/s
Iteration   5: 1719.464 ops/s


in happy path linear search is slower, but its constant no matter whether method exists or not
and in the case I saw, there were many lookups for methods that didn't exist

PR #1213, opened 2024-07-03
Prefetch declared methods only once

wendigo:


wendigo:
I've cleaned this class. It's now much easier to read.

martint (approved):

PR #1217, opened 2024-07-09
Introduce pluggable secrets provider

Praveen2112:
This PR introduces a framework for injecting custom secrets provider which allows us to add secrets to  the configurations dynamically. The configurations be resolved are of a pattern like ${resolver_name:key_for_resolving_config_property} - We have also introduced SecretProviderFactory and SecretProvider which allows us to resolve configuration in runtime. We have migrated the environment loader to this framework

### DIFF
--- a/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
@@ -17,6 +17,7 @@ package io.airlift.bootstrap;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableList.Builder;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSortedMap;
 import com.google.inject.Binder;
 import com.google.inject.Guice;
@@ -30,10 +31,14 @@ import io.airlift.configuration.ConfigurationInspector.ConfigAttribute;
 import io.airlift.configuration.ConfigurationInspector.ConfigRecord;
 import io.airlift.configuration.ConfigurationModule;
 import io.airlift.configuration.WarningsMonitor;
+import io.airlift.configuration.secrets.SecretsPluginManager;
+import io.airlift.configuration.secrets.SecretsResolver;
+import io.airlift.configuration.secrets.env.EnvironmentVariableSecretProvider;
 import io.airlift.log.Logger;
 import io.airlift.log.Logging;
 import io.airlift.log.LoggingConfiguration;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.ArrayList;
@@ -47,7 +52,7 @@ import java.util.TreeMap;
 import static com.google.common.base.Preconditions.checkState;
 import static io.airlift.configuration.ConfigurationLoader.getSystemProperties;
 import static io.airlift.configuration.ConfigurationLoader.loadPropertiesFrom;
-import static io.airlift.configuration.ConfigurationUtils.replaceEnvironmentVariables;
+import static io.airlift.configuration.TomlConfiguration.createTomlConfiguration;
 import static java.lang.String.format;
 
 /**
@@ -72,9 +77,11 @@ public class Bootstrap
     private Map<String, String> optionalConfigurationProperties;
     private boolean initializeLogging = true;
     private boolean quiet;
+    private boolean loadSecretsPlugins;
 
     private State state = State.UNINITIALIZED;
     private ConfigurationFactory configurationFactory;
+    private SecretsResolver secretsResolver;
 
     public Bootstrap(Module... modules)
     {
@@ -134,6 +141,12 @@ public class Bootstrap
         return this;
     }
 
+    public Bootstrap loadSecretsPlugins()
+    {
+        this.loadSecretsPlugins = true;
+        return this;
+    }
+
     /**
      * Validate configuration and return used properties.
      */
@@ -145,6 +158,19 @@ public class Bootstrap
         Logging logging = null;
         if (initializeLogging) {
             logging = Logging.initialize();
+        }
+
+        this.secretsResolver = new SecretsResolver(ImmutableMap.of("env", new EnvironmentVariableSecretProvider()));
+
+        if (loadSecretsPlugins) {
+            log.info("Loading secrets plugins");
+            String secretsConfigFile = System.getProperty("secretsConfig");
+            if (secretsConfigFile != null) {
+                SecretsPluginManager secretsPluginManager = new SecretsPluginManager(createTomlConfiguration(new File(secretsConfigFile)));
+                secretsPluginManager.installPlugins();
+                secretsPluginManager.load();
+                this.secretsResolver = secretsPluginManager.getSecretsResolver();
+            }
         }
 
         Map<String, String> requiredProperties;
@@ -176,15 +202,12 @@ public class Bootstrap
         properties.putAll(requiredProperties);
         properties.putAll(getSystemProperties());
 
-        // replace environment variables in property values
+        // replace secrets in property values
         List<Message> errors = new ArrayList<>();
-        properties = replaceEnvironmentVariables(properties, System.getenv(), (key, error) -> {
+        properties = ImmutableSortedMap.copyOf(secretsResolver.getResolvedConfiguration(properties, (key, error) -> {
             unusedProperties.remove(key);
-            errors.add(new Message(error));
-        });
-
-        // create configuration factory
-        properties = ImmutableSortedMap.copyOf(properties);
+            errors.add(new Message(error.getMessage()));
+        }));
 
         List<Message> warnings = new ArrayList<>();
         configurationFactory = new ConfigurationFactory(properties, warning -> warnings.add(new Message(warning)));
@@ -258,6 +281,7 @@ public class Bootstrap
         moduleList.add(Binder::requireExplicitBindings);
         moduleList.add(Binder::requireExactBindingAnnotations);
 
+        moduleList.add(binder -> binder.bind(SecretsResolver.class).toInstance(secretsResolver));
         moduleList.addAll(modules);
 
         // create the injector
--- a/bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrapWithSecretsProvider.java
+++ b/bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrapWithSecretsProvider.java
@@ -1,0 +1,138 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.bootstrap;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.inject.Injector;
+import io.airlift.configuration.Config;
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static io.airlift.configuration.ConfigBinder.configBinder;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+final class TestBootstrapWithSecretsProvider
+{
+    @Test
+    void testBootstrapWithDefaultSecretsProvider()
+    {
+        Bootstrap bootstrap = new Bootstrap(binder -> configBinder(binder).bindConfig(FooConfig.class))
+                .setRequiredConfigurationProperties(ImmutableMap.of("foo.value", "${ENV:TEST_KEY}"));
+
+        Injector injector = bootstrap.initialize();
+
+        assertThat(injector.getInstance(FooConfig.class).getValue()).isEqualTo("test_value");
+    }
+
+    @Test
+    void testBootstrapWithEnvironmentSecretsProviderDisabled()
+            throws Exception
+    {
+        Path configurationPluginDirectory = Files.createTempDirectory(null);
+
+        File configurationResolverFile = createConfigurationResolverFile("secrets-plugins-dir=\"%s\"".formatted(configurationPluginDirectory));
+
+        System.setProperty("secretsConfig", configurationResolverFile.getAbsolutePath());
+
+        Bootstrap bootstrap = new Bootstrap(binder -> configBinder(binder).bindConfig(FooConfig.class))
+                .loadSecretsPlugins()
+                .setRequiredConfigurationProperties(ImmutableMap.of("foo.value", "${ENV:TEST_KEY}"));
+
+        assertThatThrownBy(() -> bootstrap.initialize())
+                .hasMessageContaining("No secret provider for key 'env'");
+    }
+
+    @Test
+    void testBootstrapWithEnvironmentSecretsProviderEnabled()
+            throws Exception
+    {
+        Path configurationPluginDirectory = Files.createTempDirectory(null);
+
+        File configurationResolverFile = createConfigurationResolverFile("""
+                secrets-plugins-dir="%s
+                
+                [env]
+                secrets-provider.name="env"
+                """.formatted(configurationPluginDirectory));
+
+        System.setProperty("secretsConfig", configurationResolverFile.getAbsolutePath());
+
+        Bootstrap bootstrap = new Bootstrap(binder -> configBinder(binder).bindConfig(FooConfig.class))
+                .loadSecretsPlugins()
+                .setRequiredConfigurationProperties(ImmutableMap.of("foo.value", "${ENV:TEST_KEY}"));
+
+        Injector injector = bootstrap.initialize();
+
+        assertThat(injector.getInstance(FooConfig.class).getValue()).isEqualTo("test_value");
+    }
+
+    @Test
+    void testBootstrapWithEnvironmentSecretsProviderWithDifferentNamespace()
+            throws Exception
+    {
+        Path configurationPluginDirectory = Files.createTempDirectory(null);
+
+        File configurationResolverFile = createConfigurationResolverFile("""
+                secrets-plugins-dir="%s
+                
+                [multi]
+                secrets-provider.name="env"
+                """.formatted(configurationPluginDirectory));
+
+        System.setProperty("secretsConfig", configurationResolverFile.getAbsolutePath());
+
+        Bootstrap bootstrap = new Bootstrap(binder -> configBinder(binder).bindConfig(FooConfig.class))
+                .loadSecretsPlugins()
+                .setRequiredConfigurationProperties(ImmutableMap.of("foo.value", "${MULTI:TEST_KEY}"));
+
+        Injector injector = bootstrap.initialize();
+
+        assertThat(injector.getInstance(FooConfig.class).getValue()).isEqualTo("test_value");
+    }
+
+    private File createConfigurationResolverFile(String configurationFile)
+            throws Exception
+    {
+        File tomlFile = File.createTempFile("config_resolver", ".toml");
+        tomlFile.deleteOnExit();
+
+        try (FileWriter writer = new FileWriter(tomlFile)) {
+            writer.append(configurationFile);
+        }
+
+        return tomlFile;
+    }
+
+    public static class FooConfig
+    {
+        private String value;
+
+        public String getValue()
+        {
+            return value;
+        }
+
+        @Config("foo.value")
+        public FooConfig setValue(String value)
+        {
+            this.value = value;
+            return this;
+        }
+    }
+}
--- a/configuration/src/main/java/io/airlift/configuration/TomlConfiguration.java
+++ b/configuration/src/main/java/io/airlift/configuration/TomlConfiguration.java
@@ -1,0 +1,87 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.configuration;
+
+import com.google.common.collect.Iterators;
+import org.tomlj.Toml;
+import org.tomlj.TomlArray;
+import org.tomlj.TomlTable;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.joining;
+
+public class TomlConfiguration
+{
+    public static TomlConfiguration createTomlConfiguration(File file)
+    {
+        checkArgument(file.getName().endsWith(".toml"), "TOML files must end with '.toml'");
+        try {
+            return new TomlConfiguration(Toml.parse(file.toPath()));
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private final TomlTable tomlTable;
+
+    public TomlConfiguration(TomlTable tomlTable)
+    {
+        this.tomlTable = requireNonNull(tomlTable, "tomlTable is null");
+    }
+
+    public Map<String, String> getParentConfiguration()
+    {
+        return this.tomlTable.entryPathSet().stream()
+                .filter(entry -> entry.getKey().size() == 1)
+                .collect(toImmutableMap(
+                        entry -> Iterators.getOnlyElement(entry.getKey().iterator()),
+                        entry -> tomlValueToString(entry.getValue())));
+    }
+
+    public Set<String> getNamespaces()
+    {
+        return this.tomlTable.entrySet().stream()
+                .filter(entry -> entry.getValue() instanceof TomlTable)
+                .map(Entry::getKey)
+                .collect(toImmutableSet());
+    }
+
+    public Map<String, String> getNamespaceConfiguration(String namespace)
+    {
+        checkArgument(tomlTable.contains(namespace), "Namespace %s not found", namespace);
+        return this.tomlTable.getTable(namespace).dottedEntrySet().stream()
+                .collect(toImmutableMap(Entry::getKey, entry -> tomlValueToString(entry.getValue())));
+    }
+
+    private static String tomlValueToString(Object object)
+    {
+        if (object instanceof TomlArray tomlArray) {
+            return tomlArray.toList().stream()
+                    .map(TomlConfiguration::tomlValueToString)
+                    .collect(joining(","));
+        }
+        return object.toString();
+    }
+}
--- a/configuration/src/main/java/io/airlift/configuration/secrets/SecretsPluginManager.java
+++ b/configuration/src/main/java/io/airlift/configuration/secrets/SecretsPluginManager.java
@@ -1,0 +1,204 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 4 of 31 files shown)


Praveen2112:
cc: @electrum

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -178,7 +210,7 @@ public Set<String> configure()
 
         // replace environment variables in property values
         List<Message> errors = new ArrayList<>();
-        properties = replaceEnvironmentVariables(properties, System.getenv(), (key, error) -> {
+        configurationResolver.getResolvedConfiguration(properties, (key, error) -> {
```
Praveen2112, line 213: @electrum Do we need to migrate the environment config resolution to this framework ?

wendigo:
@Praveen2112 can you fix failures?

Praveen2112:
@wendigo Now the CI is green

Praveen2112:
The usage of this PR is shown here - https://github.com/trinodb/trino/pull/22633

wendigo:
I'm not super convinced that airlift should have SPI. I think that we could add a way to plug-in an interface implementation into bootstrap like this:

```
Bootstrap app = new Bootstrap(...);

Injector injector = app
                .doNotInitializeLogging()
                .setConfigurationInterpolator(interpolator)
                .setRequiredConfigurationProperties(config)
                .initialize();
```

Where interpolator is:

```
interface ConfigurationInterpolator
{
        Optional<String> interpolate(String namespace, String key);
}
```

Ideally I'd like to provide a pre-configured interpolator to Trino connectors through the `ConnectorContext` that is initialized and injected through the engine.

Praveen2112:
> Ideally I'd like to provide a pre-configured interpolator to Trino connectors through the ConnectorContext that is initialized and injected through the engine.

But how would we make them pluggable or have some custom implementation in the future - Won't we need classloader isolation for its implementation ? Would it be a part of Trino's plugin ? The problem is the scope of this pluggable configuration provider is not specific to Trino connectors and its ecosystem (like password-authenticator or resource-group.manager) but would be applied for `config.properties` for the TrinoServer as well. If we have it as a part of trino spi - then we might need to load the plugins twice - Once during the startup for configuration resolution and then once for normal execution -

wendigo:
Yes. And we would add an SPI in Trino

Praveen2112:
Yes and we have a version which adds a SPI in Trino as well - https://github.com/trinodb/trino/pull/22621 - But there were a few questions on how to pass the properties for the same - We had it propagated via `config.properties` - But we might need a separate file for configuring this `ConfigurationResolver` - how to configure them during the startup.

Praveen2112:
I'm fine with both the options - @electrum Can you guide us on which direction to take - We have a skeleton PR which tracks both the approaches - Having the changes in Airlift and having the changes in Trino - Both has its own set of complexity.

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
electrum: That seems like a good idea. It will probably simplify the code and is a good test for the framework.

electrum:
The reason to do it in Airlift via an SPI is to allow pluggable providers with classloader isolation that can handle interpolation for the main server configuration. For example, we'll have a plugin for AWS Secrets Manager that uses the AWS SDK. Using a plugin loader similar to Trino will prevent leaking the plugin's version of the AWS SDK into the main application's classpath.

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -147,6 +159,26 @@ public Set<String> configure()
             logging = Logging.initialize();
         }
 
+        if (loadConfigResolver) {
+            log.info("Loading configuration resolver");
+            Map<String, String> configurationResolverProperties = Collections.emptyMap();
+            String configResolverFile = System.getProperty("configResolver");
+            if (configResolverFile != null) {
```
electrum, line 166: `configurationResolver` will be null if the config property is not set, which seems wrong

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -258,6 +290,8 @@ public Injector initialize()
         moduleList.add(Binder::requireExplicitBindings);
         moduleList.add(Binder::requireExactBindingAnnotations);
 
+        moduleList.add(binder -> binder.bind(ConfigurationResolver.class)
+                .toInstance(requireNonNull(configurationResolver, "configurationResolver is null")));
```
electrum, line 294: See above. We shouldn't check for null here.

config-spi/src/main/java/io/airlift/spi/ThreadContextClassLoader.java
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.spi;
+
+import java.io.Closeable;
+
+public class ThreadContextClassLoader
```
electrum, line 18: Make this a package-private class in the `configuration` module

configuration/src/main/java/io/airlift/configuration/EnvironmentVariableConfigurationValueResolver.java
```diff
@@ -0,0 +1,34 @@
...
+import io.airlift.spi.config.InvalidConfigurationResolverException;
+
+import static java.lang.String.format;
+
+public class EnvironmentVariableConfigurationValueResolver
+        implements ConfigurationValueResolver
+{
+    @Override
+    public String resolveConfigurationValue(String key)
+            throws InvalidConfigurationResolverException
```
electrum, line 26: This exception isn't thrown here

config-spi/src/main/java/io/airlift/spi/config/ConfigurationValueResolver.java
```diff
@@ -0,0 +1,20 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.spi.config;
+
+public interface ConfigurationValueResolver
+{
+    String resolveConfigurationValue(String key)
+            throws InvalidConfigurationResolverException;
```
electrum, line 19: No need to declare unchecked exceptions. Add a Javadoc instead.

config-spi/src/main/java/io/airlift/spi/config/InvalidConfigurationResolverException.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.spi.config;
+
+public class InvalidConfigurationResolverException
```
electrum, line 16: How about `ConfigurationResolverException`? This sounds like the resolver itself is invalid, but this is simplify a failure to resolve configuration.

wendigo:
@electrum yes, but we can add this SPI in Trino rather than in the airlift

configuration/src/main/java/io/airlift/configuration/resolver/ConfigurationResolverManager.java
```diff
@@ -0,0 +1,200 @@
...
+    public void load()
+    {
+        listFiles(installedConfigurationResolver).stream()
+                .filter(File::isDirectory)
+                .forEach(file -> loadConfigurationResolvers(file.getAbsolutePath(), () -> createClassLoader(file.getName(), buildClassPath(file))));
+        ImmutableMap.Builder<String, ConfigurationValueResolver> builder = ImmutableMap.builder();
+        for (File configFile : configFiles) {
+            builder.put(getNameWithoutExtension(configFile.getName()), loadConfigProvider(configFile.getAbsoluteFile()));
+        }
+        builder.put("env", new EnvironmentVariableConfigurationValueResolver());
```
electrum, line 89: We should consider making `env` be a configurable plugin, just like any other plugin. Some people may want to disable it for security purposes. Or if you're using a different provider, you might not want or need environment variables.

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
electrum: We'll want to allow configuring multiple providers, as people may have secrets in multiple systems. While this is rare in the common case, it is likely to be useful or necessary during a migration from one system to another.

What do you think about making this file be TOML similar to #1189? Then we could support multiple providers in a single file.

electrum: Another issue... do we need to support environment variable interpolation in the `config_resolver` file? I have a feeling that this will be needed, unfortunately.

Praveen2112:
@wendigo We have tried it a similar approach (https://github.com/trinodb/trino/pull/22621) but in order to have in Trino we have two main set of issues - 
1. Where would we store the configuration property for this `config_resolver` - It could be a part of `config.properties` - but we need to remove the properties which are used for `config_resolver` - We might need modify the `launcher.py` to add any new additional property file. 
2. Copying the `config-plugin` to a different location during the startup 

In this case we might need some sort of changes in airlift and some in the Trino side. Plus this configuration resolver doesn't fully fit under Plugin lifecycle in case of Trino - These are mostly static during the startup

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
Praveen2112: > We'll want to allow configuring multiple providers, as people may have secrets in multiple systems. While this is rare in the common case, it is likely to be useful or necessary during a migration from one system to another.

We currently handle multiple systems - In the configResolver file we track the list of config providers we need to use. But yes using TOML would make it easy to configure.

Praveen2112: > configurationResolver will be null if the config property is not set, which seems wrong. 

We would be replacing it with a default - implementation which is based on environment variables -

Praveen2112: `configuration-resolver.config-files` allow us to inject multiple files into the system and each file could point to a specific config-provider - so we could fetch credentials from multiple system and multiple instances of same system.

Praveen2112: > What do you think about making this file be TOML similar to https://github.com/airlift/airlift/pull/1189? Then we could support multiple providers in a single file.
We could try - But I saw @martint had a few comments regarding its conversion to `Map<String, String>` - How are we planning to represent them ?

configuration/src/main/java/io/airlift/configuration/resolver/ConfigurationResolverManager.java
Praveen2112: But won't it contradict with this comment - https://github.com/airlift/airlift/pull/1217#discussion_r1672323394. We could migrate them to `configuration-resolvers` plugin which captures a set of predefined `ConfigPlugin` -

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
Praveen2112: I missed an initialization there -

Praveen2112:
@electrum Thanks a lot for the review. AC

configuration/src/main/java/io/airlift/configuration/resolver/ConfigurationResolverManager.java
Praveen2112: But would it be backward compatible ? Like in order to use the `env` we need to add an entry to the `config-resolver.properties` - Won't it break for the existing Trino deployment

Praveen2112:
@electrum Based on the previous discussion we have migrated the configuration to TOML file and have made the `EnvironmentVariableConfigurationResolver` pluggable (via Config - so it would be used of configured)

wendigo:
@Praveen2112 how the TOML is plugged in into existing Configuration binding? Doesn't seem to be a case

Praveen2112:
The toml file path is passed via `configResolver` system property like how we handle for `config.properties`

electrum:
Typo "fot" in commit title

Praveen2112:
@electrum Based on our previous conversation - Have migrated the code from Configuration resolver to Secret provider

Praveen2112:
@electrum Gentle ping for this PR

Praveen2112:
@electrum Gentle ping

configuration/src/main/java/io/airlift/configuration/secrets/SecretsResolver.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private Optional<String> resolveConfiguration(String configurationValue)
+    {
+        StringBuilder replacedPropertyValue = new StringBuilder();
+        Matcher matcher = PATTERN.matcher(configurationValue);
+        while (matcher.find()) {
+            String configProviderName = matcher.group(1).toLowerCase(ENGLISH);
+            SecretProvider secretProvider = secretProviders.get(configProviderName);
+            if (secretProvider != null) {
```
electrum, line 67: This should fail if the secret provider does not exist

configuration/src/main/java/io/airlift/configuration/secrets/SecretsResolver.java
```diff
@@ -0,0 +1,75 @@
...
+                String keyName = matcher.group("key");
```
electrum, line 68: It's strange to mix named and numbered groups. Use names or numbers for both.

configuration/src/main/java/io/airlift/configuration/secrets/env/EnvironmentVariableSecretProvider.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class EnvironmentVariableSecretProvider
+        implements SecretProvider
+{
+    @Override
+    public String generateSecretValue(String key)
+    {
+        String envValue = System.getenv().get(key);
+        if (envValue == null) {
+            throw new SecretProviderException(format("References unset environment variable '%s'", key));
```
electrum, line 29:
```java
throw new RuntimeException("Environment variable is not set: " + key);

File: configuration/src/main/java/io/airlift/configuration/secrets/SecretsResolver.java [hunk 12]
```diff
@@ -0,0 +1,75 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static java.util.Locale.ENGLISH;
+import static java.util.Objects.requireNonNull;
+import static java.util.regex.Matcher.quoteReplacement;
+
+public class SecretsResolver
+{
+    private static final Pattern PATTERN = Pattern.compile("\\$\\{([a-zA-Z][a-zA-Z0-9_-]*):(?<key>[^}]+?)}");
```
electrum, line 32: I'm thinking we should enforce the same naming pattern for keys. We can relax the restriction later if needed, but allowing spaces, control characters or other arbitrary characters seems like a bad idea.

configuration/src/test/java/io/airlift/configuration/secrets/TestSecretsResolver.java
```diff
@@ -0,0 +1,127 @@
...
+                        "key3", "{prefix:key}"));
+    }
+
+    @Test
+    public void testSecretsResolutionWithUnknownResolver()
+    {
+        SecretsResolver secretsResolver = new SecretsResolver(
+                ImmutableMap.of("prefix", new PrefixedSecretProvider()));
+
+        assertThat(secretsResolver.getResolvedConfiguration(ImmutableMap.of("key", "${unknown_key:key}")))
```
electrum, line 81: This should fail

secrets-plugin/pom.xml
```diff
@@ -0,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.airlift</groupId>
+        <artifactId>airlift</artifactId>
+        <version>251-SNAPSHOT</version>
```
electrum, line 8: This needs to be updated to the latest

secrets-plugin/src/main/java/io/airlift/secrets/keystore/KeystoreSecretProvider.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public String generateSecretValue(String key)
+    {
+        try {
+            return readEntity(keyStore, key, keystorePassword);
+        }
+        catch (GeneralSecurityException e) {
+            throw new SecretProviderException(e.getMessage(), e);
```
electrum, line 48: No need to pass both the message and the cause. Passing the cause will fill in the message.

secrets-plugin/src/main/java/io/airlift/secrets/keystore/KeystoreSecretProviderFactory.java
```diff
@@ -0,0 +1,57 @@
...
+    public String getName()
+    {
+        return "keystore";
+    }
+
+    @Override
+    public SecretProvider createSecretProvider(Map<String, String> config)
+    {
+        Bootstrap app = new Bootstrap(
+                new Module() {
```
electrum, line 41: I think this can be a lambda

secrets-plugin/src/main/java/io/airlift/secrets/keystore/KeystoreUtils.java
```diff
@@ -0,0 +1,54 @@
...
+        return keyStore;
+    }
+
+    public static String readEntity(KeyStore keyStore, String entityAlias, String password)
+            throws GeneralSecurityException
+    {
+        SecretKeyEntry secretKeyEntry = (SecretKeyEntry) keyStore.getEntry(entityAlias, new KeyStore.PasswordProtection(password.toCharArray()));
+
+        if (secretKeyEntry == null) {
+            throw new SecretProviderException("Key not found '%s'".formatted(entityAlias));
```
electrum, line 45:
```java
throw new RuntimeException("Key not found in keystore: " + entityAlias);

File: secrets-plugin/src/test/java/io/airlift/secrets/keystore/TestKeystoreSecretProvider.java [hunk 18]
```diff
@@ -0,0 +1,93 @@
...
+
+import static java.io.File.createTempFile;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
+
+@TestInstance(PER_CLASS)
+final class TestKeystoreSecretProvider
+{
+    private static final String KEYSTORE_PASSWORD = "password";
```
electrum, line 35: I would inline this

electrum (approved):

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -178,13 +204,10 @@ public Set<String> configure()
 
         // replace environment variables in property values
         List<Message> errors = new ArrayList<>();
-        properties = replaceEnvironmentVariables(properties, System.getenv(), (key, error) -> {
+        properties = ImmutableSortedMap.copyOf(secretsResolver.getResolvedConfiguration(properties, (key, error) -> {
             unusedProperties.remove(key);
             errors.add(new Message(error));
```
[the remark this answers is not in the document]
Praveen2112: I have migrated it to `errors.add(new Message(error.getMessage(), error);`

Praveen2112:
@electrum Thanks a lot for the review - I have addressed most of the comments except 1 - What would be our fallback mechanism if the key for a config provider doesn't exist -
1. We could fail explicitly
2. Skip resolving them.
We could go for option 2 as a config of a similar pattern could be used by other system

bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrapWithSecretsProvider.java
```diff
@@ -0,0 +1,140 @@
...
+
+        System.setProperty("secretsConfig", configurationResolverFile.getAbsolutePath());
+
+        Bootstrap bootstrap = new Bootstrap(binder -> configBinder(binder).bindConfig(TestConfig.class))
+                .loadSecretsPlugin()
+                .setRequiredConfigurationProperties(ImmutableMap.of("test.value", "${ENV:TEST_KEY}"));
+
+        Injector injector = bootstrap.initialize();
+
+        assertThat(injector.getInstance(TestConfig.class).getValue()).isEqualTo("${ENV:TEST_KEY}");
```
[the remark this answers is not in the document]
Praveen2112: Or should we escape them ?

Praveen2112:
@electrum Addressed all the comments - Now if the secret provider is not available we would throw an error. We have captured and throw the error via `Consumer`

secrets-spi/src/main/java/io/airlift/spi/secrets/SecretProvider.java
```diff
@@ -0,0 +1,25 @@
...
+
+public interface SecretProvider
+{
+    /**
+     * Resolves a ConfigurationValue based on the given key
+     *
+     * @return resolved value
+     * @throws SecretProviderException if key cannot be resolved.
+     */
+    String generateSecretValue(String key);
```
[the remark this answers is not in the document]
electrum: That sounds good to me

configuration/src/main/java/io/airlift/configuration/TomlConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Set;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Verify.verify;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.joining;
+
+public class TomlConfiguration
```
electrum, line 35: Make this package private for now

configuration/src/main/java/io/airlift/configuration/TomlConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+{
+    public static TomlConfiguration createTomlConfiguration(String path)
+    {
+        verify(path.endsWith(".toml"), "TOML files must end with '.toml'");
```
electrum, line 39: This should use `checkArgument()` since it is validating an argument

configuration/src/main/java/io/airlift/configuration/TomlConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Verify.verify;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.joining;
+
+public class TomlConfiguration
+{
+    public static TomlConfiguration createTomlConfiguration(String path)
```
electrum, line 37: Make this take `File` directly

bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrapWithSecretsProvider.java
```diff
@@ -0,0 +1,140 @@
...
+
+        Bootstrap bootstrap = new Bootstrap(binder -> configBinder(binder).bindConfig(FooConfig.class))
+                .loadSecretsPlugins()
+                .setRequiredConfigurationProperties(ImmutableMap.of("foo.value", "${MULTI:TEST_KEY}"));
+
+        Injector injector = bootstrap.initialize();
+
+        assertThat(injector.getInstance(FooConfig.class).getValue()).isEqualTo("test_value");
+    }
+
```
electrum, line 108: Remove extra blank line

bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrapWithSecretsProvider.java
```diff
@@ -0,0 +1,140 @@
...
+        File tomlFile = File.createTempFile("config_resolver", ".toml");
+        tomlFile.deleteOnExit();
+
+        try (FileWriter writer = new FileWriter(tomlFile)) {
+            writer.append(configurationFile);
+        }
+
+        return tomlFile;
+    }
+
```
electrum, line 122: Remove extra blank line

configuration/src/main/java/io/airlift/configuration/TomlConfiguration.java
Praveen2112: I'm not sure if we could use it as package private as we use them in `io.airlift.configuration.secrets.SecretsPluginManager`

Praveen2112:
@electrum Except this [comment](https://github.com/airlift/airlift/pull/1217#discussion_r1699300191) all others were addressed

secrets-plugin/src/main/java/io/airlift/secrets/keystore/KeystoreUtils.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.crypto.SecretKeyFactory;
+import javax.crypto.spec.PBEKeySpec;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.security.KeyStore;
+import java.security.KeyStore.SecretKeyEntry;
+
+public final class KeystoreUtils
```
dain, line 25: I don't think we need this utility class.  Let's just move the code to KeystoreSecretProvider

secrets-plugin/src/test/java/io/airlift/secrets/keystore/TestKeystoreSecretProvider.java
```diff
@@ -0,0 +1,91 @@
...
+{
+    private File keystoreFile;
+
+    private KeystoreSecretProvider secretProvider;
+
+    @BeforeAll
+    public void setup()
+            throws Exception
+    {
+        keystoreFile = createTempFile(System.getProperty("java.io.tmpdir"), "keystore");
```
dain, line 43: Does airlift have a testing temp directory utility like Trino has?

pom.xml
```diff
@@ -60,6 +60,7 @@
         <module>openmetrics</module>
         <module>packaging</module>
         <module>sample-server</module>
+        <module>secrets-plugin</module>
```
dain, line 63: So is the idea that we would put all plugins in one module?  We can always change this later, but I would imagine that for additional plugins we would want separate modules, and if so maybe this should be the secrets-keystore-plugin.

dain (commented):
some minor comments but otherwise looks good

configuration/src/main/java/io/airlift/configuration/secrets/SecretsPluginManager.java
```diff
@@ -0,0 +1,205 @@
...
+    @VisibleForTesting
+    void installSecretsPlugin(SecretsPlugin secretsPlugin)
+    {
+        secretsPlugin.getSecretProviderFactories()
+                .forEach(this::addSecretProviderFactory);
+    }
+
+    private void addSecretProviderFactory(SecretProviderFactory secretProviderFactory)
+    {
+        verify(CONFIG_PROVIDER_NAME_PATTERN.matcher(secretProviderFactory.getName()).matches(), "Secret provider name '%s' doesn't match pattern '%s'", secretProviderFactory.getName(), CONFIG_PROVIDER_NAME_PATTERN);
```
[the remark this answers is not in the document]
Praveen2112: But it would break with the `${ENV:...}` which we apply today right ?

pom.xml
Praveen2112: No - But for airlift we could try to have all the secrets-plugin in one place -

secrets-plugin/src/test/java/io/airlift/secrets/keystore/TestKeystoreSecretProvider.java
Praveen2112: I don't think we have any - Even in most on trino we use `deleteOnExit`

Praveen2112:
@dain Thanks a lot for the review - Except this comment https://github.com/airlift/airlift/pull/1217#discussion_r1699364480 most of them are addressed.

secrets-plugin/src/test/java/io/airlift/secrets/keystore/TestKeystoreSecretProvider.java
Praveen2112: Or should we use a different API here ?

configuration/src/main/java/io/airlift/configuration/secrets/SecretsPluginManager.java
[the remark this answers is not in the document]
dain: Are you sure it would break that.  The matcher is only used to validate the name of the factory. All names are lowered, so I assume this is all case insensitive.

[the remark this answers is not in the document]
Praveen2112: Sorry I was confused with the patter which matches the `${ENV:...}` - You are right - we could modify it to work only with lower case

Praveen2112:
@dain All the comments were addressed

secrets-plugin/src/test/java/io/airlift/secrets/keystore/TestKeystoreSecretProvider.java
Praveen2112: Thank you @electrum !! Migrated to `TestFile`

Praveen2112:
@dain / @electrum All the comments were addressed.

PR #1218, opened 2024-07-09
Reduce cpu and memory alloc around configuration

oskar-szwajkowski:
Use short-circuit when there is no key in methodMap
Pull up metadataCache to be static field in ConfigurationFactory

I could see this static cache being behind some ENV variable, but its static memory footprint should be below 1MB per class-loader (tested with galaxy-trino object store connector)


configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -137,7 +138,7 @@ public ConfigurationFactory(Map<String, String> properties, WarningsMonitor warn
     {
         this.properties = ImmutableMap.copyOf(properties);
         this.warningsMonitor = warningsMonitor;
-        this.monitor = monitor;
+        ConfigurationFactory.monitor = monitor;
```
electrum, line 141: We shouldn't set a static field from a constructor. It's confusing and not thread safe.

Instead, we should refactor the tests so that `Problems.Monitor` is not needed. It appears that the tests could access `Problems` directly, but you'll need to try refactoring it and see what it looks like.

electrum (changes requested):

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
oskar-szwajkowski: Changed it so that it's not static anymore.
I did try to remove monitor but it require more code changes that I wanted to fit in this PR

wendigo (approved):

oskar-szwajkowski:
@electrum Can you review it?

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -84,6 +84,15 @@ public class ConfigurationFactory
     private static final Validator VALIDATOR;
 
     private static final Splitter VALUE_SPLITTER = Splitter.on(",").omitEmptyStrings().trimResults();
+    private static final LoadingCache<Class<?>, ConfigurationMetadata<?>> metadataCache = CacheBuilder.newBuilder()
```
electrum, line 87: Name this `METADATA_CACHE` as it's a constant

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -84,6 +84,15 @@ public class ConfigurationFactory
...
+            .build(new CacheLoader<>()
```
electrum, line 88: This can simplify to
```java
.build(CacheLoader.from(ConfigurationMetadata::getConfigurationMetadata));

File: configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java [hunk 4]
```diff
@@ -286,6 +287,13 @@ <T> void registerConfigDefaults(ConfigDefaultsHolder<T> holder)
         registeredDefaultConfigs.put(holder.getConfigKey(), holder);
     }
 
+    @VisibleForTesting
+    static void cleanMetadataCache()
```
electrum, line 291: Do we need this if we get rid of `Problems.Monitor`?

wendigo:
@oskar-szwajkowski can you fix failures?

configuration/src/test/java/io/airlift/configuration/TestConfigurationFactory.java
```diff
@@ -174,12 +168,11 @@ public void testConfigurationThroughDeprecatedConfig()
...
         properties.put("string-b", "this is b");
-        TestMonitor monitor = new TestMonitor();
-        Injector injector = createInjector(properties, monitor, binder -> configBinder(binder).bindConfig(DeprecatedConfigPresent.class));
+        TestWarningsMonitor warningsMonitor = new TestWarningsMonitor();
+        Injector injector = createInjector(properties, binder -> configBinder(binder).bindConfig(DeprecatedConfigPresent.class), warningsMonitor);
         DeprecatedConfigPresent deprecatedConfigPresent = injector.getInstance(DeprecatedConfigPresent.class);
-        monitor.assertNumberOfErrors(0);
-        monitor.assertNumberOfWarnings(1);
-        monitor.assertMatchingWarningRecorded("Configuration property 'string-a' is deprecated and should not be used");
+        assertThat(warningsMonitor.messages).hasSize(1);
```
electrum, line 174: Add a `messages()` getter

configuration/src/main/java/io/airlift/configuration/ConfigurationMetadata.java
```diff
@@ -72,13 +60,13 @@ static <T> ConfigurationMetadata<T> getConfigurationMetadata(Class<T> configClas
...
 
-    private ConfigurationMetadata(Class<T> configClass, Monitor monitor)
+    private ConfigurationMetadata(Class<T> configClass)
     {
         if (configClass == null) {
             throw new NullPointerException("configClass is null");
         }
 
-        this.problems = new Problems(monitor);
+        this.problems = new Problems();
```
electrum, line 69: Move this initialization to the field declaration

electrum (approved):

oskar-szwajkowski:
> @oskar-szwajkowski can you fix failures?

Done, can you merge & release?

configuration/src/test/java/io/airlift/configuration/TestMonitor.java
```diff
@@ -18,17 +18,18 @@
...
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.fail;
 
 class TestMonitor
         implements Problems.Monitor
 {
-    private final List<Message> errors = new ArrayList<>();
```
wendigo, line 30: Why? Seems unrelated. Changes order of error messages.

oskar-szwajkowski: this class gets deleted in later commit, but when I've added changes to Problems in commit no. 2, then it sometimes contained duplicate messages

configuration/src/test/java/io/airlift/configuration/TestConfigurationFactory.java
```diff
@@ -451,16 +436,16 @@ public void testUsedPropertiesWithFailure()
...
 
-        ConfigurationFactory configurationFactory = new ConfigurationFactory(properties, null, new TestMonitor());
+        ConfigurationFactory configurationFactory = new ConfigurationFactory(properties);
         configurationFactory.registerConfigurationClasses(ImmutableList.of(binder -> configBinder(binder).bindConfig(AnnotatedSetter.class)));
         List<Message> messages = configurationFactory.validateRegisteredConfigurationProvider();
         assertMessagesMatch(messages, ImmutableList.of(".*Invalid value 'invalid' for type boolean \\(property 'boolean-value'\\).*AnnotatedSetter.*"));
         assertThat(configurationFactory.getUsedProperties()).isEqualTo(properties.keySet());
     }
 
-    private static Injector createInjector(Map<String, String> properties, TestMonitor monitor, Module module)
```
wendigo, line 461: do we need to switch order here? Makes diff bigger than it could be

wendigo (approved):

PR #1219, opened 2024-07-09
Introduce streaming execution for streaming responses

Randgalt:
Introduce `executeStreaming` which allows callers
to stream response content in whatever manner they want. The HttpClient does not close response streams or connection. It is the responsibility of the caller to close the `StreamingResponse` so that the request is completed.

`StreamingResponse` is written in a way that maps well to Jakarta's `StreamingOutput` class. Combined with an `AsyncResponse` async streaming responses, piped request-to-responses, etc. are possible.

`StreamingResponse` implements `Closeable` to ensure/indicate that the caller is responsible for finalizing/closing the response.


http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -672,7 +674,19 @@ public <T, E extends Exception> T execute(Request request, ResponseHandler<T, E>
...
+    @Override
+    public StreamingResponse executeStreaming(Request request)
+    {
+        request = applyRequestFilters(request);
+
+        Span span = startSpan(request);
+        request = injectTracing(request, span);
+
+        StreamingResponseHandler responseHandler = new StreamingResponseHandler(span);
+        return doExecute(request, responseHandler, span, responseHandler);
```
wendigo, line 686: Do we need this StreamingResponseHandler? Can we do it without it?

Randgalt: I guess we could inline it in `executeStreaming`. Would you prefer that?

wendigo: Yes. We wanted to avoid using that abstraction for this purpose as it doesn't fit

Randgalt: @wendigo reworked the code - PTAL

http-client/src/main/java/io/airlift/http/client/StreamingResponse.java
```diff
@@ -0,0 +1,6 @@
+package io.airlift.http.client;
+
+public interface StreamingResponse
+        extends Response, AutoCloseable
```
electrum, line 4: Let's make this `extends Closeable` and override `close()` here to not throw any checked exception. The production implementation doesn't throw checked exceptions, and it doesn't seem useful since the caller can't do anything with the exception (it means some cleanup failed after the caller is done with the response).

Thinking about this more: under what cases would close throw?

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -602,6 +608,29 @@ public void testResponseStatusCode()
...
+        try (CloseableTestHttpServer server = newServer()) {
+            server.servlet().setResponseBody(LARGE_CONTENT);
+
+            Request request = prepareGet()
+                    .setUri(server.baseURI())
+                    .build();
+
+            executeRequest(server, request).ifPresent(streamingResponse -> {
+                try (streamingResponse) {
+                    String responseString = new String(ByteStreams.toByteArray(streamingResponse.getInputStream()), UTF_8);
```
electrum, line 624: Can use `readAllBytes()`
```java
String responseString = new String(streamingResponse.getInputStream().readAllBytes(), UTF_8);

File: http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java [hunk 4]
```diff
@@ -602,6 +608,29 @@ public void testResponseStatusCode()
...
+                    assertThat(responseString).isEqualTo(LARGE_CONTENT);
+                }
+                catch (Exception e) {
+                    throw new RuntimeException(e);
```
electrum, line 628: After the change to `StreamingResponse`:
```suggestion
                catch (IOException e) {
                    throw new UncheckedIOException(e);
```

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -877,104 +906,82 @@ public void testDowngradedHttpProtocolUsed()
...
+                protected void service(HttpServletRequest request, HttpServletResponse response)
+                        throws IOException
                 {
-                    throw propagate(request, exception);
-                }
+                    switch (request.getPathInfo()) {
+                        case "/get" -> {
+                            response.setIntHeader("Content-Length", content.length());
+                            response.setStatus(200);
+                            ByteStreams.copy(new ByteArrayInputStream(content.getBytes(UTF_8)), response.getOutputStream());
```
electrum, line 944: Can use `transferTo()`
```suggestion
                            new ByteArrayInputStream(content.getBytes(UTF_8)).transferTo(response.getOutputStream());
```

electrum: Actually, why not do
```suggestion
                            response.getOutputStream().write(content.getBytes(UTF_8));
```

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -877,104 +906,82 @@ public void testDowngradedHttpProtocolUsed()
...
-                                    }
-                                }
-                                catch (IOException e) {
-                                    throw new UncheckedIOException(e);
-                                }
-                                return response.getStatusCode();
+                            Request pipeRequest = prepareGet().setUri(locals.server.baseURI().resolve("get")).build();
+                            StreamingResponse streamingResponse = httpClient.executeStreaming(pipeRequest);
+
+                            Request putRequest = preparePut().setUri(locals.server.baseURI().resolve("put")).setBodyGenerator(streamingBodyGenerator(streamingResponse.getInputStream())).build();
```
electrum, line 955: Wrap builder calls for readability

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -877,104 +906,82 @@ public void testDowngradedHttpProtocolUsed()
...
-                        assertThat(server.servlet().getRequestMethod()).isEqualTo("PUT");
-                        assertThat(server.servlet().getRequestUri()).isEqualTo(uri);
-                        assertThat(server.servlet().getRequestHeaders(CONTENT_TYPE)).isEqualTo(ImmutableList.of("x-test"));
-                        assertThat(server.servlet().getRequestBytes()).isEqualTo(content.getBytes());
-                    }
-                    catch (Exception e) {
-                        throw new RuntimeException(e);
-                    }
+                            catch (Exception e) {
+                                throw new RuntimeException(e);
```
electrum, line 960:
```suggestion
                            catch (IOException e) {
                                throw new UncheckedIOException(e);
```

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -877,104 +906,82 @@ public void testDowngradedHttpProtocolUsed()
...
+                            }
+                            response.setStatus(204);
+                        }
+
+                        case "/put" -> {
+                            // save the PUT content into a local for validation after the test
+                            locals.putContent = new String(ByteStreams.toByteArray(request.getInputStream()), UTF_8);
```
electrum, line 967:
```suggestion
                            locals.putContent = new String(request.getInputStream().readAllBytes(), UTF_8);
```

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -877,104 +906,82 @@ public void testDowngradedHttpProtocolUsed()
...
-                    executorService.submit(() -> httpClient.execute(request, responseHandler)).get(30, SECONDS);
-                }
+            try (CloseableTestHttpServer server = new CloseableTestHttpServer(getScheme(), new TestingHttpServer(keystore, pipeServlet), HashMultiset.create(), new EchoServlet())) {
+                locals.server = server;
+
+                Request request = preparePut().setUri(locals.server.baseURI().resolve("pipe")).build();
+                executeRequest(server, request, createStatusResponseHandler());
+
+                assertThat(locals.putContent)
+                        .isNotNull()
```
electrum, line 983: The `isNotNull()` seems unnecessary

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyHttpsClientPem.java
```diff
@@ -44,14 +57,19 @@ public <T, E extends Exception> T executeRequest(CloseableTestHttpServer server,
...
+        applyCertAndPem(config);
 
         try (JettyHttpClient client = server.createClient(config)) {
             return client.execute(request, responseHandler);
         }
     }
 
+    private static void applyCertAndPem(HttpClientConfig config)
+    {
+        config.setKeyStorePath(getResource("client.pem").getPath())
```
electrum, line 69: Nit: the chaining here makes this harder to read

electrum (approved):
All of the callbacks in `JettyHttpClient` make the code hard to reason about, but I'm not sure there is a better way. We could follow up on this if there are ideas on how to improve it.

http-client/src/main/java/io/airlift/http/client/StreamingResponse.java
Randgalt: It's OK. For example, in the test the Guava `Closer` can throw. But we can wrap that in `UncheckedIOException`

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
Randgalt: The test's abstract `executeRequest()` throws Exception.

PR #1220, opened 2024-07-11
Enable JSON format in console

andythsu:


log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -101,9 +101,6 @@ private Logging()
             ROOT.removeHandler(handler);
         }
 
-        enableConsole();
```
andythsu, line 104: instead of enabling console at the constructor level, we can wait until `configure()` step to determine which kind of console to enable

mosabua:
More context in https://trinodb.slack.com/archives/C059KUNPTSP/p1719430408571069
as discussed with @dain @mattstep

mosabua:
If/when we merge this and upgrade airlift in Trino we have to update the log properties docs accordingly.

log-manager/src/main/java/io/airlift/log/Logging.java
dain: Without digging in at all, I expect this will break something in some very subtle way.  The logging code is super prickly.  Instead, I suggest we just reconfigure the console after the initial startup

dain: BTW, that will likely cause issues where the console contains non-json messages, but that **might** be unavoidable if there are things logging before we are invoked.  If we really need to avoid that, then I expect we will need a system property since configuration is available much later in the startup.

andythsu: I'm not exactly sure how deep down this consoleHandler overrides. If it overrides all the way to jvm console then we'll need to keep it here. If it only overrides by the time we instantiate the logging class then I think we are fine...

mattstep:
We should add a configuration property for log.console-format that specifies the format for the console explicitly. This is better than just inferring it since the file/tcp output is likely to go through some kind of log processing infrastructure whereas the console logs will go through whatever infrastructure is built into the compute infrastructure executing Trino (e.g. Kubernetes where you want kubectl logs to look pretty along with using fluentbit/d to forward structured logs to your observability stack).

andythsu:
> We should add a configuration property for `log.console-format` that specifies the format for the console explicitly. This is better than just inferring it since the file/tcp output is likely to go through some kind of log processing infrastructure whereas the console logs will go through whatever infrastructure is built into the compute infrastructure executing Trino (e.g. Kubernetes where you want `kubectl logs` to look pretty along with using fluentbit/d to forward structured logs to your observability stack).

sure. And I'm assuming `log.console-format` should support both TEXT and JSON right?

electrum:
> And I'm assuming log.console-format should support both TEXT and JSON right?

Yes, and should default to `TEXT`.

To restate what @mattstep said: it's a feature to allow a different format for each one. You can have `JSON` for log collection systems and `TEXT` for humans.

andythsu:
should we get rid of 
https://github.com/airlift/airlift/blob/70f07d7cbb2ecd56a434585ac1236f75a73653a1/log-manager/src/main/java/io/airlift/log/Logging.java#L104-L105
in the constructor and let `Configure` do the job? But we may lose the logging during initialization

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -276,8 +286,10 @@ public synchronized void configure(LoggingConfiguration config)
             }
         }
 
-        if (!config.isConsoleEnabled()) {
-            disableConsole();
+        boolean isLoggingJsonInConsole = config.getConsoleFormat().equals(Format.JSON);
+
+        if (config.isConsoleEnabled() && isLoggingJsonInConsole) {
+            enableJsonConsole(logAnnotations);
         }
```
dain, line 293: I think this is the wrong place for this code.  Instead in `configure`, replace:

```java
        if (!config.isConsoleEnabled()) {
            disableConsole();
        }
```

```java
        if (config.isConsoleEnabled()) {
            ROOT.removeHandler(consoleHandler);
            // todo do we want logAnnotations on the console output
            consoleHandler = new OutputStreamHandler(System.err, config.getConsoleFormat().createFormatter(logAnnotations));
            ROOT.addHandler(consoleHandler);
        }
        else {
            disableConsole();
        }
```

dain: @mattstep and @electrum Do we want the `logAnnotations` sent to the console output. 

In cases where you are doing console logging piped into a log collector, I could see you want that.  Generally, if you are only using one log style, I think we should add the annotations; otherwise, I likely would not want them on the console.

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -153,6 +154,15 @@ public synchronized void disableConsole()
...
     }
 
+    private synchronized void enableJsonConsole(Map<String, String> logAnnotations)
+    {
+        // disabling any existing console allow adding json console
+        disableConsole();
+        log.info("Enabling stderr output in JSON format");
+        consoleHandler = new OutputStreamHandler(System.err, new JsonFormatter(logAnnotations));
+        ROOT.addHandler(consoleHandler);
+    }
```
dain, line 164: With the change below, you would not need this method.

log-manager/src/main/java/io/airlift/log/Logging.java
```diff
@@ -142,6 +142,7 @@ private static void redirectStdStreams()
 
     private synchronized void enableConsole()
     {
+        log.info("Enabling stderr output");
```
dain, line 145: I would not add this log message.  It also happens before the appender is  added, so it would not go anwhere.

@electrum I think we should inline this method. It is private and only called in the constructor.

dain (changes requested):
I added some comments on where to put the console enablement.

There are also some questions for David and Matt about how we want some items to work.

log-manager/src/main/java/io/airlift/log/Logging.java
andythsu: removed it

log-manager/src/main/java/io/airlift/log/Logging.java
andythsu: For us we are doing console logging piped into a log collector in trino-gateway, so having an option to include `logAnnotations` is beneficial for us

log-manager/src/test/java/io/airlift/log/TestOutputStreamHandler.java
```diff
@@ -0,0 +1,71 @@
...
+        logRecord.setThrown(testException);
+        handler.publish(logRecord);
+
+        String consoleMessage = out.toString();
+        Map<String, Object> jsonMap = mapJsonCodec(String.class, Object.class).fromJson(consoleMessage);
+        JsonRecord jsonRecord = jsonCodec(JsonRecord.class).fromJson(consoleMessage);
+
+        assertThat(consoleMessage).as("Log lines should end with newline").endsWith("\n");
+
+        assertThat(jsonRecord.getTimestamp().truncatedTo(NANOS)).as("Ensure timestamps between the original LogRecord and Json are equal to the nano").isEqualTo(logRecord.getInstant().truncatedTo(NANOS));
```
wendigo, line 60: nit: wrap this line

assertThat(...)
.as(...)
.isEqualTo(...)

log-manager/src/test/java/io/airlift/log/TestOutputStreamHandler.java
```diff
@@ -0,0 +1,71 @@
...
+import static com.google.common.base.Throwables.getStackTraceAsString;
+import static io.airlift.json.JsonCodec.jsonCodec;
+import static io.airlift.json.JsonCodec.mapJsonCodec;
+import static io.airlift.log.Level.fromJulLevel;
+import static java.time.temporal.ChronoUnit.NANOS;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class TestOutputStreamHandler
+{
```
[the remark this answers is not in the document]
andythsu: the `testLoggingToJsonFormat` tests the `jsonRecord`. Do you want a test that shows the `jsonRecord` ?

wendigo:
```
Exception: java.lang.StackOverflowError thrown from the UncaughtExceptionHandler in thread "log-buffer-drainer"

Exception: java.lang.StackOverflowError thrown from the UncaughtExceptionHandler in thread "log-buffer-drainer"
```

a-narsudinov:
Hi guys! It's really exciting that someone has already decided to make it possible to use the JSON format in the console logger. What's the status of this PR? I hope it will be merged and used in Trino :^)

wendigo:
Superseded by https://github.com/airlift/airlift/pull/1389. I've added @andythsu as an author of this change. This is now updated in Trino and will be released along with the 471 version.

PR #1223, opened 2024-07-26
HTTP client additional settings

wendigo:


wendigo:
ping @electrum @dain

wendigo:
These changes should go in regardless of the HTTP1/2. Connection sweeper is built-in so we don't need to keep it but in order to make it work, we need a destination idle timeout which this PR is adding

electrum (approved):

PR #1224, opened 2024-07-29
Add per request limits to http client

wendigo:


wendigo:
Needed to have a single shared http client instance for all internal communication

http-client/src/main/java/io/airlift/http/client/Request.java
```diff
@@ -221,6 +246,8 @@ public static Builder fromRequest(Request request)
         private Optional<HttpVersion> version = Optional.empty();
         private boolean followRedirects = true;
         private boolean preserveAuthorizationOnRedirect;
+        private Duration requestTimeout = Duration.ZERO;
```
electrum, line 262: It seems better to make these `Optional`. We actually allow zero for both of these in `HttpClientConfig` (though I'm not sure what that means in the context of a request timeout).

http-client/src/main/java/io/airlift/http/client/Request.java
```diff
@@ -316,6 +330,18 @@ public Builder setIdleTimeout(Duration timeout)
             return this;
         }
 
+        public Builder setMaxContentLength(Optional<DataSize> maxContentLength)
+        {
+            this.maxContentLength = maxContentLength;
```
electrum, line 335: `requireNonNull`

http-client/src/main/java/io/airlift/http/client/Request.java
```diff
@@ -316,6 +330,18 @@ public Builder setIdleTimeout(Duration timeout)
             return this;
         }
 
+        public Builder setMaxContentLength(Optional<DataSize> maxContentLength)
```
electrum, line 333: Do we need this method? We don't have it for `setVersion()`. It's fine if we need it.

electrum (approved):

PR #1226, opened 2024-07-30
Add separate modes for X-Forwarded processing

wendigo:
These are:
- ACCEPT: process X-Forwarded-Header(s).
- DROP: remove X-Forwarded-Header(s) from the request. Application won't receive these headers.
- REJECT: reject X-Forwarder-Header(s) if present. This results in HTTP 406 if X-Forwarder-* header is present.

The existing configuration `http-server.process-forwarded` previously accepting boolean values maps to new modes:
- true => ACCEPT
- false => REJECT

Fixes https://github.com/trinodb/trino/issues/22865

### DIFF
--- a/http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
@@ -25,8 +25,10 @@ import io.airlift.units.MinDataSize;
 import jakarta.validation.constraints.Min;
 import jakarta.validation.constraints.NotNull;
 
+import static io.airlift.http.server.HttpServerConfig.ProcessForwardedMode.REJECT;
 import static io.airlift.units.DataSize.Unit.KILOBYTE;
 import static io.airlift.units.DataSize.Unit.MEGABYTE;
+import static java.util.Locale.ENGLISH;
 import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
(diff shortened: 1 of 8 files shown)


http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -495,4 +497,20 @@ public HttpServerConfig setHttp2StreamIdleTimeout(Duration http2StreamIdleTimeou
...
+        ACCEPT,
+        REJECT,
+        DROP;
+
+        public static ProcessForwardedMode fromString(String value)
+        {
+            return switch (value.toUpperCase(ENGLISH)) {
+                case "TRUE" -> ACCEPT;
+                case "FALSE" -> REJECT;
+                default -> valueOf(value.toUpperCase(ENGLISH));
```
wendigo, line 512: otherwise just use the enum name

http-server/src/main/java/io/airlift/http/server/DropForwardedRequestCustomizer.java
```diff
@@ -0,0 +1,46 @@
...
+import org.eclipse.jetty.http.HttpField;
+import org.eclipse.jetty.http.HttpFields;
+import org.eclipse.jetty.server.HttpConfiguration;
+import org.eclipse.jetty.server.Request;
+
+import java.util.Set;
+
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+
+public class DropForwardedRequestCustomizer
```
wendigo, line 12: This customizer removes X-Forwarded- headers from the request. I followed similar customizer to see how to mutate headers and wrap request without copying all the fields.

electrum:
Let's name this `IGNORE` as `DROP` sounds like it would drop the request

electrum (approved):

huw0:
Thanks @wendigo

PR #1235, opened 2024-08-02
Log warning when Major GC fails to recover significant memory

wweiss-starburst:
There are instances where processes spin in GC loops when Major GCs fail to recover enough memory. By providing a warning before this condition is reached, it should be possible to act on the warning by taking a JFR before the JVM completely runs out of memory.

### DIFF
--- a/stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
+++ b/stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
@@ -13,7 +13,9 @@
  */
 package io.airlift.stats;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 import io.airlift.log.Logger;
 import io.airlift.units.DataSize;
@@ -35,11 +37,14 @@ import java.lang.management.ManagementFactory;
 import java.lang.management.MemoryUsage;
 import java.util.Collection;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
+import static com.sun.management.GarbageCollectionNotificationInfo.GARBAGE_COLLECTION_NOTIFICATION;
 import static io.airlift.units.DataSize.succinctBytes;
 import static java.lang.Math.max;
+import static java.lang.management.ManagementFactory.getPlatformMBeanServer;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
@@ -55,9 +60,16 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 public class JmxGcMonitor
         implements GcMonitor
 {
+    @VisibleForTesting
+    static final double FRACTION_OF_MAX_HEAP_TO_TRIGGER_WARN = 0.8;
+    private static final double MINIMUM_PERCENTAGE_OF_HEAP_RECLAIMED = 10.0;
+    private static final String MAJOR_ZGC_NAME = "ZGC Major Cycles";
+    private static final Set<String> ZGC_MBEANS = ImmutableSet.of("ZGC Cycles", MAJOR_ZGC_NAME, "ZGC Minor Cycles");
+
     private final Logger log = Logger.get(JmxGcMonitor.class);
 
     private final NotificationListener notificationListener = (notification, ignored) -> onNotification(notification);
+    private final NotificationListener zgcNotificationListener = (notification, ignored) -> onZgcNotification(notification);
 
     private final AtomicLong majorGcCount = new AtomicLong();
     private final AtomicLong majorGcTime = new AtomicLong();
@@ -74,11 +86,21 @@ public class JmxGcMonitor
         for (GarbageCollectorMXBean mbean : ManagementFactory.getGarbageCollectorMXBeans()) {
             ObjectName objectName = mbean.getObjectName();
             try {
-                ManagementFactory.getPlatformMBeanServer().addNotificationListener(
-                        objectName,
-                        notificationListener,
-                        null,
-                        null);
+                // Don't register listeners for ZGC Pauses because it does not report memory usage statistics
+                if (ZGC_MBEANS.contains(mbean.getName())) {
+                    getPlatformMBeanServer().addNotificationListener(
+                            objectName,
+                            zgcNotificationListener,
+                            null,
+                            null);
+                }
+                else {
+                    getPlatformMBeanServer().addNotificationListener(
+                            objectName,
+                            notificationListener,
+                            null,
+                            null);
+                }
             }
             catch (JMException e) {
                 throw new RuntimeException("Unable to add GC listener", e);
@@ -92,7 +114,14 @@ public class JmxGcMonitor
         for (GarbageCollectorMXBean mbean : ManagementFactory.getGarbageCollectorMXBeans()) {
             ObjectName objectName = mbean.getObjectName();
             try {
-                ManagementFactory.getPlatformMBeanServer().removeNotificationListener(objectName, notificationListener);
+                if (mbean.getName().startsWith("ZGC")) {
+                    if (!mbean.getName().equals("ZGC Pauses")) {
+                        getPlatformMBeanServer().removeNotificationListener(objectName, zgcNotificationListener);
+                    }
+                }
+                else {
+                    getPlatformMBeanServer().removeNotificationListener(objectName, notificationListener);
+                }
             }
             catch (JMException ignored) {
                 log.warn("Failed to unregister GC listener");
@@ -128,24 +157,11 @@ public class JmxGcMonitor
 
     private synchronized void onNotification(Notification notification)
     {
-        if ("com.sun.management.gc.notification".equals(notification.getType())) {
+        if (GARBAGE_COLLECTION_NOTIFICATION.equals(notification.getType())) {
             GarbageCollectionNotificationInfo info = new GarbageCollectionNotificationInfo((CompositeData) notification.getUserData());
 
             if (info.isMajorGc()) {
-                majorGcCount.incrementAndGet();
-                majorGcTime.addAndGet(info.getDurationMs());
-                majorGc.add(info.getDurationMs(), MILLISECONDS);
-
-                // assume that major GCs stop the application
-                long applicationRuntime = max(0, info.getStartTime() - lastGcEndTime);
-                lastGcEndTime = info.getEndTime();
-
-                log.info(
-                        "Major GC: application %sms, stopped %sms: %s -> %s",
-                        applicationRuntime,
-                        info.getDurationMs(),
-                        info.getBeforeGcTotal(),
-                        info.getAfterGcTotal());
+                logMajorGc(info);
             }
             else if (info.isMinorGc()) {
                 minorGc.add(info.getDurationMs(), MILLISECONDS);
@@ -158,6 +174,68 @@ public class JmxGcMonitor
                         info.getAfterGcTotal());
             }
         }
+    }
+
+    private synchronized void onZgcNotification(Notification notification)
+    {
+        if (GARBAGE_COLLECTION_NOTIFICATION.equals(notification.getType())) {
+            GarbageCollectionNotificationInfo info = new GarbageCollectionNotificationInfo((CompositeData) notification.getUserData());
+            if (info.isMajorZgc()) {
+                logMajorGc(info);
+            }
+            else {
+                minorGc.add(info.getDurationMs(), MILLISECONDS);
+
+                // assumption that minor GCs run currently, so we do not print stopped or application time
+                log.debug(
+                        "Minor GC: duration %sms: %s -> %s",
+                        info.getDurationMs(),
+                        info.getBeforeGcTotal(),
+                        info.getAfterGcTotal());
+            }
+        }
+    }
+
+    private void logMajorGc(GarbageCollectionNotificationInfo info)
+    {
+        majorGcCount.incrementAndGet();
+        majorGcTime.addAndGet(info.getDurationMs());
+        majorGc.add(info.getDurationMs(), MILLISECONDS);
+
+        // assume that major GCs stop the application
+        long applicationRuntime = max(0, info.getStartTime() - lastGcEndTime);
+        lastGcEndTime = info.getEndTime();
+
+        double percentHeapReclaimed = percentOfMaxHeapReclaimed(
+                ManagementFactory.getMemoryMXBean().getHeapMemoryUsage().getMax(),
+                info.getBeforeGcTotal().toBytes(),
+                info.getAfterGcTotal().toBytes());
+        if (percentHeapReclaimed < MINIMUM_PERCENTAGE_OF_HEAP_RECLAIMED) {
+            log.warn(
+                    "Major GC: application %sms, stopped %sms: %s -> %s (only recovered %.1f%% of heap memory)",
+                    applicationRuntime,
+                    info.getDurationMs(),
+                    info.getBeforeGcTotal(),
+                    info.getAfterGcTotal(),
+                    percentHeapReclaimed);
+            return;
+        }
+
+        log.info(
+                "Major GC: application %sms, stopped %sms: %s -> %s",
+                applicationRuntime,
+                info.getDurationMs(),
+                info.getBeforeGcTotal(),
+                info.getAfterGcTotal());
+    }
+
+    static double percentOfMaxHeapReclaimed(long maxHeapMemoryUsage, double totalBeforeGcMemory, double totalAfterGcMemory)
+    {
+        if (maxHeapMemoryUsage <= 0 || (totalBeforeGcMemory / maxHeapMemoryUsage) < FRACTION_OF_MAX_HEAP_TO_TRIGGER_WARN) {
+            // There is no max heap defined or the heap is not close to the max yet
+            return 100.0;
+        }
+        return (1 - (totalAfterGcMemory / totalBeforeGcMemory)) * 100;
     }
 
     private static class GarbageCollectionNotificationInfo
@@ -232,6 +310,11 @@ public class JmxGcMonitor
         public boolean isMajorGc()
         {
             return gcAction.equalsIgnoreCase(MAJOR_GC_NAME);
+        }
+
+        public boolean isMajorZgc()
+        {
+            return gcName.equalsIgnoreCase(MAJOR_ZGC_NAME);
         }
 
         @Override
--- a/stats/src/test/java/io/airlift/stats/TestJmxGcMonitor.java
+++ b/stats/src/test/java/io/airlift/stats/TestJmxGcMonitor.java
@@ -3,6 +3,8 @@ package io.airlift.stats;
 import io.airlift.units.Duration;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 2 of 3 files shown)


stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -65,9 +65,16 @@ public class JmxGcMonitor
 
     private final TimeStat minorGc = new TimeStat();
 
+    private final long heapMemoryLimit;
+
+
     @GuardedBy("this")
     private long lastGcEndTime = System.currentTimeMillis();
 
+    public JmxGcMonitor() {
```
wendigo, line 74: formatting is off

wendigo (approved):

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -146,6 +153,13 @@ private synchronized void onNotification(Notification notification)
                         info.getDurationMs(),
                         info.getBeforeGcTotal(),
                         info.getAfterGcTotal());
+                if (info.getBeforeGcTotal().toBytes() / heapMemoryLimit > .8) {
+                    // Only warn of GC issues when the Heap is close to the limit
+                    double percentRecovered = 1 - info.getBeforeGcTotal().toBytes() / info.getAfterGcTotal().toBytes();
+                    if (percentRecovered < .1) {
+                        log.warn("Major GC only recovered %s% of heap memory", percentRecovered);
```
dain, line 160: My memory is when you are in the case that you are only freeing up a few percent memory, the GC basically runs constantly. We may want to rate limit the messages to not spam the logs.

wendigo: yeah:

```
private final RateLimiter rateLimiter = RateLimiter.create(0.1);

if (!rateLimiter.tryAcquire()) {
    return;
}
```

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: Will fix. Thanks.

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: Alternatively, we could replace the info log with a warn log and report as usual.

wweiss-starburst: As it is, we already spam the log when we get into a GC loop. It is just that we don't make progress.

wendigo: this will generate noise in the logs. rate limiting is a better way and we use it already in other places (see: HTTP client diagnostics)

wweiss-starburst: Yes. When I investigated an issue with a coordinator that would become unresponsive, there were lots of misleading logs but the GC logs showed that the VM was effectively in a GC loop. Unfortunately, whenever we saw this, it was usually too late.

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -65,9 +65,16 @@ public class JmxGcMonitor
...
     private final TimeStat minorGc = new TimeStat();
 
+    private final long heapMemoryLimit;
+
     @GuardedBy("this")
     private long lastGcEndTime = System.currentTimeMillis();
 
+    public JmxGcMonitor()
+    {
+        this.heapMemoryLimit = ManagementFactory.getMemoryMXBean().getHeapMemoryUsage().getMax();
```
electrum, line 75: This can move to the initializer

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -140,6 +147,20 @@ private synchronized void onNotification(Notification notification)
                 long applicationRuntime = max(0, info.getStartTime() - lastGcEndTime);
                 lastGcEndTime = info.getEndTime();
 
+                if (info.getBeforeGcTotal().toBytes() / heapMemoryLimit > .8) {
+                    // Only warn of GC issues when the Heap is close to the limit
+                    double percentRecovered = 1 - info.getBeforeGcTotal().toBytes() / info.getAfterGcTotal().toBytes();
```
electrum, line 152: Nit: add parens around division

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -140,6 +147,20 @@ private synchronized void onNotification(Notification notification)
                 long applicationRuntime = max(0, info.getStartTime() - lastGcEndTime);
                 lastGcEndTime = info.getEndTime();
 
+                if (info.getBeforeGcTotal().toBytes() / heapMemoryLimit > .8) {
```
electrum, line 150: Nit: we prefer to write decimals with leading zero `0.8`

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -140,6 +147,20 @@ private synchronized void onNotification(Notification notification)
...
+                    // Only warn of GC issues when the Heap is close to the limit
+                    double percentRecovered = 1 - info.getBeforeGcTotal().toBytes() / info.getAfterGcTotal().toBytes();
+                    if (percentRecovered < .1) {
```
electrum, line 153: `0.1`

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -140,6 +147,20 @@ private synchronized void onNotification(Notification notification)
...
+                        log.warn(
+                                "Major GC: application %sms, stopped %sms: %s -> %s (only recovered %s% of heap memory)",
+                                applicationRuntime,
+                                info.getDurationMs(),
+                                info.getBeforeGcTotal(),
+                                info.getAfterGcTotal(),
+                                percentRecovered);
```
electrum, line 160: This is a fraction/ratio not a percentage, so need to multiply here

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -140,6 +147,20 @@ private synchronized void onNotification(Notification notification)
                 long applicationRuntime = max(0, info.getStartTime() - lastGcEndTime);
                 lastGcEndTime = info.getEndTime();
 
+                if (info.getBeforeGcTotal().toBytes() / heapMemoryLimit > .8) {
+                    // Only warn of GC issues when the Heap is close to the limit
+                    double percentRecovered = 1 - info.getBeforeGcTotal().toBytes() / info.getAfterGcTotal().toBytes();
+                    if (percentRecovered < .1) {
+                        log.warn(
+                                "Major GC: application %sms, stopped %sms: %s -> %s (only recovered %s% of heap memory)",
```
electrum, line 155: Probably want to round here. Also, need to use `%%` to render as `%`

`only recovered %.1f%% of heap memory`

electrum (dismissed):

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
electrum: Actually, I don't think this should be a constant, as the heap size can change. This should be fetched at the time of usage.

wweiss-starburst: I am not sure what to do here. It is not uncommon for Major GCs to occur before the heap has reached it's upper bound. As best I can tell, Max can be [unspecified](https://docs.oracle.com/javase/8/docs/api/java/lang/management/MemoryUsage.html), in which case Max will return -1. So perhaps what should be done is only use this when it is defined. Thoughts?

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: done

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: done

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: done

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: good catch.

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: sounds good.

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
electrum: We should call this every time, rather than making it a constant.

wweiss-starburst: PR updated.

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -140,6 +141,22 @@ private synchronized void onNotification(Notification notification)
                 long applicationRuntime = max(0, info.getStartTime() - lastGcEndTime);
                 lastGcEndTime = info.getEndTime();
 
+                if (Optional.of(ManagementFactory.getMemoryMXBean().getHeapMemoryUsage().getMax())
```
electrum, line 144: `getMax()` returns `long` so using `Optional` doesn’t make sense here

wweiss-starburst: fixed

wweiss-starburst:
PR has been updated to support the Z collector. Note that when Generational is not enabled, the logging is noisy because it does not distinguish major and minor collections.

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -132,22 +159,29 @@ private synchronized void onNotification(Notification notification)
...
             }
-            else if (info.isMinorGc()) {
+        }
+    }
+
+    private synchronized void onZgcNotification(Notification notification)
+    {
+        if ("com.sun.management.gc.notification".equals(notification.getType())) {
+            GarbageCollectionNotificationInfo info = new GarbageCollectionNotificationInfo((CompositeData) notification.getUserData());
+            if (!zgcSupportsMajor || info.isMajorZgc()) {
```
electrum, line 181: Let's remove this flag and have `isMajorZgc()` check for both `ZGC Cycles` and `ZGC Major Cycles`

stats/src/test/java/io/airlift/stats/TestJmxGcMonitor.java
```diff
@@ -25,4 +26,15 @@ public void test()
             gcMonitor.stop();
         }
     }
+
+    @Test
+    public void testWarnTrigger()
+    {
+        // Don't warn when max heap is not specified
+        assertThat(JmxGcMonitor.percentOfMaxHeapReclaimed(-1, 100.0, 100.0)).isEqualTo(100.0);
```
electrum, line 34: Static import

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -92,7 +112,14 @@ public void stop()
         for (GarbageCollectorMXBean mbean : ManagementFactory.getGarbageCollectorMXBeans()) {
             ObjectName objectName = mbean.getObjectName();
             try {
-                ManagementFactory.getPlatformMBeanServer().removeNotificationListener(objectName, notificationListener);
+                if (mbean.getName().startsWith("ZGC")) {
+                    if (!mbean.getName().equals("ZGC Pauses")) {
```
electrum, line 116: This will miss `ZGC Major Pauses`. We can change to explicit names.

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -74,11 +81,24 @@ public void start()
         for (GarbageCollectorMXBean mbean : ManagementFactory.getGarbageCollectorMXBeans()) {
             ObjectName objectName = mbean.getObjectName();
             try {
-                ManagementFactory.getPlatformMBeanServer().addNotificationListener(
-                        objectName,
-                        notificationListener,
-                        null,
-                        null);
+                if (mbean.getName().startsWith("ZGC")) {
```
electrum, line 84: Let's change this to be explicit
```java
private static final Set<String> ZGC_MBEANS = Set.of("ZGC Cycles", "ZGC Major Cycles", "ZGC Minor Cycles");

if (ZGC_MBEANS.contains(mbean.getName())) {
```

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -92,7 +112,14 @@ public void stop()
         for (GarbageCollectorMXBean mbean : ManagementFactory.getGarbageCollectorMXBeans()) {
             ObjectName objectName = mbean.getObjectName();
             try {
-                ManagementFactory.getPlatformMBeanServer().removeNotificationListener(objectName, notificationListener);
+                if (mbean.getName().startsWith("ZGC")) {
```
electrum, line 115:
```java
if (ZGC_MBEANS.contains(mbean.getName())) {

File: stats/src/main/java/io/airlift/stats/JmxGcMonitor.java [hunk 15]
```diff
@@ -74,11 +81,24 @@ public void start()
...
-                ManagementFactory.getPlatformMBeanServer().addNotificationListener(
-                        objectName,
-                        notificationListener,
-                        null,
-                        null);
+                if (mbean.getName().startsWith("ZGC")) {
+                    // Don't bother registering ZGC Pauses because it does not report memory usage statistics
+                    if (!mbean.getName().endsWith("Pauses")) {
+                        zgcSupportsMajor |= mbean.getName().endsWith("ZGC Major");
+                        ManagementFactory.getPlatformMBeanServer().addNotificationListener(
```
electrum, line 88: Static import `getPlatformMBeanServer()` and remove the wrapping

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -132,22 +159,29 @@ private synchronized void onNotification(Notification notification)
...
                         info.getBeforeGcTotal(),
                         info.getAfterGcTotal());
             }
-            else if (info.isMinorGc()) {
+        }
+    }
+
+    private synchronized void onZgcNotification(Notification notification)
+    {
+        if ("com.sun.management.gc.notification".equals(notification.getType())) {
```
electrum, line 179: We can update this to use `com.sun.management.GarbageCollectionNotificationInfo#GARBAGE_COLLECTION_NOTIFICATION`

Also update the existing usage above

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -160,11 +194,54 @@ else if (info.isMinorGc()) {
...
+            log.warn(
+                    "Major GC: application %sms, stopped %sms: %s -> %s (only recovered %.1f%% of heap memory)",
+                    applicationRuntime,
+                    info.getDurationMs(),
+                    info.getBeforeGcTotal(),
+                    info.getAfterGcTotal(),
+                    percentHeapReclaimed);
+            return;
+        }
+        log.info(
```
electrum, line 221: Nit: add newline to separate from above

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
```diff
@@ -160,11 +194,54 @@ else if (info.isMinorGc()) {
...
+                "Major GC: application %sms, stopped %sms: %s -> %s",
+                applicationRuntime,
+                info.getDurationMs(),
+                info.getBeforeGcTotal(),
+                info.getAfterGcTotal());
+    }
+
+    static double percentOfMaxHeapReclaimed(long maxHeapMemoryUsage, double totalBeforeGcMemory, double totalAfterGcMemory)
+    {
+        if (maxHeapMemoryUsage <= 0 || totalBeforeGcMemory / maxHeapMemoryUsage < FRACTION_OF_MAX_HEAP_TO_TRIGGER_WARN) {
```
electrum, line 231: Nit: add parenthesis

electrum:
MBean names for reference

G1:
```
G1 Young Generation
G1 Concurrent GC
G1 Old Generation
```

ZGC: `-XX:+UseZGC`
```
ZGC Cycles
ZGC Pauses
```

ZGC Generational: `-XX:+UseZGC -XX:+ZGenerational`
```
ZGC Minor Cycles
ZGC Minor Pauses
ZGC Major Cycles
ZGC Major Pauses
```

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: Cycles don't return memory data. That was why I removed them.

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: My recollection is that this is meant to skip non-generational notifications  (!)

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: Done

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: done

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: done

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: done

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: done

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: done

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: Since we are using contains(), I have removed this conditional

stats/src/test/java/io/airlift/stats/TestJmxGcMonitor.java
wweiss-starburst: done

stats/src/main/java/io/airlift/stats/JmxGcMonitor.java
wweiss-starburst: Oops. Pauses don't return memory data but that is already addressed with the static set and contains().

wendigo (approved):

PR #1239, opened 2024-08-05
Remove http-server admin and jmx-http-rpc

wendigo:


wendigo:
the only reference to `jmx-http-rpc` that I found was an exclude in Trino :)

wendigo:
@electrum @dain PTAL

wendigo:
@electrum @dain PTAL

wendigo:
@electrum PTAL

dain:
I'm cool with dropping these

wendigo:
@electrum please review

http-server/src/main/java/io/airlift/http/server/HttpServerProvider.java
```diff
@@ -70,9 +67,9 @@ public HttpServerProvider(HttpServerInfo httpServerInfo,
             NodeInfo nodeInfo,
             HttpServerConfig config,
             Optional<HttpsConfig> httpsConfig,
-            @TheServlet Servlet theServlet,
-            @TheServlet Set<Filter> filters,
-            @TheServlet Set<HttpResourceBinding> resources,
+            Servlet theServlet,
+            Set<Filter> filters,
+            Set<HttpResourceBinding> resources,
```
electrum, line 72: It feels a bit strange to inject these `jakarta.servlet` directly without an annotation, but we own the injector and I can't think of any reason why application code would inject these, so I guess it's fine.

electrum (approved):
Nice! These are good cleanups.

Can you split the removal of `jmx-http-rpc` into a separate commit? It seems unrelated to the admin server.

electrum:
Future cleanup ideas:

* Remove `LoginService` binding?
* Remove `TraceTokenManager` in `HttpServerProvider` and instead bind `TraceTokenFilter` via Guice

http-server/src/main/java/io/airlift/http/server/HttpServerProvider.java
wendigo: Actually it's a preparation to reintroduce annotated Servlets, Filters along with the binder for HttpServer :)

PR #1240, opened 2024-08-07
Add dependabot configuration

wendigo:
I don't want to do that manually anymore :)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "maven"
+    directory: "/"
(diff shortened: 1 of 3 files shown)


dain (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
findinpath, line 6: why not weekly?

wendigo: Why weekly?

PR #1241, opened 2024-08-07
Improve confusing error message

wendigo:
Wrapping with [] make the error message look like the returned JSON was always an array which could be confusing to the developer.


wendigo:
@electrum @dain PTAL

wendigo:
@electrum PTAL

electrum:
We need some kind of delimiter, as there can be multiple levels of errors, nested stack traces, etc.

electrum (approved):
I somehow missed that the delimiter was changed to angle brackets.

PR #1300, opened 2024-10-19
Improve JsonSubType so that mappers on base type serialize correctly

Randgalt:
This handles JsonCode's toJson() methods

### DIFF
--- a/json/src/main/java/io/airlift/json/JsonSubType.java
+++ b/json/src/main/java/io/airlift/json/JsonSubType.java
@@ -13,11 +13,11 @@
  */
 package io.airlift.json;
 
+import com.fasterxml.jackson.annotation.JsonTypeInfo.Value;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.TreeNode;
 import com.fasterxml.jackson.core.Version;
-import com.fasterxml.jackson.databind.AnnotationIntrospector;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.JsonDeserializer;
@@ -25,9 +25,12 @@ import com.fasterxml.jackson.databind.Module;
 import com.fasterxml.jackson.databind.PropertyName;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.cfg.MapperConfig;
+import com.fasterxml.jackson.databind.introspect.Annotated;
 import com.fasterxml.jackson.databind.introspect.AnnotatedClass;
 import com.fasterxml.jackson.databind.introspect.AnnotatedMember;
+import com.fasterxml.jackson.databind.introspect.JacksonAnnotationIntrospector;
 import com.fasterxml.jackson.databind.introspect.VirtualAnnotatedMember;
+import com.fasterxml.jackson.databind.jsontype.NamedType;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.databind.node.ValueNode;
 import com.fasterxml.jackson.databind.ser.BeanPropertyWriter;
@@ -46,6 +49,9 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
+import static com.fasterxml.jackson.annotation.JsonTypeInfo.As.EXISTING_PROPERTY;
+import static com.fasterxml.jackson.annotation.JsonTypeInfo.Id.NAME;
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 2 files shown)


Randgalt:
cc @mdeady, @wendigo and @Praveen2112

Randgalt:
Alternatively we can remove `writerFor()` in all cases

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -214,6 +218,9 @@ public byte[] toJsonBytes(T instance)
             throws IllegalArgumentException
     {
         try {
+            if (isJsonSubType(mapper, instance)) {
+                return mapper.writeValueAsBytes(instance);
+            }
```
mwd410, line 223: i'm generally ok with this if it's how it has to be, but it doesn't _feel_ like it should be necessary, as `fromJson` always uses a reader, and it works just fine when reading the subtype (probably because we `addDeserializer` to the module). This just feels like it shouldn't be necessary - like there should be a way to implement a serializer that Just Works, though I could not find one.

I'm also totally cool with removing the readers and writers from the equation altogether, but that feels risky regarding downstream effects.

wendigo:
@Randgalt what's the downside of removing writerFor ?

wendigo:
@Randgalt Let's try - we can always revert :)

json/src/main/java/io/airlift/json/JsonSubType.java
```diff
@@ -53,6 +55,20 @@ public class JsonSubType
 {
     private final Set<Module> modules;
 
+    public static boolean isJsonSubType(ObjectMapper objectMapper, Object instance)
+    {
+        if (instance != null) {
```
wendigo, line 60: I'd invert this condition to reduce nesting:

if (instance == null) {
    return false;
}

json/src/main/java/io/airlift/json/JsonSubType.java
```diff
@@ -53,6 +55,20 @@ public class JsonSubType
...
+            AnnotationIntrospector annotationIntrospector = objectMapper.getSerializationConfig().getAnnotationIntrospector();
+            if (annotationIntrospector != null) {
```
wendigo, line 62: same here. return early.

json/src/main/java/io/airlift/json/JsonSubType.java
```diff
@@ -53,6 +55,20 @@ public class JsonSubType
...
+                return annotationIntrospector.allIntrospectors()
+                        .stream()
+                        .flatMap(introspector -> (introspector instanceof TypeAddingIntrospector typeAddingIntrospector) ? Stream.of(typeAddingIntrospector) : Stream.empty())
```
wendigo, line 65: do we need flatMap? isn't filter+map enough?

wendigo:
```
    public static boolean isJsonSubType(ObjectMapper objectMapper, Object instance)
    {
        if (instance == null) {
            return false;
        }
        AnnotationIntrospector annotationIntrospector = objectMapper.getSerializationConfig().getAnnotationIntrospector();
        if (annotationIntrospector == null) {
            return false;
        }

        return annotationIntrospector.allIntrospectors()
                .stream()
                .filter(TypeAddingIntrospector.class::isInstance)
                .map(TypeAddingIntrospector.class::cast)
                .anyMatch(typeAddingIntrospector -> typeAddingIntrospector.baseClass.isAssignableFrom(instance.getClass()));
    }
```

Randgalt: It's for the cast to `TypeAddingIntrospector`

wendigo: I know. the filter+map does the same without creating streams and flattening them

Randgalt:
@wendigo and @mwd410 I now have a generalized implementation that handles the same cases that using the Jackson annotations does.

mwd410:
@Randgalt i'm so glad you were able to figure out how to do this, thank you so much for the work!

mwd410 (approved):

json/src/main/java/io/airlift/json/JsonSubType.java
```diff
@@ -161,22 +167,42 @@ public void setupModule(SetupContext context)
...
+        public JsonTypeInfo.Value findPolymorphicTypeInfo(MapperConfig<?> config, Annotated ann)
+        {
+            if (ann.getRawType().equals(baseClass)) {
+                return JsonTypeInfo.Value.construct(JsonTypeInfo.Id.NAME, JsonTypeInfo.As.EXISTING_PROPERTY, propertyName, null, true, true);
+            }
+            return super.findPolymorphicTypeInfo(config, ann);
+        }
+
+        @Override
+        public List<NamedType> findSubtypes(Annotated a)
```
wendigo, line 198: s/a/annotated/

json/src/main/java/io/airlift/json/JsonSubType.java
```diff
@@ -161,22 +167,42 @@ public void setupModule(SetupContext context)
...
-        private TypeAddingIntrospector(String propertyName, Map<Class<?>, String> subClassPropertyValues)
+        private TypeAddingIntrospector(Class<?> baseClass, String propertyName, Map<Class<?>, String> subClassPropertyValues)
         {
+            this.baseClass = requireNonNull(baseClass, "baseClass is null");
             this.propertyName = requireNonNull(propertyName, "propertyName is null");
             this.subClassPropertyValues = ImmutableMap.copyOf(subClassPropertyValues);
         }
 
+        @Override
+        public JsonTypeInfo.Value findPolymorphicTypeInfo(MapperConfig<?> config, Annotated ann)
```
wendigo, line 189: s/ann/annotated

wendigo (approved):

wendigo:
Tested with Trino (product tests and some connector tests). Seems to work just fine.

PR #1331, opened 2024-12-02
Guard against negative duration

wendigo:

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
@@ -1303,8 +1303,8 @@ public class JettyHttpClient
             return;
         }
 
-        Duration responseProcessingTime = Duration.nanosSince(responseStart);
-        Duration requestProcessingTime = new Duration(responseStart - requestStart, NANOSECONDS);
+        Duration responseProcessingTime = new Duration(max(0, System.nanoTime() - responseStart), NANOSECONDS);
+        Duration requestProcessingTime = new Duration(max(0, responseStart - requestStart), NANOSECONDS);
 
         requestStats.recordResponseReceived(request.getMethod(),
                 response.getStatusCode(),


http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -1303,8 +1303,8 @@ static void recordRequestComplete(RequestStats requestStats, Request request, lo
             return;
         }
 
-        Duration responseProcessingTime = Duration.nanosSince(responseStart);
-        Duration requestProcessingTime = new Duration(responseStart - requestStart, NANOSECONDS);
+        Duration responseProcessingTime = new Duration(max(0, System.nanoTime() - responseStart), NANOSECONDS);
```
martint, line 1306: Have you seen any cases where it's negative? If duration is obtained from System.nanoTime(), the difference should never be negative.

wendigo: @martint yes we did

wendigo:
```
Caused by: java.lang.IllegalArgumentException: value is negative: -243459.0
	at io.airlift.units.Duration.<init>(Duration.java:79)
	at io.airlift.units.Duration.succinctDuration(Duration.java:66)
	at io.airlift.units.Duration.succinctNanos(Duration.java:58)
	at io.airlift.units.Duration.nanosSince(Duration.java:53)
	at io.airlift.http.client.jetty.JettyHttpClient.recordRequestComplete(JettyHttpClient.java:1306)
	at io.airlift.http.client.jetty.JettyHttpClient.lambda$buildCompletionHandler$12(JettyHttpClient.java:880)
```

wendigo:
JDK 24-ea failing due to https://github.com/openjdk/jdk/pull/22163

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
grantatspothero, line 1306: This also happens in task stats in trino (not just in the airlift client) see: https://github.com/trinodb/trino/issues/11039

If we want to guard against these issues, wouldn't it make more sense to do so in Duration itself/airlift units?

wendigo: @martint the jetty code is super simple to analyze by looking at it, it operates on delta between two System.nanoTime()-s. I can't find anything wrong there

wendigo: @martint but it is

grantatspothero:
Commenting for posterity, with @wendigo's help confirmed the negative duration issues go away when using temurin JDK (instead of corretto).

PR #1349, opened 2025-01-05
Configure Buffer pools for Jetty

wendigo:
This circumvents https://github.com/jetty/jetty.project/issues/12670

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
@@ -15,6 +15,7 @@
  */
 package io.airlift.http.client;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.net.HostAndPort;
@@ -74,6 +75,12 @@ public class HttpClientConfig
     private DataSize maxContentLength = DataSize.of(16, MEGABYTE);
     private DataSize requestBufferSize = DataSize.of(4, KILOBYTE);
     private DataSize responseBufferSize = DataSize.of(16, KILOBYTE);
+    private DataSize maxHeapMemory = max(
+            DataSize.ofBytes(Runtime.getRuntime().maxMemory() / 8),
+            DataSize.of(16, MEGABYTE));
+    private DataSize maxDirectMemory = max(
+            DataSize.ofBytes(Runtime.getRuntime().maxMemory() / 8),
+            DataSize.of(16, MEGABYTE));
     private HostAndPort socksProxy;
     private HostAndPort httpProxy;
     private boolean secureProxy;
@@ -266,6 +273,32 @@ public class HttpClientConfig
     public HttpClientConfig setResponseBufferSize(DataSize responseBufferSize)
     {
         this.responseBufferSize = responseBufferSize;
+        return this;
+    }
+
+    @MinDataSize("16MB")
+    public DataSize getMaxHeapMemory()
+    {
+        return maxHeapMemory;
+    }
+
+    @Config("http-client.max-heap-memory")
+    public HttpClientConfig setMaxHeapMemory(DataSize maxHeapMemory)
+    {
+        this.maxHeapMemory = maxHeapMemory;
+        return this;
+    }
+
+    @MinDataSize("16MB")
+    public DataSize getMaxDirectMemory()
+    {
+        return maxDirectMemory;
+    }
+
+    @Config("http-client.max-direct-memory")
+    public HttpClientConfig setMaxDirectMemory(DataSize maxDirectMemory)
+    {
+        this.maxDirectMemory = maxDirectMemory;
         return this;
     }
 
@@ -717,5 +750,14 @@ public class HttpClientConfig
             throw new ConfigurationException(ImmutableList.of(
                     new Message("http-client.http-proxy.secure can be enabled only when http-client.http-proxy is set")));
         }
+    }
+
+    @VisibleForTesting
+    static DataSize max(DataSize value, DataSize value2)
+    {
+        if (value.compareTo(value2) < 0) {
+            return value2;
+        }
+        return value;
     }
 }
--- a/http-client/src/main/java/io/airlift/http/client/jetty/BufferingResponseListener.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/BufferingResponseListener.java
@@ -15,6 +15,7 @@ import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Locale;
+import java.util.Optional;
 
 import static java.util.Objects.requireNonNull;
 
@@ -134,11 +135,8 @@ public abstract class BufferingResponseListener
      */
     public String getContentAsString()
     {
-        String encoding = this.encoding;
-        if (encoding == null) {
-            return getContentAsString(StandardCharsets.UTF_8);
-        }
-        return getContentAsString(encoding);
+        return getContentAsString(Optional.ofNullable(encoding)
+                .orElse(StandardCharsets.UTF_8.name()));
     }
 
     /**
@@ -148,11 +146,7 @@ public abstract class BufferingResponseListener
      */
     public String getContentAsString(String encoding)
     {
-        if (buffer.getLength() == 0) {
-            return null;
-        }
-        ByteBuffer output = buffer.takeByteBuffer(); // Releases a buffer
-        return new String(output.array(), 0, output.remaining(), Charset.forName(encoding));
+        return getContentAsString(Charset.forName(encoding));
     }
 
     /**
--- a/http-client/src/main/java/io/airlift/http/client/jetty/BufferingResponseListener.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/BufferingResponseListener.java
@@ -6,7 +6,8 @@ import org.eclipse.jetty.client.Result;
 import org.eclipse.jetty.http.HttpFields;
 import org.eclipse.jetty.http.HttpHeader;
 import org.eclipse.jetty.http.HttpMethod;
-import org.eclipse.jetty.util.BufferUtil;
+import org.eclipse.jetty.io.ByteBufferAccumulator;
+import org.eclipse.jetty.io.ByteBufferPool;
 
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
@@ -15,20 +16,22 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Locale;
 
+import static java.util.Objects.requireNonNull;
+
 public abstract class BufferingResponseListener
         implements Response.Listener
 {
+    private final ByteBufferAccumulator buffer;
     private final int maxLength;
-    private ByteBuffer buffer;
     private String mediaType;
     private String encoding;
 
     /**
      * Creates an instance with a default maximum length of 2 MiB.
      */
-    public BufferingResponseListener()
+    public BufferingResponseListener(ByteBufferPool byteBufferPool)
     {
-        this(2 * 1024 * 1024);
+        this(byteBufferPool, 2 * 1024 * 1024);
     }
 
     /**
@@ -36,8 +39,9 @@ public abstract class BufferingResponseListener
      *
      * @param maxLength the maximum length of the content
      */
-    public BufferingResponseListener(int maxLength)
+    public BufferingResponseListener(ByteBufferPool byteBufferPool, int maxLength)
     {
+        this.buffer = new ByteBufferAccumulator(requireNonNull(byteBufferPool, "byteBufferPool is null"), true);
         if (maxLength < 0) {
             throw new IllegalArgumentException("Invalid max length " + maxLength);
         }
@@ -95,16 +99,7 @@ public abstract class BufferingResponseListener
         if (length == 0) {
             return;
         }
-        if (length > BufferUtil.space(buffer)) {
-            int remaining = buffer == null ? 0 : buffer.remaining();
-            if (remaining + length > maxLength) {
-                response.abort(new IllegalArgumentException("Buffering capacity " + maxLength + " exceeded"));
-            }
-            int requiredCapacity = buffer == null ? length : buffer.capacity() + length;
-            int newCapacity = Math.min(Integer.highestOneBit(requiredCapacity) << 1, maxLength);
-            buffer = BufferUtil.ensureCapacity(buffer, newCapacity);
-        }
-        BufferUtil.append(buffer, content);
+        buffer.copyBuffer(content);
     }
 
     @Override
@@ -126,10 +121,10 @@ public abstract class BufferingResponseListener
      */
     public byte[] getContent()
     {
-        if (buffer == null) {
+        if (buffer.getLength() == 0) {
             return new byte[0];
         }
-        return BufferUtil.toArray(buffer);
+        return buffer.takeByteBuffer().array();
     }
 
     /**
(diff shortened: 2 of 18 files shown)


wendigo:
@martint ptal

guyco33:
I can approve that this fix solves https://github.com/trinodb/trino/issues/24572

guyco33 (approved):

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -94,6 +98,7 @@ public enum ClientCertificate
     }
 
     private static final Logger log = Logger.get(HttpServer.class);
+    private static final int DEFAULT_MAX_CAPACITY = 64 * 1024;
```
losipiuk, line 101: nit: this constant name is not meaningful. Maybe just inline.

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -139,7 +144,19 @@ public HttpServer(
             log.info("Virtual threads support is enabled");
             threadPool.setVirtualThreadsExecutor(executor);
         }
-        server = new Server(threadPool);
+
+        int maxBufferSize = toIntExact(max(
```
losipiuk, line 148: Maybe ensure that minimum is `DEFAULT_MAX_CAPACITY` so you do not shrink it too much.

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -139,7 +144,19 @@ public HttpServer(
...
+                toSafeBytes(config.getMaxRequestHeaderSize()),
+                toSafeBytes(config.getMaxResponseHeaderSize())));
+
+        ByteBufferPool byteBufferPool = new ArrayByteBufferPool.Quadratic(
+                0, // min capacity
+                maxBufferSize == -1 ? DEFAULT_MAX_CAPACITY : maxBufferSize, // max capacity
+                Integer.MAX_VALUE, // maximum number of buffers
```
losipiuk, line 155: drop the  `// maximum number of buffers` comment. It is actually max bucket size - which counts bytes - not number of objects.

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -471,4 +487,13 @@ private static ServerConnector createServerConnector(
         connector.open(channel);
         return connector;
     }
+
+    private static long toSafeBytes(DataSize dataSize)
```
losipiuk, line 491: OptionalLong?

losipiuk (commented):
nice: couple of minor comments

wendigo:
Ping @electrum @dain @martint

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -94,6 +96,9 @@ public class HttpServerConfig
 
     private boolean compressionEnabled = true;
 
+    private DataSize maxHeapMemory = DataSize.of(1, GIGABYTE);
+    private DataSize maxDirectMemory = DataSize.of(256, MEGABYTE);
```
martint, line 100: What's the size in Jetty's default configuration?

wendigo: Which seems like a lot and we didn't account for that in our heap headroom estimatons

wendigo:
```
    private long maxMemory(long maxMemory)
    {
        if (maxMemory < 0)
            return -1;
        if (maxMemory == 0)
            return Runtime.getRuntime().maxMemory() / 8;
        return maxMemory;
    }
```

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -473,6 +478,36 @@ public HttpServerConfig setCompressionEnabled(boolean compressionEnabled)
         return this;
     }
 
+    @MinDataSize("128MB")
```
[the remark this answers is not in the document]
wendigo: this is 1/8th of the 1GB heap

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
martint: My concern is that it may be too large for applications that run with much less memory. I would suggest we make the default 1/8th of the heap with a cap at that value, unless overridden via config with a specific reservation.

martint: Also, what's the corresponding default for "maxHeapMemory"?

wendigo: both heap and direct are set to 1/8th of the max memory

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -473,6 +485,36 @@ public HttpServerConfig setCompressionEnabled(boolean compressionEnabled)
...
+    @MaxDataSize("32GB")
```
losipiuk, line 489: this limit seems artificial

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -473,6 +485,36 @@ public HttpServerConfig setCompressionEnabled(boolean compressionEnabled)
...
+    @Config("http-server.max-heap-memory")
+    @ConfigHidden
+    public HttpServerConfig setMaxHeapMemory(DataSize maxHeapMemory)
+    {
+        this.maxHeapMemory = maxHeapMemory;
+        return this;
+    }
+
+    @MinDataSize("128MB")
+    @MaxDataSize("32GB")
```
losipiuk, line 504: also this

http-client/src/test/java/io/airlift/http/client/TestHttpClientConfig.java
```diff
@@ -239,4 +246,16 @@ private List<String> getJettyDefaultExcludedCiphers()
         SslContextFactory sslContextFactory = new SslContextFactory.Client();
         return Arrays.asList(sslContextFactory.getExcludeCipherSuites());
     }
+
+    @VisibleForTesting
+    static DataSize clamp(DataSize value, DataSize min, DataSize max)
```
losipiuk, line 251: drop this one if you already expose version from `HttpClientConfig`

http-client/src/main/java/io/airlift/http/client/jetty/BufferingResponseListener.java
```diff
@@ -15,29 +16,32 @@
...
     private final int maxLength;
-    private ByteBuffer buffer;
     private String mediaType;
     private String encoding;
 
     /**
      * Creates an instance with a default maximum length of 2 MiB.
      */
-    public BufferingResponseListener()
+    public BufferingResponseListener(ByteBufferPool byteBufferPool)
```
losipiuk, line 32: Can you upstream this changes to jetty?

http-client/src/main/java/io/airlift/http/client/jetty/BufferingResponseListener.java
```diff
@@ -6,7 +6,8 @@
 import org.eclipse.jetty.http.HttpFields;
 import org.eclipse.jetty.http.HttpHeader;
 import org.eclipse.jetty.http.HttpMethod;
-import org.eclipse.jetty.util.BufferUtil;
+import org.eclipse.jetty.io.ByteBufferAccumulator;
```
losipiuk, line 9: Can you provide rationale for `Acquire ByteBuffer from client pool` in commit message. 
It is not obvious from the code, as end of the day we are copying the data to new buffer anyway. So why the pull.
I assume this is about preventing from OOMs due to delayed direct byte buffer memory reclaiming, right?

losipiuk (commented):
some comments. Thanks

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
wendigo: the default would reach this value if the heap size is 245 GB

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
wendigo: the default would reach this value if the heap size is 245 GB

http-client/src/main/java/io/airlift/http/client/jetty/BufferingResponseListener.java
wendigo: https://github.com/jetty/jetty.project/issues/12687

http-client/src/main/java/io/airlift/http/client/jetty/BufferingResponseListener.java
wendigo: Done

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
wendigo: Done

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
losipiuk: Do you suggest 256GB should be enough for everybody?

http-client/src/test/java/io/airlift/http/client/TestHttpClientConfig.java
wendigo: Dropped

wendigo:
I've run 1M queries on the local cluster with these changes applied and it's stable:

Screenshot 2025-01-09 at 18 38 47

wendigo:
Screenshot 2025-01-09 at 18 39 54

guyco33:
I've also tested the recent updates with significant production load and the cluster is stable

wendigo:
@guyco33 thanks for confirming this!

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
martint: Do we even need a max allowed value here?

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -74,6 +75,14 @@ public class HttpClientConfig
     private DataSize maxContentLength = DataSize.of(16, MEGABYTE);
     private DataSize requestBufferSize = DataSize.of(4, KILOBYTE);
     private DataSize responseBufferSize = DataSize.of(16, KILOBYTE);
+    private DataSize maxHeapMemory = clamp(
+            DataSize.ofBytes(Runtime.getRuntime().maxMemory() / 8),
+            DataSize.of(128, MEGABYTE),
```
martint, line 80: What's the minimum value that will possibly work. E.g., if you run an application with a heap of 100 MB, what does jetty do by default?

wendigo: it will assign 1/8th of the memory

wendigo: but 100 mb isn't enough for the airlift based application to even load I think

wendigo: I can remove the @MaxSize but I think that it should be clamped either way. 32 GB for the http-server, even a busy one, is enough

wendigo: I've dropped @MaxDataSize and lowered @MinDataSize and default value to 16 MB. This is extremely low value that is highly unlikely to occur in any real-world deployment scenario as this would imply 128 MB heap which is not enough to start most Java applications.

martint: >  which is not enough to start most Java applications.

It depends on what application you're talking about. Simple well-written clients can run with limited memory. The Trino CLI can run with as little as 30MB, for instance.

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -94,6 +96,14 @@ public class HttpServerConfig
 
     private boolean compressionEnabled = true;
 
+    private DataSize maxHeapMemory = max(
+            DataSize.ofBytes(Runtime.getRuntime().maxMemory() / 8),
+            DataSize.of(16, MEGABYTE));
```
martint, line 101: This is different from what I had suggested earlier. This sets up a floor on the max amount of memory of 16MB, and scales above that at 1/8th of the heap size. The issue is with applications that run with much smaller heap sizes.

I would prefer we don't place a lower bound on the max amount of memory:
* Let Jetty pick the memory setting with whichever heuristic it wants to use. 
* Allow for overriding the heuristic by setting a hard-coded max size via configuration.

We could place an upper limit on the max amount of memory to use if no config is provided (as I mentioned in my earlier comment), but I don't think that's necessary at this point -- that's something we can add in the future if warranted.

http-client/src/main/java/io/airlift/http/client/jetty/BufferingResponseListener.java
```diff
@@ -126,10 +121,10 @@ public String getEncoding()
      */
     public byte[] getContent()
     {
-        if (buffer == null) {
+        if (buffer.getLength() == 0) {
             return new byte[0];
         }
-        return BufferUtil.toArray(buffer);
+        return buffer.takeByteBuffer().array();
```
martint, line 128: This may return more data than what's represented by the ByteBuffer bounds, no? Is that what we want here?

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
wendigo: Droprd

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
wendigo: trino-cli isn't using jetty client

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
wendigo: The heuristic is 1/8th of heap size which is a value replicated here. What's the ask here?

martint: The main ask is to not put an artificial floor on the max amount of memory. In this code, the value will range from [16MB, ∞), scaling at 1/8th of heap size, which means any application using this library will use, in the worst case, at least 16MB.

Taking a step back, my understanding of what we're trying to achieve here is to be able to limit the amount of memory used by jetty buffers, which can end up using too much memory when heap sizes are large.

Summarizing from my two earlier comments, I can see a couple of ways of setting this up in a way that doesn't sacrifice the ability to run smaller applications:
1. Allow the heap memory to range from [1, x] at 1/8th of heap, where `x` is a user specified value, or ∞ if the configuration is omitted. This was my earlier comment, but I realize it's can be a little confusing to users. The value is not the max heap size, but a cap on the max heap size that jetty would calculate.
2. Allow the max heap memory to range from [1, ∞] at 1/8th of heap, unless the user specifies a concrete max via config. This can be achieved either by replicating Jetty's heuristic here, or by instantiating Jetty with a concrete vs default value later, depending on whether the config is present or absent.

http-client/src/main/java/io/airlift/http/client/jetty/BufferingResponseListener.java
wendigo: takeByteBuffer does the copy to a new sized byte buffer

PR #1353, opened 2025-01-09
Set default encoding for the error handler in HTTP module

ksobolew:
This makes handling of all errors consistent, where we show a plain text error message. For most errors this gets handled by the context, but some errors, like a malformed URI, get processed even before we know which context to forward to, and for those errors the default of `text/html` was used.


ksobolew:
cc: @wendigo I guess? :) (I can't even assign reviewers for some reason, a CI is not running?)

wendigo (approved):

wendigo:
@ksobolew needed approval to run CI

PR #1354, opened 2025-01-09
Add option to retrieve used properties metadata from Bootstrap

piotrrzysko:
This PR introduces the following changes:
* Adds the `Bootstrap.configureAndGetUsedProperties()` method, which is analogous to the existing `Bootstrap.configure()` but returns more information. Instead of returning just a set of used property names, it provides metadata for the used properties.
* Adds:
    * `ConfigurationFactory.consumeProperty(ConfigPropertyMetadata)` 
    * `ConfigurationFactory.getUsedPropertiesWithMetadata()`
    * `AbstractConfigurationAwareModule.consumeProperty(ConfigPropertyMetadata)`
    
   Additionally, it deprecates the corresponding methods that operate on raw property names.
* Adds an option to suppress configuration errors and warnings, which is helpful for retrieving used properties without triggering the logging typical of a full application bootstrap.

Examples of usage:
* [Bootstrap.configureAndGetUsedProperties()]( https://github.com/trinodb/trino/blob/47db44be99415cf51d9e33929dce3f11de2f307f/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveConnectorFactory.java#L99C62-L99C91)
* [Bootstrap.suppressErrorsAndWarnings()](https://github.com/trinodb/trino/blob/47db44be99415cf51d9e33929dce3f11de2f307f/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveConnectorFactory.java#L198-L199)
* [getUsedPropertiesWithMetadata()/consumeProperty()](https://github.com/trinodb/trino/blob/47db44be99415cf51d9e33929dce3f11de2f307f/lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/FileSystemModule.java#L91)

### DIFF
--- a/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
@@ -80,6 +80,7 @@ public class Bootstrap
     private boolean initializeLogging = true;
     private boolean quiet;
     private boolean loadSecretsPlugins;
+    private boolean skipErrorReporting;
 
     private State state = State.UNINITIALIZED;
     private ConfigurationFactory configurationFactory;
@@ -146,6 +147,12 @@ public class Bootstrap
     public Bootstrap loadSecretsPlugins()
     {
         this.loadSecretsPlugins = true;
+        return this;
+    }
+
+    public Bootstrap skipErrorReporting()
+    {
+        this.skipErrorReporting = true;
         return this;
     }
 
@@ -241,7 +248,7 @@ public class Bootstrap
         }
 
         // If there are configuration errors, fail-fast to keep output clean
-        if (!errors.isEmpty()) {
+        if (!skipErrorReporting && !errors.isEmpty()) {
             throw new ApplicationConfigurationException(errors, warnings);
         }
 
@@ -251,7 +258,7 @@ public class Bootstrap
         }
 
         // Log any warnings
-        if (!warnings.isEmpty()) {
+        if (!skipErrorReporting && !warnings.isEmpty()) {
             StringBuilder message = new StringBuilder();
             message.append("Configuration warnings\n");
             message.append("==========\n\n");
--- a/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
@@ -25,6 +25,7 @@ import com.google.inject.Injector;
 import com.google.inject.Module;
 import com.google.inject.Stage;
 import com.google.inject.spi.Message;
+import io.airlift.configuration.ConfigPropertyMetadata;
 import io.airlift.configuration.ConfigurationFactory;
 import io.airlift.configuration.ConfigurationInspector;
 import io.airlift.configuration.ConfigurationInspector.ConfigAttribute;
@@ -50,6 +51,7 @@ import java.util.Set;
 import java.util.TreeMap;
 
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static io.airlift.configuration.ConfigurationLoader.getSystemProperties;
 import static io.airlift.configuration.ConfigurationLoader.loadPropertiesFrom;
 import static io.airlift.configuration.TomlConfiguration.createTomlConfiguration;
@@ -150,7 +152,7 @@ public class Bootstrap
     /**
      * Validate configuration and return used properties.
      */
-    public Set<String> configure()
+    public Set<ConfigPropertyMetadata> configure()
     {
         checkState(state == State.UNINITIALIZED, "Already configured");
         state = State.CONFIGURED;
@@ -229,7 +231,10 @@ public class Bootstrap
 
         // at this point all config file properties should be used
         // so we can calculate the unused properties
-        unusedProperties.keySet().removeAll(configurationFactory.getUsedProperties());
+        Set<String> usedProperties = configurationFactory.getUsedProperties().stream()
+                .map(ConfigPropertyMetadata::name)
+                .collect(toImmutableSet());
+        unusedProperties.keySet().removeAll(usedProperties);
 
         for (String key : unusedProperties.keySet()) {
             errors.add(new Message(format("Configuration property '%s' was not used", key)));
--- a/bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrap.java
+++ b/bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrap.java
@@ -15,14 +15,19 @@
  */
 package io.airlift.bootstrap;
 
+import com.google.inject.Binder;
 import com.google.inject.ConfigurationException;
 import com.google.inject.Inject;
 import com.google.inject.ProvisionException;
 import com.google.inject.Scopes;
 import com.google.inject.spi.Message;
+import io.airlift.configuration.AbstractConfigurationAwareModule;
 import io.airlift.configuration.Config;
+import io.airlift.configuration.ConfigPropertyMetadata;
+import io.airlift.configuration.ConfigSecuritySensitive;
 import org.junit.jupiter.api.Test;
 
+import static io.airlift.configuration.ConditionalModule.conditionalModule;
 import static io.airlift.configuration.ConfigBinder.configBinder;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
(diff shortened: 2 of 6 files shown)


configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -546,6 +568,7 @@ private Object getInjectedValue(AttributeMetadata attribute, ConfigurationMetada
     {
         String name = prefix + injectionPoint.getProperty();
         usedProperties.add(name);
```
wendigo, line 570: Can we have a single list instead?

piotrrzysko: Unfortunately not, as the older approach only stores property names and is not aware of whether a property is sensitive (see `ConfigurationFactory#consumeProperty(String)`).

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -130,18 +132,38 @@ public Map<String, String> getProperties()
 
     /**
      * Marks the specified property as consumed.
+     *
+     * @deprecated Use {@link #consumeProperty(ConfigPropertyMetadata)} instead
      */
+    @Deprecated
     public void consumeProperty(String property)
```
wendigo, line 139: There is only a single usage of this API in Trino and it's a case when one Bootstrap.configure() is passed to another one. I think that we can break compatibility here to make the code more consistent

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -147,10 +149,35 @@ public Bootstrap loadSecretsPlugins()
...
+        return this;
+    }
+
     /**
      * Validate configuration and return used properties.
+     *
+     * @deprecated Use {@link #configureAndGetUsedProperties} instead
      */
+    @Deprecated
     public Set<String> configure()
```
wendigo, line 164: Let's break compatibility here and return metadata

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
wendigo: We can break compatibility here. There is a single use-case in Trino where we use it and it's based on values returned from other Bootstrap.

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -147,10 +150,16 @@ public Bootstrap loadSecretsPlugins()
         return this;
     }
 
+    public Bootstrap suppressErrorsAndWarnings()
```
wendigo, line 153: nit: skipErrorReporting

configuration/src/main/java/io/airlift/configuration/ConfigPropertyMetadata.java
```diff
@@ -0,0 +1,11 @@
+package io.airlift.configuration;
+
+public record ConfigPropertyMetadata(String name, boolean securitySensitive)
+        implements Comparable<ConfigPropertyMetadata>
+{
+    @Override
+    public int compareTo(ConfigPropertyMetadata that)
+    {
+        return this.name.compareTo(that.name);
```
wendigo, line 9: nit: name.compareTo(that.name), this. qualifier is redundant

configuration/src/test/java/io/airlift/configuration/TestConfigurationFactory.java
```diff
@@ -440,15 +441,15 @@ public void testUsedPropertiesWithFailure()
...
+        assertThat(configurationFactory.getUsedProperties().stream().map(ConfigPropertyMetadata::name).collect(toImmutableSet())).isEqualTo(properties.keySet());
     }
 
     private static Injector createInjector(Map<String, String> properties, Module module, WarningsMonitor warningsMonitor)
     {
         ConfigurationFactory configurationFactory = new ConfigurationFactory(properties, warningsMonitor);
         configurationFactory.registerConfigurationClasses(ImmutableList.of(module));
         List<Message> messages = configurationFactory.validateRegisteredConfigurationProvider();
-        assertThat(configurationFactory.getUsedProperties()).hasSameElementsAs(properties.keySet());
+        assertThat(configurationFactory.getUsedProperties().stream().map(ConfigPropertyMetadata::name)).hasSameElementsAs(properties.keySet());
```
wendigo, line 452: you can do: 

`assertThat(configurationFactory.getUsedProperties())
   .extracting(ConfigPropertyMetadata::name)
   .hasSameElementsAs()`

wrap this for improved readability

wendigo: here and elsewhere

bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrap.java
```diff
@@ -128,15 +135,94 @@ public void testSeparateConfigureAndInitialize()
...
+    public void testSuppressErrorsAndWarnings()
+    {
+        Bootstrap bootstrap = new Bootstrap(binder -> {
+            configBinder(binder).bindConfig(FooConfig.class);
+            binder.bind(FooInstance.class).asEagerSingleton();
+        });
+        bootstrap.setOptionalConfigurationProperty("foo.enabled", "shouldBeBoolean");
+        bootstrap.suppressErrorsAndWarnings();
+
+        assertThatCode(bootstrap::configure).doesNotThrowAnyException();
```
wendigo, line 220: I don't like this assertion.

If the bootstrap.configure() will throw the test will fail so it's already validated.

wendigo (approved):
LGTM % review comments

wendigo:
@piotrrzysko amend first commit's message as it's outdated

bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrap.java
piotrrzysko: I can remove it, but then we will have something like this:
```java
    @Test
    public void testSkipErrorReporting()
    {
        Bootstrap bootstrap = new Bootstrap(binder -> {
            configBinder(binder).bindConfig(FooConfig.class);
            binder.bind(FooInstance.class).asEagerSingleton();
        });
        bootstrap.setOptionalConfigurationProperty("foo.enabled", "shouldBeBoolean");
        bootstrap.skipErrorReporting();

        bootstrap.configure();
        
        assertThatThrownBy(bootstrap::initialize)
                .isInstanceOf(CreationException.class)
                .hasMessageContaining("Invalid value 'shouldBeBoolean' for type boolean (property 'foo.enabled')");
    }
```

I know that this assertion doesn't do much in practice, but it emphasises what the expectations are: `configure` should not fail, `initialize` should.

wendigo: comment would do

wendigo: let's leave it as-is

configuration/src/test/java/io/airlift/configuration/TestConfigurationFactory.java
```diff
@@ -426,7 +426,7 @@ public void testUsedProperties()
         ConfigurationFactory configurationFactory = new ConfigurationFactory(properties);
         configurationFactory.registerConfigurationClasses(ImmutableList.of(binder -> configBinder(binder).bindConfig(AnnotatedSetter.class)));
         configurationFactory.validateRegisteredConfigurationProvider();
-        assertThat(configurationFactory.getUsedProperties()).hasSameElementsAs(ImmutableSet.of("string-value", "boolean-value"));
+        assertThat(configurationFactory.getUsedProperties()).extracting(ConfigPropertyMetadata::name).hasSameElementsAs(ImmutableSet.of("string-value", "boolean-value"));
```
wendigo, line 429: nit: wrap method calls

configuration/src/test/java/io/airlift/configuration/TestConfigurationFactory.java
```diff
@@ -440,15 +440,15 @@ public void testUsedPropertiesWithFailure()
...
+        assertThat(configurationFactory.getUsedProperties()).extracting(ConfigPropertyMetadata::name).hasSameElementsAs(properties.keySet());
     }
 
     private static Injector createInjector(Map<String, String> properties, Module module, WarningsMonitor warningsMonitor)
     {
         ConfigurationFactory configurationFactory = new ConfigurationFactory(properties, warningsMonitor);
         configurationFactory.registerConfigurationClasses(ImmutableList.of(module));
         List<Message> messages = configurationFactory.validateRegisteredConfigurationProvider();
-        assertThat(configurationFactory.getUsedProperties()).hasSameElementsAs(properties.keySet());
+        assertThat(configurationFactory.getUsedProperties()).extracting(ConfigPropertyMetadata::name).hasSameElementsAs(properties.keySet());
```
wendigo, line 451: nit: wrap method calls

wendigo (approved):

PR #1360, opened 2025-01-15
Honor java.net.preferIPv6Address configuration in NodeInfo

jalpan-randeri:
This commit add support for honoring the preference of selecting ipv6 over ipv4 ip address in the internal address selection

 - [Y] Git commit messages follow https://cbea.ms/git-commit/.
 - [Y] All automated tests are passing.

### DIFF
--- a/node/src/main/java/io/airlift/node/NodeInfo.java
+++ b/node/src/main/java/io/airlift/node/NodeInfo.java
@@ -31,10 +31,13 @@ import java.net.InetAddress;
 import java.net.NetworkInterface;
 import java.net.SocketException;
 import java.net.UnknownHostException;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
+import java.util.function.Function;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
 import static com.google.common.base.Preconditions.checkArgument;
@@ -58,6 +61,7 @@ public class NodeInfo
     private final InetAddress bindIp;
     private final long startTime = System.currentTimeMillis();
     private final Map<String, String> annotations;
+    private static boolean preferIpv6ForTest;
 
     public NodeInfo(String environment)
     {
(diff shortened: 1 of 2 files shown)


jalpan-randeri:
@martint  can you review this ?

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -58,6 +61,7 @@ public class NodeInfo
     private final InetAddress bindIp;
     private final long startTime = System.currentTimeMillis();
     private final Map<String, String> annotations;
+    private static boolean preferIpv6ForTest;
```
martint: This approach to testing is problematic.
1. It introduces non-thread-safe state in a class that is meant to be thread-safe
2. It adds mutable global state, which is harder to reason about.
3. It add testing code in the main execution path, which adds complexity and potential for things breaking.

A better way of modeling it would be to pass the preference in the constructor of this object (e.g., via the config object)

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -301,13 +305,72 @@ private static String findInternalAddress(AddressSource addressSource)
...
+        // check all up network interfaces for a good v6 address
+        for (InetAddress address : goodAddresses) {
+            if (isV6Address(address)) {
+                return Optional.of(address);
+            }
+        }
+        return Optional.empty();
+    }
+
+    private static Optional<InetAddress> findLocalAddress(List<InetAddress> goodAddresses)
```
martint: Rename `goodAddresses` to `candidates`

martint (changes requested):

node/src/main/java/io/airlift/node/NodeInfo.java
jalpan-randeri: Make sense. Thank you for feedback. Fixed it

node/src/main/java/io/airlift/node/NodeInfo.java
jalpan-randeri: Fixed

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -300,14 +307,59 @@ private static String findInternalAddress(AddressSource addressSource)
...
+            searchOrder.add(NodeInfo::findIpv4address);
+            searchOrder.add(NodeInfo::findIpv6address);
+        }
+
+        List<InetAddress> goodAddresses = getGoodAddresses();
+        return searchOrder.stream()                             // list of sources
+            .map(source -> source.apply(goodAddresses))         // potential results
+            .filter(Optional::isPresent)                        // select only valid results
+            .findFirst()                                        // take the first value
+            .orElse(Optional.empty())                     // we could not find any source
```
martint: These comments don't provide any more info than what can be inferred from reading the code. Remove them.

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -300,14 +307,59 @@ private static String findInternalAddress(AddressSource addressSource)
...
+            .orElse(null);                                // it is most likely that this is a disconnected developer machine
```
martint, line 330: This comment is good. It provides context that is not evident from the code.

martint (commented):
A couple additional of minor comments. Otherwise, it looks good.

node/src/main/java/io/airlift/node/NodeInfo.java
jalpan-randeri: Done

node/src/main/java/io/airlift/node/NodeInfo.java
jalpan-randeri: Done

node/src/main/java/io/airlift/node/NodeConfig.java
```diff
@@ -46,6 +46,7 @@ public class NodeConfig
     private String configSpec;
     private AddressSource internalAddressSource = AddressSource.IP;
     private String annotationFile;
+    private boolean preferIpv6Address = "true".equalsIgnoreCase(System.getenv("java.net.preferIPv6Address"));
```
xkrogen, line 49: @jalpan-randeri shouldn't this be checking `System.getProperty()`, not `System.getenv()`?

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -300,14 +307,59 @@ private static String findInternalAddress(AddressSource addressSource)
...
+        // check all up network interfaces for a good v6 address
+        for (InetAddress address : candidates) {
+            if (isV6Address(address)) {
+                return Optional.of(address);
+            }
+        }
+        return Optional.empty();
+    }
+
+    private static Optional<InetAddress> findLocalAddress(List<InetAddress> candidates)
```
xkrogen, line 355: @jalpan-randeri couple of comments/questions:
- Should this be called `findLocalIpv4Address`? It only looks for IPv4, right?
- `candidates` is not used in this function
- Why do we only look for an IPv4 local addr? Why not IPv6 as well (if `preferIpv6Address == true`)?

node/src/main/java/io/airlift/node/NodeInfo.java
lukasz-stec, line 367: If `InetAddress.getLocalHost` fails with `UnknownHostException` (can happen if the hostname cannot be resolved to ip), this catch will return `Optional.of(127.0.0.1)`, making the `findInternalIp` logic prefer this address over others in case of ipv4. That is incorrect behaviour as we could still detect ip based on the network interfaces. It was working like that before this change.

wendigo: Can you submit a bugfix?

lukasz-stec: Here is a try https://github.com/airlift/airlift/pull/1735. I don't know how to test it automatically, though. I tested this in a Docker container manually, and it works as expected.

PR #1366, opened 2025-01-24
Map jaxrs json parsing exceptions to 400s not 500s

grantatspothero:
Still testing not ready for review.


grantatspothero:
Approach does not work, there is no way to distinguish Jackson exceptions that happen during the JAXRS json mapping phase vs Jackson exceptions that happen during the middle of a request.

`ExceptionMapper<JacksonException>` is too broad, we need some extra context to distinguish where the jackson exception is coming from.

PR #1368, opened 2025-01-28
Fix double calling of PostConstruct/PreDestroy methods

Randgalt:
Guice sometimes double-provisions objects. In particular when `OptionalBinder` is used. This causes the LifeCycleManager to call `@PostConstruct` and  `@PreDestroy` methods twice.

To fix, store objects that have either `@PostConstruct` or `@PreDestroy` in a concurrent identity-based map to prevent invoking these methods more than once.

### DIFF
--- a/bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
@@ -39,6 +39,7 @@ public final class LifeCycleManager
 {
     private final Logger log = Logger.get(LifeCycleManager.class);
     private final AtomicReference<State> state = new AtomicReference<>(State.LATENT);
+    private final ConcurrentWeakIdentitySet startedInstances = new ConcurrentWeakIdentitySet();
     private final Queue<Object> managedInstances = new ConcurrentLinkedQueue<>();
     private final LifeCycleMethodsMap methodsMap;
     private final AtomicReference<Thread> shutdownHook = new AtomicReference<>();
@@ -208,9 +209,6 @@ public final class LifeCycleManager
         State currentState = state.get();
         checkState((currentState != State.STOPPING) && (currentState != State.STOPPED), "life cycle is stopped");
         startInstance(instance);
-        if (methodsMap.get(instance.getClass()).hasFor(PreDestroy.class)) {
-            managedInstances.add(instance);
-        }
     }
 
     private void stopInstance(Object obj, LifeCycleStopFailureHandler handler)
@@ -233,6 +231,22 @@ public final class LifeCycleManager
     {
         log.debug("Starting %s", obj.getClass().getName());
         LifeCycleMethods methods = methodsMap.get(obj.getClass());
+
+        if (!methods.hasFor(PostConstruct.class) && !methods.hasFor(PreDestroy.class)) {
+            // no need to track in startedInstances or managedInstances
+            return;
+        }
+
+        // Guice can double provision instances (in particular with Optional binding).
+        // Protect against calling post-construct and pre-destroy methods more than once
+        if (!startedInstances.add(obj)) {
+            return;
+        }
+
+        if (methods.hasFor(PreDestroy.class)) {
+            managedInstances.add(obj);
+        }
+
         for (Method postConstruct : methods.methodsFor(PostConstruct.class)) {
             log.debug("- invoke %s::%s()", postConstruct.getDeclaringClass().getName(), postConstruct.getName());
             try {
--- a/bootstrap/src/main/java/io/airlift/bootstrap/ConcurrentWeakIdentitySet.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/ConcurrentWeakIdentitySet.java
@@ -1,0 +1,66 @@
+package io.airlift.bootstrap;
+
+import com.google.common.collect.Sets;
+
+import java.lang.ref.ReferenceQueue;
+import java.lang.ref.WeakReference;
+import java.util.Set;
+
+import static java.lang.System.identityHashCode;
+
+class ConcurrentWeakIdentitySet
+{
+    private final Set<Wrapper> set = Sets.newConcurrentHashSet();
+    private final ReferenceQueue<Object> queue = new ReferenceQueue<>();
+
+    // copied/modified from WeakHashMap implementation
+    private static class Wrapper
+            extends WeakReference<Object>
+    {
+        private final int id;
+
+        private Wrapper(Object o, ReferenceQueue<Object> queue)
+        {
+            super(o, queue);
+
+            // Asserting that "id" refers to an object's address in memory and that no
+            // two objects can have the same identityHashCode.
+            // This may not be true for all VM implementations.
+            id = identityHashCode(o);
+        }
+
+        @Override
+        public boolean equals(Object o)
+        {
+            return (this == o) || ((o instanceof Wrapper wrapper) && (wrapper.id == id));
+        }
+
+        @Override
+        public int hashCode()
+        {
+            return id;
+        }
+    }
+
+    int size()
+    {
+        removeStaleEntries();
+
+        return set.size();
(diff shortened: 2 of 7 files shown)


Randgalt:
cc @electrum and @wendigo

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -233,7 +235,27 @@ private void startInstance(Object obj)
     {
         log.debug("Starting %s", obj.getClass().getName());
         LifeCycleMethods methods = methodsMap.get(obj.getClass());
-        for (Method postConstruct : methods.methodsFor(PostConstruct.class)) {
+
+        Collection<Method> postConstructMethods = methods.methodsFor(PostConstruct.class);
+        boolean hasPostConstructMethods = !postConstructMethods.isEmpty();
+        boolean hasPreDestroyMethod = methodsMap.get(obj.getClass()).hasFor(PreDestroy.class);
+
+        if (!hasPostConstructMethods && !hasPreDestroyMethod) {
```
wendigo, line 243: just: 
```
       if (!methods.hasFor(PostConstruct.class) && !methods.hasFor(PreDestroy.class)) {
            // no need to track in startedInstances or managedInstances
            return;
        }
```

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -233,7 +235,27 @@ private void startInstance(Object obj)
...
+        // Protect against calling post-construct and pre-destroy methods more than once
+        if (startedInstances.putIfAbsent(identityHashCode(obj), obj) != null) {
+            return;
+        }
+
+        if (hasPreDestroyMethod) {
+            managedInstances.add(obj);
+        }
+
+        for (Method postConstruct : postConstructMethods) {
```
wendigo, line 258:
```suggestion
       for (Method postConstruct : methods.methodsFor(PostConstruct.class)) {
```

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -233,7 +235,27 @@ private void startInstance(Object obj)
...
+        boolean hasPreDestroyMethod = methodsMap.get(obj.getClass()).hasFor(PreDestroy.class);
+
+        if (!hasPostConstructMethods && !hasPreDestroyMethod) {
+            // no need to track in startedInstances or managedInstances
+            return;
+        }
+
+        // Guice can double provision instances (in particular with Optional binding).
+        // Protect against calling post-construct and pre-destroy methods more than once
+        if (startedInstances.putIfAbsent(identityHashCode(obj), obj) != null) {
```
wendigo, line 250: Can this be a set instead? do we need to track values?

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -39,6 +43,7 @@ public final class LifeCycleManager
 {
     private final Logger log = Logger.get(LifeCycleManager.class);
     private final AtomicReference<State> state = new AtomicReference<>(State.LATENT);
+    private final Map<Integer, Object> startedInstances = new ConcurrentHashMap<>();
```
wendigo, line 46:
```suggestion
    private final Set<Integer> startedInstances = new ConcurrentSkipListSet<>();
````

File: bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java (hunk 3 above)
Comment (Randgalt) [reply bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java:250]: I was concerned about not holding on to the reference. Are Object IDs re-used in the VM? I was worried that an object could get garbage collected and then its Object ID would get re-used resulting in a false positive for the map. Is that possible?

File: bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java [hunk 5]
```diff
@@ -233,7 +235,27 @@ private void startInstance(Object obj)
...
+        if (!hasPostConstructMethods && !hasPreDestroyMethod) {
+            // no need to track in startedInstances or managedInstances
+            return;
+        }
+
+        // Guice can double provision instances (in particular with Optional binding).
+        // Protect against calling post-construct and pre-destroy methods more than once
+        if (startedInstances.putIfAbsent(identityHashCode(obj), obj) != null) {
+            return;
+        }
```
wendigo, line 252:
```suggestion
        if (!startedInstances.add(identityHashCode(obj))) {
            return;
        }
```

wendigo:
Good catch @Randgalt. How did you found that?

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
Randgalt: Why call `methods.methodsFor(PostConstruct.class)` twice? Are you sure the optimizer will hoist this?

Randgalt:
> Good catch @Randgalt. How did you found that?

I have the need to use an Optional binder with an object that has `PostConstruct`. This object starts a thread and happens to check if it's already been started.

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
wendigo: this doesn't matter, the number of such annotated objects is small, I prefer readable and concise code here

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
wendigo: I don't think that this is possible

wendigo: The identityHashCode is derived from the memory address of the living object. In theory this could be reused but no two living objects can have the same address (obviously). Since the LifecycleManager lives as long as the objects that are managed by it, these will be garbage collected together so there is no risk that this will break.

Randgalt: Per offline conversation I've created `ConcurrentWeakIdentitySet` to manage the started instances. PTAL @wendigo

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -39,6 +39,7 @@ public final class LifeCycleManager
 {
     private final Logger log = Logger.get(LifeCycleManager.class);
     private final AtomicReference<State> state = new AtomicReference<>(State.LATENT);
+    private final ConcurrentWeakIdentitySet startedInstances = new ConcurrentWeakIdentitySet();
```
wendigo, line 42: isn't that enough:
`private final Set<Object> startedInstances = newSetFromMap(synchronizedMap(new WeakHashMap<>()));` ?

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
Randgalt: FYI - this bug DB entry (and a few others) explain that a JVM is not required to have unique IDs for objects: https://bugs.java.com/bugdatabase/view_bug?bug_id=JDK-8071434

wendigo (approved):

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
losipiuk: would work as long as you put "identity based" wrappers there. Not original objects.

Randgalt: I considered that but this was equivalent and simpler.

PR #1371, opened 2025-01-31
Add support for `Path` as configuration value type

ksobolew:
Just like support for `File` is indirect via matching a `String`-accepting constructor, support for `Path` is indirect via matching a static `of` factory method (available on `Path` since Java 11). The quirk is that this is a vararg method, which takes "at least one" `String`: `of(String, String...)`. So we have to treat such methods as `String`-accepting as well.

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
@@ -627,7 +627,7 @@ public class ConfigurationFactory
         if (stringAcceptingMethods.get("fromString") instanceof Method fromString) {
             if (type.isSubtypeOf(fromString.getGenericReturnType())) {
                 try {
-                    return fromString.invoke(null, value);
+                    return invokeFactory(fromString, value);
                 }
                 catch (ReflectiveOperationException e) {
                     return null;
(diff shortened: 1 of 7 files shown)


configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -683,7 +683,20 @@ private static Object coerce(TypeToken<?> type, String value)
...
+                }
+                catch (ReflectiveOperationException e) {
+                    return null;
+                }
+            }
+        }
+
+        // Look for a static of(String) method
+        Method of = stringAcceptingMethods.get("of");
+        if (of != null) {
```
wendigo, line 696: I think that for readability we can replace these with instanceof check like that:

```
       if (stringAcceptingMethods.get("of") instanceof Method of) {
            if (type.isSubtypeOf(of.getGenericReturnType())) {
                try {
                    return invokeFactory(of, value);
                }
                catch (ReflectiveOperationException e) {
                    return null;
                }
            }
        }
```

wendigo: This won't catch a null value. If you can apply this technique to other similar places in that class, that would be great.

wendigo (approved):

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
ksobolew: That's an interesting idea...

ksobolew: (We don't have Error Prone here, though)

PR #1373, opened 2025-02-01
Expand @Deprecated warning rendering in configuration

wendigo:

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
@@ -477,7 +477,7 @@ public class ConfigurationFactory
         }
 
         if (injectionPoint.getSetter().isAnnotationPresent(Deprecated.class)) {
-            problems.addWarning("Configuration property '%s' is deprecated and should not be used", prefix + injectionPoint.getProperty());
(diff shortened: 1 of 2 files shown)


configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -477,7 +477,19 @@ private <T> void setConfigProperty(T instance, AttributeMetadata attribute, Stri
...
-            problems.addWarning("Configuration property '%s' is deprecated and should not be used", prefix + injectionPoint.getProperty());
+            Deprecated deprecated = injectionPoint.getSetter().getAnnotation(Deprecated.class);
+
+            String deprecationNotice = "is deprecated";
+            if (!deprecated.since().isBlank()) {
+                deprecationNotice = deprecationNotice + " since " + deprecated.since();
+            }
+
+            if (deprecated.forRemoval()) {
+                problems.addWarning("Configuration property '%s' %s and will be removed in the future", prefix + injectionPoint.getProperty(), deprecationNotice);
```
electrum, line 488: Since the conditional is just changing the last past, I'd make it a ternary operation to pick the string, similar to how notice is handled. Then the rest of the warning code won't be duplicated.

PR #1374, opened 2025-02-03
Make BufferedHandler flushable

pettyjamesm:
Adds a new method `requestFullFlush()`, which allows a caller to request explicitly that all messages present in the buffer at the time of calling be flushed and written to the output and to be notified of that completion via a returned `CompletableFuture`

### DIFF
--- a/log-manager/src/main/java/io/airlift/log/BufferedHandler.java
+++ b/log-manager/src/main/java/io/airlift/log/BufferedHandler.java
@@ -4,7 +4,9 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.HashMultiset;
 import com.google.common.collect.ImmutableMultiset;
 import com.google.common.collect.Multiset;
+import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.RateLimiter;
+import com.google.common.util.concurrent.SettableFuture;
 import com.google.errorprone.annotations.ThreadSafe;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 import jakarta.annotation.Nullable;
@@ -28,6 +30,7 @@ import java.util.logging.LogRecord;
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.util.concurrent.Futures.nonCancellationPropagating;
 import static io.airlift.concurrent.Threads.daemonThreadsNamed;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Comparator.comparing;
@@ -41,7 +44,7 @@ import static java.util.logging.ErrorManager.WRITE_FAILURE;
 import static java.util.stream.Collectors.joining;
 
 @ThreadSafe
-class BufferedHandler
+public class BufferedHandler
         extends Handler
 {
     public interface DropSummaryFormatter
@@ -49,8 +52,9 @@ class BufferedHandler
         String formatDropSummary(Multiset<String> dropCountBySource);
     }
 
-    private static final MessageAndSource TERMINAL_MESSAGE = new MessageAndSource(new byte[] {}, "");
+    private static final byte[] EMPTY_BYTES = new byte[0];
 
+    private final MessageAndSource terminalMessage = new MessageAndSource(EMPTY_BYTES, "", SettableFuture.create());
     private final ExecutorService bufferDrainExecutor = newSingleThreadExecutor(daemonThreadsNamed("log-buffer-drainer"));
     private final AtomicBoolean initialized = new AtomicBoolean();
     private final AtomicBoolean inputClosed = new AtomicBoolean();
@@ -70,6 +74,10 @@ class BufferedHandler
     private final Multiset<String> dropCountBySource = HashMultiset.create();
     @GuardedBy("queueDrainLock")
     private boolean terminalMessageDequeued;
+    // This field holds any pending flush signal set on the most recently dequeued message
+    @Nullable
+    @GuardedBy("queueDrainLock")
+    private SettableFuture<Void> flushedSignal;
 
     public BufferedHandler(MessageOutput messageOutput, Formatter formatter, ErrorManager errorManager)
     {
@@ -129,6 +137,7 @@ class BufferedHandler
             try {
                 Multiset<String> dropSnapshot = ImmutableMultiset.of();
                 MessageAndSource message = null;
+                SettableFuture<Void> flushedSignal = null;
                 try {
                     // Extract work to do
                     queueDrainLock.lock();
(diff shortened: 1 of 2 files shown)


wendigo:
@pettyjamesm how this would be used for example in Trino to make sure that all of the loggers are flushed?

pettyjamesm:
I'm not sure how common this is going to be outside of our environment- but we have a use-case where we want to make sure that all log messages enqueued before a given point in time are fully written to the output, but without shutting down the logging system entirely (by closing the handlers).

In that situation, the usage would looks something like:
```
for (Handler handler : ROOT.getHandlers()) {
  if (handler instanceof BufferedHandler bufferedHandler) {
    bufferedHandler.flushFullyAndAwait(timeout, SECONDS);
  }
}
```

This is what necessitated making the class `public`.

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -368,6 +390,39 @@ public void flush()
...
+        synchronized (fullFlushMutex) {
+            queueDrainLock.lock();
+            try {
+                if (inputClosed.get() || terminalMessageDequeued) {
+                    return false;
+                }
+                if (!queue.offer(FLUSH_MESSAGE)) {
+                    return false;
+                }
+                return recordsFlushed.await(timeout, unit);
```
pettyjamesm, line 414: This needs to awake the background thread before awaiting the condition, will fix shortly

erichwang (commented):
It appears any call to flushFullyAndAwait  seems to block until any flush in the system happens, not necessarily from the context of the call that submitted it.

I would recommend doing this:
Change `MessageAndSource` to have an `Optional<CountDownLatch>` field that is only set when the request is a flush call (and there should be no payload set in those cases). When that specific flush message is dequeued, call `flushMessageOutputSafe` to flush the underlying medium, and  just call countDown() on the countdown latch to unblock any waiters on it.

pettyjamesm:
@erichwang - I've made the suggested changes, with some adaptation for edge case handling. Specifically, it made more sense to use `CompletableFuture<Void>` instead of `CountDownLatch` since there are some scenarios around termination and log message dropping where multiple flush listeners would need to be coalesced.

erichwang:
Had a chat with @dain who can help provide some suggestions here

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -70,6 +73,9 @@ public interface DropSummaryFormatter
     private final Multiset<String> dropCountBySource = HashMultiset.create();
     @GuardedBy("queueDrainLock")
     private boolean terminalMessageDequeued;
+    @Nullable
+    @GuardedBy("queueDrainLock")
+    private CompletableFuture<Void> flushedSignal;
```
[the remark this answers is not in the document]
pettyjamesm: Ok, updated the API to use `ListenableFuture` / `SettableFuture` instead.

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -368,6 +423,27 @@ public void flush()
         flushMessageOutputSafe();
     }
 
+    /**
+     * Requests a full flush of all currently queued messages from the background thread. If the handler has already been closed,
+     * this method will return an empty {@link Optional}. Otherwise, the returned {@link Optional} will contain a {@link CompletableFuture}
+     * that will complete when the background flush has completed.
+     */
+    public Optional<CompletableFuture<Void>> requestFullFlush()
```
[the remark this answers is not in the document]
erichwang: But maybe you can make it meaningful in trivial sense?

For example, if you give the terminal message a flush marker, then any flush requested after close can just default to the flush future of the terminal message? Does that satisfy what you are trying to do? In that case the call to flush would be correct right? Or are you trying to do something else?

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -368,6 +419,27 @@ public void flush()
         flushMessageOutputSafe();
     }
 
+    /**
+     * Requests a full flush of all currently queued messages from the background thread. If the handler has already been closed,
+     * this method will return an empty {@link Optional}. Otherwise, the returned {@link Optional} will contain a {@link ListenableFuture}
+     * that will complete when the background flush has completed.
+     */
+    public Optional<ListenableFuture<Void>> requestFullFlush()
```
dain, line 427: Why use an optional here.  In the empty case just return `Futures.getDoneFuture(null)`

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -238,15 +260,44 @@ private MessageAndSource queuePollFirst()
         }
 
         MessageAndSource message = queue.pollFirst();
```
dain, line 262: The `queuePollFirst` method should be annotated as `@GuardedBy("queueDrainLock")`

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -144,13 +152,17 @@ private void bufferDrainLoop()
                             // Note: queuePollFirst() will return null and set terminalMessageDequeued when encountering TERMINAL_MESSAGE.
                             message = queuePollFirst();
 
-                            if (!dropSnapshot.isEmpty() || message != null) {
+                            flushedSignal = this.flushedSignal;
+                            this.flushedSignal = null;
```
dain, line 156: add a comment saying that this code will handle the flush signal, so move it to a local variable.

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -70,6 +74,9 @@ public interface DropSummaryFormatter
     private final Multiset<String> dropCountBySource = HashMultiset.create();
     @GuardedBy("queueDrainLock")
     private boolean terminalMessageDequeued;
+    @Nullable
+    @GuardedBy("queueDrainLock")
+    private SettableFuture<Void> flushedSignal;
```
dain, line 79: Add comment saying that is is a pending flush signal for a dequeued message.

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -368,6 +419,27 @@ public void flush()
...
+    public Optional<ListenableFuture<Void>> requestFullFlush()
+    {
+        queueDrainLock.lock();
+        try {
+            if (inputClosed.get() || terminalMessageDequeued) {
+                return Optional.empty();
+            }
+            SettableFuture<Void> flushedSignal = SettableFuture.create();
+            queueInsert(new MessageAndSource(EMPTY_BYTES, "", flushedSignal));
+            return Optional.of(flushedSignal);
```
dain, line 436: this should be wrapped in `Futures.nonCancellationPropagating(flushedSignal)` to prevent a caller from disrupting a future shared by another caller (since they are linked with `setFuture`)

dain (commented):
Some comments.

Eric suggest that a way to clean up this code would be to give the terminal message a flush marker too, and then any flush submitted after the terminal message just defaults to the terminal message flush marker

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
pettyjamesm: The idea was that the caller could distinguish between a flush that was respected (and processed immediately) vs a flush that was ignored, but I guess the question would be "what are you gonna do when you realize your flush was ignored, log something?" - I've removed the `Optional` wrapping.

pettyjamesm:
Updated the PR based on feedback, thanks @dain and @erichwang

log-manager/src/main/java/io/airlift/log/BufferedHandler.java
```diff
@@ -368,13 +416,39 @@ public void flush()
...
+    {
+        SettableFuture<Void> flushedSignal;
+        queueDrainLock.lock();
+        try {
+            if (inputClosed.get() || terminalMessageDequeued) {
+                flushedSignal = requireNonNull(terminalMessage.flushSignal(), "terminalMessage flush signal must not be null");
+            }
+            else {
+                flushedSignal = SettableFuture.create();
+                queueInsert(new MessageAndSource(EMPTY_BYTES, "", flushedSignal));
```
dain, line 434: This calls `queueInsert` while holding the lock.  That method does a notify outside of the lock, so I think we need to do the insert outside of the drain lock.

dain (approved):

PR #1377, opened 2025-02-08
Disable ENV interpolation during bootstrap

wendigo:

### DIFF
--- a/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
@@ -78,6 +78,7 @@ public class Bootstrap
     private Map<String, String> requiredConfigurationProperties;
     private Map<String, String> optionalConfigurationProperties;
     private boolean initializeLogging = true;
+    private boolean envInterpolation = true;
     private boolean quiet;
     private boolean loadSecretsPlugins;
     private boolean skipErrorReporting;
@@ -137,6 +138,11 @@ public class Bootstrap
         return withInitializeLogging(false);
     }
 
+    public Bootstrap disableEnvInterpolation()
+    {
+        return withEnvInterpolation(false);
+    }
+
     public Bootstrap quiet()
     {
         return withQuiet(true);
@@ -176,6 +182,12 @@ public class Bootstrap
         return this;
     }
 
+    public Bootstrap withEnvInterpolation(boolean interpolate)
+    {
+        this.envInterpolation = interpolate;
+        return this;
+    }
+
     /**
      * Validate configuration and return used properties.
      */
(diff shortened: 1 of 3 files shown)


jklamer (approved):

configuration/src/main/java/io/airlift/configuration/secrets/NoOpSecretsResolver.java
```diff
@@ -0,0 +1,25 @@
...
+{
+    public NoOpSecretsResolver()
+    {
+        super(Map.of());
+    }
+
+    @Override
+    public Map<String, String> getResolvedConfiguration(Map<String, String> properties)
+    {
+        return properties;
```
electrum, line 17: I don't think this is the correct behavior. If we enable ENV processing, then we will fail for any other provider. With this `NoOpSecretsResolver`, we disable interpolation entirely, including for ENV. Given that this is a backwards incompatible change, it seems better to fail for all providers.

So instead of `NoOpSecretsResolver`, we can use `new SecretsResolver(Map.of())`.

bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrap.java
```diff
@@ -225,6 +225,33 @@ public void testSkipErrorReporting()
...
     }
 
+    @Test
+    public void testEnableEnvInterpolation()
+    {
+        FooConfig interpolatedConfig = new Bootstrap(binder -> {
+            configBinder(binder).bindConfig(FooConfig.class);
+            binder.bind(FooInstance.class).asEagerSingleton();
+        })
+                .setOptionalConfigurationProperty("foo.password", "${ENV:FOO_PASSWORD}")
```
oskar-szwajkowski, line 235: could you add a test with `${env:FOO_PASSWORD}` as well?

as it seems that resolver is case insensitive actually, which could be not expected but test will present it

wendigo: done

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -78,6 +78,7 @@ private enum State { UNINITIALIZED, CONFIGURED, INITIALIZED }
     private Map<String, String> requiredConfigurationProperties;
     private Map<String, String> optionalConfigurationProperties;
     private boolean initializeLogging = true;
+    private boolean envInterpolation;
```
martint, line 81: This should be enabled by default to maintain backward compatibility. Environment interpolation is a feature of Airlift that's useful when building services (the primary use case). Any other cases should explicitly opt out.

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -189,7 +201,8 @@ public Set<ConfigPropertyMetadata> configure()
             logging = Logging.initialize();
         }
 
-        this.secretsResolver = new SecretsResolver(ImmutableMap.of("env", new EnvironmentVariableSecretProvider()));
+        this.secretsResolver = envInterpolation ? new SecretsResolver(ImmutableMap.of("env", new EnvironmentVariableSecretProvider())) :
+                new SecretsResolver(ImmutableMap.of());
```
martint, line 205: Format like this:

```suggestion
        this.secretsResolver = envInterpolation ? 
                new SecretsResolver(ImmutableMap.of("env", new EnvironmentVariableSecretProvider())) :
                new SecretsResolver(ImmutableMap.of());
```

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
wendigo: Done

martint (approved):

PR #1384, opened 2025-02-13
Use SdkTracerProvider to create a Tracer

wendigo:


tracing/src/main/java/io/airlift/tracing/OpenTelemetryModule.java
```diff
@@ -101,9 +102,12 @@ public SdkTracerProvider createTracerProvider(NodeInfo nodeInfo, Set<SpanProcess
 
     @Provides
     @Singleton
-    public Tracer createTracer(OpenTelemetry openTelemetry)
+    public Tracer createTracer(Set<SpanProcessor> spanProcessors, SdkTracerProvider tracerProvider)
     {
-        return openTelemetry.getTracer(serviceName);
+        if (spanProcessors.isEmpty()) {
+            return noopTracer();
+        }
```
martint, line 109: This doesn't look like just a cleanup. What's the difference in behavior with the old version?

wendigo: None. It's just more direct. In both versions noop tracer will be created on empty span processors

martint (approved):

PR #1390, opened 2025-02-19
Make command line log output colored if supported

wendigo:
This assigns different colors to different log levels and is only supported if we are running in the interactive, non-dumb terminal with NO_COLORS unset.

### DIFF
--- a/log-manager/src/main/java/io/airlift/log/TerminalColors.java
+++ b/log-manager/src/main/java/io/airlift/log/TerminalColors.java
@@ -1,0 +1,109 @@
+package io.airlift.log;
+
+import java.io.PrintWriter;
+
+import static io.airlift.log.TerminalColors.Color.BLUE;
+import static io.airlift.log.TerminalColors.Color.GREEN;
+import static io.airlift.log.TerminalColors.Color.RED;
+import static io.airlift.log.TerminalColors.Color.WHITE;
+import static io.airlift.log.TerminalColors.Color.YELLOW;
+
+public class TerminalColors
+{
+    private static final boolean isColorSupported = detectIsColorSupported();
+    private static final String ANSI_RESET = "\033[0m";
+
+    private TerminalColors() {}
+
+    public enum Color
+    {
+        WHITE("\033[37m"),
+        RED("\033[31m"),
+        GREEN("\033[32m"),
+        YELLOW("\033[33m"),
+        BLUE("\033[34m"),
+        PURPLE("\033[35m"),
+        CYAN("\033[36m"),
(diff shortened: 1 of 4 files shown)


wendigo:
It allows for easier visual inspection of the logged messages

Screenshot 2025-02-19 at 10 41 49

wendigo:
Screenshot 2025-02-19 at 10 47 53

log-manager/src/main/java/io/airlift/log/TerminalColors.java
```diff
@@ -0,0 +1,109 @@
...
+            {
+                writer.close();
+            }
+        };
+    }
+
+    private static boolean detectIsColorSupported()
+    {
+        // Non-interactive terminal
+        if (System.console() == null) {
```
martint, line 91: What happens when there is a terminal that satisfies all the "color supported" conditions below but logs are being written to a file?

log-manager/src/main/java/io/airlift/log/TerminalColors.java
```diff
@@ -0,0 +1,109 @@
...
+    public static String colored(String text, Level level)
+    {
+        return switch (level) {
+            case OFF -> text;
+            case TRACE -> colored(text, WHITE);
+            case DEBUG -> colored(text, BLUE);
+            case INFO -> colored(text, GREEN);
+            case WARN -> colored(text, YELLOW);
+            case ERROR -> colored(text, RED);
+        };
```
martint, line 59: This choice of colors may not be appropriate for some terminal color themes.

martint (commented):
Nice! I like the idea.

log-manager/src/main/java/io/airlift/log/TerminalColors.java
wendigo: I took the colors from the recommended ones for command line tools. Maven uses the same set of colors for example

log-manager/src/main/java/io/airlift/log/TerminalColors.java
wendigo: If you redirect the stream System.console will be null

martint: This line (https://github.com/airlift/airlift/blob/41ca358713a915be92d3c188470544caac5b11aa/log-manager/src/main/java/io/airlift/log/Logging.java#L287), creates a `StaticFormatter` if format is configured to TEXT.

wendigo: Done, fixed

martint (approved):

PR #1395, opened 2025-02-21
Include error detail when config class can't be created

wendigo:

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
@@ -472,7 +472,7 @@ public class ConfigurationFactory
             if (e instanceof InvocationTargetException && e.getCause() != null) {
                 e = e.getCause();
             }
-            throw exceptionFor(e, "Error creating instance of configuration class [%s]", configurationMetadata.getConfigClass().getName());
+            throw exceptionFor(e, "Error creating instance of configuration class [%s], caused by %s: %s", configurationMetadata.getConfigClass().getName(), e.getClass().getSimpleName(), e.getMessage());
         }
     }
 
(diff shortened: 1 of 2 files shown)


lshrinivas (approved):

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -472,7 +472,7 @@ private static <T> T newInstance(ConfigurationMetadata<T> configurationMetadata)
             if (e instanceof InvocationTargetException && e.getCause() != null) {
                 e = e.getCause();
             }
-            throw exceptionFor(e, "Error creating instance of configuration class [%s]", configurationMetadata.getConfigClass().getName());
+            throw exceptionFor(e, "Error creating instance of configuration class [%s], caused by %s: %s", configurationMetadata.getConfigClass().getName(), e.getClass().getSimpleName(), e.getMessage());
```
martint, line 475: The cause should already be recorded in the underlying cause of the exception, so this should be unnecessary. Is that not working as expected?

wendigo: @martint it's not. Guice will not display this cause when formatting errors

wendigo: This is due to a strange condition that if the underlying root cause is a single error, cause won't be rendered. I can point to the relevant code. This makes it hard to understand why configuration has failed to build due to a generic, non actionable error message

martint: > that if the underlying root cause is a single error, cause won't be rendered

Sounds like we should fix that.

wendigo: @martint within guice. We can copy whole formatting logic

PR #1409, opened 2025-03-11
Extract configuration testing package to a standalone module

nineinchnick:
Thanks to this, transitive testing dependencies are not added as runtime dependencies, especially the relatively large (9MB) ByteBuddy library.

I tested this manually in Trino in this branch: https://github.com/trinodb/trino/compare/master...nineinchnick:trino:config-deps

* It speeds up the build by 10s (2m 16s vs 2m 26s). I guess because there's less copying.
* It also makes the final artifact slightly smaller, the `core/trino-server/target/trino-server-473-SNAPSHOT-hardlinks` directory shrinks from 887M to 870M. It removes all occurrences of `net.bytebuddy_byte-buddy-1.17.2.jar` from the final artifact. Trino uses hardlinks, it shouldn't affect the size much.
* It also does **not** change the startup speed. I checked it by running the Trino container locally. It starts in around 7s and uses about 700MiB of memory right after startup, on both master and the other branch.


nineinchnick:
@wendigo @electrum PTAL, you can close this PR if you think this is not needed.

bootstrap/pom.xml
```diff
@@ -54,6 +54,12 @@
         </dependency>
 
         <!-- for testing -->
+        <dependency>
+            <groupId>io.airlift</groupId>
+            <artifactId>configuration-testing</artifactId>
```
nineinchnick, line 59: I can't move this to the `testing` package because it's used by the `configuration` package, and it would create a circular dependency.

SemionPar (approved):

wendigo:
Alternatively this class could be moved to test-jar which doesn't require a new module: https://github.com/airlift/airlift/pull/1410

electrum:
Test-jar is a hack that doesn’t work well for consumers as it can’t have separate dependencies. Using a different module is better.

wendigo:
We already have a testing package in airlift

wendigo (approved):

PR #1410, opened 2025-03-11
Move ConfigAssertions to test-jar

wendigo:


wendigo:
This makes modules pulling configuration independent of byte-buddy that is used only in testing.

PR #1437, opened 2025-04-02
Allow binding SslContextFactory only to a specific HTTP client via annotation

xkrogen:
Currently `HttpClientBinder` allows for the ability for users to bind their own `SslContextFactory.Client` to control the SSL behavior of the client. However, unlike most of the relevant bindings, the `SslContextFactory` is only pulled from a binding _without_ any annotation, so all `HttpClient`s created by the binder must share the same `SslContextFactory`. This is not always desirable; sometimes it is needed for different clients to have different SSL contexts.

This PR retains the current behavior of pulling the `SslContextFactory` from any binding without the annotation, but makes it so that if there is a binding present _with_ the annotation, it will override the non-annotated version.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
@@ -70,8 +70,10 @@ public class HttpClientModule
         // bind the configuration
         configBinder(binder).bindConfig(HttpClientConfig.class, annotation, name);
 
-        // Allow users to bind their own SslContextFactory
+        // Allow users to bind their own SslContextFactory, pulling a globally-bound
+        // SslContextFactory if one is not bound with the specific annotation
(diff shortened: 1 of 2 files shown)


xkrogen:
It looks like I don't have the ability to set labels here, so I can't satisfy the third point of the checklist.

cc @wendigo @martint

wendigo:
@xkrogen what's the use case here?

xkrogen:
Generally, the use case is that you may need different SSL contexts for different clients. For example in our environment, some endpoints are signed using a standard public CA, and some endpoints are signed using a private internal CA, and we use different truststores when accessing those different endpoints. Similarly you may need to present different certificates (different keystores) to use different identities depending on which endpoint you are talking to.

More concretely, we are leveraging this capacity in the Trino Gateway to use different auth depending on which source we are communicating with for metrics gathering, query proxying, etc.

wendigo (approved):

PR #1441, opened 2025-04-09
Allow NodeInfo annotations to be set via a comma separated key value string

ikashperskyi:
Adds an easy way of creating NodeInfo in memory without needing a physical file for annotations. Useful when having more than one Metrics endpoint with different labels exposed.

### DIFF
--- a/node/src/main/java/io/airlift/node/NodeConfig.java
+++ b/node/src/main/java/io/airlift/node/NodeConfig.java
@@ -15,15 +15,19 @@
  */
 package io.airlift.node;
 
+import com.google.common.base.Splitter;
 import com.google.common.net.InetAddresses;
 import io.airlift.configuration.Config;
 import io.airlift.configuration.DefunctConfig;
 import io.airlift.configuration.LegacyConfig;
 import io.airlift.configuration.validation.FileExists;
+import jakarta.validation.constraints.AssertTrue;
 import jakarta.validation.constraints.NotNull;
 import jakarta.validation.constraints.Pattern;
 
 import java.net.InetAddress;
+import java.util.Map;
+import java.util.Optional;
 
 @DefunctConfig({"http-server.ip", "jetty.ip"})
 public class NodeConfig
@@ -34,6 +38,9 @@ public class NodeConfig
     public static final String ENV_REGEXP_ERROR = "should match " + ENV_REGEXP;
     public static final String POOL_REGEXP = "[a-z0-9][_a-z0-9]*";
     public static final String POOL_REGEXP_ERROR = "should match " + POOL_REGEXP;
+    public static final Splitter.MapSplitter ANNOTATION_SPLITTER = Splitter.on(",")
+            .omitEmptyStrings()
+            .withKeyValueSeparator("=");
 
     private String environment;
     private String pool = "general";
@@ -47,6 +54,7 @@ public class NodeConfig
     private AddressSource internalAddressSource = AddressSource.IP;
     private String annotationFile;
     private boolean preferIpv6Address = "true".equalsIgnoreCase(System.getenv("java.net.preferIPv6Address"));
+    private Map<String, String> annotations;
 
     @NotNull
     @Pattern(regexp = ENV_REGEXP, message = ENV_REGEXP_ERROR)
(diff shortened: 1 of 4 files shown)


node/src/main/java/io/airlift/node/NodeConfig.java
```diff
@@ -46,6 +46,7 @@ public class NodeConfig
     private String configSpec;
     private AddressSource internalAddressSource = AddressSource.IP;
     private String annotationFile;
+    private String annotationString;
```
wendigo, line 49: just: `annotations`

node/src/main/java/io/airlift/node/NodeConfig.java
```diff
@@ -201,4 +202,16 @@ public NodeConfig setAnnotationFile(String annotationFile)
         this.annotationFile = annotationFile;
         return this;
     }
+
+    public String getAnnotationString()
```
wendigo, line 206: `getAnnotations`

node/src/main/java/io/airlift/node/NodeConfig.java
```diff
@@ -201,4 +202,16 @@ public NodeConfig setAnnotationFile(String annotationFile)
...
+    {
+        return annotationString;
+    }
+
+    @Config("node.annotation-string")
```
wendigo, line 211: `node.annotations`

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -140,6 +143,10 @@ public NodeInfo(String environment,
             this.externalAddress = this.internalAddress;
         }
 
+        if (annotationFile != null && annotationString != null) {
+            throw new IllegalArgumentException("Only one of annotationFile or annotationString should be set, but not both.");
```
wendigo, line 147: This should be validated in the config class itself:

```
@AssertTrue("Only one of node.annotations or node.annotations-file can be set")
public boolean isConfigurationValid()
{
   return (annotationFile != null) ^ (annotations != null);
}
```

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -148,6 +155,9 @@ public NodeInfo(String environment,
                 throw new UncheckedIOException(e);
             }
         }
+        else if (annotationString != null) {
+            this.annotations = ImmutableMap.copyOf(replaceEnvironmentVariables(Splitter.on(",").withKeyValueSeparator("=").split(annotationString)));
```
wendigo, line 159: Splitter should be moved to configuration class and used there. Field should be `Map<String, String>` instead of a `String`.

`replaceEnvironmentVariables` is redundant - it's called by the Bootstrap. Remove

ikashperskyi: In that case reading and parsing the file should also be done by the config object for consistency.

ikashperskyi: Configs seem to be designed with one input translating to one output. 

With that in mind it makes sense to change NodeInfo annotations input type to Map and move the file loading to the config file.

node/src/main/java/io/airlift/node/NodeInfo.java
brybacki: What if someone creates NodeInfo on the fly, without going through config? I think constructor is the best place for such precondition checks.

brybacki:
What is the intended use case? Are you going to create a specific NodeInfo on the fly from default template NodeInfo?

node/src/main/java/io/airlift/node/NodeInfo.java
electrum: They should be checked in both places

node/src/main/java/io/airlift/node/NodeInfo.java
electrum: I can see how this seems inconsistent, but there's a different way to look at this, which I hope helps you understand how we think about this.

The purpose of the config class is to parse and validate the config properties. For the map property, the property is a map, so the parsing and validation belongs in the config class. For the file property, the property is the path to a file, so the config class validates that the file exists. At that point, the config class has completed its job: it provided a valid path to a file. Doing something with the file is outside the scope of the config property system.

PR #1496, opened 2025-07-23
Add some extension points to SecretsResolver

ksobolew:
* Extract a method to get a single secret value
  * This is a narrower version of (private) `resolveConfiguration` and the (public) `getResolvedConfiguration`; the latter two take a value or a whole `Map` of values with potentially many secrets references, parse them and invoke a relevant secrets provider for each encountered reference. The new method explicitly takes the provider name and a single key, which is much more focused and also sometimes useful.
* Provide a way to use a different source of secrets resolvers
  * We don't always have the ability to package the secrets resolvers into a neat `Map` - sometimes we'd like to decide what to do only once we encounter a specific provider name.


configuration/src/main/java/io/airlift/configuration/secrets/SecretsResolver.java
```diff
@@ -30,11 +31,16 @@ public class SecretsResolver
 {
     private static final Pattern PATTERN = Pattern.compile("\\$\\{([a-zA-Z][a-zA-Z0-9_-]*):(?<key>[^}]+?)}");
 
-    private final Map<String, SecretProvider> secretProviders;
+    private final Function<String, SecretProvider> secretProvidersFactory;
 
     public SecretsResolver(Map<String, SecretProvider> secretProviders)
     {
-        this.secretProviders = ImmutableMap.copyOf(requireNonNull(secretProviders, "secretProviders is null"));
+        this(ImmutableMap.copyOf(requireNonNull(secretProviders, "secretProviders is null"))::get);
```
wendigo: nit: there is no need for rnn check here - ImmutableMap.copyOf checks for null

ksobolew: It does. It's a matter of a more or less informative error message, mostly. But the odds of it mattering are low :)

ksobolew: Removed

PR #1540, opened 2025-09-08
Expose Jetty client mbeans

wendigo:

### DIFF
--- a/http-server/src/main/java/io/airlift/http/server/HttpServer.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpServer.java
@@ -123,7 +123,7 @@ public class HttpServer
             boolean enableLegacyUriCompliance,
             boolean enableCaseSensitiveHeaderCache,
             ClientCertificate clientCertificate,
-            MBeanServer mbeanServer,
+            Optional<MBeanServer> mbeanServer,
             Optional<SslContextFactory.Server> maybeSslContextFactory,
             Optional<ByteBufferPool> byteBufferPool)
             throws IOException
@@ -135,6 +135,7 @@ public class HttpServer
         requireNonNull(servlet, "servlet is null");
         requireNonNull(maybeSslContextFactory, "maybeSslContextFactory is null");
         requireNonNull(clientCertificate, "clientCertificate is null");
+        requireNonNull(mbeanServer, "mbeanServer is null");
 
         checkArgument(!config.isHttpsEnabled() || maybeHttpsConfig.isPresent(), "httpsConfig must be present when HTTPS is enabled");
 
@@ -165,9 +166,9 @@ public class HttpServer
 
         this.sslContextFactory = maybeSslContextFactory;
 
-        if (mbeanServer != null) {
+        if (mbeanServer.isPresent()) {
             // export jmx mbeans if a server was provided
-            MBeanContainer mbeanContainer = new MBeanContainer(mbeanServer);
+            MBeanContainer mbeanContainer = new MBeanContainer(mbeanServer.orElseThrow());
             server.addBean(mbeanContainer);
         }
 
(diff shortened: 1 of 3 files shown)


http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -777,6 +778,19 @@ public HttpClientConfig setUseVirtualThreads(boolean useVirtualThreads)
         return this;
     }
 
+    public boolean isExposeJettyMBeans()
+    {
+        return exposeJettyMBeans;
+    }
+
+    @ConfigHidden
+    @Config("http-client.expose-jetty-mbeans")
```
dain: Don't we normally we name these something like `xyz.feature.enabled`?  So in this case it would be `http-client.mbean.enabled`.  I'm not sure I  line the name "mbean" for the feature either.  @electrum, thoughts?

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -238,6 +240,7 @@ public JettyHttpClient(
             Iterable<? extends HttpRequestFilter> requestFilters,
             OpenTelemetry openTelemetry,
             Tracer tracer,
+            MBeanServer mbeanServer,
```
dain: Can we make this `Optional<MBeanServer>`?

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -299,6 +303,13 @@ protected void configure(SelectableChannel selectable)
 
         httpClient.setConnectBlocking(config.isConnectBlocking());
 
+        if (config.isExposeJettyMBeans() && mbeanServer != null) {
```
dain: Should it be an error, if you asked for mbeans to be enabled and the server is not bound?

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
wendigo: It's nullable in HttpServer. I wanted to keep it consistent. I can make it optional here and there

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
wendigo: Makes sense!

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
martint: +1

Also, why wouldn’t this always be enabled? More configuration options, more complexity.

wendigo: @martint this exposes a lot of information including potentially sensitive one. I can expose it by default without a toggle. Thoughts?

wendigo:
@martint ptal

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -299,6 +303,13 @@ protected void configure(SelectableChannel selectable)
 
         httpClient.setConnectBlocking(config.isConnectBlocking());
 
+        if (mbeanServer.isPresent()) {
+            // export jmx mbeans if a server was provided
+            MBeanContainer mbeanContainer = new MBeanContainer(mbeanServer.orElseThrow());
+            mbeanContainer.setDomain("org.eclipse.jetty.http.client." + name);
```
martint: Where does `"org.eclipse.jetty.http.client." + name` show up? How does it affect the mbean names?

wendigo: CleanShot 2025-09-09 at 12 29 27@2x

wendigo: Alternative is: 

CleanShot 2025-09-09 at 12 35 37@2x

wendigo: I'm checking whether I can name a client (instead of `oejc_HttpClient@identityHashCode`)

wendigo: It seems I can: 
CleanShot 2025-09-09 at 12 37 09@2x

PR #1543, opened 2025-09-09
Improve TimeDistribution congestion

wendigo:

Supersedes https://github.com/airlift/airlift/pull/323

### DIFF
--- a/stats/src/main/java/io/airlift/stats/TimeDistribution.java
+++ b/stats/src/main/java/io/airlift/stats/TimeDistribution.java
@@ -175,14 +175,6 @@ public class TimeDistribution
         return merged;
     }
 
-    private double convertToUnit(long nanos)
-    {
-        if (nanos == Long.MAX_VALUE || nanos == Long.MIN_VALUE) {
-            return Double.NaN;
-        }
-        return nanos * 1.0 / unit.toNanos(1);
-    }
-
     private double convertToUnit(double nanos)
     {
         return convertToUnit(nanos, (double) unit.toNanos(1));


wendigo:
cc @martint

stats/src/main/java/io/airlift/stats/TimeDistribution.java
```diff
@@ -140,6 +154,26 @@ public Map<Double, Double> getPercentiles()
         return result;
     }
 
+    private DecayTDigest mergeAndGetIfNeeded()
+    {
+        long now = System.nanoTime();
+        if (now - lastMerge > MERGE_THRESHOLD_NANOS) {
+            synchronized (this) {
+                lastMerge = now;
```
raunaqmorarka: can we set this after the loop, to avoid including the merge time here ?

stats/src/main/java/io/airlift/stats/TimeDistribution.java
```diff
@@ -166,6 +200,7 @@ public TimeDistributionSnapshot snapshot()
         double max;
         double[] quantiles;
         synchronized (this) {
+            DecayTDigest digest = mergeAndGetIfNeeded();
```
raunaqmorarka: should we force a merge here to avoid missing latest updates ?
i think a user of this method may expect latest state

stats/src/main/java/io/airlift/stats/TimeDistribution.java
```diff
@@ -140,6 +154,26 @@ public Map<Double, Double> getPercentiles()
         return result;
     }
 
+    private DecayTDigest mergeAndGetIfNeeded()
+    {
+        long now = System.nanoTime();
+        if (now - lastMerge > MERGE_THRESHOLD_NANOS) {
```
martint: This changes semantics. All the methods that return stats will now be stale up to 1s. If we're going to do this, the staleness period needs to be *much* lower.

wendigo: I've based that on your previous PR from 10 years back :)

stats/src/main/java/io/airlift/stats/TimeDistribution.java
wendigo: Makes sense, added!

stats/src/main/java/io/airlift/stats/TimeDistribution.java
wendigo: Is 100 ms good enough?

stats/src/main/java/io/airlift/stats/Distribution.java
```diff
@@ -176,16 +175,6 @@ public Map<Double, Double> getPercentiles()
         return result;
     }
 
-    public synchronized List<Double> getPercentiles(List<Double> percentiles)
```
losipiuk: How to you know those are not used give this is a library :)

stats/src/main/java/io/airlift/stats/TimeDistribution.java
wendigo: Done!

stats/src/main/java/io/airlift/stats/Distribution.java
wendigo: I've checked the usage sites

raunaqmorarka (approved):

stats/src/main/java/io/airlift/stats/TimeDistribution.java
```diff
@@ -27,9 +30,13 @@ public class TimeDistribution
     }
 
     private final double alpha;
+    private final Object[] locks = new Object[STRIPES];
+    @GuardedBy("locks")
```
losipiuk, line 32: I understand the intention - but I doubt it will be properly understood by tools like intellij/errorprone

stats/src/main/java/io/airlift/stats/TimeDistribution.java
```diff
@@ -140,6 +155,30 @@ public Map<Double, Double> getPercentiles()
         return result;
     }
 
+    private DecayTDigest mergeAndGetIfNeeded()
+    {
+        return mergeAndGetIfNeeded(false);
+    }
+
+    private DecayTDigest mergeAndGetIfNeeded(boolean forceMerge)
```
losipiuk, line 159: Maybe you want to have this more complex implementation separate. Contention is not always a big deal. And probably you do not want to pay for this extra complexity if it is not.

stats/src/main/java/io/airlift/stats/TimeDistribution.java
wendigo: It's recognized by errorprone :)

wendigo:
cc @martint

stats/src/main/java/io/airlift/stats/TimeDistribution.java
```diff
@@ -52,63 +60,70 @@ public TimeDistribution(double alpha, TimeUnit unit)
...
+        }
         total = new DecayCounter(alpha);
         this.unit = unit;
     }
 
-    public synchronized void add(long value)
+    public void add(long value)
     {
-        digest.add(value);
+        int segment = identityHashCode(Thread.currentThread()) % STRIPES;
```
martint: Why not just base it on thread ID?

wendigo: I'll revert to threadId

PR #1545, opened 2025-09-10
Fix TimeDistribution partial handling

wendigo:
1) Do not reset merged state unless explicit reset() is called 2) Merge while getting avg
3) Delay merging until initial delay is passed
4) Add test

### DIFF
--- a/stats/src/main/java/io/airlift/stats/TimeDistribution.java
+++ b/stats/src/main/java/io/airlift/stats/TimeDistribution.java
@@ -1,5 +1,7 @@
 package io.airlift.stats;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Ticker;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 import org.weakref.jmx.Managed;
 
@@ -7,6 +9,7 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
+import static com.google.common.base.Ticker.systemTicker;
 import static com.google.common.base.Verify.verify;
 import static java.lang.Math.floorMod;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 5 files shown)


stats/src/main/java/io/airlift/stats/TimeDistribution.java
```diff
@@ -160,14 +164,15 @@ private DecayTDigest mergeAndGetIfNeeded(boolean forceMerge)
...
         synchronized (this) {
             if (forceMerge || System.nanoTime() - lastMerge > MERGE_THRESHOLD_NANOS) {
-                merged = new DecayTDigest(TDigest.DEFAULT_COMPRESSION, alpha);
                 for (int i = 0; i < STRIPES; i++) {
                     synchronized (locks[i]) {
                         merged.merge(partials[i]);
                         // Reset the partial
                         partials[i] = new DecayTDigest(TDigest.DEFAULT_COMPRESSION, alpha);
                     }
                 }
```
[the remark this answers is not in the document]
wendigo: That's the whole point dear @copilot ;)

stats/src/main/java/io/airlift/stats/TimeDistribution.java
```diff
@@ -159,15 +169,16 @@ private DecayTDigest mergeAndGetIfNeeded()
     private DecayTDigest mergeAndGetIfNeeded(boolean forceMerge)
     {
         synchronized (this) {
-            if (forceMerge || System.nanoTime() - lastMerge > MERGE_THRESHOLD_NANOS) {
-                merged = new DecayTDigest(TDigest.DEFAULT_COMPRESSION, alpha);
+            if (forceMerge || mergeThreshold == MERGE_IMMEDIATE || System.nanoTime() - lastMerge > mergeThreshold) {
```
martint: You can simplify the expression to `System.nanoTime() - lastMerge >= mergeThreshold` and get rid of `mergeThreshold == MERGE_IMMEDIATE`

stats/src/main/java/io/airlift/stats/TimeDistribution.java
```diff
@@ -35,7 +38,9 @@ public class TimeDistribution
     private DecayTDigest merged;
     @GuardedBy("this")
     private long lastMerge;
+    private final long mergeThreshold;
```
martint: Am I interpreting it right that this is just for testing purposes? In that case, use a Ticker to represent time instead of System.nanoTime() and make sure the tests pass an appropriate ticker and advance time as necessary.

wendigo: To merge immediate when getting values too. But maybe it is not needed

stats/src/main/java/io/airlift/stats/TimeDistribution.java
wendigo: Done

stats/src/main/java/io/airlift/stats/TimeDistribution.java
wendigo: I've passed ticker instead which makes it more testable :)

PR #1552, opened 2025-09-12
Allocate partials only when needed

wendigo:

### DIFF
--- a/stats/src/main/java/io/airlift/stats/TimeDistribution.java
+++ b/stats/src/main/java/io/airlift/stats/TimeDistribution.java
@@ -11,6 +11,7 @@ import java.util.concurrent.TimeUnit;
 
 import static com.google.common.base.Ticker.systemTicker;
 import static com.google.common.base.Verify.verify;
+import static java.lang.Math.clamp;
 import static java.lang.Math.floorMod;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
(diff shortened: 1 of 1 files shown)


stats/src/main/java/io/airlift/stats/TimeDistribution.java
```diff
@@ -22,7 +22,7 @@ public class TimeDistribution
     static final long MERGE_THRESHOLD_NANOS = MILLISECONDS.toNanos(100);
     private static final double[] SNAPSHOT_QUANTILES = new double[] {0.5, 0.75, 0.9, 0.95, 0.99};
     private static final double[] PERCENTILES;
-    private static final int STRIPES = 16;
+    private static final int STRIPES = 8;
```
raunaqmorarka: What's the rationale for decreased number of stripes ?

wendigo: Memory usage - we now need to keep at most N stripes + 1 TDigests

raunaqmorarka: Can we make this `Runtime.getRuntime().availableProcessors()` ?

wendigo: We can I guess

wendigo: I've clamped it to [2, 16] range

PR #1583, opened 2025-10-02
Add option to preserve content decoder

oneonestar:
Add option to preserve content decoder.

Airlift Http Client by default disable compression by removing content decoder factories.
In Trino Gateway, we are forwarding request from client to Trino using Airlift Http Client.
We would like to preserve the encoding and allow the communication between Client, Gateway and Trino to use compression.

Relates to: https://github.com/trinodb/trino-gateway/pull/765

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
@@ -93,6 +93,7 @@ public class HttpClientConfig
     private Optional<Duration> tcpKeepAliveIdleTime = Optional.empty();
     private boolean strictEventOrdering;
     private boolean useVirtualThreads;
+    private boolean preserveContentDecoderFactories;
 
     /**
      * This property is initialized with Jetty's default excluded ciphers list.
@@ -774,6 +775,19 @@ public class HttpClientConfig
     public HttpClientConfig setUseVirtualThreads(boolean useVirtualThreads)
     {
         this.useVirtualThreads = useVirtualThreads;
+        return this;
+    }
+
+    public boolean isPreserveContentDecoderFactories()
+    {
+        return preserveContentDecoderFactories;
+    }
+
+    @ConfigHidden
+    @Config("http-client.preserve-content-decoder-factories")
+    public HttpClientConfig setPreserveContentDecoderFactories(boolean preserveContentDecoderFactories)
+    {
+        this.preserveContentDecoderFactories = preserveContentDecoderFactories;
         return this;
     }
 
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
@@ -359,7 +359,9 @@ public class JettyHttpClient
 
             // remove the GZIP encoding from the client
             // TODO: there should be a better way to to do this
-            httpClient.getContentDecoderFactories().clear();
+            if (!config.isPreserveContentDecoderFactories()) {
+                httpClient.getContentDecoderFactories().clear();
+            }
         }
         catch (Exception e) {
             if (e instanceof InterruptedException) {
--- a/http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
+++ b/http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
@@ -19,6 +19,7 @@ import org.junit.jupiter.api.Timeout;
 import org.junit.jupiter.api.parallel.Execution;
 
 import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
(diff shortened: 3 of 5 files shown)


oneonestar:
https://jetty.org/docs/jetty/12.1/programming-guide/client/http.html#content-response
```
HttpClient httpClient = new HttpClient();
httpClient.start();
// Remove all the default response content decoders.
httpClient.getContentDecoderFactories().clear();
```

Looks like the `.clear()` is the official way to do it. I rebased the PR, PTAL.

mosabua:
Or @electrum

electrum (approved):

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -709,6 +712,39 @@ public void testCompressionIsDisabled()
...
+    {
+        HttpClientConfig clientConfig = createClientConfig().setPreserveContentDecoderFactories(true);
+        try (CloseableTestHttpServer server = newServer(); JettyHttpClient client = server.createClient(clientConfig)) {
+            Request request = prepareGet()
+                    .setUri(server.baseURI())
+                    .setHeader(ACCEPT_ENCODING, "gzip")
+                    .build();
+
+            StringResponse response = client.execute(request, createStringResponseHandler());
+            Thread.sleep(1000);
```
electrum: Why do we need to sleep here?

electrum:
@wendigo do you want to take a look?

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -357,9 +357,10 @@ protected void configure(SelectableChannel selectable)
         try {
             httpClient.start();
 
-            // remove the GZIP encoding from the client
-            // TODO: there should be a better way to to do this
-            httpClient.getContentDecoderFactories().clear();
```
findepi, line 362: This comes from https://github.com/airlift/airlift/pull/201 (https://github.com/airlift/airlift/pull/201/changes/6100c437681491ffb143f2389d6048eef4c80cf8). But that was a while ago. Maybe we can drop those lines altogether? Then we don't need a new config.

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -778,6 +779,19 @@ public HttpClientConfig setUseVirtualThreads(boolean useVirtualThreads)
         return this;
     }
 
+    public boolean isPreserveContentDecoderFactories()
+    {
+        return preserveContentDecoderFactories;
+    }
+
+    @ConfigHidden
+    @Config("http-client.preserve-content-decoder-factories")
```
findepi, line 788: So far the HTTP Client Config tries to model semantics, without directly exposing underlying Jetty client. Of course, tuning knobs such as `http-client.max-direct-memory` may need to be configured with knowledge of the internals, but it's clear what they can and cannot impact.

This one is different. It impacts behavior on HTTP level, but it's unclear how it does that. It's unclear can be the effect of setting this config. 

Is there a better way to frame this problem? Like "enable gzipped response" or "allow compressed response for uncompressed request"?

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
oneonestar: That PR didn’t explain why it was needed. IIUC, it was to disable compression between workers when sending data for performance reasons.

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
oneonestar: Removed.

http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
oneonestar: https://github.com/jetty/jetty.project/blob/8fe9356a33b51d48a5a372f501e76a96f5ef56da/jetty-core/jetty-client/src/main/java/org/eclipse/jetty/client/ContentDecoder.java#L129-L133
`httpClient.getContentDecoderFactories().clear();` disables two default features provided by Jetty:

* It stops sending the `Accept-Encoding: gzip` header in requests.
* It disables decoding of `gzip` response content.
(`br and zstd` are also added if the corresponding Jetty plugins are on the classpath.)

How about `http-client.enable-compression` defaulting to `false`?

findepi: > How about `http-client.enable-compression` defaulting to `false`?

sgtm!

wendigo: What about this: https://github.com/airlift/airlift/pull/1660 ?

PR #1584, opened 2025-10-02
Allow skipping attributes in assertDefaults

chenjian2664:

### DIFF
--- a/configuration-testing/src/test/java/io/airlift/configuration/testing/TestConfigAssertions.java
+++ b/configuration-testing/src/test/java/io/airlift/configuration/testing/TestConfigAssertions.java
@@ -515,6 +515,35 @@ public class TestConfigAssertions
     }
 
     @Test
+    public void testRecordedDefaultsSkippingAttributes()
+    {
+        ConfigAssertions.assertRecordedDefaults(ConfigAssertions.recordDefaults(Config1.class)
+                .setBoxedBooleanOption(null)
+                .setBoxedByteOption(null)
+                .setBoxedDoubleOption(null)
+                .setBoxedFloatOption(null)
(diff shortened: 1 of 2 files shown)


configuration-testing/src/test/java/io/airlift/configuration/testing/TestConfigAssertions.java
```diff
@@ -514,6 +514,35 @@ public void testRecordedDefaultsOneOfEverything()
...
+                .setPathOption(null)
+                .setMyEnumList(ImmutableList.of())
+                .setMyEnumSet(ImmutableSet.of())
+                .setMyIntegerList(ImmutableList.of())
+                .setMyPathList(ImmutableList.of())
+                .setShortOption((short) 0)
+                .setStringOption(null)
+                .setValueClassOption(null)
+                .setOptionalValueClassOption(Optional.empty())
+                .setOptionalPathOption(Optional.empty()), ImmutableSet.of("BooleanOption", "MyEnumSecondOption"));
```
wendigo, line 543: Move ImmutableSet to next line

wendigo (approved):

martint:
Can you give me an example? Not sure I understand how skipping options when testing defaults matters in that case.

chenjian2664:
@martint We want to add a new property but deprecate an existing one (not remove it). but the new and old property conflict, they must not be set at the same time. In test `testDefaults`, we'd like to use a similar mechanism to `assertFullMapping ` that allows skipping specific attributes

martint:
Do the properties conflict in their default settings?

chenjian2664:
@martint No.
Just found recordDefaults won't triggle the validation, seems we may not need this

PR #1601, opened 2025-10-14
Track memory allocations in http server/client

wendigo:

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
@@ -126,6 +126,7 @@ public class HttpClientConfig
     private boolean verifyHostname = true;
     private Optional<String> httpProxyUser = Optional.empty();
     private Optional<String> httpProxyPassword = Optional.empty();
+    private boolean trackMemoryAllocations;
 
     public boolean isVerifyHostname()
     {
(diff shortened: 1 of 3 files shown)


http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -554,6 +555,20 @@ public HttpServerConfig setNotifyRemoteAsyncErrors(boolean notifyRemoteAsyncErro
         return this;
     }
 
+    public boolean isTrackMemoryAllocations()
+    {
+        return trackMemoryAllocations;
+    }
+
+    @Config("http-server.track-memory-allocations")
```
martint, line 563: Why make it configurable? What’s the downside of using the tracking buffer pool?

wendigo: It's using more memory, we need this only in cases when we debug why Jetty is failing to either release memory or does double release (as in the compression bug that I've reported).

wendigo: Hence it is @ConfigHidden

wendigo: > A variant of ArrayByteBufferPool that tracks buffer acquires/releases, useful to identify buffer leaks.
Use getLeaks() when the system is idle to get the ArrayByteBufferPool.Tracking.TrackedBuffers that have been leaked, which contain the stack trace information of where the buffer was acquired.

Among other things, Tracking captures allocation stack traces (which is expensive in itself)

PR #1611, opened 2025-10-21
Add automatic release notes

wendigo:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install java
         uses: actions/setup-java@v4
         with:
-          java-version: '22'
+          java-version: '25'
           distribution: 'temurin'
           gpg-private-key: ${{ secrets.JRELEASER_GPG_SECRET_KEY }}
           gpg-passphrase: MAVEN_GPG_PASSPHRASE


.github/workflows/release.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           MAVENCENTRAL_USERNAME: ${{ secrets.JRELEASER_NEXUS2_MAVEN_CENTRAL_USERNAME }}
           MAVENCENTRAL_PASSWORD: ${{ secrets.JRELEASER_NEXUS2_MAVEN_CENTRAL_TOKEN }}
         run:
-          ./mvnw njord:publish -Poss-release,oss-stage -Ddetails -Dfull
```
findepi, line 87: > Do not output details while publishing


but _why_?

.github/workflows/release.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Install java
         uses: actions/setup-java@v4
         with:
-          java-version: '22'
+          java-version: '25'
```
findepi, line 29: is 25 the min supported version? i know it's not... so _why_?

wendigo: It will be a new baseline soon

.github/workflows/release.yml
wendigo: It doesn't provide any value

wendigo: It's a leftover from the times this workflow was built

.github/workflows/release.yml
wendigo: Added a rationale to the commit

.github/workflows/release.yml
wendigo: Added a rationale to the commit

PR #1618, opened 2025-10-29
Allow disabling merging with system properties

wendigo:

### DIFF
--- a/configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
+++ b/configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
@@ -32,7 +32,14 @@ public final class ConfigurationLoader
 {
     private ConfigurationLoader() {}
 
+    @Deprecated
     public static Map<String, String> loadProperties()
+            throws IOException
+    {
+        return loadProperties(true);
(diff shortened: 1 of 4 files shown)


configuration/src/main/java/io/airlift/configuration/ConfigurationLoader.java
```diff
@@ -32,16 +32,25 @@ public final class ConfigurationLoader
...
+        return loadProperties(true);
+    }
+
+    public static Map<String, String> loadProperties(boolean useSystemProperties)
+            throws IOException
     {
         Map<String, String> result = new TreeMap<>();
         String configFile = System.getProperty("config");
         if (configFile != null) {
             result.putAll(loadPropertiesFrom(configFile));
```
findepi, line 48: it seems not to make much sense for this method, when `! useSystemProperties` to still use `System.getProperty("config")`, but ignore all other properties.
It seems that this method is inherently tied to system properties and useless otherwise.

findepi: - https://github.com/airlift/airlift/pull/1668

PR #1634, opened 2025-11-11
Improve codebase

wendigo:


configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
```diff
@@ -82,27 +84,11 @@
 import static java.util.function.Function.identity;
 
 public class ConfigurationFactory
+        implements AutoCloseable
```
electrum: Let's make this `Closeable` and remove the checked exception from the `close` method as `ValidatorFactory.close` doesn't throw any checked exceptions.

electrum (approved):

configuration/src/main/java/io/airlift/configuration/ConfigurationFactory.java
wendigo: Done

PR #1638, opened 2025-11-13
[MCP Server] Support server-to-client notifications during a request

Randgalt:
The reference MCP server does not support server-to-client messages. This change wraps the servlet objects sent to the reference implementation's Servlet so that we can change the response to an SSE response for the purpose of supporting server-to-client messages. The response is changed only if a message is actually sent.

Added tests for this.

Note: this takes advantage of inner knowledge of how `HttpServletStatelessServerTransport` works. Hopefully the MCP team will add native support for this in the future. We've opened PRs with possible implementations and have started GitHub discussions about future features.

### DIFF
--- a/mcp/src/main/java/io/airlift/mcp/McpRequestContext.java
+++ b/mcp/src/main/java/io/airlift/mcp/McpRequestContext.java
@@ -1,0 +1,10 @@
+package io.airlift.mcp;
+
+import jakarta.servlet.http.HttpServletRequest;
+
+public interface McpRequestContext
+{
+    HttpServletRequest request();
+
+    void sendProgress(double progress, double total, String message);
+}
--- a/mcp/src/main/java/io/airlift/mcp/reference/ReferenceRequestContextProvider.java
+++ b/mcp/src/main/java/io/airlift/mcp/reference/ReferenceRequestContextProvider.java
@@ -1,0 +1,75 @@
+package io.airlift.mcp.reference;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.inject.Inject;
+import io.airlift.mcp.McpRequestContext;
+import io.airlift.mcp.handler.RequestContextProvider;
+import io.airlift.mcp.model.JsonRpcRequest;
+import io.airlift.mcp.model.ProgressNotification;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.util.Optional;
+import java.util.OptionalDouble;
+
+import static io.airlift.mcp.McpException.exception;
+import static java.util.Objects.requireNonNull;
+
+public class ReferenceRequestContextProvider
+        implements RequestContextProvider
+{
+    private final ObjectMapper objectMapper;
+
+    @Inject
+    public ReferenceRequestContextProvider(ObjectMapper objectMapper)
+    {
+        this.objectMapper = requireNonNull(objectMapper, "objectMapper is null");
+    }
+
+    @Override
+    public McpRequestContext get(HttpServletRequest request, HttpServletResponse response, Optional<Object> progressToken)
+    {
+        return new McpRequestContext()
+        {
+            @Override
+            public HttpServletRequest request()
+            {
+                return request;
+            }
+
+            @SuppressWarnings("SwitchStatementWithTooFewBranches")
+            @Override
+            public void sendProgress(double progress, double total, String message)
+            {
+                Optional<Object> appliedProgressToken = progressToken.map(token -> switch (token) {
+                    case Number value -> Optional.of(value.longValue());
+                    default -> progressToken;
+                });
+
+                ProgressNotification notification = new ProgressNotification(appliedProgressToken, message, OptionalDouble.of(progress), OptionalDouble.of(total));
+                sendNotification("notifications/progress", Optional.of(notification));
+            }
+
+            private void sendNotification(String method, Optional<Object> params)
+            {
+                try {
+                    PrintWriter writer = response.getWriter();
+                    if (!(writer instanceof SsePrintWriter ssePrintWriter)) {
+                        throw exception("Expected writer to be an instance of SsePrintWriter but got %s".formatted(writer.getClass().getName()));
+                    }
+                    JsonRpcRequest<?> notification = params
+                            .map(values -> JsonRpcRequest.buildNotification(method, values))
+                            .orElseGet(() -> JsonRpcRequest.buildNotification(method));
+                    String json = objectMapper.writeValueAsString(notification);
+                    ssePrintWriter.writeMessage(json);
+                    ssePrintWriter.flush();
+                }
+                catch (IOException e) {
+                    throw exception(e);
+                }
+            }
+        };
+    }
+}
--- a/mcp/src/main/java/io/airlift/mcp/reference/SsePrintWriter.java
+++ b/mcp/src/main/java/io/airlift/mcp/reference/SsePrintWriter.java
@@ -1,0 +1,62 @@
+package io.airlift.mcp.reference;
+
+import java.io.PrintWriter;
+import java.io.Writer;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicLong;
+
+import static io.airlift.mcp.McpException.exception;
+import static java.util.Objects.requireNonNull;
+
+class SsePrintWriter
+        extends PrintWriter
+{
+    private final AtomicLong nextId = new AtomicLong();
+    private final Runnable upgradeHandler;
+    private boolean hasBeenUpgraded;
+
+    SsePrintWriter(Writer out, Runnable upgradeHandler)
+    {
+        super(out, false);
+        this.upgradeHandler = requireNonNull(upgradeHandler, "upgradeHandler is null");
+    }
+
+    @Override
+    public void write(String data)
+    {
+        if (hasBeenUpgraded) {
+            writeMessage(data);
+        }
+        else {
+            // no messages were sent, keep it as a standard JSON response
+            super.write(data);
+        }
+    }
+
+    public void writeMessage(String data)
+    {
+        writeMessage(data, Optional.of(Long.toString(nextId.getAndIncrement())));
+    }
+
+    public void writeMessage(String data, Optional<String> messageId)
+    {
+        if (!hasBeenUpgraded) {
+            hasBeenUpgraded = true;
+            // this will change response to an SSE response
+            upgradeHandler.run();
+        }
+
+        messageId.ifPresent(id -> super.write("id: " + encode(id) + "\n"));
+
+        super.write("data: " + encode(data) + "\n\n");
+        if (checkError()) {
+            throw exception("Error writing SSE message");
+        }
+    }
+
+    private String encode(String str)
+    {
+        // Escape newlines and carriage returns for SSE compliance
+        return str.replace("\n", "\\n").replace("\r", "\\r");
+    }
+}
--- a/mcp/src/main/java/io/airlift/mcp/reference/SseResponseWrapper.java
+++ b/mcp/src/main/java/io/airlift/mcp/reference/SseResponseWrapper.java
@@ -1,0 +1,51 @@
+package io.airlift.mcp.reference;
+
+import jakarta.servlet.http.HttpServletResponse;
+import jakarta.servlet.http.HttpServletResponseWrapper;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+
+/**
+ * <p>
+ *     Intercept output from {@link io.modelcontextprotocol.server.transport.HttpServletStatelessServerTransport}. This class
+ *     has deep internal knowledge of how {@link io.modelcontextprotocol.server.transport.HttpServletStatelessServerTransport} works.
+ *     {@link io.modelcontextprotocol.server.transport.HttpServletStatelessServerTransport} will think it's outputting a single
+ *     JSON RPC response but this class, via {@link SsePrintWriter}, will convert that output into an SSE stream which allows
+ *     us to support server-to-client notifications during request processing. NOTE: if {@link io.modelcontextprotocol.server.transport.HttpServletStatelessServerTransport}
+ *     should change, then this class and {@link SsePrintWriter} will need to change.
+ * </p>
+ *
+ * <p>
+ *     Note: I've been trying to get traction in the MCP project for native support with this PR and others.
+ *     <a href="https://github.com/modelcontextprotocol/java-sdk/pull/472">pull/472</a>. Additionally, I've started this discussion:
+ *     <a href="https://github.com/modelcontextprotocol/java-sdk/discussions/665">discussions/665</a>
+ * </p>
+ */
+class SseResponseWrapper
+        extends HttpServletResponseWrapper
+{
+    private SsePrintWriter ssePrintWriter;
+
+    SseResponseWrapper(HttpServletResponse response)
+    {
+        super(response);
+    }
+
+    @Override
+    public PrintWriter getWriter()
+            throws IOException
+    {
+        int status = getStatus();
+        boolean useSse = (status >= SC_OK) && (status < SC_MULTIPLE_CHOICES);
+        if (useSse) {
+            if (ssePrintWriter == null) {
+                ssePrintWriter = new SsePrintWriter(super.getWriter(), () -> super.setContentType("text/event-stream"));
+            }
+
+            return ssePrintWriter;
+        }
+
+        return super.getWriter();
+    }
+}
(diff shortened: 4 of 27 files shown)


mcp/src/main/java/io/airlift/mcp/reflection/ReflectionHelper.java
```diff
@@ -67,6 +69,10 @@ static List<MethodParameter> parseParameters(Method method, Optional<? extends C
                         return HttpRequestParameter.INSTANCE;
                     }
 
+                    if (parameter.getType().equals(McpRequestContext.class)) {
+                        return McpRequestContextParameter.INSTANCE;
+                    }
```
codyzwief, line 74: Do we want to consider deprecating the `HttpServletRequest` parameter since it's contained in the `McpRequestContext`?

mcp/src/main/java/io/airlift/mcp/McpRequestContext.java
```diff
@@ -0,0 +1,19 @@
...
+    HttpServletRequest request();
+
+    default void sendNotification(String method)
+    {
+        sendNotification(method, Optional.empty());
+    }
+
+    <T> void sendNotification(String method, Optional<T> params);
+
+    void sendProgress(double progress, double total, String message);
```
codyzwief, line 9: Is this intended to be a [progress notification](https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/progress)? If so, could we add the `progressToken` param here to comply with the spec?

mcp/src/main/java/io/airlift/mcp/McpRequestContext.java
```diff
@@ -0,0 +1,19 @@
...
+public interface McpRequestContext
+{
+    HttpServletRequest request();
+
+    default void sendNotification(String method)
+    {
+        sendNotification(method, Optional.empty());
+    }
+
+    <T> void sendNotification(String method, Optional<T> params);
```
codyzwief: What type of notification is this? Is it a [logging notification](https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging)? If so, would it be possible to pass an object representing the fields in the notification (such as what's in `McpSchema.LoggingMessageNotification`)?

mcp/src/main/java/io/airlift/mcp/model/ProgressNotificationLong.java
```diff
@@ -0,0 +1,15 @@
+package io.airlift.mcp.model;
+
+import java.util.OptionalDouble;
+
+import static java.util.Objects.requireNonNull;
+
+public record ProgressNotificationLong(long progressToken, String message, OptionalDouble progress, OptionalDouble total)
```
codyzwief: Just curious, why this object instead of just using `String.valueOf(long)` as the progress token?

mcp/src/test/java/io/airlift/mcp/TestMcp.java
```diff
@@ -399,4 +444,46 @@ private JsonResponse<JsonRpcResponse<?>> rpcCall(String identityHeader, JsonRpcR
 
         return httpClient.execute(request, createFullJsonResponseHandler(jsonCodec(new TypeToken<>() {})));
     }
+
+    private JsonResponse<JsonRpcResponse<?>> rpcCallSse(String identityHeader, JsonRpcRequest<?> jsonrpcRequest)
+    {
```
codyzwief, line 461: Do you think it makes sense to wire up a test using the `io.modelcontextprotocol.client.McpSyncClient` as a way to validate the notifications that get sent over the event stream? That class has a way to pass in a `loggingConsumer` and `progressConsumer` to validate what was received, and additionally can set logging levels as well to test that functionality.

mcp/src/main/java/io/airlift/mcp/reference/SseResponseWrapper.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.PrintWriter;
+
+/**
+ * Intercept output from {@link io.modelcontextprotocol.server.transport.HttpServletStatelessServerTransport}. This class
+ * has deep internal knowledge of how {@link io.modelcontextprotocol.server.transport.HttpServletStatelessServerTransport} works.
+ * {@link io.modelcontextprotocol.server.transport.HttpServletStatelessServerTransport} will think it's outputting a single
+ * JSON RPC response but this class, via {@link SsePrintWriter}, will convert that output into an SSE stream which allows
+ * us to support server-to-client notifications during request processing. NOTE: if {@link io.modelcontextprotocol.server.transport.HttpServletStatelessServerTransport}
+ * should change, then this class and {@link SsePrintWriter} will need to change.
+ */
```
codyzwief, line 24: Can we add a link to the PR you've opened so we have a reference to how we'll need to adjust this class if/when it gets accepted by the community?

mcp/src/main/java/io/airlift/mcp/reference/ReferenceRequestContextProvider.java
```diff
@@ -0,0 +1,73 @@
...
+                sendNotification("notifications/progress", Optional.of(notification));
+            }
+
+            @Override
+            public <T> void sendNotification(String method, Optional<T> params)
+            {
+                try {
+                    PrintWriter writer = response.getWriter();
+                    if (!(writer instanceof SsePrintWriter ssePrintWriter)) {
+                        throw new IOException("Response writer is not an SsePrintWriter");
```
codyzwief: Should this be an `McpException.exception()` instance?

mcp/src/main/java/io/airlift/mcp/reference/ReferenceRequestContextProvider.java
```diff
@@ -0,0 +1,73 @@
...
+                    }
+                    JsonRpcRequest<?> notification = params.map(p -> JsonRpcRequest.buildNotification(method, p)).orElseGet(() -> JsonRpcRequest.buildNotification(method));
+                    String json = objectMapper.writeValueAsString(notification);
+                    ssePrintWriter.writeMessage(json);
+                }
+                catch (IOException e) {
+                    throw new UncheckedIOException(e);
```
codyzwief: or maybe this is the `McpException.exception()`

mcp/src/main/java/io/airlift/mcp/reference/ReferenceRequestContextProvider.java
```diff
@@ -0,0 +1,73 @@
...
+
+            @Override
+            public <T> void sendNotification(String method, Optional<T> params)
+            {
+                try {
+                    PrintWriter writer = response.getWriter();
+                    if (!(writer instanceof SsePrintWriter ssePrintWriter)) {
+                        throw new IOException("Response writer is not an SsePrintWriter");
+                    }
+                    JsonRpcRequest<?> notification = params.map(p -> JsonRpcRequest.buildNotification(method, p)).orElseGet(() -> JsonRpcRequest.buildNotification(method));
```
codyzwief: The `McpSchema` has a separate `JSONRPCNotification` object that might be useful to create a mapping for for this instead of reusing `JsonRpcRequest`. WDYT?

mcp/src/main/java/io/airlift/mcp/reference/SseResponseWrapper.java
```diff
@@ -0,0 +1,43 @@
...
+    {
+        super(response);
+    }
+
+    @Override
+    public PrintWriter getWriter()
+            throws IOException
+    {
+        int status = getStatus();
+        boolean useSse = (status >= 200) && (status < 300);
```
codyzwief: Just want to confirm because I didn't know this until now. During request processing, before response completion, the status defaults to a 200 until either the resource method completes or an exception is thrown? So in the middle of request processing this should always return an event streaming response.

mcp/src/main/java/io/airlift/mcp/reference/SsePrintWriter.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        messageId.ifPresent(id -> super.write("id: " + encode(id) + "\n"));
+
+        try {
+            out.write("data: " + encode(data) + "\n\n");
+            out.flush();
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
```
codyzwief: `McpException.exception()` here?

mcp/src/main/java/io/airlift/mcp/reflection/ReflectionHelper.java
Randgalt: I thought there would be little weight in still supporting it. These injections are intended for convenience.

mcp/src/main/java/io/airlift/mcp/McpRequestContext.java
Randgalt: It is included. See here: https://github.com/airlift/airlift/pull/1638/files#diff-c8012c09829d6b66f652913225f1881ab9a32733c8ee244f9a08b72bd5a99171R33. It's really an implementation detail so let's have Airlift manage it.

mcp/src/main/java/io/airlift/mcp/McpRequestContext.java
Randgalt: The spec is open-ended on this. They could be anything. However, I now think this should be handled when the work on MCP sessions is completed. I'll remove this for now.

mcp/src/main/java/io/airlift/mcp/model/ProgressNotificationLong.java
Randgalt: See the spec: https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/progress#progress-flow

> Progress tokens MUST be a string or integer value

This actually matters. In my testing if I changed to a String the MCP Inspector stopped processing progress notifications. I'm going to do some more testing though.

mcp/src/test/java/io/airlift/mcp/TestMcp.java
Randgalt: Good idea. Let's do that as a separate PR. OK?

mcp/src/main/java/io/airlift/mcp/reference/ReferenceRequestContextProvider.java
Randgalt: A `JSONRPCNotification` is merely a `JsonRpcRequest` without an ID. It seemed unnecessary to me. We could add it though if people want it. Note: we already have `JsonRpcRequest.buildNotification()`

mcp/src/main/java/io/airlift/mcp/reference/SseResponseWrapper.java
Randgalt: This bit of code wraps the real `Writer` with a `SsePrintWriter`. `SsePrintWriter` will only upgrade the response to SSE if `writeMessage()` is called. Otherwise the response remains standard.

mcp/src/main/java/io/airlift/mcp/reference/SsePrintWriter.java
Randgalt: I'll double check all of these.

mcp/src/main/java/io/airlift/mcp/McpRequestContext.java
codyzwief: Got it, SGTM. Only other comment here would be that both `total` and `message` are optional fields in the spec. So we should at a minimum make `total` an optional or boxed double.

Randgalt: I thought that enforcing these values is better even though the spec says it's optional.

mcp/src/test/java/io/airlift/mcp/TestMcp.java
```diff
@@ -399,4 +444,46 @@ private JsonResponse<JsonRpcResponse<?>> rpcCall(String identityHeader, JsonRpcR
...
+            BufferedReader reader = new BufferedReader(new InputStreamReader(response.getInputStream()));
+
+            while (true) {
+                String line = reader.readLine();
+                if (line == null) {
+                    break;
+                }
+
+                if (line.startsWith("data: ")) {
+                    lastRequestEvents.add(line.substring(6));
```
findepi: 6 -> "data: ".length

mcp/src/test/java/io/airlift/mcp/TestMcp.java
```diff
@@ -383,6 +403,31 @@ public void testGetMcpReturns405()
         assertThat(response.getStatusCode()).isEqualTo(SC_METHOD_NOT_ALLOWED);
     }
 
+    @Test
+    public void testProgress()
+            throws JsonProcessingException
+    {
+        CallToolRequest callToolRequest = new CallToolRequest("progress", ImmutableMap.of());
+        JsonRpcRequest<?> jsonrpcRequest = JsonRpcRequest.buildRequest(1, "tools/call", callToolRequest);
```
findepi:
```suggestion
        JsonRpcRequest<?> jsonRpcRequest = JsonRpcRequest.buildRequest(1, "tools/call", callToolRequest);
```

mcp/src/test/java/io/airlift/mcp/TestMcp.java
```diff
@@ -399,4 +444,46 @@ private JsonResponse<JsonRpcResponse<?>> rpcCall(String identityHeader, JsonRpcR
...
+        Request request = preparePost().setUri(baseUri)
+                .addHeader("Content-Type", "application/json")
+                .addHeader("Accept", "application/json,text/event-stream")
+                .addHeader(IDENTITY_HEADER, identityHeader)
+                .setBodyGenerator(JsonBodyGenerator.jsonBodyGenerator(jsonCodec(new TypeToken<JsonRpcRequest<?>>() {}), jsonrpcRequest))
+                .build();
+
+        lastRequestEvents.clear();
+
+        try (StreamingResponse response = httpClient.executeStreaming(request)) {
```
findepi, line 473: how does this work with network retries?

mcp/src/test/java/io/airlift/mcp/TestMcp.java
```diff
@@ -399,4 +444,46 @@ private JsonResponse<JsonRpcResponse<?>> rpcCall(String identityHeader, JsonRpcR
...
+                if (line.startsWith("data: ")) {
+                    lastRequestEvents.add(line.substring(6));
+                }
+            }
+
+            if (lastRequestEvents.isEmpty()) {
+                throw new UncheckedIOException(new IOException("No events received from MCP server"));
+            }
+
+            return new JsonResponse<>(response.getStatusCode(), response.getHeaders(), responseCodec, lastRequestEvents.getLast().getBytes(UTF_8));
```
findepi, line 495: Since this method feeds into `this.lastRequestEvents` queue, it should not return anything.

findepi:
>The reference MCP server does not support server-to-client messages. This change wraps the servlet objects sent to the reference implementation's Servlet so that we can change the response to an SSE response for the purpose of supporting server-to-client messages. The response is changed only if a message is actually sent.

is it still beneficial to use the reference MCP server implementation?
or the use of it works against us?

mcp/src/test/java/io/airlift/mcp/TestMcp.java
Randgalt: SSE are HTTP requests with chunked responses - as a single request/response. If I understand your question, a retry would require the client to disconnect and reconnect.

mcp/src/test/java/io/airlift/mcp/TestMcp.java
Randgalt: It still returns a response. The SSE protocol is a bunch of optional interim messages with a final response.

mcp/src/test/java/io/airlift/mcp/TestMcp.java
findepi: > SSE are HTTP requests with chunked responses

this i know 

but shouldn't there be some way to reconnect and re-receive notifications that were sent back recently?
otherwise it's unreliable => less useful

> If I understand your question, a retry would require the client to disconnect and reconnect.

for sure, yes.

mcp/src/test/java/io/airlift/mcp/TestMcp.java
findepi: it's more java method design question, not a protocol question

mcp/src/test/java/io/airlift/mcp/TestMcp.java
Randgalt: The protocol allows clients to reconnect with a header that will replay messages. But, this is optional. Also, it would require storing messages in a backing store. The messages that would be sent inline in this case are not important. They are usually pings and progress messages that aren't worth replaying. Important messages are handled by the new "task" protocol that was just introduced. There are separate PRs in Airlift for that.

mcp/src/test/java/io/airlift/mcp/TestMcp.java
Randgalt: The events queue is _optional_. The response is upgraded to SSE only if the tool/prompt/etc. decides to send events. If that doesn't happen it's a standard `application/json` response. So the code is appropriate.

codyzwief:
@wendigo / @findepi  would it be possible to review/merge this one soon?

wendigo:
I've applied some nitpicks and going to merge it as soon as CI is done

PR #1640, opened 2025-11-14
Don't use local date when generating SSL certificate

aalbu:
Basing certificate dates on local date can lead to a situation where the certificate is created with a notBefore date in the future.

### DIFF
--- a/security/pom.xml
+++ b/security/pom.xml
@@ -14,6 +14,7 @@
 
     <properties>
         <air.main.basedir>${project.parent.basedir}</air.main.basedir>
+        <air.test.timezone>Asia/Kathmandu</air.test.timezone>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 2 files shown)


security/pom.xml
```diff
@@ -63,4 +63,18 @@
...
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <systemPropertyVariables>
+                        <user.timezone>Asia/Kathmandu</user.timezone>
```
wendigo: No need to do that this way.

Set property `<air.test.timezone>`

wendigo: Add a comment why this is needed

PR #1641, opened 2025-11-15
[MCP Server] Support MCP completions

Randgalt:
- Added two new annotations and general handler support for MCP completions.
- Added a few examples in the test endpoints
- Added tests and doc

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -17,13 +17,11 @@ variations of MCP servers defined by the standard. This module supports:
 - Ping [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/ping)
 - Structured content [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content)
 - Progress notifications [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/progress)
+- Completions [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/completion)
 
 It uses the [MCP reference Java SDK](https://github.com/modelcontextprotocol/java-sdk) as its internal implementation.
 This implementation is very limited at does not support:
 
-- Completions [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/completion)
-- `_meta` field [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/basic#meta)
-- `context` field [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/changelog) in `CompletionRequest`
 - Sessions [(see spec)](https://modelcontextprotocol.io/docs/concepts/transports#session-management)
 - Cancellation [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/cancellation)
 - List changed events [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle#initialization)
@@ -34,10 +32,20 @@ This implementation is very limited at does not support:
 - Roots [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/client/roots)
 - Sampling [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/client/sampling)
 
-## Creating tools, prompts, and resources declaratively
+## Creating tools, prompts, resources, and completions declaratively
 
 ```java
 // in some class...
+
+import com.google.common.collect.ImmutableList;
+import io.airlift.mcp.McpDescription;
+import io.airlift.mcp.McpPrompt;
+import io.airlift.mcp.McpPromptCompletion;
+import io.airlift.mcp.McpResource;
+import io.airlift.mcp.McpResourceTemplate;
+import io.airlift.mcp.McpResourceTemplateCompletion;
+import io.airlift.mcp.McpTool;
+import io.airlift.mcp.model.ResourceTemplateValues;
 
 @McpTool(name = "add", description = "Adds two numbers")
 public int addTwoNumbers(
@@ -53,6 +61,18 @@ public String greeting(@McpDescription("Name of the person to greet") String nam
     return "Hello, " + name + "!";
 }
 
+@McpPromptCompletion(name = "greeting")
+public List<String> nameCompletions(CompleteArgument argument)
+{
+    if (argument.name().equals("name")) {
+        return ImmutableList.of("Jordan", "Rita", "Bobby", "Oliver", "Olive", "Steve")
+                .stream()
+                .filter(name -> name.toLowerCase().startsWith(argument.value().toLowerCase()))
+                .collect(toImmutableList());
+    }
+    return ImmutableList.of();
+}
+
 @McpResource(name = "example1", uri = "file://example1.txt", description = "This is example1 resource.", mimeType = "text/plain")
 public ResourceContents resource()
 {
@@ -63,6 +83,18 @@ public ResourceContents resource()
 public ResourceContents resourceTemplate(ReadResourceRequest request, ResourceTemplateValues templateValues)
 {
     return new ResourceContents("foo2", "file://example1.txt", "text/plain", "This is the content of file://example1.txt");
+}
+
+@McpResourceTemplateCompletion(uriTemplate = "file:{path1}/{path2}")
+public List<String> resourceTemplateCompletion(CompleteArgument argument)
+{
+    if (argument.name().equals("id")) {
+        return ImmutableList.of("manny", "moe", "jack")
+                .stream()
+                .filter(uri -> uri.toLowerCase().startsWith(argument.value().toLowerCase()))
+                .collect(toImmutableList());
+    }
+    return ImmutableList.of();
 }
 ```
 
@@ -79,7 +111,7 @@ Module module = McpModule.builder()
 binder.install(module);
 ```
 
-## Creating tools, prompts, and resources programmatically
+## Creating tools, prompts, resources, and completions programmatically
 
 Example of creating a tool programmatically:
 
@@ -113,7 +145,7 @@ npx @modelcontextprotocol/inspector
 A browser should open with the MCP Inspector tool. Set the "Transport Type" to
 "Streamable HTTP". Change the URL to `http://localhost:8888/mcp` and click "Connect".
 
-## Allowed parameters for declarative tools, prompts, and resources
+## Allowed parameters for declarative tools, prompts, resources, and completions
 
 #### Tools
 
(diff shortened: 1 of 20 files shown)


Randgalt:
attn @codyzwief

mcp/src/main/java/io/airlift/mcp/McpCompletionType.java
codyzwief, line 1: This is unused

mcp/src/test/java/io/airlift/mcp/LocalServer.java
```diff
@@ -24,6 +25,7 @@ private LocalServer() {}
     private static final Logger log = Logger.get(LocalServer.class);
 
     public static void main(String[] args)
+            throws JsonProcessingException
```
codyzwief: exception not thrown, can be removed

mcp/src/test/java/io/airlift/mcp/TestMcp.java
```diff
@@ -122,6 +129,8 @@ public TestMcp()
         httpClient = injector.getInstance(Key.get(HttpClient.class, ForTest.class));
         baseUri = injector.getInstance(HttpServerInfo.class).getHttpUri().resolve("/mcp");
         objectMapper = injector.getInstance(ObjectMapper.class);
+
+        jsonCodecFactory = new JsonCodecFactory(() -> objectMapper);
```
codyzwief, line 133: Curious: Why this change, and why the need to specify a specific `jsonCodecFactory` instead of the static methods?

mcp/src/main/java/io/airlift/mcp/reference/Mapper.java
```diff
@@ -227,6 +235,38 @@ static McpSchema.PromptArgument mapPromptArgument(Prompt.Argument ourArgument)
...
+            Optional<CompleteContext> completeContext = Optional.ofNullable(theirCompleteRequest.context())
+                    .map(theirContext -> new CompleteContext(theirContext.arguments()));
+
+            CompleteRequest completeRequest = new CompleteRequest(completeReference, completeArgument, completeContext);
+            McpRequestContext requestContext = buildMcpRequestContext(context, Optional.ofNullable(theirCompleteRequest.meta()), requestContextProvider);
+
+            CompleteResult completeResult = ourCompletionHandler.complete(requestContext, completeRequest);
+
+            Integer total = completeResult.completion().total().orElse(completeResult.completion().values().size());
+            Boolean hasMore = completeResult.completion().hasMore().orElse(false);
```
codyzwief: In the [spec](https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/completion#completeresult) there's a max amount of 100 values in the completion result. Do we want to add any validation for this?

Additionally, do we want to support a case where `values` may have more (so `hasMore` is true) but the total number is unknown?

mcp/src/main/java/io/airlift/mcp/reference/ReferenceModule.java
```diff
@@ -74,11 +98,70 @@ public McpStatelessSyncServer buildServer(
...
+    private static void buildCompletions(StatelessSyncSpecification builder, RequestContextProvider requestContextProvider, Set<PromptEntry> prompts, Set<ResourceTemplateEntry> resourceTemplates, Set<CompletionEntry> completions)
+    {
+        if (completions.isEmpty()) {
+            return;
+        }
+
+        Set<String> promptNames = prompts.stream()
+                .map(entry -> entry.prompt().name())
+                .collect(toImmutableSet());
+        Set<String> usedPromptNames = new HashSet<>();
```
codyzwief, line 116: nit: would it make sense to use `ImmutableSet.Builder`?

mcp/src/main/java/io/airlift/mcp/reference/ReferenceModule.java
```diff
@@ -74,11 +98,70 @@ public McpStatelessSyncServer buildServer(
...
+                .map(missingPromptCompletion -> new SyncCompletionSpecification(new McpSchema.PromptReference(McpSchema.PromptReference.TYPE, missingPromptCompletion),
+                        (_, _) -> emptyResult))
+                .forEach(mappedCompletions::add);
+
+        // the reference implementation requires that all resource templates have a completion, so add empty completions for any missing resource templates
+        Sets.difference(resourceTemplateUris, usedResourceTemplateUris)
+                .stream()
+                .map(missingResourceTemplatetCompletion -> new SyncCompletionSpecification(new McpSchema.ResourceReference(McpSchema.ResourceReference.TYPE, missingResourceTemplatetCompletion),
+                        (_, _) -> emptyResult))
+                .forEach(mappedCompletions::add);
```
codyzwief, line 159: Do we want to consider throwing an exception in this case? Will the reference implementation advertise a bunch of prompt/template completions because of this that consumers of this library may not be aware of?

mcp/src/test/java/io/airlift/mcp/TestingEndpoints.java
```diff
@@ -58,12 +61,30 @@ public void throwsException()
         throw exception("this ain't good");
     }
 
+    @McpPrompt(name = "age", description = "What is your age?")
+    public String age(@McpDescription("What is your age?") String age)
+    {
+        return "You are " + age + "years old.";
+    }
```
codyzwief, line 68: Maybe I missed it, but I didn't see a test for this prompt? Did you mean to add it as part of this change?

mcp/src/main/java/io/airlift/mcp/model/CompleteResult.java
```diff
@@ -0,0 +1,27 @@
...
+import static java.util.Objects.requireNonNull;
+
+public record CompleteResult(CompleteCompletion completion)
+{
+    public CompleteResult
+    {
+        requireNonNull(completion, "completion is null");
+    }
+
+    public record CompleteCompletion(List<String> values, OptionalInt total, Optional<Boolean> hasMore)
```
codyzwief: Two questions:

1. From what I can tell, the `values` do not need to be Strings. Can this be made generic in some way?
2. The [spec](https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/completion#completeresult) says `total` is optional, but `hasMore` is not. Do we want to change `hasMore` to be a little-b boolean?

mcp/src/test/java/io/airlift/mcp/TestMcp.java
Randgalt: We need the `JsonSubType` bindings that are done in the reference module.

mcp/src/main/java/io/airlift/mcp/reference/Mapper.java
Randgalt: I would have hoped that the reference SDK handled this. tbh - we're not getting much value from that library. I'll take a look at this today or tomorrow.

mcp/src/main/java/io/airlift/mcp/reference/ReferenceModule.java
Randgalt: No - we're checking for duplicates so we want the behavior of `add()` returning `false` if it already exists.

mcp/src/test/java/io/airlift/mcp/TestingEndpoints.java
Randgalt: I do a lot of manual testing with the MCP Inspector. I was using this for that. I can remove if people want.

mcp/src/main/java/io/airlift/mcp/reference/ReferenceModule.java
Randgalt: From a usability point of view I thought adding the NOP completion was easier. Also, this restriction may change in the future.

mcp/src/main/java/io/airlift/mcp/model/CompleteResult.java
Randgalt: Here's the schema definition from the spec: https://modelcontextprotocol.io/specification/2025-11-25/schema#completeresult

```
interface CompleteResult {
  _meta?: { [key: string]: unknown };
  completion: { hasMore?: boolean; total?: number; values: string[] };
  [key: string]: unknown;
}
```

`hasMore` and `total` are optional and `values` is `string[]`.

mcp/README.md
```diff
@@ -176,3 +208,15 @@ A browser should open with the MCP Inspector tool. Set the "Transport Type" to
...
+#### Completions (for prompts and resource templates)
+
+- Parameters can be:
+    - `HttpServletRequest`
+    - `McpRequestContext`
+    - An Identity instance (via [McpIdentityMapper](src/main/java/io/airlift/mcp/McpIdentityMapper.java))
+    - [CompleteArgument](src/main/java/io/airlift/mcp/model/CompleteRequest.java)
+    - [CompleteContext](src/main/java/io/airlift/mcp/model/CompleteRequest.java)
+- Returns either:
+    - [CompleteCompletion](src/main/java/io/airlift/mcp/model/CompleteResult.java)
```
codyzwief, line 221: nit: these links need to be updated

mcp/src/main/java/io/airlift/mcp/McpCompletionType.java
codyzwief: This is still unused. Is it worth removing?

mcp/src/main/java/io/airlift/mcp/model/CompleteResult.java
codyzwief: Ah, good call, I missed this.

mcp/src/main/java/io/airlift/mcp/reference/Mapper.java
codyzwief: Just wondering about this comment, if we want to implement it until the bespoke implementation exists. Would it be worthwhile to create a test or two which test for large numbers of completions?

Randgalt: I'll see what I can do

mcp/README.md
Randgalt: They're actually correct - you can't link to a nest class from Markdown (that I know of)

mcp/src/main/java/io/airlift/mcp/reference/Mapper.java
Randgalt: I've now added a check for max 100 values

mcp/src/main/java/io/airlift/mcp/model/CompleteResult.java
```diff
@@ -0,0 +1,33 @@
...
+    {
+        public CompleteCompletion
+        {
+            values = ImmutableList.copyOf(values);
+            requireNonNull(total, "total is null");
+            requireNonNull(hasMore, "hasMore is null");
+
+            checkArgument(values.size() <= MAX_COMPLETIONS, "values exceeds max completions");
```
wendigo, line 29: can it be == 0 ?

Randgalt: Yes - that's allowed

PR #1645, opened 2025-11-18
Disable outbound buffering

wendigo:
Jetty buffers output when writing so Jersey buffering is redundant as it will be handled by http server.

### DIFF
--- a/jaxrs/src/main/java/io/airlift/jaxrs/JaxrsResourceConfigProvider.java
+++ b/jaxrs/src/main/java/io/airlift/jaxrs/JaxrsResourceConfigProvider.java
@@ -13,6 +13,7 @@ import java.util.Set;
 
 import static io.airlift.jaxrs.BinderUtils.qualifiedKey;
 import static java.util.Objects.requireNonNull;
+import static org.glassfish.jersey.CommonProperties.OUTBOUND_CONTENT_LENGTH_BUFFER;
 import static org.glassfish.jersey.server.ServerProperties.RESPONSE_SET_STATUS_OVER_SEND_ERROR;
 
 public class JaxrsResourceConfigProvider
@@ -37,6 +38,12 @@ public class JaxrsResourceConfigProvider
     {
         Set<Object> singletons = injector.getInstance(qualifiedKey(qualifier, new TypeLiteral<>() {}));
         return new JaxrsResourceConfig(singletons)
-                .setProperties(ImmutableMap.of(RESPONSE_SET_STATUS_OVER_SEND_ERROR, "true"));
+                .setProperties(ImmutableMap.of(
+                        RESPONSE_SET_STATUS_OVER_SEND_ERROR, "true",
+                        // Jetty http server buffers output when writing which makes Jersey server-side buffering redundant.
+                        // For small responses, allocating 8KB buffer is wasteful. For large responses, Jetty will buffer
+                        // as needed. Having Content-Length for small responses is not critical as in the HTTP/2
+                        // Content-Length can be inferred from DATA frame length.
+                        OUTBOUND_CONTENT_LENGTH_BUFFER, "0"));
     }
 }


jaxrs/src/main/java/io/airlift/jaxrs/JaxrsResourceConfigProvider.java
```diff
@@ -37,6 +38,9 @@ public ResourceConfig get()
     {
         Set<Object> singletons = injector.getInstance(qualifiedKey(qualifier, new TypeLiteral<>() {}));
         return new JaxrsResourceConfig(singletons)
-                .setProperties(ImmutableMap.of(RESPONSE_SET_STATUS_OVER_SEND_ERROR, "true"));
+                .setProperties(ImmutableMap.of(
+                        RESPONSE_SET_STATUS_OVER_SEND_ERROR, "true",
+                        // Jetty http server buffers output when writing which makes Jersey server-side buffering redundant
+                        OUTBOUND_CONTENT_LENGTH_BUFFER, "-1"));
```
findepi: from https://eclipse-ee4j.github.io/jersey.github.io/apidocs/2.35/jersey/org/glassfish/jersey/CommonProperties.html#OUTBOUND_CONTENT_LENGTH_BUFFER

> An integer value that defines the buffer size used to buffer the outbound message entity in order to determine its size and set the value of HTTP "Content-Length" header.
If the entity size exceeds the configured buffer size, the buffering would be cancelled and the entity size would not be determined. Value less or equal to zero disable the buffering of the entity at all.

1. after this change, is the "Content-Length" header still set? the docs suggest it won't be
2. they document "0" as canonical way to disable buffering

wendigo: Content-Length is added by Jetty as well (if it doesn't exist and can be computed). That itself doesn't matter for http/2 where content length is part of the frame.

wendigo: I was looking at the actual usage in the code, rather than javadocs: https://github.com/eclipse-ee4j/jersey/blob/efbd59bf2a400c72cf67552e6e8c4a582c3e732b/connectors/jetty-connector/src/test/java/org/glassfish/jersey/jetty/connector/TimeoutTest.java#L187

wendigo:
```
        if (bufferSize <= 0) {
            this.directWrite = true;
            this.buffer = null;
        } else {
            directWrite = false;
            buffer = new ByteArrayOutputStream(bufferSize);
        }
```

wendigo: Default buffer size is 8K which is a waste for most of the small responses and doesn't bring buffering benefits for larger ones (as Jetty will handle buffering too)

jaxrs/src/main/java/io/airlift/jaxrs/JaxrsResourceConfigProvider.java
```diff
@@ -37,6 +38,9 @@ public ResourceConfig get()
     {
         Set<Object> singletons = injector.getInstance(qualifiedKey(qualifier, new TypeLiteral<>() {}));
         return new JaxrsResourceConfig(singletons)
-                .setProperties(ImmutableMap.of(RESPONSE_SET_STATUS_OVER_SEND_ERROR, "true"));
+                .setProperties(ImmutableMap.of(
+                        RESPONSE_SET_STATUS_OVER_SEND_ERROR, "true",
+                        // Jetty http server buffers output when writing which makes Jersey server-side buffering redundant
```
findepi: capture https://github.com/airlift/airlift/pull/1645#discussion_r2538358493 as a code comment too

findepi (approved):

PR #1646, opened 2025-11-19
Do not buffer response on async http execution

wendigo:

### DIFF
--- a/http-client/src/test/java/io/airlift/http/client/jetty/TestAsyncJettyHttpClient.java
+++ b/http-client/src/test/java/io/airlift/http/client/jetty/TestAsyncJettyHttpClient.java
@@ -49,13 +49,4 @@ public class TestAsyncJettyHttpClient
             return executeAsync(client, request, responseHandler);
         }
     }
-
-    protected void testPutMethodWithStreamingBodyGenerator(boolean largeContent)
-            throws Exception
-    {
-        // don't test with async clients as they buffer responses and the LARGE content is too big
-        if (!largeContent) {
-            super.testPiped(largeContent);
-        }
-    }
 }


http-client/src/main/java/io/airlift/http/client/HttpClientConfig.java
```diff
@@ -59,6 +59,7 @@
         "http.authentication.krb5.keytab",
         "http.authentication.krb5.use-canonical-hostname",
         "http-client.buffer-pool-type",
+        "http-client.max-content-length"
```
findepi: > Deprecate http-client.max-content-length

Deprecate means "keep but deprecated, perhaps triggering a warning when used"
rename the commit to "Remove ..."

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -248,8 +245,6 @@ public JettyHttpClient(
         requireNonNull(config, "config is null");
         requireNonNull(requestFilters, "requestFilters is null");
         requireNonNull(httpStatusListeners, "httpStatusListeners is null");
-
-        maxContentLength = config.getMaxContentLength();
```
findepi, line 252: > Do not buffer response on async http execution

This removes usage of the `config.getMaxContentLength()` config, but removal of the config comes in a separate commit. Why? Changes are confusing to me, as the config removal is comes with another set of changes in this class.

http-client/src/main/java/io/airlift/http/client/ResponseHandlerUtils.java
```diff
@@ -20,6 +21,9 @@ public static RuntimeException propagate(Request request, Throwable exception)
         if (exception instanceof IOException) {
             throw new UncheckedIOException("Failed communicating with server: " + urlFor(request), (IOException) exception);
         }
+        if (exception instanceof ExecutionException e) {
```
findepi, line 24: ExecutionException is important bridge between for async exceptions allowing to understand what failed (the inner exception stacktrace) and where failed, where it was called (outer exception stacktrace).

It's therefore preferred not to strip ExecutionException out of exception causal chain.
if it has to be stripped away (eg to produce exception of the right type to the caller), please add a code comment explaining the need.

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -698,6 +695,7 @@ public StreamingResponse executeStreaming(Request request)
             }
         };
 
+        final Request finalRequest = request;
```
findepi, line 699: drop `final`

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -729,7 +727,7 @@ public long getBytesRead()
                 @Override
                 public InputStream getInputStream()
                 {
-                    return response.jettyResponse.getInputStream();
+                    return lengthLimited(response.jettyResponse.getInputStream(), finalRequest.getMaxContentLength());
```
findepi, line 731: The response comes from `internalExecute` which already applied the `lengthLimited` logic.

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -746,6 +744,13 @@ public void close()
         };
     }
 
+    private InputStream lengthLimited(InputStream inputStream, Optional<DataSize> maxContentLength)
+    {
+        return maxContentLength
+                .map(dataSize -> limiting(inputStream, dataSize.toBytes()))
+                .orElseGet(() -> limiting(inputStream, this.maxContentLength.toBytes()));
```
findepi, line 752:
```
lengthLimit = maxContentLength.orElse(this.maxContentLength);
return limiting(inputStream, lengthLimit.toBytes());
```

findepi: Also, it would be better to take `RequestContext` in `lengthLimited`. this would make it clearer what `Optional<DataSize>` param is. And that the caller is not suppose to OR it with `this.maxContentLength` before the call

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -915,23 +920,21 @@ public <T, E extends Exception> HttpResponseFuture<T> executeAsync(Request reque
         Span span = startSpan(request);
         request = injectTracing(request, span);
 
-        RequestContext jettyRequest = buildRequestContext(request);
-
-        request.getMaxContentLength()
-                .ifPresent(maxRequestContentLength -> verify(maxRequestContentLength.compareTo(maxContentLength) <= 0, "maxRequestContentLength must be less than or equal to maxContentLength"));
```
findepi, line 921: This logic was a good safety measure. Shouldn't we keep it?

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -960,6 +963,35 @@ private Span startSpan(Request request)
                 .startSpan();
     }
 
+    private <T, E extends Exception> Runnable processResponse(RequestContext context, InputStreamResponseListener listener, JettyResponseFuture<T, E> future)
```
findepi, line 972: `processAsyncResponse`?

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -960,6 +963,35 @@ private Span startSpan(Request request)
...
+    {
+        return () -> {
+            try {
+                Response response = listener.get(context.request.getTimeout(), MILLISECONDS);
+                try (InputStream stream = lengthLimited(listener.getInputStream(), context.maxContentLength())) {
+                    future.completed(response, stream);
+                }
+                catch (IOException e) {
+                    requestLogger.log(context.info(), ResponseInfo.failed(Optional.of(response), Optional.of(e)));
+                    future.failed(e);
```
findepi, line 982: Can this be any other unchecked exception?

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitInputStream.java
```diff
@@ -0,0 +1,76 @@
...
+            checkLimit();
+        }
+        return skipped;
+    }
+
+    private void checkLimit()
+            throws IOException
+    {
+        if (bytesLeft < 0) {
+            throw new IOException("Response size exceeded limit of " + maxSize);
```
findepi, line 65: Response -> Stream

the class itself doesn't know it's used for responses only and it doesn't have to be

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitInputStream.java
```diff
@@ -0,0 +1,76 @@
...
+    private final DataSize maxSize;
+    private long bytesLeft;
+
+    public ThrowingLimitInputStream(InputStream in, long maxBytes)
+    {
+        super(in);
+        if (maxBytes < 0) {
+            throw new IllegalArgumentException("maxBytes must be non-negative");
+        }
+        this.maxSize = DataSize.succinctBytes(maxBytes);
```
findepi, line 21: Just take `DataSize` in the constructor. The caller has it.

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitInputStream.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len)
+            throws IOException
+    {
+        int n = in.read(b, off, len);
+        if (n > 0) {
+            bytesLeft -= n;
+            checkLimit();
```
findepi, line 44: combine into single call
```
bytesRead(n)
```

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitInputStream.java
```diff
@@ -0,0 +1,76 @@
...
+            throws IOException
+    {
+        if (bytesLeft < 0) {
+            throw new IOException("Response size exceeded limit of " + maxSize);
+        }
+    }
+
+    static InputStream limiting(InputStream in, long maxBytes)
+    {
+        if (maxBytes == Long.MAX_VALUE) {
```
findepi, line 71: use `Optional.empty` to configure lack of any limit

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitInputStream.java
```diff
@@ -0,0 +1,76 @@
+package io.airlift.http.client.jetty;
+
+import io.airlift.units.DataSize;
+
+import java.io.FilterInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+public final class ThrowingLimitInputStream
+        extends FilterInputStream
```
findepi, line 10: I recommend against subclassing `FilterInputStream` because it's unclear which methods needs to be overridden to implement bookkeeping correctly. the `InputStream` interface is very broad and we don't want to implement all of them, do we?

For example, `skip` is overridden and `skipNBytes`  is not. It behaves correctly because `skipNBytes` implementation gets inherited from `InputStream` and uses `skip`. Is there anything preventing `FilterInputStream` from implementing ``skipNBytes`` too?
The jdoc says "FilterInputStrea itself simply overrides select methods of InputStream", which sounds like no guarantee this won't change.

Also, this class does not implement `mark` and `reset`, which can be used to rewind the stream. Thus the accounting becomes incorrect if those are used.

I recommend to inherit from `InputStream` and override all methods relevant for performance. This would guarantee bookkeeping is correct.

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -625,6 +627,30 @@ public void testExecuteStreaming()
...
+            executeRequest(server, request).ifPresent(streamingResponse -> {
+                try (streamingResponse) {
+                    String responseString = new String(streamingResponse.getInputStream().readAllBytes(), UTF_8);
+                    assertThat(responseString).isEqualTo(LARGE_CONTENT);
+                }
+                catch (IOException e) {
+                    assertThat(e).hasMessageContaining("Response size exceeded limit of 10B");
+                }
+            });
+        }
```
findepi, line 652: `ifPresent(streamingResponse -> { assert }`  and if not present then no assert?

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public <T, E extends Exception> T executeRequest(AbstractHttpClientTest.CloseableTestHttpServer server, Request request, ResponseHandler<T, E> responseHandler)
+            throws Exception
+    {
+        return executeRequest(server, createClientConfig(), request, responseHandler);
+    }
+
+    @SuppressWarnings("unchecked")
```
findepi, line 46: Redundant suppression?

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    public void testConnectTimeout()
+            throws Exception
+    {
+        // Async client doesn't work properly with timeouts as this is expected for the caller to put timeout on the Future.get
+        doTestConnectTimeout(true);
+    }
+
+    @SuppressWarnings("unchecked")
```
findepi, line 38: Redundant suppression?

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
```diff
@@ -0,0 +1,55 @@
...
+    {
+        return Optional.empty();
+    }
+
+    @Override
+    public void testConnectTimeout()
+            throws Exception
+    {
+        // Async client doesn't work properly with timeouts as this is expected for the caller to put timeout on the Future.get
+        doTestConnectTimeout(true);
```
findepi, line 35: That comment is worrisome. There should be two timeouts: on the request itself and the caller's on future.get

also, do we test that caller's `future.cancel` gets propagated?

PR #1649, opened 2025-11-20
[MCP Server] MCP Session support

Randgalt:
Support MCP sessions

Adds general support for MCP sessions via a new session controller.
An in-memory version of the controller is provided that is suitable
for testing or non-HA/fault-tolerant use. MCP session support in
Airlift is optional. If enabled, it's expected that a `SessionController`
implementation backed by a database or distributed hash map will
be used.

Additionally, MCP logging is implemented which, currently, requires
sessions. The MCP roadmap shows that this will change in the future
however, this implementation can easily conform to any future changes
because the logging-level currently stored in the session is not
exposed in Airlift MCP.

Finally, tests using the reference MCP SDK are introduced. These
test sessions with a conforming client.

### DIFF
--- a/mcp/src/main/java/io/airlift/mcp/sessions/SessionController.java
+++ b/mcp/src/main/java/io/airlift/mcp/sessions/SessionController.java
@@ -7,7 +7,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.BiFunction;
+import java.util.function.Function;
 import java.util.function.UnaryOperator;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 public interface SessionController
 {
@@ -45,6 +48,28 @@ public interface SessionController
     void deleteSession(SessionId sessionId);
 
     /**
+     * <p>
+     *     Block until the given condition returns {@code true} for the value associated with the given key in the session,
+     *     or until the timeout expires.
+     * </p>
+     *
+     * <p>
+     *     IMPORTANT: this is optional behavior. If the implementation does not support waiting for session changes,
+     *     the method should simply sleep for the duration of the timeout and return (as the default does).
+     * </p>
+     *
+     * <p>
+     *     The condition receives an optional value for the key. If the value is {@code empty()}, the key is not present in the session,
+     *     has been deleted, etc.
+     * </p>
+     */
+    default <T> void blockUntilCondition(SessionId sessionId, SessionValueKey<T> key, Duration timeout, Function<Optional<T>, Boolean> condition)
+            throws InterruptedException
+    {
+        MILLISECONDS.sleep(timeout.getSeconds() * 1000);
+    }
+
+    /**
      * @return the value associated with the given key in the session, or {@link Optional#empty()}
      * if the key is not present in the session or if the session ID is invalid
      */
@@ -66,8 +91,8 @@ public interface SessionController
      * </p>
      *
      * <p>
-     *     {@code updater} receives an {@link Optional} containing the existing value (or {@link Optional#empty()} if no value is present)
-     *     and should return an {@link Optional} containing the new value to set (or {@link Optional#empty()} to remove the value).
+     * {@code updater} receives an {@link Optional} containing the existing value (or {@link Optional#empty()} if no value is present)
+     * and should return an {@link Optional} containing the new value to set (or {@link Optional#empty()} to remove the value).
      * </p>
      *
      * @return the computed value or {@code empty()} if that's the computed value or if the session ID is invalid
--- a/mcp/src/main/java/io/airlift/mcp/sessions/MemorySessionController.java
+++ b/mcp/src/main/java/io/airlift/mcp/sessions/MemorySessionController.java
@@ -17,10 +17,13 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
 import java.util.function.UnaryOperator;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static io.airlift.mcp.sessions.SessionConditionUtil.waitForCondition;
 import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 public class MemorySessionController
         implements SessionController
@@ -37,6 +40,7 @@ public class MemorySessionController
     {
         private final Map<Class<?>, Map<String, Object>> values = new ConcurrentHashMap<>();
         private final Optional<Duration> ttl;
+        private final Signal signal = new Signal();
         private volatile Instant lastUsage = Instant.now();
 
         private Session(Optional<Duration> ttl)
@@ -62,6 +66,8 @@ public class MemorySessionController
             lastUsage = Instant.now();
 
             typeMap(key.type()).put(key.name(), value);
+
+            signal.signalAll();
         }
 
         private <T> Optional<T> compute(SessionValueKey<T> key, UnaryOperator<Optional<T>> updater)
@@ -73,6 +79,8 @@ public class MemorySessionController
                 return updater.apply(existing).orElse(null);
             });
 
+            signal.signalAll();
+
             return Optional.ofNullable(computed).map(key.type()::cast);
         }
 
@@ -81,6 +89,8 @@ public class MemorySessionController
             lastUsage = Instant.now();
 
             typeMap(key.type()).remove(key.name());
+
+            signal.signalAll();
         }
 
         private boolean canBeExpired(Instant now)
@@ -177,6 +187,22 @@ public class MemorySessionController
                     return true;
                 })
                 .orElse(false);
+    }
+
+    @Override
+    public <T> void blockUntilCondition(SessionId sessionId, SessionValueKey<T> key, Duration timeout, Function<Optional<T>, Boolean> condition)
+            throws InterruptedException
+    {
+        waitForCondition(this, sessionId, key, timeout, condition, maxWait -> {
+            Optional<Session> maybeSession = getSession(sessionId);
+            if (maybeSession.isEmpty()) {
+                MILLISECONDS.sleep(maxWait.toMillis());
+                return;
+            }
+
+            Session session = maybeSession.get();
+            session.signal.waitForSignal(maxWait.toMillis(), MILLISECONDS);
+        });
     }
 
     @Override
--- a/mcp/src/main/java/io/airlift/mcp/sessions/SessionConditionUtil.java
+++ b/mcp/src/main/java/io/airlift/mcp/sessions/SessionConditionUtil.java
@@ -1,0 +1,40 @@
+package io.airlift.mcp.sessions;
+
+import com.google.common.base.Stopwatch;
+
+import java.time.Duration;
+import java.util.Optional;
+import java.util.function.Function;
+
+public class SessionConditionUtil
+{
+    private SessionConditionUtil() {}
+
+    public interface WaitProc
+    {
+        void waitForKey(Duration maxWait)
+                throws InterruptedException;
+    }
+
+    public static <T> void waitForCondition(SessionController sessionController, SessionId sessionId, SessionValueKey<T> key, Duration timeout, Function<Optional<T>, Boolean> condition, WaitProc waitProc)
+            throws InterruptedException
+    {
+        long timeoutMsRemaining = timeout.toMillis();
+        while (timeoutMsRemaining > 0) {
+            Stopwatch stopwatch = Stopwatch.createStarted();
+
+            Optional<T> value = sessionController.getSessionValue(sessionId, key);
+            if (condition.apply(value)) {
+                break;
+            }
+
+            timeoutMsRemaining -= stopwatch.elapsed().toMillis();
+
+            if (timeoutMsRemaining > 0) {
+                stopwatch.reset().start();
+                waitProc.waitForKey(Duration.ofMillis(timeoutMsRemaining));
+                timeoutMsRemaining -= stopwatch.elapsed().toMillis();
+            }
+        }
+    }
+}
--- a/mcp/src/main/java/io/airlift/mcp/sessions/Signal.java
+++ b/mcp/src/main/java/io/airlift/mcp/sessions/Signal.java
@@ -1,0 +1,35 @@
+package io.airlift.mcp.sessions;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class Signal
+{
+    private final Lock lock = new ReentrantLock();
+    private final Condition condition = lock.newCondition();
+
+    public void signalAll()
+    {
+        lock.lock();
+        try {
+            condition.signalAll();
+        }
+        finally {
+            lock.unlock();
+        }
+    }
+
+    public boolean waitForSignal(long timeout, TimeUnit unit)
+            throws InterruptedException
+    {
+        lock.lock();
+        try {
+            return condition.await(timeout, unit);
+        }
+        finally {
+            lock.unlock();
+        }
+    }
+}
--- a/mcp/src/test/java/io/airlift/mcp/TestMemorySessionController.java
+++ b/mcp/src/test/java/io/airlift/mcp/TestMemorySessionController.java
@@ -11,10 +11,14 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.function.Function;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Future;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.IntStream;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.util.concurrent.Executors.newVirtualThreadPerTaskExecutor;
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class TestMemorySessionController
@@ -43,8 +47,8 @@ public class TestMemorySessionController
         List<TypeA> aValues = aKeys.stream().map(key -> new TypeA(key.name())).collect(toImmutableList());
         List<TypeB> bValues = bKeys.stream().map(key -> new TypeB(key.name())).collect(toImmutableList());
 
-        assertThat(list(controller, sessionId, TypeA.class, TypeA::name)).containsExactlyInAnyOrderElementsOf(aValues);
-        assertThat(list(controller, sessionId, TypeB.class, TypeB::name)).containsExactlyInAnyOrderElementsOf(bValues);
+        assertThat(list(controller, sessionId, TypeA.class)).containsExactlyInAnyOrderElementsOf(aValues);
+        assertThat(list(controller, sessionId, TypeB.class)).containsExactlyInAnyOrderElementsOf(bValues);
     }
 
     @Test
@@ -69,19 +73,104 @@ public class TestMemorySessionController
         assertThat(success).isFalse();
     }
 
-    private <T> List<T> list(SessionController controller, SessionId sessionId, Class<T> type, Function<T, String> mapper)
+    @Test
+    public void testConditions()
+            throws InterruptedException
+    {
+        SessionController controller = new MemorySessionController();
+        SessionId sessionId = controller.createSession(Optional.empty(), Optional.empty());
+
+        SessionValueKey<String> key = new SessionValueKey<>("key", String.class);
+
+        Semaphore semaphore = new Semaphore(0);
+
+        // wait for key to appear
+
+        newVirtualThreadPerTaskExecutor().execute(() -> {
+            try {
+                assertThat(semaphore.tryAcquire(5, TimeUnit.SECONDS)).isTrue();
+                controller.setSessionValue(sessionId, key, "value");
+            }
+            catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+        });
+
+        controller.blockUntilCondition(sessionId, key, Duration.ofSeconds(1), Optional::isPresent);
+
+        CountDownLatch latch1 = new CountDownLatch(1);
+        Future<Void> future = newVirtualThreadPerTaskExecutor().submit(() -> {
+            latch1.countDown();
+            controller.blockUntilCondition(sessionId, key, Duration.ofSeconds(1), Optional::isPresent);
+            return null;
+        });
+
+        assertThat(latch1.await(5, TimeUnit.SECONDS)).isTrue();
+        semaphore.release();
+
+        assertThat(future)
+                .succeedsWithin(5, TimeUnit.SECONDS);
+
+        // wait for change
+
+        CountDownLatch latch2 = new CountDownLatch(1);
+        future = newVirtualThreadPerTaskExecutor().submit(() -> {
+            latch2.countDown();
+            controller.blockUntilCondition(sessionId, key, Duration.ofSeconds(1), value -> value.isPresent() && value.get().equals("newValue"));
+            return null;
+        });
+
+        assertThat(latch1.await(5, TimeUnit.SECONDS)).isTrue();
+        TimeUnit.MILLISECONDS.sleep(50); // ensure the wait is in progress
+        controller.setSessionValue(sessionId, key, "newValue");
+
+        assertThat(future)
+                .succeedsWithin(5, TimeUnit.SECONDS);
+
+        // wait for computed change
+
+        CountDownLatch latch3 = new CountDownLatch(1);
+        future = newVirtualThreadPerTaskExecutor().submit(() -> {
+            latch3.countDown();
+            controller.blockUntilCondition(sessionId, key, Duration.ofSeconds(1), value -> value.isPresent() && value.get().equals("computed"));
+            return null;
+        });
+
+        assertThat(latch1.await(5, TimeUnit.SECONDS)).isTrue();
+        TimeUnit.MILLISECONDS.sleep(50); // ensure the wait is in progress
+        Optional<String> computed = controller.computeSessionValue(sessionId, key, _ -> Optional.of("computed"));
+        assertThat(computed).contains("computed");
+
+        assertThat(future)
+                .succeedsWithin(5, TimeUnit.SECONDS);
+
+        // wait for deletion
+
+        CountDownLatch latch4 = new CountDownLatch(1);
+        future = newVirtualThreadPerTaskExecutor().submit(() -> {
+            latch4.countDown();
+            controller.blockUntilCondition(sessionId, key, Duration.ofSeconds(1), Optional::isEmpty);
+            return null;
+        });
+
+        assertThat(latch1.await(5, TimeUnit.SECONDS)).isTrue();
+        TimeUnit.MILLISECONDS.sleep(50); // ensure the wait is in progress
+        controller.deleteSessionValue(sessionId, key);
+
+        assertThat(future)
+                .succeedsWithin(5, TimeUnit.SECONDS);
+    }
+
+    private <T> List<T> list(SessionController controller, SessionId sessionId, Class<T> type)
     {
         int pageSize = 12;
 
         List<T> results = new ArrayList<>();
         Optional<String> lastName = Optional.empty();
         do {
-            List<T> thisList = controller.listSessionValues(sessionId, type, pageSize, lastName)
-                    .stream()
-                    .map(Map.Entry::getValue)
-                    .collect(toImmutableList());
-            results.addAll(thisList);
-            lastName = (thisList.size() < pageSize) ? Optional.empty() : Optional.of(mapper.apply(thisList.getLast()));
+            List<Map.Entry<String, T>> thisList = controller.listSessionValues(sessionId, type, pageSize, lastName);
+            thisList.forEach(entry -> results.add(entry.getValue()));
+            lastName = (thisList.size() < pageSize) ? Optional.empty() : Optional.of(thisList.getLast().getKey());
         }
         while (lastName.isPresent());
         return results;


Randgalt:
> @Randgalt can you rebase your MCP PRs?

I'll do it soon

wendigo:
Thanks @Randgalt

Randgalt:
done

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
```diff
@@ -0,0 +1,189 @@
...
+        super.destroy();
+    }
+
+    @Override
+    protected void doGet(HttpServletRequest request, HttpServletResponse response)
+            throws IOException
+    {
+        boolean isMcpEndpoint = request.getRequestURI().endsWith(mcpEndpoint);
+        response.sendError(isMcpEndpoint ? SC_METHOD_NOT_ALLOWED : SC_NOT_FOUND);
+    }
```
codyzwief: We may want to consider overriding `doOptions` at some point. I know when testing MCP servers locally, some clients can tip over if not configured to respond to CORS preflight requests. Not necessary for this PR though.

mcp/src/main/java/io/airlift/mcp/reference/ReferenceModule.java
```diff
@@ -33,6 +29,7 @@ public class ReferenceModule
     public void configure(Binder binder)
     {
         binder.bind(ReferenceServer.class).asEagerSingleton();
+        binder.bind(ReferenceServerTransport.class).asEagerSingleton();
```
codyzwief: Do we want to consider making a new module for using this custom transport implementation, possibly using a different `McpModule.Mode`? Or a way of configuring which transport is used in the `McpModule`? I don't have a strong opinion on it, just throwing it out there.

mcp/src/test/java/io/airlift/mcp/TestMcp.java
```diff
@@ -189,7 +189,7 @@ public void testInvalidRpcRequests()
         FullJsonResponseHandler.JsonResponse<Object> response = httpClient.execute(request, createFullJsonResponseHandler(jsonCodec(new TypeToken<>() {})));
         assertThat(response.getStatusCode()).isEqualTo(400);
         assertThat(response.getResponseBody())
-                .isEqualTo("{\"message\":\"Both application/json and text/event-stream required in Accept header\"}");
+                .isEqualTo("{\"code\":-32600,\"message\":\"Both application/json and text/event-stream required in Accept header\"}");
```
codyzwief: Nit: for the code here, can we make sure it matches the appropriate `JsonRpcErrorCode` instead of just the integer code?

mcp/src/main/java/io/airlift/mcp/handler/RequestContextProvider.java
```diff
@@ -8,5 +8,14 @@
 
 public interface RequestContextProvider
 {
-    McpRequestContext get(HttpServletRequest request, HttpServletResponse response, Optional<Object> progressToken);
+    interface MessageWriter
+    {
+        void writeMessage(String data);
+
+        void writeMessage(String data, Optional<String> messageId);
```
codyzwief: Why do we want both of these? Which use cases does the server want to manually specify the message ID?

mcp/src/main/java/io/airlift/mcp/internal/InternalMessageWriter.java
```diff
@@ -0,0 +1,83 @@
...
+                writeMessage(data);
+            }
+            else {
+                // no messages were sent, keep it as a standard JSON response
+                writer.write(data);
+            }
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
```
codyzwief, line 43: Does it make sense to log exceptions (client disconnects, network blips etc)?

mcp/src/main/java/io/airlift/mcp/internal/InternalMessageWriter.java
```diff
@@ -0,0 +1,83 @@
...
+    public void writeMessage(String data, Optional<String> messageId)
+    {
+        if (hasBeenUpgraded.compareAndSet(false, true)) {
+            response.setContentType("text/event-stream");
+        }
+
+        try {
+            PrintWriter writer = response.getWriter();
+            messageId.ifPresent(id -> writer.write("id: " + encode(id) + "\n"));
+            writer.write("data: " + encode(data) + "\n\n");
```
codyzwief, line 62: should these be formatted as JSON?

mcp/src/main/java/io/airlift/mcp/reference/ReferenceMessageWriter.java
```diff
@@ -0,0 +1,83 @@
...
+    {
+        try {
+            response.getWriter().flush();
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private String encode(String str)
```
codyzwief: nit: static

mcp/src/main/java/io/airlift/mcp/sessions/SessionController.java
```diff
@@ -0,0 +1,60 @@
...
+import jakarta.servlet.http.HttpServletRequest;
+
+import java.util.Optional;
+import java.util.function.BiFunction;
+import java.util.function.UnaryOperator;
+
+public interface SessionController
+{
+    /**
+     * Create a mew sessopm and return its ID
```
codyzwief: nit: typos

mcp/src/test/java/io/airlift/mcp/TestMcpWithSessions.java
```diff
@@ -0,0 +1,133 @@
...
+    private final Closer closer = Closer.create();
+    private final Client client1;
+    private final Client client2;
+    private final TestingServer testingServer;
+
+    private record Client(McpSyncClient mcpClient, List<String> logs, List<String> progress) {}
+
+    public TestMcpWithSessions()
+    {
+        McpIdentityMapper identityMapper = _ -> new McpIdentity.Authenticated<>("yep");
```
codyzwief: Since this class is now using the MCP client to test, would it be possible to test with authentication enabled? Specifically, I'm thinking of the following cases:

* initialize should fail if the header is invalid, so the client doesn't even start up, with the appropriate error code
* If a session is not found then we should throw the appropriate error.
* Some of the session management cases from https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management such as
  * > The server MAY terminate the session at any time, after which it MUST respond to requests containing that session ID with HTTP 404 Not Found.
  *  > Clients that no longer need a particular session (e.g., because the user is leaving the client application) SHOULD send an HTTP DELETE to the MCP endpoint with the Mcp-Session-Id header, to explicitly terminate the session.

mcp/src/main/java/io/airlift/mcp/sessions/SessionController.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    SessionId createSession(HttpServletRequest request);
+
+    /**
+     * @return {@code true} if the session ID is valid, {@code false} otherwise
+     */
+    boolean validateSession(SessionId sessionId);
```
codyzwief, line 43: What will this be used for?

mcp/src/main/java/io/airlift/mcp/sessions/SessionController.java
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * Delete the session or do nothing if the session ID is invalid
+     */
+    void deleteSession(SessionId sessionId);
```
codyzwief, line 48: Suppose a request comes in to delete a session (or write any sort of value), but the authenticated identity of the request isn't allowed to do that. How would a consumer of this library plug into this implementation to perform the check that this request is allowed to delete this session?

One other question, for any implementation, is the expectation that servers need to implement their own stale session cleanup mechanism?

mcp/src/main/java/io/airlift/mcp/reference/ReferenceRequestContextProvider.java
```diff
@@ -51,6 +60,18 @@ public void sendProgress(double progress, double total, String message)
                 sendNotification("notifications/progress", Optional.of(notification));
             }
 
+            @Override
+            public void sendLog(LoggingLevel level, Optional<String> logger, Optional<Object> data)
+            {
+                SessionId sessionId = requireSessionId(request);
```
codyzwief: This method eventually filters down and throws an `INVALID_REQUEST` McpException. IMO, if a server tries to send logging but doesn't configure the `SessionController`, that's an internal error instead. Do you think that could be modified?

mcp/src/main/java/io/airlift/mcp/reference/ReferenceRequestContextProvider.java
```diff
@@ -51,6 +60,18 @@ public void sendProgress(double progress, double total, String message)
...
+                LoggingLevel sessionLoggingLevel = sessionController.flatMap(controller -> controller.getSessionValue(sessionId, LOGGING_LEVEL))
+                        .orElseThrow(() -> exception("Session is invalid"));
```
codyzwief: same here

mcp/src/main/java/io/airlift/mcp/reference/ReferenceFilter.java
```diff
@@ -63,30 +63,35 @@ protected void doFilter(HttpServletRequest request, HttpServletResponse response
...
+                        log.error(error.cause(), "An error was thrown during MCP authentication");
+                        throw error.cause();
+                    }
                 }
             }
+            catch (McpException mcpException) {
+                log.debug(mcpException, "Request: %s", request.getRequestURI());
+                JsonRpcErrorDetail errorDetail = mcpException.errorDetail();
+                throw new McpError(new McpSchema.JSONRPCResponse.JSONRPCError(errorDetail.code(), errorDetail.message(), errorDetail.data()));
+            }
```
codyzwief: The `catch` block here doesn't actually include the identity mapper doing its thing, just errors thrown by the `Error` case. Could this catch instead be built into the `Error` case if that's intentional?

mcp/src/main/java/io/airlift/mcp/McpRequestContext.java
```diff
@@ -1,10 +1,25 @@
...
+    default void sendLog(LoggingLevel level, String message)
+    {
+        sendLog(level, Optional.empty(), Optional.of(message));
+    }
+
+    default void sendLog(LoggingLevel level, Optional<String> logger, Optional<Object> data)
+    {
+        // only implemented when sessions are configured
+
+        throw new UnsupportedOperationException();
```
codyzwief, line 25: This should probably throw an `McpException`. WDYT?

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
```diff
@@ -141,21 +157,28 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
...
     {
-        JSONRPCResponse rpcResponse = mcpHandler
-                .handleRequest(transportContext, rpcRequest)
-                .contextWrite(ctx -> ctx.put(KEY, transportContext))
-                .block();
+        Mono<JSONRPCResponse> rpcResponseMono = switch (rpcRequest.method()) {
+            case METHOD_INITIALIZE -> sessionController.map(controller -> handleInitialize(controller, transportContext, rpcRequest)).orElseGet(() -> mcpHandler.handleRequest(transportContext, rpcRequest));
+            case METHOD_LOGGING_SET_LEVEL -> sessionController.map(controller -> handleSetLoggingLevel(controller, transportContext, rpcRequest)).orElseGet(() -> mcpHandler.handleRequest(transportContext, rpcRequest));
+            default -> mcpHandler.handleRequest(transportContext, rpcRequest);
+        };
```
codyzwief: Should we add a todo to support [cancellations](https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/cancellation)?

Randgalt: I had a look at cancellation and it doesn't seem to be supported anywhere. But, we can do it ourselves if we need to. The reference SDK doesn't support it nor the Inspector.

mcp/src/main/java/io/airlift/mcp/reference/ReferenceModule.java
Randgalt: I don't think users of this library will care what the underlying implementation is. They might have a reason to use a custom, app-specific implementation, but I don't think Airlift needs to support more than one. A separate module would be overkill IMO.

mcp/src/main/java/io/airlift/mcp/handler/RequestContextProvider.java
Randgalt: The ID in SSE can be used to correlate responses. A caller of this method might have the need to set it.

mcp/src/main/java/io/airlift/mcp/internal/InternalMessageWriter.java
Randgalt: I believe there are filters, etc. that get installed that do that. Excessive logging can cause its own problems. But, I don't have a strong feeling on this.

mcp/src/main/java/io/airlift/mcp/internal/InternalMessageWriter.java
Randgalt: For the SSE spec data is simply data. The caller should format the data however they want.

mcp/src/main/java/io/airlift/mcp/sessions/SessionController.java
Randgalt: Sessions can expire or be invalid. All we get is a header with an opaque value. It needs to be validated.

mcp/src/main/java/io/airlift/mcp/sessions/SessionController.java
Randgalt: Here's what the spec says:

> Clients that no longer need a particular session (e.g., because the user is leaving the client application) SHOULD send an HTTP DELETE to the MCP endpoint with the MCP-Session-Id header, to explicitly terminate the session.
The server MAY respond to this request with HTTP 405 Method Not Allowed, indicating that the server does not allow clients to terminate sessions.

So, the session implementation should throw a 405 if it's not allowed.

mcp/src/main/java/io/airlift/mcp/reference/Mapper.java
```diff
@@ -85,7 +89,24 @@ static McpSchema.Role mapRole(Role ourRole)
         return McpSchema.Role.valueOf(ourRole.name().toUpperCase());
     }
 
-    static McpStatelessServerFeatures.SyncResourceSpecification mapResource(Resource ourResource, ResourceHandler ourHandler)
+    static McpRequestContext buildMcpRequestContext(McpTransportContext mcpTransportContext, Optional<Map<String, Object>> meta, RequestContextProvider requestContextProvider)
+    {
```
wendigo:
```
        Optional<Object> progressToken = meta.flatMap(m -> switch (m.get("progressToken")) {
            case null -> Optional.empty();
            case Number number -> Optional.of(number.longValue());
            case Object obj -> Optional.of(String.valueOf(obj));
        });

```

mcp/src/main/java/io/airlift/mcp/reference/ReferenceRequestContextProvider.java
```diff
@@ -0,0 +1,73 @@
...
+            {
+                return request;
+            }
+
+            @SuppressWarnings("SwitchStatementWithTooFewBranches")
+            @Override
+            public void sendProgress(double progress, double total, String message)
+            {
+                Optional<Object> appliedProgressToken = progressToken.map(token -> switch (token) {
+                    case Number n -> Optional.of(n.longValue());
```
wendigo: case Number number

mcp/src/main/java/io/airlift/mcp/reference/ReferenceRequestContextProvider.java
```diff
@@ -0,0 +1,73 @@
...
+                    default -> progressToken;
+                });
+
+                ProgressNotification notification = new ProgressNotification(appliedProgressToken, message, OptionalDouble.of(progress), OptionalDouble.of(total));
+                sendNotification("notifications/progress", Optional.of(notification));
+            }
+
+            private void sendNotification(String method, Optional<Object> params)
+            {
+                try {
```
wendigo: should we close writer here?

If so `try (PrintWriter writer = response.getWriter())`

mcp/src/main/java/io/airlift/mcp/reference/ReferenceRequestContextProvider.java
```diff
@@ -0,0 +1,73 @@
...
+                    PrintWriter writer = response.getWriter();
+                    if (!(writer instanceof SsePrintWriter ssePrintWriter)) {
+                        throw exception("Response writer is not an SsePrintWriter");
+                    }
+                    JsonRpcRequest<?> notification = params.map(p -> JsonRpcRequest.buildNotification(method, p)).orElseGet(() -> JsonRpcRequest.buildNotification(method));
```
wendigo: don't use one letter variable names,

`params.map(values -> ...`

mcp/src/main/java/io/airlift/mcp/McpMetadata.java
```diff
@@ -8,7 +8,7 @@
 
 public record McpMetadata(String uriPath, Implementation implementation, Optional<String> instructions, boolean tools, boolean prompts, boolean resources)
 {
-    public static final String CONTEXT_REQUEST_KEY = McpMetadata.class.getName();
+    public static final String CONTEXT_REQUEST_KEY = McpMetadata.class.getName() + ".request";
```
wendigo: can we just se "mcp.request" ?

mcp/src/main/java/io/airlift/mcp/reference/Mapper.java
```diff
@@ -85,7 +89,24 @@ static McpSchema.Role mapRole(Role ourRole)
...
+        Optional<Object> progressToken = meta.flatMap(m -> {
```
wendigo: Can we represent token as an class/record or a sealed interface?

```
record ProgressToken(Object value)
{
}
```

or 

```
    sealed interface ProgressToken
    {
        record StringProgressToken(String token) implements ProgressToken
        {
        }

        record LongProgressToken(long token) implements ProgressToken
        {
        }
    }
```

`Object` tells nothing

mcp/src/main/java/io/airlift/mcp/McpMetadata.java
Randgalt: Using the class name is much safer. Http request/response attributes are shared and there is no namespacing. Usng the class name creates a quasi-namespace.

mcp/src/main/java/io/airlift/mcp/reference/Mapper.java
Randgalt: It's an opaque value per the spec. It can be a number or a string. Regardless, it's supposed to be passed unchanged back to the client. It's the client's context.

mcp/src/main/java/io/airlift/mcp/reference/SseResponseWrapper.java
```diff
@@ -0,0 +1,51 @@
...
+    {
+        super(response);
+    }
+
+    @Override
+    public PrintWriter getWriter()
+            throws IOException
+    {
+        int status = getStatus();
+        boolean useSse = (status >= SC_OK) && (status < SC_MULTIPLE_CHOICES);
```
wendigo: redundant ()

just:
```
boolean useSse = status >= SC_OK && status < SC_MULTIPLE_CHOICES;
````
```

mcp/src/main/java/io/airlift/mcp/reflection/Predicates.java
```diff
@@ -21,7 +22,7 @@
 
 public interface Predicates
 {
-    Predicate<MethodParameter> isHttpRequest = methodParameter -> methodParameter instanceof HttpRequestParameter;
+    Predicate<MethodParameter> isHttpRequestOrContext = methodParameter -> (methodParameter instanceof HttpRequestParameter) || (methodParameter instanceof McpRequestContextParameter);
```
wendigo: redundant () 

just:

```
methodParameter instanceof HttpRequestParameter || methodParameter instanceof McpRequestContextParameter;
```

mcp/src/main/java/io/airlift/mcp/reference/ReferenceModule.java
wendigo: I agree with Jordan here. We don't want to build general-purpose library but something tailored to our needs so we don't need to overcomplicate it (beyond the complexity of the specification)

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
```diff
@@ -0,0 +1,189 @@
...
+import static jakarta.servlet.http.HttpServletResponse.SC_NOT_FOUND;
+import static jakarta.servlet.http.HttpServletResponse.SC_OK;
+import static jakarta.servlet.http.HttpServletResponse.SC_SERVICE_UNAVAILABLE;
+import static jakarta.ws.rs.core.HttpHeaders.ACCEPT;
+import static jakarta.ws.rs.core.MediaType.APPLICATION_JSON;
+import static jakarta.ws.rs.core.MediaType.SERVER_SENT_EVENTS;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Objects.requireNonNull;
+
+// based off of io.modelcontextprotocol.server.transport.HttpServletStatelessServerTransport.java
```
wendigo: based on ? :)

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
```diff
@@ -0,0 +1,189 @@
...
+    @Override
+    public Mono<Void> closeGracefully()
+    {
+        return Mono.fromRunnable(() -> isClosing = true);
+    }
+
+    @Override
+    public void destroy()
+    {
+        closeGracefully().block();
```
wendigo: just `isClosing = true` ? Do we need Mono here? :)

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
```diff
@@ -0,0 +1,189 @@
...
+    protected void doPost(HttpServletRequest request, HttpServletResponse response)
+            throws IOException
+    {
+        String requestURI = request.getRequestURI();
+        if (!requestURI.endsWith(mcpEndpoint)) {
+            response.sendError(SC_NOT_FOUND);
+            return;
+        }
+
+        if (isClosing) {
```
wendigo: can we check it earlier? as a first statement?

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
Randgalt: We need to call `closeGracefully()` and it returns a `Mono` immediately. So, yeah. My vote is stop using the reference SDK altogether. It's not giving us much. I posted a straw man implementation that removes it.

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
```diff
@@ -0,0 +1,189 @@
...
+
+        McpTransportContext transportContext = McpTransportContext.create(ImmutableMap.of(CONTEXT_REQUEST_KEY, request));
+
+        String accept = request.getHeader(ACCEPT);
+        if (accept == null || !(accept.contains(APPLICATION_JSON) && accept.contains(SERVER_SENT_EVENTS))) {
+            responseError(response, SC_BAD_REQUEST, invalidRequest("Both application/json and text/event-stream required in Accept header"));
+            return;
+        }
+
+        try {
```
wendigo:
```
try (BufferedReader reader = request.getReader()) {
```

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
```diff
@@ -0,0 +1,189 @@
...
+            BufferedReader reader = request.getReader();
+            String body = reader.lines().collect(Collectors.joining("\n"));
```
wendigo: Please add `TODO: use Reader.readAllLines once on JDK 25` as a reminder for myself ;)

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
Randgalt: That would be incorrect. We should return not found if it isn't an MCP request.

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
```diff
@@ -0,0 +1,189 @@
...
+            responseError(response, SC_INTERNAL_SERVER_ERROR, internalError("Unexpected error: " + e.getMessage()));
+        }
+    }
+
+    private void handleRpcNotification(McpTransportContext transportContext, HttpServletResponse response, JSONRPCNotification jsonrpcNotification)
+    {
+        mcpHandler.handleNotification(transportContext, jsonrpcNotification)
+                .contextWrite(ctx -> ctx.put(KEY, transportContext))
+                .block();
+        response.setStatus(SC_ACCEPTED);
```
wendigo: can we set status first and then handle?

mcp/src/main/java/io/airlift/mcp/reference/Mapper.java
Randgalt: And what if it's a double? This is too much overhead for something we will never interrogate.

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
```diff
@@ -0,0 +1,189 @@
...
+        JSONRPCResponse rpcResponse = mcpHandler
+                .handleRequest(transportContext, rpcRequest)
+                .contextWrite(ctx -> ctx.put(KEY, transportContext))
+                .block();
+
+        response.setContentType(APPLICATION_JSON);
+        response.setCharacterEncoding(UTF_8.name());
+        response.setStatus(SC_OK);
+
+        String jsonResponseText = mcpJsonMapper.writeValueAsString(rpcResponse);
```
wendigo: inline jsonResponseTest

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
```diff
@@ -0,0 +1,189 @@
...
+
+        PrintWriter writer = response.getWriter();
```
wendigo: `try (PrintWriter writer = response.getWriter())`

if this is not the case - add a comment that

mcp/src/main/java/io/airlift/mcp/handler/RequestContextProvider.java
```diff
@@ -8,5 +8,14 @@
 
 public interface RequestContextProvider
 {
-    McpRequestContext get(HttpServletRequest request, HttpServletResponse response, Optional<Object> progressToken);
+    interface MessageWriter
+    {
+        void writeMessage(String data);
```
wendigo: can this delegate to other write?

```
default void writeMessage(String data)
{
    writeMessage(data, Optional.empty());
}
```

mcp/src/main/java/io/airlift/mcp/internal/InternalMessageWriter.java
```diff
@@ -0,0 +1,83 @@
...
+
+    ReferenceMessageWriter(HttpServletResponse response)
+    {
+        this.response = requireNonNull(response, "response is null");
+    }
+
+    void write(String data)
+    {
+        try {
+            PrintWriter writer = response.getWriter();
```
wendigo, line 32: `try (PrintWriter writer = response.getWriter())`

mcp/src/main/java/io/airlift/mcp/internal/InternalMessageWriter.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Override
+    public void writeMessage(String data, Optional<String> messageId)
+    {
+        if (hasBeenUpgraded.compareAndSet(false, true)) {
+            response.setContentType("text/event-stream");
+        }
+
+        try {
+            PrintWriter writer = response.getWriter();
```
wendigo, line 60: add a comment that this writer can't be closed here

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
```diff
@@ -118,7 +123,7 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
 
             JSONRPCMessage message = deserializeJsonRpcMessage(mcpJsonMapper, body);
             switch (message) {
-                case JSONRPCRequest rpcRequest -> handleRpcRequest(transportContext, response, rpcRequest);
+                case JSONRPCRequest rpcRequest -> handleRpcRequest(transportContext, response, rpcRequest, messageWriter);
```
wendigo: i think that messageWriter should go after transportContext but before response and request

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
```diff
@@ -141,21 +157,28 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
...
     {
-        mcpHandler.handleNotification(transportContext, jsonrpcNotification)
-                .contextWrite(ctx -> ctx.put(KEY, transportContext))
-                .block();
+        Mono<Void> mono = switch (rpcNotification.method()) {
+            case METHOD_NOTIFICATION_INITIALIZED -> Mono.empty();  // do nothing - just acknowledge
+            default -> mcpHandler.handleNotification(transportContext, rpcNotification);
+        };
+
+        mono.contextWrite(ctx -> ctx.put(KEY, transportContext)).block();
```
wendigo: what's the point of having reactive API and using it in blocking fashion?

wendigo (approved):

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
Randgalt: The sync code came later and they just wrapped all the Mono stuff with sync wrappers. It's not a very good general purpose library tbh. It feels like internal code that they open sourced.

mcp/src/main/java/io/airlift/mcp/reference/ReferenceRequestContextProvider.java
Randgalt: @wendigo I thought if was the container's responsibility to close the writer, no? We are already flushing.

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
Randgalt: @wendigo same as other. Are we supposed to close the writer? I always thought that that was the container's responsibility.

mcp/src/main/java/io/airlift/mcp/handler/RequestContextProvider.java
Randgalt: It does internally. I'd rather not expose that in the interface as it's an implementation detail.

mcp/src/main/java/io/airlift/mcp/reference/ReferenceFilter.java
Randgalt: It also catches errors from `transport.service(request, response);`. The intent is to catch any McpException and convert it a proper error. I took another look at it and it seems correct to me. What is is missing?

mcp/src/main/java/io/airlift/mcp/McpRequestContext.java
Randgalt: This code should never execute. There must always be an `McpRequestContext` implementation. So, I think the UOE is correct here.

mcp/src/test/java/io/airlift/mcp/TestMcpWithSessions.java
Randgalt: Oof - good call. I completely missed a bunch of things regarding sessions. It's all in now.

mcp/src/main/java/io/airlift/mcp/sessions/SessionController.java
Randgalt: > is the expectation that servers need to implement their own stale session cleanup mechanism?

Yes. It would be hard to write a general purpose cleaner. For a DB, it can be done with one query, for example or maybe Row TTLs, etc.

mcp/src/test/java/io/airlift/mcp/TestMcp.java
codyzwief, line 1: One thing IMO that might be helpful is to treat this as an abstract base class, and only the tests that are supported in all configurations are run here, so that way we can test the same MCP functionality with different combinations of what's wired up. For example, it would be great if we had a test class that wired up an `identityMapper` but no `sessionController`, and finally, have both the `identityMapper` and `sessionController` wired up. The latter would test logging + notifications, and be able to test some of the session deletion functionality. IMO this is ideal because it is self-documenting what needs to be configured to support which parts of the MCP spec. WDYT? Do you think that's worth the effort?

mcp/src/main/java/io/airlift/mcp/reference/ReferenceServerTransport.java
Randgalt: FYI - there is now a cancellation PR in this series

mcp/src/test/java/io/airlift/mcp/TestMcp.java
Randgalt: There's something similar in a PR later in this series. I can pull it down to this if you want though.

mcp/src/main/java/io/airlift/mcp/McpIdentityMapper.java
```diff
@@ -1,9 +1,16 @@
...
 
 import io.airlift.mcp.model.McpIdentity;
+import io.airlift.mcp.sessions.SessionId;
 import jakarta.servlet.http.HttpServletRequest;
 
+import java.util.Optional;
+
 public interface McpIdentityMapper
 {
-    McpIdentity map(HttpServletRequest request);
```
wendigo, line 8: I don't like the changed API. SessionId is extracted from request so why can't McpIdentityMapper implementation handle that?

Randgalt: It's more of a convenience. But, sure, I can return to the previous.

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -125,19 +133,29 @@ protected void doFilter(HttpServletRequest request, HttpServletResponse response
...
+            case DELETE -> handleMcpDeleteRequest(request, response);
             default -> response.setStatus(SC_NOT_FOUND);
         }
     }
 
-    private void handleMcpPostRequest(HttpServletRequest request, HttpServletResponse response)
+    private void handleMcpDeleteRequest(HttpServletRequest request, HttpServletResponse response)
+    {
+        sessionController.ifPresentOrElse(controller -> {
+            SessionId sessionId = requireSessionId(request);
```
wendigo: you can inline this variable in the call

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -207,6 +228,18 @@ private void handleRpcRequest(HttpServletRequest request, HttpServletResponse re
         messageWriter.flushMessages();
     }
 
+    private void validateSession(HttpServletRequest request, JsonRpcRequest<?> rpcRequest)
+    {
+        if (!METHOD_INITIALIZE.equals(rpcRequest.method())) {
+            sessionController.ifPresent(controller -> {
+                SessionId sessionId = requireSessionId(request);
```
wendigo: inline variable

mcp/src/main/java/io/airlift/mcp/internal/InternalMcpServer.java
```diff
@@ -175,14 +195,22 @@ public void removeCompletion(CompleteReference reference)
         completions.remove(completionKey(reference));
     }
 
-    InitializeResult initialize(InitializeRequest initializeRequest)
+    InitializeResult initialize(HttpServletResponse response, McpIdentity.Authenticated<?> authenticated, InitializeRequest initializeRequest)
     {
+        boolean sessionsEnabled = sessionController.map(controller -> {
+            SessionId sessionId = controller.createSession(Optional.of(authenticated), Optional.of(sessionTimeout));
```
wendigo, line 201: can session be created for unauthenticated client?

mcp/src/main/java/io/airlift/mcp/internal/InternalMcpServer.java
```diff
@@ -263,11 +291,21 @@ CompleteResult completionComplete(HttpServletRequest request, InternalMessageWri
...
-        McpRequestContext requestContext = new InternalRequestContext(objectMapper, request, messageWriter, progressToken(completeRequest));
+        McpRequestContext requestContext = new InternalRequestContext(objectMapper, sessionController, request, messageWriter, progressToken(completeRequest));
 
         return completionEntry.handler().complete(requestContext, completeRequest);
     }
 
+    Object setLoggingLevel(HttpServletRequest request, SetLevelRequest setLevelRequest)
+    {
+        SessionController localSessionController = sessionController.orElseThrow(() -> exception(INVALID_REQUEST, "set logging level not supported"));
+        SessionId sessionId = requireSessionId(request);
```
wendigo: inline variable

mcp/src/main/java/io/airlift/mcp/internal/InternalRequestContext.java
```diff
@@ -51,6 +61,37 @@ public void sendProgress(double progress, double total, String message)
...
+                .orElseThrow(() -> exception("Session is invalid"));
+        if (level.level() >= sessionLoggingLevel.level()) {
+            LoggingMessageNotification logNotification = new LoggingMessageNotification(level, logger, data);
+            internalSendMessage(NOTIFICATION_MESSAGE, Optional.of(logNotification));
+        }
+    }
+
+    static SessionId requireSessionId(HttpServletRequest request)
+    {
+        return optionalSessionId(request).orElseThrow(() -> exception("Missing MCP_SESSION_ID header in request"));
```
wendigo: can you use const here?

wendigo (approved):

mcp/src/main/java/io/airlift/mcp/internal/InternalMcpServer.java
Randgalt: IMO we don't want that. Auth is on the outer edge of everything. If you're not authenticated nothing should work. That said, there are some proposals in the MCP spec to allow metadata queries that are not authenticated. We can cross that when those spec changes land. cc @codyzwief

mcp/src/main/java/io/airlift/mcp/internal/InternalRequestContext.java
Randgalt: For the message?

wendigo: for the `MCP_SESSION_ID` header name

PR #1650, opened 2025-11-20
[MCP Server] MCP tasks support

Randgalt:
🎥 [Video of sessions and tasks combining to enable elicitations and sampling](https://drive.google.com/file/d/120A4K3yJCLTaX8oAQyq9MuZ_FDlAqqHt/view?usp=share_link)

Docs: https://github.com/Randgalt/airlift/blob/jordanz/mcp-tasks/mcp/tasks.md

Early support for MCP tasks

Tasks solve a number of issues in earlier versions of the MCP spec that make it
very difficult for horizontally scaled MCP server implementations to work well.

This commit adds support for the draft specification of Tasks adding
a backwards compatible implementation using MCP sessions. It is assumed
that a future version of the reference Java SDK will add direct support
for tasks in some manner. With this implementation in Airlift we can
start to use tasks but still work with older clients. When the Java
SDK adds support we can check when a client connects if they support
tasks directly and, if so, substitute the Java SDK's implementation.


mcp/src/main/java/io/airlift/mcp/model/CallToolRequest.java
```diff
@@ -8,18 +8,24 @@
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static java.util.Objects.requireNonNull;
 
-public record CallToolRequest(String name, Map<String, Object> arguments, Optional<Map<String, Object>> meta)
+public record CallToolRequest(String name, Map<String, Object> arguments, Optional<Map<String, Object>> meta, Optional<TaskMetadata> task)
```
He-Pin, line 11: Nice, just read you implemented it without the mcp java sdk

mcp/src/main/java/io/airlift/mcp/model/TaskHint.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+import static java.util.Locale.ROOT;
+
+public enum TaskHint
+{
+    NEVER,
+    OPTIONAL,
+    ALWAYS;
```
He-Pin: taskSupport

mcp/src/main/java/io/airlift/mcp/model/CallToolRequest.java
Randgalt: We use the reference SDK as an implementation detail

He-Pin: That's excellent. I think your decision allows for rapid implementation and a high degree of autonomy. I actually hope to achieve this entirely based on Pekko.

mcp/src/main/java/io/airlift/mcp/reference/ReferenceModule.java
```diff
@@ -30,36 +34,32 @@ public class ReferenceModule
...
+        ServerCapabilities serverCapabilities = new ServerCapabilities(
                 null,
                 null,
-                metadata.prompts() ? new McpSchema.ServerCapabilities.PromptCapabilities(false) : null,
-                metadata.resources() ? new McpSchema.ServerCapabilities.ResourceCapabilities(false, false) : null,
-                metadata.tools() ? new McpSchema.ServerCapabilities.ToolCapabilities(false) : null);
+                sessionsEnabled ? new LoggingCapabilities() : null,
+                metadata.prompts() ? new PromptCapabilities(false) : null,
+                metadata.resources() ? new ResourceCapabilities(false, false) : null,
+                metadata.tools() ? new ToolCapabilities(false) : null);
```
He-Pin, line 67: will need a TaskCapabilities

PR #1652, opened 2025-11-24
Do not stop http server on shutdown

wendigo:
This registers a shutdown hook thread that interferes with already registered shutdown hook thread managed by LifeCycle.


http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -168,7 +168,6 @@ public HttpServer(
 
         server = new Server(threadPool, createScheduler(name + "-scheduler"), createByteBufferPool(maxBufferSize, config));
         server.setName(name);
-        server.setStopAtShutdown(true);
```
findepi, line 171: `server.setStopAtShutdown(false);` and a code comment would be nicer

findepi (approved):

PR #1659, opened 2025-11-27
Add HttpClient.executeAsync that accepts Executor

wendigo:

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpClient.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClient.java
@@ -18,6 +18,7 @@ package io.airlift.http.client;
 import com.google.common.util.concurrent.ListenableFuture;
 
 import java.io.Closeable;
+import java.util.concurrent.Executor;
 
 public interface HttpClient
         extends Closeable
@@ -25,13 +26,24 @@ public interface HttpClient
     <T, E extends Exception> T execute(Request request, ResponseHandler<T, E> responseHandler)
             throws E;
 
-    <T, E extends Exception> HttpResponseFuture<T> executeAsync(Request request, ResponseHandler<T, E> responseHandler);
-
     /**
-     * Executes the given request and returns a response stream.
+     * Executes request asynchronously running {@link ResponseHandler} on the client's http thread pool.
+     * Blocks client's http thread while reading.
+     *
+     * @deprecated {@link HttpClient#executeAsync(Executor, Request, ResponseHandler)} instead as this variant makes does
+     * not buffer response while processing and makes it explicit which thread is used while calling ResponseHandler.
+     *
      * <p>
      * <b>Note:</b> {@link Request#getMaxResponseContentLength} is ignored.
      */
+    @Deprecated
+    <T, E extends Exception> HttpResponseFuture<T> executeAsync(Request request, ResponseHandler<T, E> responseHandler);
+
+    /**a
+     * Executes the given request and returns a response stream. The response handler is executed using the provided executor.
+     */
+    <T, E extends Exception> HttpResponseFuture<T> executeAsync(Executor executor, Request request, ResponseHandler<T, E> responseHandler);
+
     StreamingResponse executeStreaming(Request request);
 
     RequestStats getStats();
--- a/http-client/src/main/java/io/airlift/http/client/testing/TestingHttpClient.java
+++ b/http-client/src/main/java/io/airlift/http/client/testing/TestingHttpClient.java
@@ -3,6 +3,7 @@ package io.airlift.http.client.testing;
 import com.google.common.util.concurrent.ForwardingListenableFuture;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningExecutorService;
+import io.airlift.concurrent.ExecutorServiceAdapter;
 import io.airlift.http.client.HttpClient;
 import io.airlift.http.client.Request;
 import io.airlift.http.client.RequestStats;
@@ -11,6 +12,7 @@ import io.airlift.http.client.ResponseHandler;
 import io.airlift.http.client.StreamingResponse;
 import io.airlift.units.Duration;
 
+import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
@@ -42,6 +44,7 @@ public class TestingHttpClient
     }
 
     @Override
+    @Deprecated
     public <T, E extends Exception> HttpResponseFuture<T> executeAsync(Request request, ResponseHandler<T, E> responseHandler)
     {
         requireNonNull(request, "request is null");
@@ -50,6 +53,19 @@ public class TestingHttpClient
 
         AtomicReference<String> state = new AtomicReference<>("SENDING_REQUEST");
         ListenableFuture<T> future = executor.submit(() -> execute(request, responseHandler, state));
+
+        return new TestingHttpResponseFuture<>(future, state);
+    }
+
+    @Override
+    public <T, E extends Exception> HttpResponseFuture<T> executeAsync(Executor executor, Request request, ResponseHandler<T, E> responseHandler)
+    {
+        requireNonNull(request, "request is null");
+        requireNonNull(responseHandler, "responseHandler is null");
+        checkState(!closed.get(), "client is closed");
+
+        AtomicReference<String> state = new AtomicReference<>("SENDING_REQUEST");
+        ListenableFuture<T> future = listeningDecorator(ExecutorServiceAdapter.from(executor)).submit(() -> execute(request, responseHandler, state));
 
         return new TestingHttpResponseFuture<>(future, state);
     }
--- a/http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
+++ b/http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
@@ -1,0 +1,45 @@
+package io.airlift.http.client.jetty;
+
+import io.airlift.http.client.AbstractHttpClientTest;
+import io.airlift.http.client.HttpClientConfig;
+import io.airlift.http.client.Request;
+import io.airlift.http.client.ResponseHandler;
+import io.airlift.http.client.StreamingResponse;
+
+import java.util.Optional;
+import java.util.concurrent.Executors;
+
+import static io.airlift.concurrent.Threads.virtualThreadsNamed;
+
+public class TestJettyAsyncExecutorHttpClient
+        extends AbstractHttpClientTest
+{
+    @Override
+    protected HttpClientConfig createClientConfig()
+    {
+        return new HttpClientConfig()
+                .setHttp2Enabled(false);
+    }
+
+    @Override
+    public Optional<StreamingResponse> executeStreamingRequest(CloseableTestHttpServer server, Request request)
+    {
+        return Optional.empty();
+    }
+
+    @Override
+    public <T, E extends Exception> T executeRequest(CloseableTestHttpServer server, Request request, ResponseHandler<T, E> responseHandler)
+            throws Exception
+    {
+        return executeRequest(server, createClientConfig(), request, responseHandler);
+    }
+
+    @Override
+    public <T, E extends Exception> T executeRequest(CloseableTestHttpServer server, HttpClientConfig config, Request request, ResponseHandler<T, E> responseHandler)
+            throws Exception
+    {
+        try (JettyHttpClient client = server.createClient(config)) {
+            return executeAsync(Executors.newThreadPerTaskExecutor(virtualThreadsNamed("async-executor-pool#v")), client, request, responseHandler);
+        }
+    }
+}
--- a/discovery/src/main/java/io/airlift/discovery/client/HttpDiscoveryAnnouncementClient.java
+++ b/discovery/src/main/java/io/airlift/discovery/client/HttpDiscoveryAnnouncementClient.java
@@ -39,11 +39,14 @@ import java.net.UnknownHostException;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CancellationException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
 import static com.google.common.util.concurrent.Futures.immediateFailedFuture;
 import static com.google.common.util.concurrent.Futures.immediateFuture;
+import static io.airlift.concurrent.Threads.virtualThreadsNamed;
 import static io.airlift.http.client.HttpUriBuilder.uriBuilderFrom;
 import static io.airlift.http.client.JsonBodyGenerator.jsonBodyGenerator;
 import static io.airlift.http.client.Request.Builder.prepareDelete;
@@ -54,6 +57,8 @@ import static java.util.Objects.requireNonNull;
 public class HttpDiscoveryAnnouncementClient
         implements DiscoveryAnnouncementClient
 {
+    private final Executor responseExecutor = Executors.newThreadPerTaskExecutor(virtualThreadsNamed("discovery-announcement-response#v"));
+
     private static final MediaType MEDIA_TYPE_JSON = MediaType.create("application", "json");
     private static final Logger log = Logger.get(HttpDiscoveryAnnouncementClient.class);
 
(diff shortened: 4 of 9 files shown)


http-client/src/main/java/io/airlift/http/client/HttpClient.java
```diff
@@ -25,6 +25,10 @@ public interface HttpClient
     <T, E extends Exception> T execute(Request request, ResponseHandler<T, E> responseHandler)
             throws E;
 
+    /**
+     * Executes request asynchronously running responseHandler on the client's http thread pool.
+     */
+    @Deprecated
```
findepi: add `@deprecated` javadoc tag explaining why it's deprecated and what to use instead.

This generally feels better if it's part of `Add HttpClient.executeAsync that accepts Executor` commit.

wendigo: I'll squash

http-client/src/main/java/io/airlift/http/client/HttpClient.java
```diff
@@ -31,6 +32,11 @@ <T, E extends Exception> T execute(Request request, ResponseHandler<T, E> respon
     @Deprecated
     <T, E extends Exception> HttpResponseFuture<T> executeAsync(Request request, ResponseHandler<T, E> responseHandler);
 
+    /**
+     * Executes request asynchronously running responseHandler on the provided executor.
```
findepi: It also does the blocking read on that executor.

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len)
+            throws IOException
+    {
+        int read = delegate.read(b, off, len);
+        if (read > 0) {
+            bytesLeft -= read;
+            checkLimit();
```
findepi: extract "bookkeeping and check" as `processedBytes(n)`

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @Override
+    public byte[] readAllBytes()
+            throws IOException
+    {
+        byte[] read = delegate.readAllBytes();
+        if (read.length > 0) {
+            bytesLeft -= read.length;
+            checkLimit();
```
findepi: This is too late to apply the limit.

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+        return read;
+    }
+
+    @Override
+    public byte[] readNBytes(int len)
```
findepi: unnecessary? what does the default impl do?

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
```diff
@@ -0,0 +1,178 @@
...
+        if (read.length > 0) {
+            bytesLeft -= read.length;
+            checkLimit();
+        }
+
+        return read;
+    }
+
+    @Override
+    public int readNBytes(byte[] buffer, int off, int len)
```
findepi: unnecessary? what does the default impl do?

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
```diff
@@ -0,0 +1,178 @@
...
+        long skipped = delegate.skip(n);
+        if (skipped > 0) {
+            bytesLeft -= skipped;
+            checkLimit();
+        }
+        return skipped;
+    }
+
+    @Override
+    public int read(byte[] buffer)
```
findepi: unnecessary? what does the default impl do?

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
```diff
@@ -0,0 +1,178 @@
...
+            checkLimit();
+        }
+        return b;
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len)
+            throws IOException
+    {
+        int read = delegate.read(b, off, len);
```
findepi: compare len with bytesleft before reading

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
```diff
@@ -0,0 +1,178 @@
...
+        if (read > 0) {
+            bytesLeft -= read;
+            checkLimit();
+        }
+
+        return read;
+    }
+
+    @Override
+    public void skipNBytes(long n)
```
findepi, line 78: unnecessary? what does the default impl do?

http-client/src/main/java/io/airlift/http/client/HttpClient.java
wendigo: Added `@deprecated {@link }`

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
```diff
@@ -0,0 +1,178 @@
...
+        delegate.mark(readlimit);
+    }
+
+    @Override
+    public void reset()
+            throws IOException
+    {
+        delegate.reset();
+        bytesLeft = maxBytes.toBytes();
+    }
```
findepi: AFAIR, that's incorrect. 
reset() resets to the last mark(), so bytesLeft should be restored from mark time

i suggest not to support mark & reset at all.

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
```diff
@@ -0,0 +1,178 @@
...
+    public boolean markSupported()
+    {
+        return delegate.markSupported();
+    }
+
+    @Override
+    public long transferTo(OutputStream out)
+            throws IOException
+    {
+        return delegate.transferTo(out);
```
findepi: Bypassing limits?

http-client/src/main/java/io/airlift/http/client/testing/TestingHttpClient.java
```diff
@@ -42,6 +44,7 @@ public TestingHttpClient(Processor processor, ExecutorService executor)
     }
 
     @Override
+    @SuppressWarnings("deprecation")
```
findepi: or mark it `@Deprecated` like done in `JettyHttpClient`?

http-client/src/main/java/io/airlift/http/client/HttpClient.java
```diff
@@ -18,15 +18,27 @@
...
 public interface HttpClient
         extends Closeable
 {
     <T, E extends Exception> T execute(Request request, ResponseHandler<T, E> responseHandler)
             throws E;
 
+    /**
+     * Executes request asynchronously running responseHandler on the client's http thread pool.
+     *
+     * @deprecated {@link HttpClient#executeAsync(Executor, Request, ResponseHandler)} instead.
```
findepi: It doesn't say _why_ it's deprecated. I know that you know, but users won't.

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
```diff
@@ -0,0 +1,55 @@
...
+    public Optional<StreamingResponse> executeRequest(AbstractHttpClientTest.CloseableTestHttpServer server, Request request)
+    {
+        return Optional.empty();
+    }
+
+    @Override
+    public void testConnectTimeout()
+            throws Exception
+    {
+        // Async client doesn't work properly with timeouts as this is expected for the caller to put timeout on the Future.get
```
findepi: This should be fixed. Async client should still enforce timeouts if it was configured to do so.

If we're unable to do so in async API, i would rather see it deleted -- the current implementation is basically sync API + implicit executor usage, so not very async anyway. The caller can do almost the same thing.

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
```diff
@@ -0,0 +1,55 @@
...
+        doTestConnectTimeout(true);
+    }
+
+    @SuppressWarnings("unchecked")
```
findepi: what's unchecked here?

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public <T, E extends Exception> T executeRequest(AbstractHttpClientTest.CloseableTestHttpServer server, Request request, ResponseHandler<T, E> responseHandler)
+            throws Exception
+    {
+        return executeRequest(server, createClientConfig(), request, responseHandler);
+    }
+
+    @SuppressWarnings("unchecked")
```
findepi: what's unchecked here?

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
```diff
@@ -0,0 +1,55 @@
...
+    public void testConnectTimeout()
+            throws Exception
+    {
+        // Async client doesn't work properly with timeouts as this is expected for the caller to put timeout on the Future.get
+        doTestConnectTimeout(true);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public <T, E extends Exception> T executeRequest(AbstractHttpClientTest.CloseableTestHttpServer server, Request request, ResponseHandler<T, E> responseHandler)
```
findepi: import `CloseableTestHttpServer`

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
wendigo: I would need to methods in that case - one that checks the limit before executing operation that consumes n bytes and then the decrement by the actual bytes read

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
wendigo: Removed

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
wendigo: This is a copy of a test. Not a net new code.

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
wendigo: Copy&paste

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
wendigo: This isn't correct. Timeouts are enforced (read/write/connect) but once you get an input stream there is no way to track how long caller consumes it. Hence the timeout should be applied on the future

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
```diff
@@ -0,0 +1,102 @@
...
+
+import io.airlift.units.DataSize;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+import static java.util.Objects.requireNonNull;
+
+public class ThrowingLimitingInputStream
```
findepi: Would this class use a unit test?

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
```diff
@@ -0,0 +1,102 @@
...
+            processedBytes(1);
+        }
+        return b;
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len)
+            throws IOException
+    {
+        checkRemainingBytes(len);
```
findepi: this is not correct
input stream: 20 bytes
limited input stream: 20 bytes limit
`read(4mb_buffer, 0, 4m)` should succeed returning 20 bytes

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
```diff
@@ -0,0 +1,102 @@
...
+            throws IOException
+    {
+        delegate.close();
+    }
+
+    @Override
+    public long transferTo(OutputStream out)
+            throws IOException
+    {
+        return delegate.transferTo(out);
```
findepi: without any limits?

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    private void processedBytes(long n)
+    {
+        bytesLeft -= n;
```
findepi, line 106: no checks here that bytesLeft is non-negative?

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
findepi: there is no reason to copy&paste wrong code, if we can do better.
i am not asking to fix original code, but  that would also be nice

http-client/src/test/java/io/airlift/http/client/jetty/TestJettyAsyncExecutorHttpClient.java
findepi: I agree that the timeout should be applied on the future by the caller
but i still believe the HTTP request timeout should be enforced by the HTTP request client.

i would expect Jetty to do it behind the scenes, when it processes response chunks that make up the InputStream.

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
wendigo: Read 20 and throw EOF?

findepi: read 20 and return int 20

wendigo: And then do what on the next read?

findepi: input stream: 20 bytes
limited input stream: 20 bytes limit

in such case, both input stream and limited input stream should behave exactly the same for the caller
and then it's trivial to test in a unit test with ByteArrayIS or FileIS, or both.
(ByteArrayIS has special characteristics that is always satisfies read requests to their fullest, which normal IS doesn't have, but it's not something possible to observe with just 20 byte input)

> And then do what on the next read?

the test will tell you

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
findepi: make package private

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
wendigo: If you want it to limit the amount of bytes, without throwing, I can just use `com.google.common.io.LimitInputStream`

http-client/src/main/java/io/airlift/http/client/jetty/ThrowingLimitingInputStream.java
findepi: fixed (this and few other remaining) in 
- https://github.com/airlift/airlift/pull/1709

PR #1660, opened 2025-11-27
Add explicit http server/client compression support

wendigo:


http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -442,6 +473,19 @@ protected void configure(SelectableChannel selectable)
...
+    {
+        compression.setByteBufferPool(httpClient.getByteBufferPool());
+        try {
+            compression.start();
+            closer.register(() -> closeQuietly(compression));
+            return Optional.of(new CompressionContentDecoderFactory(compression));
+        }
+        catch (Exception e) {
+            return Optional.empty();
+        }
```
oneonestar, line 486: When does this happen and why are we ignoring it silently?

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -349,17 +365,32 @@ protected void configure(SelectableChannel selectable)
...
-
-            // remove the GZIP encoding from the client
-            // TODO: there should be a better way to to do this
+            // clear all content decoder factories first
             httpClient.getContentDecoderFactories().clear();
+
+            if (config.isDeflateCompressionEnabled()) {
+                createContentDecoderFactory(new GzipCompression())
+                        .ifPresent(httpClient.getContentDecoderFactories()::put);
+            }
```
oneonestar, line 383: Deflate and gzip are different.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Encoding#deflate
```
Content-Encoding: gzip
Content-Encoding: deflate
```

> [deflate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Encoding#deflate)
Using the [zlib](https://en.wikipedia.org/wiki/Zlib) structure (defined in [RFC 1950](https://datatracker.ietf.org/doc/html/rfc1950)) with the [deflate](https://en.wikipedia.org/wiki/Deflate) compression algorithm (defined in [RFC 1951](https://datatracker.ietf.org/doc/html/rfc1951)).

> [gzip](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Encoding#gzip)
A format using the [Lempel-Ziv coding](https://en.wikipedia.org/wiki/LZ77_and_LZ78#LZ77) (LZ77), with a 32-bit CRC. This is the original format of the UNIX gzip program.

mosabua:
Ping @wendigo ..

Chaho12:
@wendigo knock knock~

PR #1666, opened 2025-12-03
[API] Fix renamed UNWRAP_INTO_PARENT

Randgalt:
`UNWRAP_INTO_PARENT` was renamed to `IS_UNWRAPPED`. However, it was
never removed and a reference to it still existed in `SchemaBuilder`.
This caused the generated OpenApi to be incorrect with respect to
unwrapped components.

-------

Changing a component from unwrapped would not trigger an error from the
compatibility tool. Fixed this and added a test to prove it,
`CompatibilityTesterTests`. This test should be expanded in the future
to check more cases.

NOTE: this will require modifying existing compatibility files


api/src/main/java/io/airlift/api/model/ModelResource.java
```diff
@@ -73,6 +73,13 @@ public ModelResource withModifier(ModelResourceModifier modifier)
         return new ModelResource(type, name, openApiName, description, components, resourceType, containerType, newModifiers, quotas, limitedValues, supportsIdLookup, polyResource, enumDescriptions);
     }
 
+    public ModelResource withoutModifier(ModelResourceModifier modifier)
```
oskar-szwajkowski: duplicate of `withModifierRemoved`

api/src/test/java/io/airlift/api/compatibility/CompatibilityTesterTests.java
```diff
@@ -0,0 +1,109 @@
...
+            ApiCompatibilityTester.newDefaultInstance(tempDir.toString()).test(unwrappedApi.modelServices(), true, foundErrors::addAll);
+            // first pass creates the file
+            assertThat(foundErrors).hasSize(1).matches(l -> l.getFirst().contains("Created new compatibility file"));
+            foundErrors.clear();
+            ApiCompatibilityTester.newDefaultInstance(tempDir.toString()).test(unwrappedApi.modelServices(), true, foundErrors::addAll);
+            // second pass should be no errors
+            assertThat(foundErrors).isEmpty();
+
+            ApiCompatibilityTester.newDefaultInstance(tempDir.toString()).test(removeUnwrapped(unwrappedApi.modelServices()), true, foundErrors::addAll);
+            assertThat(foundErrors).isNotEmpty();
```
oskar-szwajkowski, line 49: is the error message here deterministic? can we put it here explicitly so that it serves as context that we are checking unwrapped -> non unwrapped incompatibility?

oskar-szwajkowski: if its non deterministic then perhaps `contains` assertion would do

api/src/test/java/io/airlift/api/compatibility/CompatibilityTesterTests.java
```diff
@@ -0,0 +1,109 @@
...
+        return new ModelResponse(modelResponse.status(), modelResponse.responseClass(), resoure);
+    }
+
+    private ModelResource removeUnwrapped(ModelResource modelResource)
+    {
+        List<ModelResource> components = modelResource.components().stream().map(resource -> resource.withoutModifier(IS_UNWRAPPED)).collect(toImmutableList());
+        return modelResource.withComponents(components);
+    }
+
+    private void cleanTempDirectory(Path tempDirectory)
```
oskar-szwajkowski: nit: `MoreFiles.deleteRecursively();` can be used instead

api/src/test/java/io/airlift/api/compatibility/CompatibilityTesterTests.java
Randgalt: The exact message isn't really important. Comparing exact against the error is brittle (I had to make many manual changes to tests that compared error messages when porting from the private library). Asserting that there is an error is good here.

oskar-szwajkowski (approved):

PR #1667, opened 2025-12-04
Make configuration quiet with system property

findepi:
Introduce a system property for controlling default value of `Bootstrap.quiet`. This can be useful in situations where multiple components using Airlift are loaded in short succession, and logging all configuration properties is too verbose. For example, in tests, or when components are loaded dynamically and ad-hoc. This is a system property, so that a system using Airlift can log its initial configuration and then toggle the system property value.

### DIFF
--- a/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
@@ -56,6 +56,7 @@ import static io.airlift.bootstrap.FuzzyMatcher.findSimilar;
 import static io.airlift.configuration.ConfigurationLoader.getSystemProperties;
 import static io.airlift.configuration.ConfigurationLoader.loadPropertiesFrom;
 import static io.airlift.configuration.TomlConfiguration.createTomlConfiguration;
+import static java.lang.Boolean.parseBoolean;
 import static java.lang.String.format;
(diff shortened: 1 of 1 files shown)


bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -84,7 +86,7 @@ private enum State { UNINITIALIZED, CONFIGURED, INITIALIZED }
     private boolean initializeLogging = true;
     private boolean envInterpolation = true;
     private boolean useSystemProperties = true;
-    private boolean quiet;
+    private boolean quiet = parseBoolean(System.getProperty("airlift.quiet"));
```
wendigo, line 88: I don't like it. Why not an ENV_VARIABLE which is easier to pass?

findepi: from PR desc

>  This is a system property, so that a system using Airlift can log its initial configuration and then toggle the system property value.

findepi: we could also do both, but then we would need to determine their precedence, etc.
also, `parseBoolean` is suitable for handling system properties (which are Java specific). for env vars the parse logic would be different.

PR #1683, opened 2025-12-12
Introduce ApiParsedResponse to aid in getting response errors

### LINKED ISSUE #1681: [API Builder] The `Invalid id` message never makes it into the API response
The method `io.airlift.api.ApiId#toInternal` throws a bad-request exception when the ID is invalid:
https://github.com/airlift/airlift/blob/9ac55f38cc01c60e19a795585c95787ac453007c/api/src/main/java/io/airlift/api/ApiId.java#L57

Unfortunately, the descriptive message passed to `badRequest()` is dropped by Jakarta and never makes it into the API response.  I'm attaching the call stack below, but the gist is that `jakarta.ws.rs.WebApplicationException#computeExceptionMessage` ignores `response.context.entity` and constructs the message as a generic description of the HTTP status.  As a result, the user sending the wrong ID in their API request never gets to find out.  The response they get is
```
{"code":400,"message":"HTTP 400 Bad Request","status":"INVALID_ARGUMENT","detailsType":"BadRequest","details":{"fieldViolations":[]}}
```

Image

Randgalt:
`ApiParsedResponse` can be used by clients to get detailed response messages from APIs that follow the `ApiException` naming conventions and add a JSON payload to responses. `ApiException` methods work as expected and are the preferred response mechanism.

@wendigo what do you think of this approach?

Closes #1681

### DIFF
--- a/api/src/test/java/io/airlift/api/servertests/exceptions/ExceptionApi.java
+++ b/api/src/test/java/io/airlift/api/servertests/exceptions/ExceptionApi.java
@@ -1,0 +1,28 @@
+package io.airlift.api.servertests.exceptions;
+
+import com.google.common.collect.ImmutableList;
+import io.airlift.api.ApiCreate;
+import io.airlift.api.ApiService;
+import io.airlift.api.ApiUpdate;
+import io.airlift.api.ServiceType;
+import jakarta.ws.rs.WebApplicationException;
+
+import static io.airlift.api.responses.ApiException.badRequest;
+import static jakarta.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static jakarta.ws.rs.core.Response.status;
+
+@ApiService(type = ServiceType.class, name = "exception", description = "Throws exceptions")
+public class ExceptionApi
+{
+    @ApiUpdate(description = "Throws a known exception")
+    public void throwsKnownException()
+    {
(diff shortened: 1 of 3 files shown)


api/src/test/java/io/airlift/api/servertests/exceptions/ExceptionApi.java
```diff
@@ -0,0 +1,28 @@
...
+public class ExceptionApi
+{
+    @ApiUpdate(description = "Throws a known exception")
+    public void throwsKnownExceptionUpdate()
+    {
+        throw badRequest("This is the message", ImmutableList.of("field1", "field2"));
+    }
+
+    @ApiCreate(description = "Throws an unknown exception", quotas = "dummy")
+    public void throwsUnknownExceptionUpdate()
```
dekimir: I wonder if `Update` is the right suffix for this method's name, given that it's not an `@ApiUpdate`?

dekimir (commented):
Parsing the response sounds reasonable.  Though I'm still confused why `curl` [doesn't print](https://github.com/airlift/airlift/issues/1681#issue-3721033098) the presumably present fields `description` and `fields`.

api/src/test/java/io/airlift/api/servertests/exceptions/ExceptionApi.java
Randgalt: auto generated names - I'll change

PR #1691, opened 2025-12-22
Deprecate ConditionalModule and SwitchModule

kokosing:
ConditionalModule and SwitchModule provides little value and they make
code unnecessarily complex.

Instead of using opaque abstraction, they can be replaced with regular java control syntax.

Consider ConditionalModule:

    install(conditionalModule(
            SomeConfig.class,
            config -> config.getSomeOption().equals("x"),
            binder -> binder.bind(String.class).toInstance("X")));

can be replaced with:

    SomeConfig testConfig = buildConfigObject(SomeConfig.class);
    if (testConfig.getSomeOption().equals("x")) {
	    binder.bind(String.class).toInstance("X");
    }

Consider SwitchModule:

    install(switchModule(
	    SomeConfig.class,
	    SomeConfig::getSomeOption,
	    value -> switch (value) {
		  case "x" -> binder -> binder.bind(String.class).toInstance("X");
		  default -> throw new IllegalArgumentException("Unsupported value: " + value);
	    }));

can be replaced with:

    SomeConfig someConfig = buildConfigObject(SomeConfig.class);
    switch (someConfig.getSomeOption()) {
	  case "x" -> binder.bind(String.class).toInstance("X");
	  default -> throw new IllegalArgumentException("Unsupported value: " + someConfig.getSomeOption());
    }

### DIFF
--- a/configuration/src/test/java/io/airlift/configuration/TestConditionalModule.java
+++ b/configuration/src/test/java/io/airlift/configuration/TestConditionalModule.java
@@ -13,17 +13,17 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Supplier;
 
-import static io.airlift.configuration.ConditionalModule.conditionalModule;
 import static org.assertj.core.api.Assertions.assertThat;
 
+@SuppressWarnings("deprecation")
 public class TestConditionalModule
 {
     @Test
     public void testConditionalModule()
     {
         Supplier<AbstractConfigurationAwareModule> moduleXY = () -> {
-            Module moduleX = conditionalModule(SomeConfig.class, config -> config.getSomeOption().equals("x"), binder -> binder.bind(String.class).toInstance("X"));
-            Module moduleY = conditionalModule(SomeConfig.class, config -> config.getSomeOption().equals("y"), binder -> binder.bind(String.class).toInstance("Y"));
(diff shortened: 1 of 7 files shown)


electrum (approved):

kokosing:
@wendigo WDYT?

configuration/src/test/java/io/airlift/configuration/TestConditionalModule.java
```diff
@@ -13,17 +13,17 @@
 import java.util.Map;
 import java.util.function.Supplier;
 
-import static io.airlift.configuration.ConditionalModule.conditionalModule;
```
kokosing, line 16: I had to avoid static imports so I could use `@SuppressWarnings("deprecation")`. If deprecation is marked on entire class then I would need to suppress the warning on import statement which is not allowed by java compiler.

findepi:
> can be replaced with:
> 
> ```
> SomeConfig testConfig = buildConfigObject(SomeConfig.class);
> if (testConfig.getSomeOption().equals("x")) {
>     binder.bind(String.class).toInstance("X");
> }
> ```

such imperative cannot work with `.bindConfigDefaults`, especially if defaults apply later, eg in sibling module

i suppose conditional modules do not  work observe defaults today, but i consider this a bug which we would want to fix.

PR #1695, opened 2025-12-26
[MCP Server] Support list change notifications and resource subscriptions

Randgalt:
Continuation of full MCP Server spec support. 

Note: the Reference Java SDK does not support any of these features. Yet-another-reason to abandon it.

This PR adds support for **List Change Notifications**, **Resource Content Subscriptions** and **Event Streams**. **List Change Notifications** and **Resource Content Subscriptions** are supported internally via existing/expected mechanisms. 

**Event Streams** are optional and enabled via setting the new `EventStreaming` value of `McpMetadata`.

## List Change Notifications

Spec:
- https://modelcontextprotocol.io/specification/2025-11-25/server/tools#list-changed-notification
- https://modelcontextprotocol.io/specification/2025-11-25/server/prompts#list-changed-notification
- https://modelcontextprotocol.io/specification/2025-11-25/server/resources#list-changed-notification

## Resource Content Subscriptions

Spec:
- https://modelcontextprotocol.io/specification/2025-11-25/server/resources#subscriptions

## Event Streams

Spec:
- https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#listening-for-messages-from-the-server

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -21,12 +21,12 @@ variations of MCP servers defined by the standard. This module supports:
 - Pagination [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/pagination)
 - Sessions [(see spec)](https://modelcontextprotocol.io/docs/concepts/transports#session-management)
 - Server-sent logging [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging)
+- List changed events [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle#initialization)
+- Subscriptions [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle#initialization)
 
 This module does not currently support:
 
 - Cancellation [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/cancellation)
-- List changed events [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle#initialization)
-- Subscriptions [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle#initialization)
 - Elicitation [(see spec)](https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation)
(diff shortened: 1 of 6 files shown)


Randgalt:
attn: @codyzwief

mcp/src/main/java/io/airlift/mcp/internal/InternalMcpServer.java
```diff
@@ -371,4 +488,12 @@ private static String completionKey(CompleteReference reference)
             case ResourceReference(var uri) -> uri;
         };
     }
+
+    @SuppressWarnings("UnstableApiUsage")
+    private static <T> String listHash(Stream<? extends T> stream)
+    {
+        Hasher hasher = Hashing.sha256().newHasher();
+        stream.forEach(item -> hasher.putString(item.toString(), UTF_8));
```
wendigo: can we hash hashCodes instead?

wendigo: This requires a resource to have a meaningful toString representation, right?

Randgalt: They're all records so they do have good `toString()`. That said, those strings have a bunch of junk in them. I'll try to improve it.

wendigo: Maybe these records should implement an interface that gives an explicit version, wdyt?

Randgalt: I don't want to burden the models with that (the models are supposed to match the spec). I'll have something in a moment.

Randgalt: @wendigo PTAL now. I'm writing them as JSON. This has several benefits:

- it won't contain any crud
- it's a normalized format
- it's guaranteed to contain the data that constitutes the state of each item

wendigo: Makes sense!

PR #1696, opened 2025-12-26
[MCP Server] Replace reference implementation with Airlift internal implementation

Randgalt:
Replace the reference Java SDK with an Airlift internal implementation

The official reference SDK is not giving us much benefit and we keep
having to work around issues with it. They are slow to improve and
are falling behind in the spec. Using our own implementation will
allow us to move faster and independently.

We will continue to use the official SDK client for testing to
help ensure compatibility/conformance.

### DIFF
--- a/mcp/src/main/java/io/airlift/mcp/McpModule.java
+++ b/mcp/src/main/java/io/airlift/mcp/McpModule.java
@@ -36,6 +36,7 @@ import java.util.function.Consumer;
 
 import static com.google.inject.Scopes.SINGLETON;
 import static com.google.inject.multibindings.Multibinder.newSetBinder;
+import static com.google.inject.multibindings.OptionalBinder.newOptionalBinder;
 import static io.airlift.configuration.ConfigBinder.configBinder;
 import static io.airlift.json.JsonSubTypeBinder.jsonSubTypeBinder;
 import static io.airlift.mcp.McpMetadata.DEFAULT;
@@ -190,6 +191,8 @@ public class McpModule
         binder.bind(McpMetadata.class).toInstance(metadata);
 
         configBinder(binder).bindConfig(McpConfig.class);
+
+        newOptionalBinder(binder, ErrorHandler.class);
 
         bindClasses(binder);
         bindTools(binder);
--- a/mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
+++ b/mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
@@ -3,20 +3,17 @@ package io.airlift.mcp.internal;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.inject.Inject;
 import io.airlift.log.Logger;
-import io.airlift.mcp.McpException;
+import io.airlift.mcp.ErrorHandler;
 import io.airlift.mcp.McpIdentityMapper;
 import io.airlift.mcp.McpMetadata;
 import io.airlift.mcp.model.CallToolRequest;
 import io.airlift.mcp.model.CompleteRequest;
 import io.airlift.mcp.model.GetPromptRequest;
 import io.airlift.mcp.model.InitializeRequest;
-import io.airlift.mcp.model.JsonRpcErrorCode;
-import io.airlift.mcp.model.JsonRpcErrorDetail;
 import io.airlift.mcp.model.JsonRpcRequest;
 import io.airlift.mcp.model.JsonRpcResponse;
 import io.airlift.mcp.model.ListRequest;
@@ -28,20 +25,17 @@ import jakarta.servlet.ServletException;
 import jakarta.servlet.http.HttpFilter;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
-import jakarta.ws.rs.WebApplicationException;
 
 import java.io.BufferedReader;
 import java.io.IOException;
-import java.io.PrintWriter;
-import java.io.UncheckedIOException;
 import java.util.Optional;
 import java.util.Set;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
-import static com.google.common.base.Throwables.throwIfInstanceOf;
 import static com.google.common.net.HttpHeaders.WWW_AUTHENTICATE;
+import static io.airlift.mcp.McpException.exception;
 import static io.airlift.mcp.model.Constants.HEADER_SESSION_ID;
 import static io.airlift.mcp.model.Constants.MCP_IDENTITY_ATTRIBUTE;
+import static io.airlift.mcp.model.Constants.MESSAGE_WRITER_ATTRIBUTE;
 import static io.airlift.mcp.model.Constants.METHOD_COMPLETION_COMPLETE;
 import static io.airlift.mcp.model.Constants.METHOD_INITIALIZE;
 import static io.airlift.mcp.model.Constants.METHOD_PING;
@@ -52,11 +46,13 @@ import static io.airlift.mcp.model.Constants.METHOD_RESOURCES_READ;
 import static io.airlift.mcp.model.Constants.METHOD_RESOURCES_TEMPLATES_LIST;
 import static io.airlift.mcp.model.Constants.METHOD_TOOLS_CALL;
 import static io.airlift.mcp.model.Constants.METHOD_TOOLS_LIST;
+import static io.airlift.mcp.model.Constants.RPC_MESSAGE_ATTRIBUTE;
 import static io.airlift.mcp.model.JsonRpcErrorCode.INVALID_REQUEST;
+import static io.airlift.mcp.model.JsonRpcErrorCode.METHOD_NOT_FOUND;
+import static io.airlift.mcp.model.JsonRpcErrorCode.PARSE_ERROR;
 import static jakarta.servlet.http.HttpServletResponse.SC_ACCEPTED;
 import static jakarta.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
 import static jakarta.servlet.http.HttpServletResponse.SC_FORBIDDEN;
-import static jakarta.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
 import static jakarta.servlet.http.HttpServletResponse.SC_METHOD_NOT_ALLOWED;
 import static jakarta.servlet.http.HttpServletResponse.SC_NOT_FOUND;
 import static jakarta.servlet.http.HttpServletResponse.SC_OK;
@@ -83,14 +79,16 @@ public class InternalFilter
     private final McpIdentityMapper identityMapper;
     private final InternalMcpServer mcpServer;
     private final ObjectMapper objectMapper;
+    private final ErrorHandler errorHandler;
 
     @Inject
-    public InternalFilter(McpMetadata metadata, McpIdentityMapper identityMapper, InternalMcpServer mcpServer, ObjectMapper objectMapper)
+    public InternalFilter(McpMetadata metadata, McpIdentityMapper identityMapper, InternalMcpServer mcpServer, ObjectMapper objectMapper, ErrorHandler errorHandler)
     {
         this.metadata = requireNonNull(metadata, "metadata is null");
         this.identityMapper = requireNonNull(identityMapper, "identityMapper is null");
         this.mcpServer = requireNonNull(mcpServer, "mcpServer is null");
         this.objectMapper = requireNonNull(objectMapper, "objectMapper is null");
+        this.errorHandler = requireNonNull(errorHandler, "errorHandler is null");
     }
 
     @Override
@@ -122,35 +120,13 @@ public class InternalFilter
                 case McpIdentity.Error error -> throw error.cause();
             }
         }
-        catch (WebApplicationException webApplicationException) {
-            throw webApplicationException;
+        catch (Throwable throwable) {
+            errorHandler.handleException(request, response, throwable);
         }
-        catch (McpException mcpException) {
-            handleMcpException(request, response, mcpException);
-        }
-        catch (IOException e) {
-            log.debug(e, "Internal I/O error. Request: %s", request.getRequestURI());
-            responseError(response, SC_BAD_REQUEST, internalError(firstNonNull(e.getMessage(), "Internal I/O error")));
-        }
-        catch (Exception e) {
-            if (Throwables.getRootCause(e) instanceof McpException mcpException) {
-                handleMcpException(request, response, mcpException);
-            }
-            else {
-                log.debug(e, "Internal error. Request: %s", request.getRequestURI());
-                responseError(response, SC_INTERNAL_SERVER_ERROR, internalError(firstNonNull(e.getMessage(), "Internal error")));
-            }
-        }
-    }
-
-    private void handleMcpException(HttpServletRequest request, HttpServletResponse response, McpException mcpException)
-    {
-        log.debug(mcpException, "Request: %s", request.getRequestURI());
-        JsonRpcErrorDetail errorDetail = mcpException.errorDetail();
-        responseError(response, SC_BAD_REQUEST, errorDetail);
     }
 
     private void handleMcpRequest(HttpServletRequest request, HttpServletResponse response, Object identity)
+            throws Exception
     {
         request.setAttribute(MCP_IDENTITY_ATTRIBUTE, identity);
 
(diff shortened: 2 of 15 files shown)


mcp/src/main/java/io/airlift/mcp/reference/ReferenceModule.java
```diff
@@ -52,35 +48,31 @@ public class ReferenceModule
...
+    // bound via method so users can override the binding if needed
+    protected void bindReferenceServer(Binder binder)
     {
-        return HttpServletStatelessServerTransport.builder()
-                .messageEndpoint(metadata.uriPath())
-                .jsonMapper(objectMapper)
-                .contextExtractor(request -> McpTransportContext.create(ImmutableMap.of(CONTEXT_REQUEST_KEY, request)))
-                .build();
+        binder.bind(ReferenceServer.class).asEagerSingleton();
     }
```
codyzwief: I'm wondering if it would be a good idea to keep the reference implementation around for backward compatibility for a version or two, but just deprecate it. I worry about a couple bugs that may exist in the new implementation, without having any ability to revert back and fix them if we run into any problems. WDYT?

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+        catch (McpException | WebApplicationException e) {
+            throw e;
+        }
+        catch (IllegalArgumentException | IOException | UncheckedIOException e) {
+            log.error("Failed to deserialize message: {}", e.getMessage());
+            responseError(response, SC_BAD_REQUEST, invalidRequest("Invalid message format"));
+        }
+        catch (Exception e) {
+            throwIfInstanceOf(e.getCause(), McpException.class);
+            log.error("Unexpected error handling message: {}", e.getMessage());
```
codyzwief: 2 nits:

1. IMO it would be good to `log.error(e, "failed to deserialize message");` here (and elsewhere in this class)
2. the log on line 205 happens after rethrowing MCP exceptions.

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -228,12 +229,12 @@ private void handleRpcRequest(HttpServletRequest request, HttpServletResponse re
...
-                case METHOD_TOOLS_LIST -> mcpServer.listTools();
+                case METHOD_TOOLS_LIST -> mcpServer.listTools(convertParams(rpcRequest, ListRequest.class));
                 case METHOD_TOOLS_CALL -> mcpServer.callTool(request, messageWriter, convertParams(rpcRequest, CallToolRequest.class));
-                case METHOD_PROMPT_LIST -> mcpServer.listPrompts();
+                case METHOD_PROMPT_LIST -> mcpServer.listPrompts(convertParams(rpcRequest, ListRequest.class));
                 case METHOD_PROMPT_GET -> mcpServer.getPrompt(request, messageWriter, convertParams(rpcRequest, GetPromptRequest.class));
-                case METHOD_RESOURCES_LIST -> mcpServer.listResources();
-                case METHOD_RESOURCES_TEMPLATES_LIST -> mcpServer.listResourceTemplates();
+                case METHOD_RESOURCES_LIST -> mcpServer.listResources(convertParams(rpcRequest, ListRequest.class));
+                case METHOD_RESOURCES_TEMPLATES_LIST -> mcpServer.listResourceTemplates(convertParams(rpcRequest, ListRequest.class));
```
codyzwief: I was testing this PR with claude code and it doesn't send any params for list requests, it will just omit the `params` entirely. This caused exceptions with this code because `convertParams` will throw if it's not present. Can this gracefully handle those list cases where params is omitted entirely and they're optional?

mcp/src/main/java/io/airlift/mcp/McpModule.java
```diff
@@ -80,13 +72,11 @@ private McpModule(Mode mode, McpMetadata metadata, Optional<IdentityMapperBindin
...
         this.resourceTemplates = ImmutableSet.copyOf(resourceTemplates);
         this.completions = ImmutableSet.copyOf(completions);
-
-        validateRoles();
     }
 
     public enum Mode
     {
-        REFERENCE_SDK,
+        STANDARD,
```
codyzwief, line 80: As mentioned in another comment, my preference would be to keep the `REFERENCE_SDK` around for another cycle or two, just in case we run into any issues upgrading. Would it be possible to just deprecate it for now, and not actually remove the reference SDK?

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Locale.ROOT;
+import static java.util.Objects.requireNonNull;
+
+public class InternalFilter
+        extends HttpFilter
+{
+    private static final Logger log = Logger.get(InternalFilter.class);
+
+    private static final Set<String> ALLOWED_HTTP_METHODS = ImmutableSet.of("GET", "POST", "DELETE");
```
codyzwief, line 76: `OPTIONS`? I know the Inspector runs some cors preflight requests

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+    protected void doFilter(HttpServletRequest request, HttpServletResponse response, FilterChain chain)
+            throws IOException, ServletException
+    {
+        if (!isMcpRequest(request)) {
+            chain.doFilter(request, response);
+            return;
+        }
+
+        identityMapper.ifPresent(mapper -> internalDoFilter(mapper, request, response));
+    }
```
codyzwief: Will this still handle the MCP request if there is no identity mapper defined? If not, do we want to make that IdentityMapper required to be defined instead of an optional?

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+            responseError(response, SC_BAD_REQUEST, internalError(firstNonNull(e.getMessage(), "Internal I/O error")));
+        }
+        catch (Exception e) {
+            if (Throwables.getRootCause(e) instanceof McpException mcpException) {
+                handleMcpException(request, response, mcpException);
+            }
+            else {
+                log.debug(e, "Internal error. Request: %s", request.getRequestURI());
+                responseError(response, SC_INTERNAL_SERVER_ERROR, internalError(firstNonNull(e.getMessage(), "Internal error")));
+            }
```
codyzwief: Two questions:

1. Why are some exceptions being logged and others not?
2. Why are we just rethrowing `WebApplicationException`s, could they or should they be wrapped in proper JSON-RPC errors?

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    private void handleMcpException(HttpServletRequest request, HttpServletResponse response, McpException mcpException)
+    {
+        log.debug(mcpException, "Request: %s", request.getRequestURI());
+        JsonRpcErrorDetail errorDetail = mcpException.errorDetail();
+        responseError(response, SC_BAD_REQUEST, errorDetail);
+    }
+
+    private void handleMpcRequest(HttpServletRequest request, HttpServletResponse response, Object identity)
```
codyzwief: nit: `handleMpcRequest` -> `handleMcpRequest`

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+    {
+        request.setAttribute(MCP_IDENTITY_ATTRIBUTE, identity);
+
+        switch (request.getMethod().toUpperCase(ROOT)) {
+            case POST -> handleMcpPostRequest(request, response);
+            case GET, DELETE -> response.setStatus(SC_METHOD_NOT_ALLOWED);
```
codyzwief, line 135: I assume this will change with the incoming session support, and with supporting notifications. Is that correct?

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+                messageWriter.write(objectMapper.writeValueAsString(rpcResponse));
+                messageWriter.flushMessages();
+            }
+            catch (JsonProcessingException e) {
+                throw new RuntimeException(e);
+            }
+        }
+        catch (Exception e) {
+            throw new RuntimeException(e);
+        }
```
codyzwief: Would it make sense to rethrow this error as a proper JSON-RPC exception with an internal failure error code?

mcp/src/main/java/io/airlift/mcp/internal/InternalMcpServer.java
```diff
@@ -0,0 +1,326 @@
...
+import static io.airlift.mcp.model.Constants.PROTOCOL_MCP_2025_06_18;
+import static io.airlift.mcp.model.JsonRpcErrorCode.INVALID_PARAMS;
+import static io.airlift.mcp.model.JsonRpcErrorCode.INVALID_REQUEST;
+import static io.airlift.mcp.model.JsonRpcErrorCode.RESOURCE_NOT_FOUND;
+import static java.util.Objects.requireNonNull;
+
+public class InternalMcpServer
+        implements McpServer
+{
+    private static final List<String> SUPPORTED_PROTOCOL_VERSIONS = ImmutableList.of(PROTOCOL_MCP_2025_06_18);
```
codyzwief, line 76: Any newer versions we could support too with this PR?

mcp/src/main/java/io/airlift/mcp/internal/PaginationUtil.java
```diff
@@ -0,0 +1,43 @@
...
+        pageSize = metadata.pageSize();
+    }
+
+    <T, L> T paginate(PaginatedRequest request, List<L> items, Function<L, String> keyMapper, BiFunction<List<L>, Optional<String>, T> resultMapper)
+    {
+        Stream<L> itemsStream = request.cursor().map(cursor -> items.stream().filter(item -> keyMapper.apply(item).compareTo(cursor) > 0))
+                .orElseGet(items::stream);
+
+        List<L> filteredList = itemsStream.sorted(comparing(keyMapper))
+                .collect(toImmutableList());
```
codyzwief, line 34: just a nit, but would it make sense to assume the list of items is already sorted / keep them sorted when adding an entry, so that we you don't have to re-sort every time we're going through pagination?

mcp/src/main/java/io/airlift/mcp/reference/ReferenceModule.java
Randgalt: It would be far too much work to retain both implementations. It would be better to add tests for any issues we find. Adding all the features of the spec that the reference implementation doesn't support is very difficult. If we won't be going forward with it it's also wasted effort.

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
Randgalt: I know what that is. I've handled it with some other fields that are optional in that way. I'll take a pass over all of the models. I'll see if I can write some tests too.

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
Randgalt: the reference implementation doesn't handle `OPTIONS`. An `OPTIONS` servlet filter can always be handled via normal Airlift injections.

mcp/src/main/java/io/airlift/mcp/McpModule.java
Randgalt: We can move it into test code maybe, but as I mention above, it's going to be a PITA to maintain and it's wasted effort.

mcp/src/main/java/io/airlift/mcp/internal/InternalMcpServer.java
Randgalt: There's a PR in this series that adds support for the latest spec.

Randgalt:
> I was testing the new implementation with claude code

@codyzwief how do you test with Claude? Do you use the MCP server you've been working on or ???

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
Randgalt: Maybe we should add an error utility or something.

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
Randgalt: For 1 - I will try to normalize that. For 2, in some cases we want a real HTTP error and not an RPC error. But, I will double check all of that.

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
Randgalt: Yes, you can see that in the PRs that come after this.

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
codyzwief: ping on this comment

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
codyzwief: ping on this comment

Randgalt:
> LGTM -- a couple comments outstanding, including a couple about consistent error handling/logging

I'm going to take another pass and normalize all exceptions, errors, logging, etc. I'll ping when it's done.

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
Randgalt: Good point. Let's make it required. I'll make that change.

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
Randgalt: @codyzwief I've reworked error handling. PTAL

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+        catch (WebApplicationException webApplicationException) {
+            throw webApplicationException;
+        }
+        catch (McpException mcpException) {
+            handleMcpException(request, response, mcpException);
+        }
+        catch (IOException e) {
+            log.debug(e, "Internal I/O error. Request: %s", request.getRequestURI());
+            responseError(response, SC_BAD_REQUEST, internalError(firstNonNull(e.getMessage(), "Internal I/O error")));
```
wendigo: should it be INTERNAL_ERROR ?

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+        }
+        catch (Exception e) {
+            if (Throwables.getRootCause(e) instanceof McpException mcpException) {
```
wendigo:
```
Throwables.getCasualChain(e)
   .filter(McpException.class::isInstance)
```

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+            else {
+                log.debug(e, "Internal error. Request: %s", request.getRequestURI());
+                responseError(response, SC_INTERNAL_SERVER_ERROR, internalError(firstNonNull(e.getMessage(), "Internal error")));
+            }
+        }
+    }
+
+    private void handleMcpException(HttpServletRequest request, HttpServletResponse response, McpException mcpException)
+    {
+        log.debug(mcpException, "Request: %s", request.getRequestURI());
```
wendigo: `Request to %s failed`

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
Randgalt: Please see later commit that re-works all error handling

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+            case GET, DELETE -> response.setStatus(SC_METHOD_NOT_ALLOWED);
+            default -> response.setStatus(SC_NOT_FOUND);
+        }
+    }
+
+    private void handleMcpPostRequest(HttpServletRequest request, HttpServletResponse response)
+    {
+        InternalMessageWriter messageWriter = new InternalMessageWriter(response);
+
+        String accept = request.getHeader(ACCEPT);
```
wendigo, line 146:
```
        List<String> accept = ImmutableList.copyOf(request.getHeaders(ACCEPT).asIterator());
        if (accept.isEmpty()|| !(accept.contains(APPLICATION_JSON) && accept.contains(SERVER_SENT_EVENTS))) {
            throw exception(INVALID_REQUEST, "Both application/json and text/event-stream required in Accept header");
        }

```

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+            responseError(response, SC_BAD_REQUEST, invalidRequest("Both application/json and text/event-stream required in Accept header"));
+            return;
+        }
+
+        response.setContentType(APPLICATION_JSON);
+        response.setCharacterEncoding(UTF_8.name());
+
+        try {
+            BufferedReader reader = request.getReader();
+            String body = reader.lines().collect(Collectors.joining("\n"));
```
wendigo: nit: static import joining

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+        if (accept == null || !(accept.contains(APPLICATION_JSON) && accept.contains(SERVER_SENT_EVENTS))) {
+            responseError(response, SC_BAD_REQUEST, invalidRequest("Both application/json and text/event-stream required in Accept header"));
+            return;
+        }
+
+        response.setContentType(APPLICATION_JSON);
+        response.setCharacterEncoding(UTF_8.name());
+
+        try {
+            BufferedReader reader = request.getReader();
```
wendigo: should reader be closed?

mcp/src/main/java/io/airlift/mcp/internal/PaginationUtil.java
Randgalt: That seems less flexible and more error-prone. These lists should be small so the overhead is light. But, I don't have a strong feeling on this.

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
wendigo: try (BufferedReader reader = request.getReader()) {

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
Randgalt: Now handled in the `ErrorHandler` and it's an INTERNAL_ERROR when appropriate.

mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -0,0 +1,324 @@
...
+    @Override
+    protected void doFilter(HttpServletRequest request, HttpServletResponse response, FilterChain chain)
+            throws IOException, ServletException
+    {
+        if (!isMcpRequest(request)) {
+            chain.doFilter(request, response);
+            return;
+        }
+
+        identityMapper.ifPresent(mapper -> internalDoFilter(mapper, request, response));
```
wendigo: what if identityMapper is absent and isMcpRequest is true ?

wendigo: Ok, it's fixed in consecutive commit.

wendigo (approved):

PR #1700, opened 2025-12-29
[MCP Server] Create a DB-based session controller for testing

Randgalt:
Ensure that the session system works with a real database

### DIFF
--- a/mcp/src/test/java/io/airlift/mcp/TestingDatabaseSessionController.java
+++ b/mcp/src/test/java/io/airlift/mcp/TestingDatabaseSessionController.java
@@ -1,0 +1,509 @@
+package io.airlift.mcp;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.util.concurrent.UncheckedExecutionException;
+import com.google.inject.Inject;
+import io.airlift.concurrent.Threads;
+import io.airlift.log.Logger;
+import io.airlift.mcp.model.McpIdentity;
+import io.airlift.mcp.sessions.SessionController;
+import io.airlift.mcp.sessions.SessionId;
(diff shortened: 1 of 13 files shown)


mcp/src/test/java/io/airlift/mcp/TestingDatabaseSessionController.java
```diff
@@ -0,0 +1,509 @@
...
+    {
+        if (!shutdownAndAwaitTermination(executorService, LISTEN_TIMEOUT_MS * 2, MILLISECONDS)) {
+            log.warn("Executor shutdown failed");
+        }
+    }
+
+    @Override
+    public SessionId createSession(Optional<McpIdentity> identity, Optional<Duration> ttl)
+    {
+        // this is merely a test implementation, so we ignore the ttl
```
wendigo, line 182: shouldn't we use identity?

wendigo: Why is identity Optional anyway? Is it allowed for it to be absent?

Randgalt: I think this is based on older implementations. I'll take another look. 

The problem with using `Identity` is that it's not typed. We'd end up with:

```
public SessionId createSession(Object identity, Optional<Duration> ttl)
```

unless we add a type parameter to `SessionController` which doesn't seem right.

Let me give this some thought (unless you have any ideas or @codyzwief)

wendigo: Why? McpIdentity is a sealed interface with 4 distinct states: Authenticated, Unauthenticated, Unauthorized and Error. Does Optional<McpIdentity> being absent means `Unauthenticated` ?

Randgalt: I thought you meant the `identity` value in `Authenticated`. This works too.

wendigo: Nope. I think that the more general approach is to allow SessionController to handle all 4 states throwing an exception if necessary

Randgalt: Remember to squash those commits

PR #1701, opened 2025-12-31
Validate skipped properties in assertFullMapping 

findepi:
Throw an error when a skipped property does not exist (e.g. has been
removed).

### DIFF
--- a/configuration-testing/src/main/java/io/airlift/configuration/testing/ConfigAssertions.java
+++ b/configuration-testing/src/main/java/io/airlift/configuration/testing/ConfigAssertions.java
@@ -16,6 +16,7 @@
 package io.airlift.configuration.testing;
 
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
 import io.airlift.configuration.ConfigurationFactory;
 import io.airlift.configuration.ConfigurationMetadata;
 import io.airlift.configuration.ConfigurationMetadata.AttributeMetadata;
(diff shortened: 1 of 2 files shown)


configuration-testing/src/main/java/io/airlift/configuration/testing/ConfigAssertions.java
```diff
@@ -116,18 +117,25 @@ public static <T> void assertFullMapping(Map<String, String> properties, T expec
...
+        // verify that every (non-deprecated and not explicitly skipped) property is tested
+        Set<String> allProperties = new HashSet<>();
         for (AttributeMetadata attribute : metadata.getAttributes().values()) {
             if (attribute.getInjectionPoint().getProperty() != null) {
-                nonSkippedProperties.add(attribute.getInjectionPoint().getProperty());
+                allProperties.add(attribute.getInjectionPoint().getProperty());
             }
         }
-        nonSkippedProperties.removeAll(skipped);
+        Set<String> invalidSkipped = new TreeSet<>(Sets.difference(skipped, allProperties));
```
chenjian2664, line 127: Do we need to use a `TreeSet` here, or just for reporting sorted properties in message?

findepi: TreeSet is just for having it sorted conveniently. 
Yes, it's only for the sake of the message.

chenjian2664: Thanks

PR #1702, opened 2025-12-31
[MCP Server] Support cancellations

Randgalt:
Support for MCP cancellation.

This will also be the foundation for upcoming task cancellations.

### DIFF
--- a/mcp/src/main/java/io/airlift/mcp/CancellationController.java
+++ b/mcp/src/main/java/io/airlift/mcp/CancellationController.java
@@ -1,0 +1,171 @@
+package io.airlift.mcp;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Supplier;
+import com.google.common.collect.Sets;
+import com.google.inject.Inject;
+import io.airlift.log.Logger;
+import io.airlift.mcp.sessions.SessionController;
+import io.airlift.mcp.sessions.SessionId;
+import io.airlift.mcp.sessions.SessionValueKey;
+import jakarta.annotation.PreDestroy;
+
+import java.time.Duration;
+import java.util.Collection;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+
+import static com.google.common.util.concurrent.MoreExecutors.shutdownAndAwaitTermination;
+import static io.airlift.concurrent.Threads.virtualThreadsNamed;
+import static io.airlift.mcp.McpException.exception;
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.Executors.newThreadPerTaskExecutor;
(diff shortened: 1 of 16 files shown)


mcp/src/main/java/io/airlift/mcp/CancellationController.java
```diff
@@ -0,0 +1,171 @@
...
+    private final Duration interval;
+    private final ExecutorService executorService;
+
+    @Inject
+    public CancellationController(McpCancellationHandler cancellationHandler, Optional<SessionController> sessionController, McpConfig mcpConfig)
+    {
+        this.cancellationHandler = requireNonNull(cancellationHandler, "cancellationHandler is null");
+        this.sessionController = requireNonNull(sessionController, "sessionController is null");
+
+        // TODO use VirtualThreadPerTaskExecutor when we move to JDK 24+
```
wendigo: we can address this todo now, JDK 24 will be dropped soon

wendigo: use `virtualThreadsNamed()`

mcp/src/main/java/io/airlift/mcp/model/CancelledNotification.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Optional;
+
+import static com.google.common.base.MoreObjects.firstNonNull;
+
+public record CancelledNotification(Object requestId, Optional<String> reason)
+{
+    public CancelledNotification
+    {
+        requestId = firstNonNull(requestId, "");
```
wendigo, line 11: can it be null actually?

mcp/src/main/java/io/airlift/mcp/McpModule.java
```diff
@@ -169,6 +173,13 @@ public Builder withSessions(Consumer<LinkedBindingBuilder<SessionController>> se
             return this;
         }
 
+        // NOTE: does nothing if sessions are not enabled
```
wendigo: should we validate if sessions are enabled?

mcp/src/main/java/io/airlift/mcp/model/CancelledNotification.java
Randgalt: Technically, the spec allows `null` request IDs. In practice we won't see them.

mcp/src/main/java/io/airlift/mcp/McpModule.java
Randgalt: Yeah - that's a good idea. I'll add that.

mcp/src/main/java/io/airlift/mcp/model/CancelledNotification.java
wendigo: Is request id controlled by the client or the server?

Randgalt: Both - it can come from any direction. For the purposes of this PR it's only from us so there will always be a request ID. But, given that this is an MCP model we should define it in a general manner.

PR #1704, opened 2026-01-01
[MCP Server] Support replayable/resumable event streams

Randgalt:
Support replayable/resumable event streams


mcp/src/main/java/io/airlift/mcp/internal/InternalMessageWriter.java
```diff
@@ -45,7 +44,7 @@ void write(String data)
     @Override
     public void writeMessage(String data)
     {
-        String messageId = Long.toString(nextId.getAndIncrement());
+        String messageId = UUID.randomUUID().toString();
```
wendigo, line 59: we could strip dashes, also we could use ULIDs that encode the timestamp

Randgalt: OK with me

PR #1735, opened 2026-01-21
Fix internal ip resolution

lukasz-stec:
Previously, `findInternalIp` would always try to base the internal address on the hostname, and if the hostname was invalid or pointed to the localhost, it would return 127.0.0.1. This is not expected as we could still use ip addresses assigned to the network interfaces to get the correct local network address.

### DIFF
--- a/node/src/main/java/io/airlift/node/NodeInfo.java
+++ b/node/src/main/java/io/airlift/node/NodeInfo.java
@@ -15,6 +15,7 @@
  */
 package io.airlift.node;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.net.InetAddresses;
@@ -102,6 +103,38 @@ public class NodeInfo
             Map<String, String> annotations,
             boolean preferIpv6Address)
     {
+        this(environment,
+                pool,
+                nodeId,
+                internalAddress,
+                bindIp,
+                externalAddress,
+                location,
+                binarySpec,
+                configSpec,
+                internalAddressSource,
+                annotationFile,
+                annotations,
+                preferIpv6Address,
+                NodeAddresses.DEFAULT);
+    }
+
+    @VisibleForTesting
+    NodeInfo(String environment,
+            String pool,
+            String nodeId,
+            String internalAddress,
+            InetAddress bindIp,
+            String externalAddress,
+            String location,
+            String binarySpec,
+            String configSpec,
+            AddressSource internalAddressSource,
+            String annotationFile,
+            Map<String, String> annotations,
+            boolean preferIpv6Address,
+            NodeAddresses networkAddresses)
+    {
         requireNonNull(environment, "environment is null");
         requireNonNull(pool, "pool is null");
         requireNonNull(internalAddressSource, "internalAddressSource is null");
@@ -128,7 +161,7 @@ public class NodeInfo
             this.internalAddress = internalAddress;
         }
         else {
-            this.internalAddress = findInternalAddress(internalAddressSource);
+            this.internalAddress = findInternalAddress(internalAddressSource, networkAddresses);
         }
 
         this.bindIp = requireNonNullElseGet(bindIp, () -> InetAddresses.fromInteger(0));
@@ -294,20 +327,20 @@ public class NodeInfo
                 .toString();
     }
 
-    private String findInternalAddress(AddressSource addressSource)
+    private String findInternalAddress(AddressSource addressSource, NodeAddresses networkAddresses)
     {
         return switch (addressSource) {
-            case IP -> InetAddresses.toAddrString(findInternalIp());
-            case IP_ENCODED_AS_HOSTNAME -> encodeAddressAsHostname(findInternalIp());
-            case HOSTNAME -> getLocalHost().getHostName();
-            case FQDN -> getLocalHost().getCanonicalHostName();
+            case IP -> InetAddresses.toAddrString(findInternalIp(networkAddresses));
+            case IP_ENCODED_AS_HOSTNAME -> encodeAddressAsHostname(findInternalIp(networkAddresses));
+            case HOSTNAME -> getLocalHost(networkAddresses).getHostName();
+            case FQDN -> getLocalHost(networkAddresses).getCanonicalHostName();
         };
     }
 
-    private InetAddress findInternalIp()
+    private InetAddress findInternalIp(NodeAddresses networkAddresses)
     {
         List<Function<List<InetAddress>, Optional<InetAddress>>> searchOrder = new ArrayList<>(3);
-        searchOrder.add(candidates -> findLocalAddress(candidates, preferIpv6Address));
+        searchOrder.add(candidates -> findLocalAddress(candidates, preferIpv6Address, networkAddresses));
         if (this.preferIpv6Address) {
             searchOrder.add(NodeInfo::findIpv6address);
             searchOrder.add(NodeInfo::findIpv4address);
@@ -317,7 +350,7 @@ public class NodeInfo
             searchOrder.add(NodeInfo::findIpv6address);
         }
 
-        List<InetAddress> goodAddresses = getGoodAddresses();
+        List<InetAddress> goodAddresses = networkAddresses.getAddresses();
         return searchOrder.stream()
             .map(source -> source.apply(goodAddresses))
             .filter(Optional::isPresent)                        // select only valid results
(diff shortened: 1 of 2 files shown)


wendigo:
cc @jalpan-randeri

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -306,14 +306,31 @@ private String findInternalAddress(AddressSource addressSource)
 
     private InetAddress findInternalIp()
     {
+        // Check if local host address is a good v4 address
+        InetAddress localAddress = null;
+        try {
+            localAddress = InetAddress.getLocalHost();
+            if (isV4Address(localAddress) && getGoodAddresses().contains(localAddress)) {
+                return localAddress;
```
electrum: This seems wrong as it's returning a v4 address even if prefer v6 is set.

Also, it's cleaner to have the logic in a separate method.

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -324,7 +341,7 @@ private InetAddress findInternalIp()
             .filter(Optional::isPresent)                        // select only valid results
             .findFirst()
             .orElse(Optional.empty())                     // we could not find any source
-            .orElse(null);                                // it is most likely that this is a disconnected developer machine
+            .orElse(localAddress);
```
electrum: This change seems correct, as `getInternalAddress()` should not return null.

We could use `InetAddress.getLocalHost()` here.

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -349,29 +366,6 @@ private static Optional<InetAddress> findIpv6address(List<InetAddress> candidate
...
-        InetAddress localAddress = null;
-        try {
-            localAddress = InetAddress.getLocalHost();
-            if (isV4Address(localAddress) && getGoodAddresses().contains(localAddress)) {
-                return Optional.of(localAddress);
-            }
-        }
-        catch (UnknownHostException ignored) {
-        }
-        if (localAddress == null) {
```
electrum, line 364: This seems like the problematic part. We should instead return `Optional.empty()` here.

node/src/main/java/io/airlift/node/NodeInfo.java
```diff
@@ -306,14 +306,31 @@ private String findInternalAddress(AddressSource addressSource)
...
+            }
+            catch (UnknownHostException e) {
+                throw new AssertionError("Could not get local ip address");
+            }
+        }
+
         List<Function<List<InetAddress>, Optional<InetAddress>>> searchOrder = new ArrayList<>(3);
         if (this.preferIpv6Address) {
             searchOrder.add(NodeInfo::findIpv6address);
-            searchOrder.add(NodeInfo::findLocalAddress);
```
electrum, line 312: If we remove the `isV4Address()` check from `findLocalAddress()`, then this should work for the search order, which seems more consistent:
```java
List<Function<List<InetAddress>, Optional<InetAddress>>> searchOrder = new ArrayList<>(3);
searchOrder.add(NodeInfo::findLocalAddress);
if (this.preferIpv6Address) {
    searchOrder.add(NodeInfo::findIpv6address);
    searchOrder.add(NodeInfo::findIpv4address);
}
else {
    searchOrder.add(NodeInfo::findIpv4address);
    searchOrder.add(NodeInfo::findIpv6address);
}

File: node/src/main/java/io/airlift/node/NodeInfo.java (hunk 1 above)
Comment (lukasz-stec) [reply node/src/main/java/io/airlift/node/NodeInfo.java]: Good catch. That does not work. I moved back to a separate method and dropped the checks

File: node/src/main/java/io/airlift/node/NodeInfo.java (hunk 4 above)
Comment (lukasz-stec) [reply node/src/main/java/io/airlift/node/NodeInfo.java:312]: I don't think it works as `findLocalAddress` always returns a non-empty value, even if the`InetAddress.getLocalHost` call fails. Perhaps, I could add it at the end of the `searchOrder` as an alternative to the `orElse` below.

File: node/src/main/java/io/airlift/node/NodeInfo.java (hunk 2 above)
Comment (lukasz-stec) [reply node/src/main/java/io/airlift/node/NodeInfo.java]: `InetAddress.getLocalHost` can throw, and previously, in this case, we would use `127.0.0.1`.

**Review by lukasz-stec — COMMENTED** [review]
@electrum I addressed the comments, but in a slightly different way than requested. See if it makes sense to you. The other option I see is to add `findLocalAddress` at the end of the `searchOorder` instead of in the `.elseGet` part.

File: node/src/main/java/io/airlift/node/NodeInfo.java (hunk 2 above)
Comment (electrum) [reply node/src/main/java/io/airlift/node/NodeInfo.java]: Sorry, I meant `getLoopbackAddress` which was probably not used originally as this code was written in the Java 1.6 days.

File: node/src/main/java/io/airlift/node/NodeInfo.java (hunk 4 above)
Comment (electrum) [reply node/src/main/java/io/airlift/node/NodeInfo.java:312]: I was thinking that we change it to return empty if it fails, rather than loopback, as I mentioned in another comment.

If we do that, then it should be simpler and consistent across both v4/v6 preferences.

Comment (lukasz-stec) [reply node/src/main/java/io/airlift/node/NodeInfo.java:312]: ok, I will do that. The `findLocalAddress` still needs to be last, as it could return localhost

**Review by lukasz-stec — COMMENTED** [review]
Comments addressed

File: node/src/main/java/io/airlift/node/NodeInfo.java (hunk 2 above)
Comment (lukasz-stec) [reply node/src/main/java/io/airlift/node/NodeInfo.java]: Makes sense. I used `getLoopbackAddress`

File: node/src/main/java/io/airlift/node/NodeInfo.java (hunk 4 above)
Comment (electrum) [reply node/src/main/java/io/airlift/node/NodeInfo.java:312]: You're right, we should fix it to not return localhost, as we still want to prefer the local address. I'll comment on that below.

File: node/src/main/java/io/airlift/node/NodeInfo.java [hunk 5]
```diff
@@ -349,27 +348,14 @@ private static Optional<InetAddress> findIpv6address(List<InetAddress> candidate
...
     }
 
-    private static Optional<InetAddress> findLocalAddress(List<InetAddress> candidates)
+    private static Optional<InetAddress> findLocalAddress()
     {
-        // Check if local host address is a good v4 address
-        InetAddress localAddress = null;
         try {
-            localAddress = InetAddress.getLocalHost();
-            if (isV4Address(localAddress) && getGoodAddresses().contains(localAddress)) {
```
electrum, line 358: Keep `candidates` as a parameter and check if it contains the local address. We should also respect the address type preference:
```java
InetAddress address = InetAddress.getLocalHost();
boolean preferred = preferIpv6Address ? isV6Address(address) : isV4Address(address);
return preferred && candidates.contains(address);

File: node/src/main/java/io/airlift/node/NodeInfo.java [hunk 6]
```diff
@@ -309,22 +309,21 @@ private InetAddress findInternalIp()
         List<Function<List<InetAddress>, Optional<InetAddress>>> searchOrder = new ArrayList<>(3);
         if (this.preferIpv6Address) {
             searchOrder.add(NodeInfo::findIpv6address);
```
electrum, line 345: When preferring v6, `findLocalAddress` is used second, as it always returns a v4 address. We can make this consistent by always using the local address first, if we fix `findLocalAddress` to take preference into account. I'll comment below.

node/src/main/java/io/airlift/node/NodeInfo.java
lukasz-stec: Eh, I missed that we do prefer the local address if it matches IPv4/IPv6 preference. 
Done.

node/src/main/java/io/airlift/node/NodeInfo.java
lukasz-stec: Done

lukasz-stec (commented):
I updated the code to match comments, but I will try to improve tests to at least have this rule of preferring local address there if possible

lukasz-stec:
I tried, but unfortunately, I don't have an easy way to add tests for the `findInternalIp`. The implementation relies on the static calls to the `NetworkInterface.getNetworkInterfaces` and `InetAddress.getLocalHost`, so to test it properly, I would either have to use mocks or wrap it all in some interface that I can replace in tests. Neither of which is usually done in airlift .
@electrum Let me know if you think doing any of this makes sense to you, or if you have some other idea how to test it.

electrum:
This change looks good to me. Thanks for making the changes.

@wendigo do you want to take a look before we merge this?

@lukasz-stec if you want to follow up with tests, that would be great. Or if you want to add tests before we merge this, that's fine too.

PR #1737, opened 2026-01-22
Support async cancellations during HTTP sync execution

findepi:
Provide a way to asynchronously cancel HTTP requests executing using
synchronous `HttpClient.execute` API.

This is similar to cancellation available in `executeAsync`. The
synchronous API has the advantage of not buffering the response data
before invoking the request handler.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/RequestLifecycleHandler.java
+++ b/http-client/src/main/java/io/airlift/http/client/RequestLifecycleHandler.java
@@ -1,0 +1,41 @@
+package io.airlift.http.client;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.SettableFuture;
+
+import static java.util.Objects.requireNonNull;
+
+public final class RequestLifecycleHandler
+{
+    public static RequestLifecycleHandler create()
+    {
+        return new RequestLifecycleHandler();
+    }
+
+    private SettableFuture<?> cancelled;
+
+    private RequestLifecycleHandler()
+    {
+        this.cancelled = SettableFuture.create();
+    }
+
+    public void cancel()
+    {
+        requireNonNull(cancelled, "handler is destroyed").set(null);
+    }
+
+    public void destroy()
+    {
+        // Prevent reuse of the lifecycle handler. Allowing reuse would require resetting listeners on the cancelled future,
+        // or otherwise we would get a memory leak.
+        cancelled = null;
+    }
+
+    /**
+     * Returns the cancellation future. Typically used by the HTTP client implementation.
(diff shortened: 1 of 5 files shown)


http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
```diff
@@ -950,6 +950,33 @@ protected void service(HttpServletRequest request, HttpServletResponse response)
...
+            throws Exception
+    {
+        RequestLifecycleHandler lifecycleHandler = RequestLifecycleHandler.create();
+        HttpServlet servlet = new HttpServlet()
+        {
+            @Override
+            protected void doGet(HttpServletRequest request, HttpServletResponse response)
+                    throws IOException
+            {
+                lifecycleHandler.cancel();
```
losipiuk, line 964: Is that a common case that thread which is handling the response data is calling cancel(). Or would it be rather some other client thread? Worth testing the alternative flow?

http-client/src/main/java/io/airlift/http/client/RequestLifecycleHandler.java
```diff
@@ -0,0 +1,41 @@
...
+    {
+        // Prevent reuse of the lifecycle handler. Allowing reuse would require resetting listeners on the cancelled future,
+        // or otherwise we would get a memory leak.
+        cancelled = null;
+    }
+
+    /**
+     * Returns the cancellation future. Typically used by the HTTP client implementation.
+     */
+    public ListenableFuture<?> cancelled()
```
losipiuk, line 37: not a big fan of `destroy` and `cancelled` being public and visible to user of http client.

http-client/src/test/java/io/airlift/http/client/AbstractHttpClientTest.java
findepi: i think it's separate thread, because there is a TCP socket in between them, i suppose

http-client/src/main/java/io/airlift/http/client/RequestLifecycleHandler.java
findepi: Good call. This was also not clear to me. In fact, `cancelled` started as package-private.
However

- philosophical: if someone choose to implement the `HttpClient` interface elsewhere, they need access to these methods. 
- technical: if `RequestLifecycleHandler` were to utilize package-private methods to communicate with `JettyHttpClient`, it would need to live in the same package. However, `HttpClient` and `JettyHttpClient` are not in one package.
   - we would need java modules to define module-private visibility of selected `RequestLifecycleHandler` methods

PR #1740, opened 2026-01-23
Reduce data copying in HTTP async JSON requests

findepi:
When performing async HTTP requests (`JettyHttpClient.executeAsync`) to get JSON data (using `JsonResponseHandler`), the response data gets copied a few times along the way.

Before the changes the flow is:

0. Jetty receives data from socket and has it in chunks
1. `JettyResponseListener` extends `BufferingResponseListener`, so it supposedly copies incoming chunks into internal buffer(s).
2. `JettyResponseListener.onComplete` asks for an `InputStream`. It gets `ByteArrayInputStream` over a fresh `byte[]` buffer.
3. `JsonResponseHandler` calls `InputStream.readAllBytes()`, which is another data copy.

This commit combines points (2) and (3) together. The `InputStream.readAllBytes()` is skipped, because `JettyResponseListener` operates directly on content bytes and exposes them to response handler.

Potential follow-ups

- Switching from `BufferingResponseListener` to `RetainingResponseListener` to avoid data copy in (1). This change needs to make sure we don't run out of Jetty buffers or anything like that.

- Switch `JsonResponseHandler` to operate directly on `InputStream`. It uses `byte[]` for error reporting though.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyResponse.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyResponse.java
@@ -8,7 +8,9 @@ import io.airlift.http.client.HttpVersion;
 import org.eclipse.jetty.client.Response;
 import org.eclipse.jetty.http.HttpFields;
 
+import java.io.ByteArrayInputStream;
 import java.io.InputStream;
+import java.util.function.LongSupplier;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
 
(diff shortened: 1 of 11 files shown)


http-client/src/main/java/io/airlift/http/client/jetty/JettyResponseListener.java
```diff
@@ -32,7 +33,11 @@ public void onComplete(Result result)
             future.failed(result.getFailure());
         }
         else {
-            future.completed(result.getResponse(), getContentAsInputStream());
+            // TODO it would be better to return response data as InputStream based on Jetty chunks, without any data copying
+            // There is no such builtin Jetty response listener to use, see https://github.com/jetty/jetty.project/issues/14373.
```
findepi, line 37: ( https://github.com/jetty/jetty.project/issues/14373 )

http-client/src/main/java/io/airlift/http/client/jetty/JettyResponse.java
```diff
@@ -8,21 +8,37 @@
...
         implements io.airlift.http.client.Response
 {
     private final Response response;
-    private final CountingInputStream inputStream;
+    private final Content content;
+    private final InputStream inputStream;
+    private final LongSupplier bytesRead;
     private final ListMultimap<HeaderName, String> headers;
 
+    public JettyResponse(Response response, byte[] content)
```
losipiuk, line 26: where is this guy called?

findepi:
```
JettyResponse.JettyResponse(Response, byte[])  (io.airlift.http.client.jetty)
  JettyResponseFuture.processResponse(Response, byte[])  (io.airlift.http.client.jetty)
    JettyResponseFuture.completed(Response, byte[])  (io.airlift.http.client.jetty)
      JettyResponseListener.onComplete(Result)  (io.airlift.http.client.jetty)
```

PR #1741, opened 2026-01-23
Allow releasing snapshot from fork

wendigo:

### DIFF
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -3,6 +3,10 @@ name: Release snapshot version
 on:
   workflow_dispatch:
     inputs:
+      repository:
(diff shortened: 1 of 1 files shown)


.github/workflows/snapshot-release.yml
```diff
@@ -3,6 +3,10 @@ name: Release snapshot version
 on:
   workflow_dispatch:
     inputs:
+      repository:
+        description: 'The repository to release snapshot release from (use with caution!)'
+        required: true
+        default: 'airlift/airlift'
```
findepi, line 9: To safely release from a fork's branch, we need to trust the fork owner. This is not the same as fork owner being an airlift committer, but it's also not unrelated.

I would rather keep the requirement that all releases, including snapshot releases, come from the repo.

PR #1743, opened 2026-01-23
Surface JSON resource serialization errors to client

findepi:
When serialization fails in the middle, it's better to return a truncated (invalid) JSON than something that could be interpreted as a valid (but incorrect) result.

- for https://github.com/trinodb/trino/issues/27993


json/src/main/java/io/airlift/json/ObjectMapperProvider.java
```diff
@@ -199,6 +200,11 @@ public ObjectMapper get()
         // use ISO dates
         objectMapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
 
+        // When serialization fails in the middle, it's better to return a truncated (invalid) JSON
+        // than something that could be interpreted as a valid (but incorrect) result.
+        // This is especially applicable to server endpoints that return JSON responses.
+        objectMapper.disable(JsonGenerator.Feature.AUTO_CLOSE_JSON_CONTENT);
```
wendigo, line 206: This is weird to be enabled by default

PR #1746, opened 2026-01-25
[MCP Server] Improve list/resource subscriptions - fix some latency/races

Randgalt:
1. Remove periodic check of resource/list changes. reconcileVersions()
is called periodically already. No need to add to that.
2. Do all resource checks inside of a `compute_session` so that we
have a transaction (for DBs). This avoids out of order reads on some
DB and ensure serializability. This introduces a small possible loss
of messages should the system crash before the messages are sent. This
is such a rare case that we can live it.
3. Improve testListChangeNotifications() so it's more stable.

### DIFF
--- a/mcp/src/test/java/io/airlift/mcp/TestingDatabaseSessionController.java
+++ b/mcp/src/test/java/io/airlift/mcp/TestingDatabaseSessionController.java
@@ -289,7 +289,7 @@ public class TestingDatabaseSessionController
                         }
                         // otherwise, another transaction inserted the value, so we need to retry
                     }
-                    else {
+                    else if (!currentValue.equals(result.get())) {
                         PreparedStatement preparedStatement = connection.prepareStatement(UPDATE_VALUE_SQL);
                         preparedStatement.setString(1, newValueJson);
                         preparedStatement.setString(2, sessionId.id());


mcp/src/main/java/io/airlift/mcp/McpConfig.java
```diff
@@ -46,19 +45,6 @@ public McpConfig setDefaultSessionTimeout(Duration defaultSessionTimeout)
         return this;
     }
 
-    @MinDuration("1ms")
-    public Duration getResourceVersionUpdateInterval()
-    {
-        return resourceVersionUpdateInterval;
-    }
-
-    @Config("mcp.resource-version.update-interval")
```
findepi, line 55: `@DefunctConfig`?

mcp/src/main/java/io/airlift/mcp/internal/InternalMcpServer.java
```diff
@@ -387,52 +388,63 @@ void reconcileVersions(HttpServletRequest request, InternalMessageWriter message
...
+                });
             });
 
             cursor = (subscriptions.size() < RECONCILE_PAGE_SIZE) ? Optional.empty() : Optional.of(subscriptions.getLast().getKey());
         }
         while (cursor.isPresent());
+
+        // ideally, we'd send these messages before updating the session state, but that would require
+        // sending them inside of a DB transaction and that isn't ideal. So, we send them after updating the session state.
+        // There is a small chance that these messages fail to send and the client will miss the notifications.
```
findepi, line 446: Should there be some form of ACK by the client?

Randgalt: No - the protocol just requires that the server send the message when the lists have changed. Frankly, I'm not happy with this right now. There's no good way to know how often to check for list changes and this code, right now, does it far too often. Of course, it only does it for clients that are connected to the server and it's not too expensive but I think it should be improved. The MCP spec is supposed to change in the future to make at least some of these notifications unnecessary (by returning versions in headers, etc.).

mcp/src/main/java/io/airlift/mcp/McpConfig.java
Randgalt: I don't think it was ever used. This is all very new code. It's only been around for like a month. Do we still need it?

mcp/src/main/java/io/airlift/mcp/internal/InternalMcpServer.java
findepi: So there is no way to implement MCP protocol without losing messages and with exactly once delivery?

mcp/src/main/java/io/airlift/mcp/McpConfig.java
findepi: Does it cost much to add it?

mcp/src/main/java/io/airlift/mcp/internal/InternalMcpServer.java
Randgalt: The protocol has a way for clients to reconnect with a header that signals the MCP server should replay messages. We support this.

mcp/src/main/java/io/airlift/mcp/internal/InternalMcpServer.java
```diff
@@ -387,52 +388,63 @@ void reconcileVersions(HttpServletRequest request, InternalMessageWriter message
...
-                    requestContext.sendMessage(NOTIFICATION_RESOURCES_UPDATED, Optional.of(new ResourcesUpdatedNotification(uri)));
-                    localSessionController.setSessionValue(sessionId, resourceVersionKey(uri), new ResourceVersion(newHash));
-                }
+
+                localSessionController.computeSessionValue(sessionId, resourceVersionKey(uri), maybeOldVersion -> {
+                    if (maybeOldVersion.map(oldVersion -> !oldVersion.version().equals(newHash)).orElse(false)) {
+                        notifications.add(new Notification(NOTIFICATION_RESOURCES_UPDATED, Optional.of(new ResourcesUpdatedNotification(uri))));
+                        return maybeOldVersion;
+                    }
+                    return Optional.of(new ResourceVersion(newHash));
```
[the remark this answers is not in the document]
Randgalt: AI is right here. I'll push a fix.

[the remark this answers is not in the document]
Randgalt: https://github.com/airlift/airlift/pull/1748

PR #1751, opened 2026-01-28
Fix outdated dependency scopes

electrum:


skeleton-server/pom.xml
```diff
@@ -89,17 +89,16 @@
             <artifactId>jmxutils</artifactId>
         </dependency>
 
-        <!-- used by tests but also needed transitively -->
         <dependency>
-            <groupId>io.airlift</groupId>
-            <artifactId>http-client</artifactId>
+            <groupId>jakarta.ws.rs</groupId>
+            <artifactId>jakarta.ws.rs-api</artifactId>
             <scope>runtime</scope>
```
findepi: Can this runtime dep have a comment why it's runtime?
i think it's a healthy hygiene.

pom.xml
```diff
@@ -86,6 +86,7 @@
         <air.check.skip-pmd>true</air.check.skip-pmd>
         <air.check.skip-jacoco>true</air.check.skip-jacoco>
         <air.check.skip-license>true</air.check.skip-license>
+        <air.check.skip-dependency-scope>true</air.check.skip-dependency-scope>
```
findepi: why so?

electrum: This plugin exists to workaround a bug (that `test` scope would override `runtime` scope) in the assembly plugin and provisio plugin. That bug has since been fixed, so the workaround and plugin should be removed.

We could actually simply remove the plugin at this point, since the check actually enforces the wrong behavior (using `runtime` scope instead of `test` scope).

skeleton-server/pom.xml
electrum: I didn't add this dependency, but the scope seems incorrect. I'll see if it can be removed.

electrum: I moved it to `test` scope and verified that the JAR is still in the tarball.

pom.xml
findepi: I didn't fully understand..  However, whenever i see `skip-xxxx>true<` in properties, i hope it's either obvious, or explained with a code comment.
So in case this isn't obvious, a code comment would be nice.

electrum: Makes sense. I'll add a TODo comment to remove when it's removed from airbase.

PR #1760, opened 2026-01-31
Set OTEL span attributes for MCP 

Randgalt:
There are now semconv defined values for MCP.
See https://github.com/open-telemetry/semantic-conventions-java/pull/396/changes#diff-1c3840320c0c5bfe484b22e6db177a87b014591107c61f9d5763f92117ddf7ff

### DIFF
--- a/mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
+++ b/mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
@@ -32,6 +32,7 @@ import io.airlift.mcp.model.SubscribeRequest;
 import io.airlift.mcp.sessions.SessionController;
 import io.airlift.mcp.sessions.SessionId;
 import io.airlift.mcp.sessions.SessionValueKey;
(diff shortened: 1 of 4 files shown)


mcp/src/main/java/io/airlift/mcp/internal/InternalFilter.java
```diff
@@ -357,6 +370,8 @@ private void handleRpcRequest(HttpServletRequest request, HttpServletResponse re
         String rpcMethod = rpcRequest.method();
         Object requestId = rpcRequest.id();
 
+        Span.current().setAttribute(MCP_METHOD_NAME, rpcMethod);
```
wendigo: Span should be access from the request:

```
    static final String REQUEST_SPAN = "airlift.trace-span";

```

Randgalt: lol. That’s what I had at first and removed it because AI told me to

PR #1761, opened 2026-01-31
Support coercing byte and char arrays in configuration

wendigo:
This is useful when the property is security sensitive and contains secrets.


dain:
`char[]` seems completely reasonable, but `byte[]` can be problematic because you don't know the encoding.  Specifically, it is UTF-8, hex, base64, etc.  Do we have a way to annotate the property to say the expected encoding?

PR #1762, opened 2026-01-31
Make TracingServletFilter extend HttpFilter

wendigo:


http-server/src/main/java/io/airlift/http/server/tracing/TracingServletFilter.java
```diff
@@ -59,14 +57,9 @@ public TracingServletFilter(OpenTelemetry openTelemetry, Tracer tracer)
     }
 
     @Override
-    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+    public void doFilter(HttpServletRequest httpRequest, HttpServletResponse httpResponse, FilterChain chain)
```
electrum: How about we keep the names “request” and “response” for the parameters

PR #1764, opened 2026-02-02
Opentelemetry: add otel.resource.attributes config

shivamvishal:


whitleykeith:
@wendigo I think this change (or the ability to customize the prefix) is needed as we have similar constraints that @shivamvishal mentioned. Our tooling also requires certain properties to be define and having that hardcoded prefix makes it functionally impossible

PR #1768, opened 2026-02-05
[MCP Server] Improve lists/resource subscriptions

Randgalt:
- Refactor list/resources changes and subscriptions into VersionsController
- Cache the resource contents via configurable period to improve performance and I/O load.
- Add `SYSTEM_LIST_VERSIONS` and `RESOURCE_VERSIONS`
to cacheable types. The assertion here is that each McpServer
instances in JVM will run `VersionsController.reconcileVersions()`
periodically which will see the same tools/prompts/resources/etc.
and the same subscribed resource contents and, so, can safely
maintain in-memory caches of these values.

### DIFF
--- a/mcp/src/main/java/io/airlift/mcp/McpConfig.java
+++ b/mcp/src/main/java/io/airlift/mcp/McpConfig.java
@@ -16,6 +16,7 @@ public class McpConfig
     private Duration eventStreamingPingThreshold = new Duration(15, SECONDS);
     private Duration eventStreamingTimeout = new Duration(5, MINUTES);
     private Duration cancellationCheckInterval = new Duration(1, SECONDS);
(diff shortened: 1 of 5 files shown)


mcp/src/main/java/io/airlift/mcp/versions/ResourceVersions.java
```diff
@@ -0,0 +1,30 @@
...
+        Map<String, String> updatedMap = ImmutableMap.<String, String>builder()
+                .putAll(uriToVersion)
+                .put(uri, version)
+                .buildKeepingLast();
+        return new ResourceVersions(updatedMap);
+    }
+
+    public ResourceVersions without(String uri)
+    {
+        Map<String, String> updatedMap = filterKeys(uriToVersion, entryUri -> !entryUri.equals(uri));
```
wendigo: inline variable in the next line

PR #1769, opened 2026-02-06
Add MCP capability filtering

codyzwief:
This commit adds the ability to additionally filter MCP operations after authentication.

For example, in some environments, some users may not be able to use some tools that others may be able to, or server administrators want to slowly roll out tools to small beta groups of users.

### DIFF
--- a/mcp/src/main/java/io/airlift/mcp/McpCapabilityFilter.java
+++ b/mcp/src/main/java/io/airlift/mcp/McpCapabilityFilter.java
@@ -1,0 +1,56 @@
+package io.airlift.mcp;
+
+import io.airlift.mcp.McpIdentity.Authenticated;
+import io.airlift.mcp.model.CompleteReference;
+import io.airlift.mcp.model.Prompt;
+import io.airlift.mcp.model.Resource;
+import io.airlift.mcp.model.ResourceTemplate;
+import io.airlift.mcp.model.Tool;
+
+import java.util.List;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
+
+/**
+ * Filter for controlling access to MCP capabilities based on authenticated identity.
+ */
+public interface McpCapabilityFilter
+{
+    boolean isAllowed(Authenticated<?> identity, Tool tool);
+
+    boolean isAllowed(Authenticated<?> identity, Prompt prompt);
+
+    boolean isAllowed(Authenticated<?> identity, Resource resource);
+
+    boolean isAllowed(Authenticated<?> identity, ResourceTemplate resourceTemplate);
+
+    boolean isAllowed(Authenticated<?> identity, CompleteReference reference);
+
+    default List<Tool> allowedTools(Authenticated<?> identity, List<Tool> tools)
+    {
+        return tools.stream()
+                .filter(tool -> isAllowed(identity, tool))
+                .collect(toImmutableList());
+    }
+
+    default List<Prompt> allowedPrompts(Authenticated<?> identity, List<Prompt> prompts)
+    {
+        return prompts.stream()
+                .filter(prompt -> isAllowed(identity, prompt))
+                .collect(toImmutableList());
+    }
+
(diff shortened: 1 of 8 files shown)


codyzwief:
cc @Randgalt

mcp/src/main/java/io/airlift/mcp/internal/InternalMcpModule.java
```diff
@@ -23,5 +23,6 @@ public void configure(Binder binder)
         newSetBinder(binder, Filter.class).addBinding().to(InternalFilter.class).in(SINGLETON);
 
         newOptionalBinder(binder, ErrorHandler.class).setDefault().to(InternalErrorHandler.class);
+        newOptionalBinder(binder, ErrorHandler.class).setDefault().to(InternalErrorHandler.class);
```
Randgalt: Mistake?

mcp/src/main/java/io/airlift/mcp/McpCapabilityFilter.java
```diff
@@ -0,0 +1,26 @@
...
+import io.airlift.mcp.model.Tool;
+
+import java.util.List;
+
+/**
+ * Filter for controlling access to MCP capabilities based on authenticated identity.
+ */
+public interface McpCapabilityFilter
+{
+    List<Tool> filterTools(Authenticated<?> identity, List<Tool> tools);
```
Randgalt: what do you think of making these filter on single tools instead of the entire list?

Randgalt: Also, what about filtering on `ToolEntry`, `PromptEntry`? It might make filtering easier.

wendigo: Isn't it better to filter by single entity rather then a whole list?w

codyzwief: I originally had it that way, but changed my mind because I thought it would be easier for downstream implementations to not have to maintain as much state. But, I don't feel strongly about that at all so I'll switch back to filtering by single entity.

> Also, what about filtering on ToolEntry, PromptEntry? It might make filtering easier.

I can do that. I chose `Tool`, `Prompt` etc just so users only have one object type to care about. But, I'll change to the XXXEntry data structures.

wendigo: I've changed this PR a bit

mcp/src/main/java/io/airlift/mcp/internal/InternalMcpServer.java
Randgalt, line 233: These need to be adjusted to filter as they're used to determine list change notifications for the current session.

mcp/src/main/java/io/airlift/mcp/McpCapabilityFilter.java
```diff
@@ -0,0 +1,56 @@
...
+import io.airlift.mcp.model.Tool;
+
+import java.util.List;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
+
+/**
+ * Filter for controlling access to MCP capabilities based on authenticated identity.
+ */
+public interface McpCapabilityFilter
```
Randgalt, line 17: Yes, this is better. 👍

mcp/src/main/java/io/airlift/mcp/internal/InternalMcpServer.java
wendigo: this requires an access to the identity, right?

Randgalt: Yes, it does. I can make the change if you want.

wendigo: I'm merging this as-is. Please follow up in a separate PR

PR #1773, opened 2026-02-10
[MCP] Ensure capability filtering occurs for list version notifications

Randgalt:
Generalize into `McpEntities` so that there are APIs to get the full/unfiltered list and a list filtered for the current
identity.

### DIFF
--- a/mcp/src/main/java/io/airlift/mcp/McpEntities.java
+++ b/mcp/src/main/java/io/airlift/mcp/McpEntities.java
@@ -1,0 +1,27 @@
+package io.airlift.mcp;
+
+import io.airlift.mcp.model.CompleteReference;
+import io.airlift.mcp.model.Prompt;
+import io.airlift.mcp.model.ReadResourceRequest;
+import io.airlift.mcp.model.Resource;
+import io.airlift.mcp.model.ResourceContents;
+import io.airlift.mcp.model.ResourceTemplate;
+import io.airlift.mcp.model.Tool;
+
+import java.util.List;
+import java.util.Optional;
+
+public interface McpEntities
+{
+    List<Tool> tools();
+
+    List<Prompt> prompts();
+
+    List<Resource> resources();
+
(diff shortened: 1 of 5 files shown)


mcp/src/main/java/io/airlift/mcp/McpCapabilityFilter.java
```diff
@@ -26,31 +22,5 @@ public interface McpCapabilityFilter
...
-                .collect(toImmutableList());
-    }
-
-    default List<ResourceTemplate> allowedResourceTemplates(Authenticated<?> identity, List<ResourceTemplate> resourceTemplates)
-    {
-        return resourceTemplates.stream()
-                .filter(resourceTemplate -> isAllowed(identity, resourceTemplate))
-                .collect(toImmutableList());
-    }
+    boolean isAllowed(Authenticated<?> identity, String resourceUri);
```
wendigo, line 25: This is a local file right? Can it be modeled as Path ?

Randgalt: It doesn't have to be a local file. For example, in the `McpApp`s extension these are made-up "ui://" references.

Randgalt: FWIW - I thought about getting rid of `boolean isAllowed(Authenticated<?> identity, Resource resource)` and `boolean isAllowed(Authenticated<?> identity, ResourceTemplate resourceTemplate)` and replacing with this. But, that's a bit complicated.

wendigo: @Randgalt can it be URI then? :)

Randgalt: It could but it's not anywhere else and the Spec always refers to them as `string`. Maybe we can come up with our own data structure. Can we do that later?

PR #1779, opened 2026-02-13
Add non-materializing/streaming JSON handler

wendigo:
It avoids copying response into a single byte[] array. Instead it's using Jackson's JsonParser that operates directly on the Response InputStream avoiding copying whole response into memory.

Jackson maintains small internal buffer while reading, while Jetty writes response bytes into pooled ByteBuffers. In 12.1.7 (jetty/jetty.project/pull/14433) Jetty will avoid final materialization to a byte array once InputStream for Response is obtained.

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/FullJsonResponseHandler.java
+++ b/http-client/src/main/java/io/airlift/http/client/FullJsonResponseHandler.java
@@ -33,6 +33,13 @@ import static java.lang.String.format;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Objects.requireNonNull;
 
+/**
+ * FullJsonResponseHandler is a {@link ResponseHandler} that creates a JSON entity from the response bytes when Content-Type
+ * is application/json. In contrast to {@link StreamingJsonResponseHandler} it always buffers a whole response, materializing
+ * it to a byte[]. This allows for retrieval of the response bytes if decoding JSON fails, but at the same time does
+ * data copying which sits in Jetty client buffers. If the response is rather large and debugabillity can be sacrified,
+ * {@link StreamingJsonResponseHandler} should be used instead.
+ */
 public class FullJsonResponseHandler<T>
         implements ResponseHandler<JsonResponse<T>, RuntimeException>
 {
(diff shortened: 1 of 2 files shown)


findepi:
Does this work with http client's synchronous, asynchronous API or both?
Should there be a test with http client?

Is this meant to address https://github.com/airlift/airlift/blob/83ba213e2e5072df62cf285a845582cb9a2771af/http-client/src/main/java/io/airlift/http/client/jetty/JettyResponseListener.java#L36-L37 ?

http-client/src/main/java/io/airlift/http/client/JsonResponse.java
```diff
@@ -0,0 +1,81 @@
...
+    T jsonValue();
+
+    Request request();
+
+    record JsonValue<T>(Request request, int statusCode, Multimap<HeaderName, String> headers, T jsonValue, long bytesRead)
+            implements JsonResponse<T>
+    {
+        public JsonValue
+        {
+            requireNonNull(jsonValue, "jsonValue is null");
```
dain, line 26: Is request nullable?

http-client/src/main/java/io/airlift/http/client/JsonResponse.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    record Exception<T>(Request request, int statusCode, Throwable throwable)
+            implements JsonResponse<T>
+    {
+        public Exception
+        {
+            requireNonNull(throwable, "throwable is null");
```
dain, line 36: Is request nullable?

http-client/src/main/java/io/airlift/http/client/JsonResponse.java
```diff
@@ -0,0 +1,81 @@
...
+            throw new IllegalStateException("Response does not contain a JSON value", throwable);
+        }
+    }
+
+    record NonJsonBytes<T>(Request request, int statusCode, Multimap<HeaderName, String> headers, byte[] body, Throwable throwable)
+            implements JsonResponse<T>
+    {
+        public NonJsonBytes
+        {
+            requireNonNull(body, "body is null");
```
dain: Is request nullable?

dain (approved):
This seems reasonable to me.

BTW, the reason we went with a JSON response design was so that we could see the json when Jackson decoding failed, and most responses were tiny.  This design makes the tradeoff in the other direction, and that seems good to give the developer the choice here.  It would be good to add javadocs to this and the existing Json response stating the tradeoff and giving a pointer to the other implementation.

http-client/src/main/java/io/airlift/http/client/JsonResponse.java
wendigo: No, fixed

http-client/src/main/java/io/airlift/http/client/JsonResponse.java
wendigo: No, fixed

PR #1782, opened 2026-02-18
Check if log is loggable before calling jul

wendigo:

### DIFF
--- a/log/src/main/java/io/airlift/log/Logger.java
+++ b/log/src/main/java/io/airlift/log/Logger.java
@@ -292,6 +292,16 @@ public class Logger
         return logger.isLoggable(INFO);
     }
 
+    public boolean isWarnEnabled()
+    {
+        return logger.isLoggable(WARNING);
+    }
+
+    public boolean isErrorEnabled()
+    {
+        return logger.isLoggable(SEVERE);
+    }
+
     private String formatMessage(String format, String level, Object[] args)
     {
         String message;
--- a/log-manager/src/test/java/io/airlift/log/TestLogging.java
+++ b/log-manager/src/test/java/io/airlift/log/TestLogging.java
@@ -89,18 +89,26 @@ public class TestLogging
         logging.setLevel("testPropagatesLevels", Level.ERROR);
         assertThat(logger.isDebugEnabled()).isFalse();
         assertThat(logger.isInfoEnabled()).isFalse();
+        assertThat(logger.isWarnEnabled()).isFalse();
+        assertThat(logger.isErrorEnabled()).isTrue();
 
         logging.setLevel("testPropagatesLevels", Level.WARN);
         assertThat(logger.isDebugEnabled()).isFalse();
         assertThat(logger.isInfoEnabled()).isFalse();
+        assertThat(logger.isWarnEnabled()).isTrue();
+        assertThat(logger.isErrorEnabled()).isTrue();
 
         logging.setLevel("testPropagatesLevels", Level.INFO);
         assertThat(logger.isDebugEnabled()).isFalse();
         assertThat(logger.isInfoEnabled()).isTrue();
+        assertThat(logger.isWarnEnabled()).isTrue();
+        assertThat(logger.isErrorEnabled()).isTrue();
 
         logging.setLevel("testPropagatesLevels", Level.DEBUG);
         assertThat(logger.isDebugEnabled()).isTrue();
         assertThat(logger.isInfoEnabled()).isTrue();
+        assertThat(logger.isWarnEnabled()).isTrue();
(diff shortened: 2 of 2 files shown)


log/src/main/java/io/airlift/log/Logger.java
```diff
@@ -221,7 +231,9 @@ public void warn(String format, Object... args)
      */
     public void error(Throwable exception, String message)
     {
-        logger.log(SEVERE, message, exception);
+        if (isErrorEnabled()) {
```
findepi: This `if` is not needed.

log/src/main/java/io/airlift/log/Logger.java
```diff
@@ -231,7 +243,9 @@ public void error(Throwable exception, String message)
      */
     public void error(String message)
     {
-        logger.severe(message);
+        if (isErrorEnabled()) {
```
findepi: This `if` is not needed.

log/src/main/java/io/airlift/log/Logger.java
```diff
@@ -169,7 +177,9 @@ public void warn(Throwable exception, String message)
      */
     public void warn(String message)
     {
-        logger.log(WARNING, message);
+        if (isWarnEnabled()) {
```
findepi: This `if` is not needed.

log/src/main/java/io/airlift/log/Logger.java
```diff
@@ -159,7 +165,9 @@ public void info(String format, Object... args)
      */
     public void warn(Throwable exception, String message)
     {
-        logger.log(WARNING, message, exception);
+        if (isWarnEnabled()) {
```
findepi: This `if` is not needed.

log/src/main/java/io/airlift/log/Logger.java
```diff
@@ -127,7 +131,9 @@ public void debug(Throwable exception, String format, Object... args)
      */
     public void info(String message)
     {
-        logger.log(INFO, message);
+        if (isInfoEnabled()) {
```
findepi: This `if` is not needed.

log/src/main/java/io/airlift/log/Logger.java
```diff
@@ -66,7 +66,9 @@ public static Logger get(String name)
      */
     public void debug(Throwable exception, String message)
     {
-        logger.log(FINE, message, exception);
+        if (isDebugEnabled()) {
```
findepi: This `if` is not needed.

log/src/main/java/io/airlift/log/Logger.java
```diff
@@ -76,7 +78,9 @@ public void debug(Throwable exception, String message)
      */
     public void debug(String message)
     {
-        logger.log(FINE, message);
+        if (isDebugEnabled()) {
```
findepi: This `if` is not needed.

wendigo: why?

wendigo: it's applied on some methods but not for all

wendigo: i.e.
```
@FormatMethod
    public void info(String format, Object... args)
    {
        if (logger.isLoggable(INFO)) {
            logger.log(INFO, formatMessage(format, "INFO", args));
        }
    }
```

wendigo: I've dropped other ifs like that instead. PTAL

findepi: > it's applied on some methods but not for all

Because it's needed in some methods but not all.

When arguments are simply passed on to the JULI, we don't need to check log level on our code. No benefit of doing so.
When arguments are not simply passed on to JULI (e.g. string formatting, or exception.getMessage()), we check log level to avoid expensive, or potentially expensive, operation before JULI code has a chance to check log level.

log/src/main/java/io/airlift/log/Logger.java
```diff
@@ -115,9 +115,7 @@ public void debug(String format, Object... args)
     @FormatMethod
     public void debug(Throwable exception, String format, Object... args)
     {
-        if (logger.isLoggable(FINE)) {
```
findepi, line 118: restore

(here & others, where if was removed)

wendigo: Why?

```
 public void log(Level level, String msg, Throwable thrown) {
        if (!isLoggable(level)) {
            return;
        }
```

findepi: > Why?

https://github.com/airlift/airlift/pull/1782#discussion_r2822859715

PR #1784, opened 2026-02-18
Restore log level guards around expensive calls

wendigo:
This avoid formatting if the log level is not enabled.
Restore log levels accidentally removed in https://github.com/airlift/airlift/pull/1782.

### DIFF
--- a/log/src/main/java/io/airlift/log/Logger.java
+++ b/log/src/main/java/io/airlift/log/Logger.java
@@ -236,7 +236,7 @@ public class Logger
      */
     public void error(String message)
     {
-        logger.severe(message);
+        logger.log(SEVERE, message);
     }
 
     /**


log/src/main/java/io/airlift/log/Logger.java
findepi, line 282: here too

wendigo: is getMessage expensive?

findepi: can it be?

wendigo: PTAL. This avoid double if-ology

log/src/main/java/io/airlift/log/Logger.java
```diff
@@ -302,16 +303,16 @@ public boolean isErrorEnabled()
...
-        String message;
-        try {
-            message = format(format, args);
-        }
-        catch (IllegalFormatException e) {
-            logger.log(SEVERE, format("Invalid format string while trying to log: %s '%s' %s", level, format, asList(args)), e);
-            message = format("'%s' %s", format, asList(args));
-        }
-        return message;
+        return () -> {
```
findepi: This is allocation unless eliminated by the JIT.

What prevents us from adding those simple IFs back?

wendigo: Isn't that the case in majority of the cases when log level is disabled?

log/src/main/java/io/airlift/log/Logger.java
```diff
@@ -277,7 +278,7 @@ public void error(Throwable exception)
      * @param args arguments for the format string
      */
     @FormatMethod
-    public void error(String format, Object... args)
+    public void error(@FormatString String format, Object... args)
```
findepi: I thought that when error-prone sees `@FormatMethod` that takes `(String, Object...)`, it can figure out which argument is the format string.

wendigo: Dropped

PR #1786, opened 2026-02-19
Support labeled stats 

grantatspothero:

### DIFF
--- a/stats/src/main/java/io/airlift/stats/labeled/LabeledStatRegistry.java
+++ b/stats/src/main/java/io/airlift/stats/labeled/LabeledStatRegistry.java
@@ -17,8 +17,8 @@ import org.weakref.jmx.MBeanExporter;
 
 /**
  * A LabeledStatRegistry is a factory for constructing labeled stats
- * There exists a single implementation GlobalLabeledStatRegistry which is intended to be a singleton per JVM
- * In future could create several registry implementations within same process with different configuration
+ * There exists a single implementation GlobalLabeledStatRegistry which is intended to be a singleton per MBeanExporter
+ * In future could create several registry implementations with different configuration
  */
 public interface LabeledStatRegistry
 {
--- a/stats/pom.xml
+++ b/stats/pom.xml
@@ -31,6 +31,12 @@
         </dependency>
 
         <dependency>
+            <groupId>com.google.inject</groupId>
+            <artifactId>guice</artifactId>
+            <classifier>classes</classifier>
+        </dependency>
+
+        <dependency>
             <groupId>io.airlift</groupId>
             <artifactId>configuration</artifactId>
         </dependency>
--- a/stats/src/main/java/io/airlift/stats/labeled/GlobalLabeledStatRegistry.java
+++ b/stats/src/main/java/io/airlift/stats/labeled/GlobalLabeledStatRegistry.java
@@ -13,6 +13,7 @@
  */
 package io.airlift.stats.labeled;
 
+import com.google.inject.Inject;
 import org.weakref.jmx.MBeanExporter;
 
 import static java.util.Objects.requireNonNull;
@@ -23,6 +24,7 @@ public class GlobalLabeledStatRegistry
     private final MBeanExporter exporter;
     private final int labeledStatMaxCardinality;
 
+    @Inject
     public GlobalLabeledStatRegistry(MBeanExporter mBeanExporter, LabeledStatConfig config)
     {
         this.exporter = requireNonNull(mBeanExporter, "mBeanExporter is null");
--- a/stats/src/main/java/io/airlift/stats/labeled/LabeledStatModule.java
+++ b/stats/src/main/java/io/airlift/stats/labeled/LabeledStatModule.java
@@ -11,23 +11,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.airlift.openmetrics;
+package io.airlift.stats.labeled;
 
 import com.google.inject.Binder;
 import com.google.inject.Module;
-import io.airlift.stats.labeled.LabeledStatModule;
+import com.google.inject.Scopes;
 
 import static io.airlift.configuration.ConfigBinder.configBinder;
-import static io.airlift.jaxrs.JaxrsBinder.jaxrsBinder;
 
-public class JmxOpenMetricsModule
+public class LabeledStatModule
         implements Module
 {
     @Override
     public void configure(Binder binder)
     {
-        binder.install(new LabeledStatModule());
-        configBinder(binder).bindConfig(MetricsConfig.class);
-        jaxrsBinder(binder).bind(MetricsResource.class);
+        configBinder(binder).bindConfig(LabeledStatConfig.class);
+        binder.bind(LabeledStatRegistry.class).to(GlobalLabeledStatRegistry.class).in(Scopes.SINGLETON);
     }
 }


openmetrics/src/main/java/io/airlift/openmetrics/types/BigCounter.java
```diff
@@ -13,25 +13,30 @@
...
 import static java.util.Objects.requireNonNull;
 
-public record BigCounter(String metricName, BigInteger value, Map<String, String> labels, String help)
+public record BigCounter(String metricName, BigInteger value, LabelSet labels, String help)
         implements Metric
 {
+    public BigCounter
+    {
+        requireNonNull(metricName, "metricName is null");
+        requireNonNull(value, "value is null");
```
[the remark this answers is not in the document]
grantatspothero: The validations were missing before, I did not change behavior as part of this commit. 

The validations should be present though, can add as a separate commit.

openmetrics/src/main/java/io/airlift/openmetrics/types/Metric.java
```diff
@@ -14,38 +14,97 @@
...
+    String getMetricExposition(boolean includeDescriptor);
 
-    static String formatSingleValuedMetric(String name, String type, String help, Map<String, String> labels, String value)
+    default String getMetricExposition()
     {
-        return TYPE_LINE_FORMAT.formatted(name, type) +
-                (Strings.isNullOrEmpty(help) ? "" : HELP_LINE_FORMAT.formatted(name, help)) +
-                VALUE_LINE_FORMAT.formatted(formatNameWithLabels(name, labels), value);
+        return getMetricExposition(true);
     }
```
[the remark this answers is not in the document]
grantatspothero: Changes are not breaking.

stats/src/main/java/io/airlift/stats/labeled/LabelSet.java
```diff
@@ -0,0 +1,122 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Wrapper around a sorted ImmutableMap of labels
+ * Provides deterministic ordering and iteration of labels, as well as equals and hashCode
+ */
+public class LabelSet
+{
+    private static final Pattern LABEL_NAME_PATTERN = Pattern.compile("[a-zA-Z_][a-zA-Z0-9_.]*");
```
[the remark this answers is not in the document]
grantatspothero: This is intentional, "." is commonly used in metric names. The regex provided by prometheus is suggested, but not required. 

Metrics systems ingest metrics with "." without problems (confirmed with AMP and datadog). 

I'll update the comment.

stats/src/main/java/io/airlift/stats/labeled/AbstractLabeledStat.java
```diff
@@ -0,0 +1,145 @@
...
+        return labeledStats.computeIfAbsent(labelSet, l -> {
+            T stat = createStat(labelSet);
+            try {
+                labeledStatRegistry.getMBeanExporter().export(objectNameGenerator.generatedNameOf(l.asMap()), stat);
+            }
+            catch (Exception e) {
+                log.error(e, "Failed to register MBean for labels %s", l);
+            }
+            return stat;
+        });
```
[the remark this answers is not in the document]
grantatspothero: is not a problem, for new stats last updated doesn't need to be updated. 

alternatively, could remove TTL support entirely for a first pass.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -292,25 +305,37 @@ private List<Metric> getMetricsRecursively(String prefix, ManagedClass managedCl
...
+            // if labeled stats are nested within another managed class, ignore them
+            default -> Optional.empty();
+        });
+    }
 
-        return Optional.empty();
+    private Optional<Metric> getMetricFromManagedClass(ManagedClass managedClass)
+    {
+        return getTarget(managedClass).flatMap(t -> switch (t) {
+            case LabeledCounterStat.CounterStat labeledCounterStat -> Optional.of(Counter.fromLabeledCounterStat(labeledCounterStat));
```
grantatspothero, line 333: TODO: need to merge nodeinfo level labels with the stat specific labels.

openmetrics/src/main/java/io/airlift/openmetrics/types/Metric.java
```diff
@@ -27,10 +27,18 @@ public interface Metric
 
     String metricName();
 
-    String getMetricExposition();
+    String getMetricExposition(boolean includeDescriptor);
 
-    static String formatSingleValuedMetric(String name, String type, String help, Map<String, String> labels, String value)
+    default String getMetricExposition()
```
losipiuk, line 30: Are there a lot of callsites? If possible lets not have this override and provide parameter explicitly everywhere where needed.

openmetrics/src/main/java/io/airlift/openmetrics/types/Metric.java
```diff
@@ -27,10 +27,18 @@ public interface Metric
...
     {
+        return getMetricExposition(true);
+    }
+
+    static String formatSingleValuedMetric(String name, String type, String help, Map<String, String> labels, String value, boolean includeDescriptor)
+    {
+        if (!includeDescriptor) {
```
losipiuk: reverse condition (drop negation)

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -331,11 +333,30 @@ private Stream<Metric> getManagedMetricsStream()
...
     }
 
-    private String managedMetricExpositions()
+    @VisibleForTesting
+    static String managedMetricExpositions(Stream<Metric> managedMetrics)
     {
         StringBuilder builder = new StringBuilder();
-
-        getManagedMetricsStream().forEach(metric -> builder.append(metric.getMetricExposition()));
+        Map<String, List<Metric>> metricFamilies = managedMetrics
```
losipiuk: You are materializing metrics here anyway - why pass Stream as argument? Maybe just pass a list to avoid mutating argument as part of calling this method.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -331,11 +333,30 @@ private Stream<Metric> getManagedMetricsStream()
...
+                .collect(Collectors.groupingBy(
+                        Metric::metricName,
+                        LinkedHashMap::new,
+                        Collectors.toList()));
+
+        // Only include metric descriptor once per metric family
+        metricFamilies.forEach((metricName, metricFamily) -> {
+            Class<?> clazz = metricFamily.get(0).getClass();
+            Optional<Metric> classMismatch = metricFamily.stream()
```
losipiuk: Maybe just build Set of all the "classes" and check if its size is 1 at the end.
Will be easier to follow I think.

openmetrics/src/main/java/io/airlift/openmetrics/types/BigCounter.java
```diff
@@ -21,12 +21,10 @@
 public record BigCounter(String metricName, BigInteger value, Map<String, String> labels, String help)
         implements Metric
 {
-    public BigCounter(String metricName, BigInteger value, Map<String, String> labels, String help)
+    public BigCounter
```
losipiuk, line 26: `Remove duplicate record constructor from metrics` LGTM

openmetrics/src/main/java/io/airlift/openmetrics/types/CompositeMetric.java
```diff
@@ -39,6 +39,12 @@ public record CompositeMetric(String metricName, Map<String, String> labels, Str
             .trimResults()
             .limit(3);
 
+    public CompositeMetric
```
losipiuk, line 43: `Add compact constructor for CompositeMetric` LGTM

openmetrics/src/main/java/io/airlift/openmetrics/types/BigCounter.java
```diff
@@ -25,6 +25,7 @@ public record BigCounter(String metricName, BigInteger value, Map<String, String
     {
         requireNonNull(metricName, "metricName is null");
         requireNonNull(value, "value is null");
+        requireNonNull(labels, "labels is null");
```
losipiuk, line 30: Should this be defensive copy?
I would assume "no" due to performance impact.
@wendigo ?

stats/src/main/java/io/airlift/stats/labeled/LabelSet.java
```diff
@@ -0,0 +1,123 @@
...
+        return canonicalLabels.hashCode();
+    }
+
+    @Override
+    public String toString()
+    {
+        return canonicalLabels.toString();
+    }
+
+    private static ImmutableMap<String, String> canonicalize(Map<String, String> labels)
```
losipiuk, line 101: nit: declare return type as `Map`

stats/src/main/java/io/airlift/stats/labeled/LabelSet.java
```diff
@@ -0,0 +1,123 @@
...
+    {
+        if (labels.isEmpty()) {
+            return ImmutableMap.of();
+        }
+
+        String[] keys = labels.keySet().toArray(new String[0]);
```
losipiuk, line 107: use labels.size() to preallocate array

losipiuk: You can also just do `labels.keySet().stream().sorted(...` and iterate over that, without explicitly materializing keys in array.

stats/src/main/java/io/airlift/stats/labeled/LabelSet.java
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Wrapper around a sorted ImmutableMap of labels
+ * Provides deterministic ordering and iteration of labels, as well as equals and hashCode
+ */
+public class LabelSet
+{
+    // suggested regex, with additional character ".": https://prometheus.io/docs/concepts/data_model/
+    private static final Pattern LABEL_NAME_PATTERN = Pattern.compile("[a-zA-Z_][a-zA-Z0-9_.]*");
+    public static final LabelSet EMPTY = new LabelSet(ImmutableMap.of());
+    public static final Comparator<LabelSet> LABEL_SET_COMPARATOR = (labelSetOne, labelSetTwo) -> {
```
losipiuk, line 36: why publci?

stats/src/main/java/io/airlift/stats/labeled/LabelSet.java
```diff
@@ -0,0 +1,123 @@
...
+
+            cmp = e1.getValue().compareTo(e2.getValue());
+            if (cmp != 0) {
+                return cmp;
+            }
+        }
+        return Integer.compare(labelsOne.size(), labelsTwo.size());
+    };
+
+    private final ImmutableMap<String, String> canonicalLabels;
```
losipiuk, line 59: declare as Map

stats/src/main/java/io/airlift/stats/labeled/LabelSet.java
losipiuk, line 36: Seems like you can write it more concise using `Comparators.lexicographical`

openmetrics/src/main/java/io/airlift/openmetrics/types/Metric.java
```diff
@@ -48,16 +46,38 @@ static String formatSingleValuedMetric(String name, String type, String help, Im
                 VALUE_LINE_FORMAT.formatted(formatNameWithLabels(name, labels), value);
     }
 
-    static String formatNameWithLabels(String name, Map<String, String> labels)
+    static String formatNameWithLabels(String name, ImmutableMap<String, String> labels)
```
losipiuk, line 49: Revert ImmutableMap to Map (do you have you intellij misconfigured somehow?)

openmetrics/src/main/java/io/airlift/openmetrics/types/Metric.java
```diff
@@ -48,16 +46,38 @@ static String formatSingleValuedMetric(String name, String type, String help, Im
...
         }
-        return NAME_WITH_LABELS_LINE_FORMAT.formatted(
-                name,
-                labels.entrySet().stream()
-                        .sorted(Map.Entry.comparingByKey())
-                        .map(e -> "%s=\"%s\"".formatted(e.getKey(), e.getValue()))
-                        .collect(Collectors.joining(",")));
+
+        // Assuming each label key and value are 2 characters each, safe under-estimate
+        StringBuilder stringBuilder = new StringBuilder(name.length() + 4 * labels.size());
```
losipiuk, line 56: I do not get the initial size and comment here

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
findepi: is this still relevant?

openmetrics/src/main/java/io/airlift/openmetrics/types/Metric.java
findepi, line 30: Should this be deprecated in favor of the new `getMetricExposition(boolean)` variant?

stats/src/main/java/io/airlift/stats/labeled/AbstractLabeledStat.java
```diff
@@ -0,0 +1,107 @@
...
+    {
+        // fast path for existing labelsets
+        T existingStat = labeledStats.get(labelSet);
+        if (existingStat != null) {
+            return existingStat;
+        }
+        if (labeledStats.size() >= labeledStatRegistry.getLabeledStatMaxCardinality() && warningRateLimiter.tryAcquire()) {
+            log.warn("High cardinality detected for metric %s: %d unique labelsets", metricName, labeledStats.size());
+        }
+        return labeledStats.computeIfAbsent(labelSet, l -> {
```
losipiuk, line 60: nit: use `_` as lamda arg and use `labelSet` in lambda body.

stats/src/main/java/io/airlift/stats/labeled/LabeledObjectNameGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+package io.airlift.stats.labeled;
+
+import org.weakref.jmx.ObjectNameBuilder;
+import org.weakref.jmx.ObjectNameGenerator;
+
+import java.util.Map;
+
+import static java.util.Objects.requireNonNull;
+
+public class LabeledObjectNameGenerator
```
losipiuk, line 23: Can you add javadoc explaining the structer of generated MBean names. How metric+labels map to textual representation of mbean.

stats/src/main/java/io/airlift/stats/labeled/LabeledObjectNameGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+    public String generatedNameOf(Class<?> type, Map<String, String> properties)
+    {
+        return new ObjectNameBuilder(metricName)
+                .withProperties(properties)
+                .build();
+    }
+
+    public String generatedNameOf(Map<String, String> properties)
+    {
+        // type ignored
```
losipiuk, line 43: why can we ignore it? expand comment

stats/src/main/java/io/airlift/stats/labeled/LabeledCounterStat.java
```diff
@@ -0,0 +1,62 @@
...
+        return new CounterStat(metricName, description, labels);
+    }
+
+    public void increment(LabelSet labelSet)
+    {
+        getOrCreate(labelSet).increment();
+    }
+
+    @ThreadSafe
+    public static class CounterStat
```
losipiuk, line 41: This name is confusing. It conflicts with top level CounterStat which has different meaning. 
Maybe just call it Value, Holder or sth. It can even be same name in each of Labeled*Stat, as it internal implementation detail.

losipiuk: Also why did you decide on going with simpler state than we have in `io.airlift.stats.CounterStat`. You are not tracking counter over decaying time windows. 
Worth explaining in `LabeledCounterStat`  javadoc what are the capabilities and maybe motivation.

losipiuk, line 41: It does not ave to be public

losipiuk: Similar comments appy to other `Labeled*Stats` classes

stats/src/main/java/io/airlift/stats/labeled/LabeledStatRegistry.java
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.stats.labeled;
+
+import org.weakref.jmx.MBeanExporter;
+
+/**
+ * A LabeledStatRegistry is a factory for constructing labeled stats
+ * There exists a single implementation GlobalLabeledStatRegistry which is intended to be a singleton per JVM
```
losipiuk: `a singleton per JVM` or per classloader?
How does it work in architecture with multiple class loader like Trino?

stats/src/main/java/io/airlift/stats/labeled/LabeledStatRegistry.java
```diff
@@ -0,0 +1,43 @@
...
+ * In future could create several registry implementations within same process with different configuration
+ */
+public interface LabeledStatRegistry
```
losipiuk, line 23: Wouldn't it be better to follow current pattern where objects mapped to statistics are fields marked with @Managed. 

You can add support for marking `LabeledCounterStat`, `LabeledGaugeStat`, ... as @Managed. 
the programatic creation of those would not be needed.

stats/src/main/java/io/airlift/stats/labeled/AbstractLabeledStat.java
```diff
@@ -0,0 +1,107 @@
...
+        this.labeledStatRegistry = requireNonNull(labeledStatRegistry, "labeledStatRegistry is null");
+        this.metricName = validateMetricName(metricName);
+        this.objectNameGenerator = new LabeledObjectNameGenerator(metricName);
+        this.description = requireNonNull(description, "description is null");
+        this.labeledStats = new ConcurrentHashMap<>();
+    }
+
+    protected abstract T createStat(LabelSet labels);
+
+    protected T getOrCreate(LabelSet labelSet)
```
losipiuk, line 50: when are stats destroyed?

stats/src/main/java/io/airlift/stats/labeled/LabelSet.java
grantatspothero: we are relying on the stable iteration order of the keys in immutablemap, which is not a part of the map interface.

openmetrics/src/main/java/io/airlift/openmetrics/types/Metric.java
grantatspothero: see above comment re: stable iteration order. 

maybe more explicit is using a list of tuples instead of a map?

stats/src/main/java/io/airlift/stats/labeled/LabeledCounterStat.java
grantatspothero: Re: why not time decaying values. 

Modern metric specs (ie: openmetrics spec) define counters as cumulative and not time decayed. 

If you wish to do time decaying you do so through the time series database at query time. The benefit is correct aggregation of values/less time series/less load on TSDB.

stats/src/main/java/io/airlift/stats/labeled/LabeledStatRegistry.java
grantatspothero: There is a single jmx registry per JVM and a separate MBeanExporter per injector/classloader. 

The LabeledStatRegistry takes an MBeanExporter (and so has the same lifecycle as the MBeanExporter)

Changing the docs to reflect this.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
grantatspothero: kept the existing stream implementation,  agree will change to list.

stats/src/main/java/io/airlift/stats/labeled/AbstractLabeledStat.java
grantatspothero: There are two types of destroy operations to consider:
1. Removing label sets, but not the stat object itself, for the purpose of reducing cardinality. 
2. Removing stat objects themselves. 

For (1):
Did not implement label set removal. OTEL suggests do not label a stat with a high cardinality attribute as you will leak memory and cause problems with your backend TSDB, so never remove label sets from memory. If we want to implement label set removal, the naive approach the [prometheus java client takes is manual expiration](https://prometheus.github.io/client_java/getting-started/labels?#expiring-unused-label-values) but it needs to be used carefully. In practice labelset removal is rarely used. 

For (2):
When MBeanExporter is destroyed stats are removed from JMX. How jmxutils/airlift does this today is registering a separate MBeanExporter per lifecycle. MBeanExporter has @PreDestroy method which unexports its associated JMX beans on shutdown. The intention is the same, create a LabeledStatRegistry per MBeanExporter. This works well for coarse grained unregistration, eg) each trino catalog gets a separate MBeanExporter, when the catalog is removed the MBeanExporter and its associated metrics are removed. More fine grained unregistration (not at the MBeanExporter level) is not supported currently. I believe more fine grained unregistration is not useful, you should never be dynamically generating metrics now that labelsets are supported.

stats/src/main/java/io/airlift/stats/labeled/LabeledStatRegistry.java
grantatspothero: We need to run `mBeanExporter.export()` per dynamic labelset , which means we need access to an MBeanExporter at runtime. 

Put another way, `@Managed` only works for static metrics not dynamic metrics, and labeled stats are dynamic metrics.

stats/src/main/java/io/airlift/stats/labeled/AbstractLabeledStat.java
losipiuk: I meant 1. How do we ensure that we would not create too many labelset which eventually would consume significant amount of memory on the Trino side. I know we should not label with high cardinality - but we should somewhat enforce we are not doing that (IIRC there is code in this PR which logs a warning right?, but it would be nicer to have something easier to notice).

stats/src/main/java/io/airlift/stats/labeled/LabeledStatRegistry.java
losipiuk: I do not see a reason why `@Managed` could not be used for dynamic metrics too. 
It just would behave differently. It would probably only bind dependencies needed by object representing stat (including MBeanExporter). 
And actual exporting would happen when you update stat value for specific label set.

stats/src/main/java/io/airlift/stats/labeled/LabeledCounterStat.java
losipiuk: Not for this PR - but should we drop decaying stuff from non-labeled metrics too?

grantatspothero: Yes, but it is a backwards incompatible change. 

Probably still need a flag to allow exporting the time decayed counters in case users have alerts/dashboards defined over the time decayed metrics.

stats/src/main/java/io/airlift/stats/labeled/LabeledStatRegistry.java
grantatspothero: Changing the behavior of `@Managed` requires changing how jmxutils works not just airlift. It is a bigger change. Would be good to get feedback from @martint

stats/src/main/java/io/airlift/stats/labeled/AbstractLabeledStat.java
grantatspothero: Examples from other systems:

Prometheus java client: does not have a cardinality limit by default, if you choose a bad label with high cardinality the client will consume lots of memory. 

Opentelemetry: [default of 2000 cardinality limit per metric](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#cardinality-limits), emits a metric `otel.metric.overflow` if the cardinality limit is reached. Additionally, prevents creating further time series if the limit has been reached. 

If we want to be safer, could both keep the log and prevent further time series from being created once the cardinality limit has been reached. This prevents high memory usage due to invalid label choices.

PR #1790, opened 2026-02-23
Add support for oneOf to OpenAPI output

electrum:


api/src/main/java/io/airlift/api/openapi/SchemaBuilder.java
```diff
@@ -361,6 +361,8 @@ private void buildPolySchema(Schema<?> schema, ModelPolyResource polyResource, B
             subSchema.addAllOfItem(mainSchemaAsRef);
             subSchema.addAllOfItem(allOfSchema);
 
+            schema.addOneOfItem(asRef(subSchema));
```
Randgalt, line 361: I need to test this - please wait to merge.

Randgalt: OK - I tested. This will break Galaxy and needs to be optional. Let me see if I can do something.

Image

Randgalt: @electrum what's the use-case for this?

Randgalt: Adding this fixes the problem:

```
            if (mode.isAllOf()) {
                // we don't save allOf/of schemas. They get merged with the parent.
                // However, keep track of them in case there are recursive references to them
                // that need to be added prior to building.
                allOfSchemas.put(subSchema.getName(), subSchema);
            }
```

We might rename `allOfSchemas` to `ofSchemas` or something. The issue is that these recursive/poly references sometimes aren't referenced anywhere else and, thus, the real model never gets defined.  Adding to `allOfSchemas` ensures that the model gets generated in the components/schemas section of OpenApi.json

electrum: Can you provide a test case for this? I added it to `buildPolySchema` but it doesn't change the OpenAPI output for the current "widget" schema.

electrum: @huberty89 found that the current change is insufficient to fix the TypeScript bindings. I'm working with him to resolve it.

Randgalt: You need to have deep recursive/poly references. Let me see.

electrum: I updated the branch with what I think you wanted -- does that look right?

Randgalt: Yes - that's correct

Randgalt: > Can you provide a test case for this? I added it to `buildPolySchema` but it doesn't change the OpenAPI output for the current "widget" schema.

https://github.com/airlift/airlift/issues/1792

electrum:
I actually don't know how this supposed be work. Looking at the diff for `widget.json`, the change seems minimal.

My process was to ask Codex to make the change and to verify that the generated TypeScript is valid. @huberty89 confirmed that the TypeScript is as expected -- it matches what we get today with the OpenAPI definition produced by Swagger annotations.

Do you have a test case for the failure you're seeing?

huberty89:
I prepared small example, with what we use today:

```java
import com.fasterxml.jackson.annotation.JsonSubTypes;
import com.fasterxml.jackson.annotation.JsonTypeInfo;
import io.swagger.v3.oas.annotations.media.DiscriminatorMapping;
import io.swagger.v3.oas.annotations.media.Schema;
import io.swagger.v3.oas.annotations.responses.ApiResponse;
import jakarta.ws.rs.GET;
import jakarta.ws.rs.Path;
import jakarta.ws.rs.Produces;

import static jakarta.ws.rs.core.MediaType.APPLICATION_JSON;

@Path("/api/v1/foo")
@Produces(APPLICATION_JSON)
public class FooResource
{
    @GET
    @Produces(APPLICATION_JSON)
    @ApiResponse(responseCode = "200", useReturnTypeSchema = true)
    public FooResponse readFoo()
    {
        return null;
    }

    @JsonTypeInfo(
            use = JsonTypeInfo.Id.NAME,
            include = JsonTypeInfo.As.PROPERTY,
            property = "responseKind")
    @JsonSubTypes({
            @JsonSubTypes.Type(value = SuccessResponse.class, name = "SUCCESS"),
            @JsonSubTypes.Type(value = FailResponse.class, name = "FAIL"),
    })
    @Schema(implementation = OpenApiAnnotation.class)
    sealed interface FooResponse
            permits FailResponse,
                    SuccessResponse
    {}

    record SuccessResponse(String data)
            implements FooResponse
    {}

    record FailResponse(String reason, int code)
            implements FooResponse
    {}

    @Schema(
            name = "FooResponse",
            oneOf = {SuccessResponse.class, FailResponse.class},
            discriminatorProperty = "responseKind",
            discriminatorMapping = {
                    @DiscriminatorMapping(value = "SUCCESS", schema = SuccessResponse.class),
                    @DiscriminatorMapping(value = "FAIL", schema = FailResponse.class)
            })
    interface OpenApiAnnotation {}
}
```

Generated schema:

```json
{
  "openapi" : "3.1.0",
  "paths" : {
    "/api/v1/foo" : {
      "get" : {
        "operationId" : "readFoo",
        "responses" : {
          "200" : {
            "content" : {
              "application/json" : {
                "schema" : {
                  "$ref" : "#/components/schemas/FooResponse"
                }
              }
            },
            "description" : "default response"
          }
        }
      }
    }
  },
  "components" : {
    "schemas" : {
      "FailResponse" : {
        "type" : "object",
        "properties" : {
          "code" : {
            "type" : "integer",
            "format" : "int32"
          },
          "reason" : {
            "type" : "string"
          }
        }
      },
      "FooResponse" : {
        "discriminator" : {
          "mapping" : {
            "FAIL" : "#/components/schemas/FailResponse",
            "SUCCESS" : "#/components/schemas/SuccessResponse"
          },
          "propertyName" : "responseKind"
        },
        "oneOf" : [ {
          "$ref" : "#/components/schemas/SuccessResponse"
        }, {
          "$ref" : "#/components/schemas/FailResponse"
        } ]
      },
      "SuccessResponse" : {
        "type" : "object",
        "properties" : {
          "data" : {
            "type" : "string"
          }
        }
      }
    }
  }
}
```

PR #1828, opened 2026-03-12
Disable trace propagation if disabled & tracing improvements

wendigo:

### DIFF
--- a/http-client/pom.xml
+++ b/http-client/pom.xml
@@ -77,6 +77,11 @@
 
         <dependency>
             <groupId>io.airlift</groupId>
+            <artifactId>tracing</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.airlift</groupId>
             <artifactId>units</artifactId>
         </dependency>
 
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
@@ -28,6 +28,7 @@ import com.google.inject.TypeLiteral;
 import io.airlift.configuration.ConfigDefaults;
 import io.airlift.http.client.jetty.JettyHttpClient;
 import io.airlift.node.NodeInfo;
+import io.airlift.tracing.TracingEnabledConfig;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.api.trace.TracerProvider;
@@ -68,6 +69,7 @@ public class HttpClientModule
 
         // bind the configuration
         configBinder(binder).bindConfig(HttpClientConfig.class, annotation, name);
+        configBinder(binder).bindConfig(TracingEnabledConfig.class);
 
         // Allow users to bind their own SslContextFactory, pulling a globally-bound
         // SslContextFactory if one is not bound with the specific annotation
@@ -136,6 +138,7 @@ public class HttpClientModule
         public HttpClient get()
         {
             HttpClientConfig config = injector.getInstance(Key.get(HttpClientConfig.class, annotation));
+            TracingEnabledConfig tracingEnabledConfig = injector.getInstance(TracingEnabledConfig.class);
             Optional<String> environment = Optional.ofNullable(nodeInfo).map(NodeInfo::getEnvironment);
             Optional<SslContextFactory.Client> sslContextFactoryAnnotated = injector.getInstance(Key.get(new TypeLiteral<>() {}, annotation));
             Optional<SslContextFactory.Client> sslContextFactoryGlobal = injector.getInstance(Key.get(new TypeLiteral<>() {}));
@@ -151,7 +154,7 @@ public class HttpClientModule
                     .addAll(injector.getInstance(Key.get(new TypeLiteral<Set<HttpStatusListener>>() {}, annotation)))
                     .build();
 
-            return new JettyHttpClient(name, config, ImmutableList.copyOf(filters), openTelemetry, tracer, environment, sslContextFactory, httpStatusListeners);
+            return new JettyHttpClient(name, config, tracingEnabledConfig, ImmutableList.copyOf(filters), openTelemetry, tracer, environment, sslContextFactory, httpStatusListeners);
         }
     }
 }
--- a/http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
@@ -31,7 +31,6 @@ import io.airlift.node.NodeInfo;
 import io.airlift.tracing.TracingEnabledConfig;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.trace.Tracer;
-import io.opentelemetry.api.trace.TracerProvider;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 
 import java.lang.annotation.Annotation;
@@ -101,8 +100,8 @@ public class HttpClientModule
         private final Class<? extends Annotation> annotation;
         private Injector injector;
         private NodeInfo nodeInfo;
-        private OpenTelemetry openTelemetry = OpenTelemetry.noop();
-        private Tracer tracer = TracerProvider.noop().get("noop");
+        private Optional<OpenTelemetry> openTelemetry = Optional.empty();
+        private Optional<Tracer> tracer = Optional.empty();
 
         private HttpClientProvider(String name, Class<? extends Annotation> annotation)
         {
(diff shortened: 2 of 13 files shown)


http-server/pom.xml
```diff
@@ -71,6 +71,10 @@
             <groupId>io.airlift</groupId>
             <artifactId>security</artifactId>
         </dependency>
+        <dependency>
```
findepi, line 75:
```suggestion

        <dependency>
```

http-server/src/main/java/io/airlift/http/server/tracing/TracingServletFilter.java
```diff
@@ -68,6 +71,10 @@ public static void updateRequestSpan(HttpServletRequest request, Consumer<Span>
     public void doFilter(HttpServletRequest request, HttpServletResponse response, FilterChain chain)
             throws IOException, ServletException
     {
+        if (!enabled) {
+            chain.doFilter(request, response);
+            return;
```
findepi: Would it be better if the "is enabled?" condition was evaluated once?
A disabled filter could perhaps not be part of the filter chain at all.

https://github.com/airlift/airlift/blob/ed7a4761333af20979e89618663bdea55e0b30ab/http-server/src/main/java/io/airlift/http/server/HttpServerModule.java#L100-L102

http-server/src/main/java/io/airlift/http/server/HttpServerModule.java
```diff
@@ -103,6 +104,7 @@ protected void setup(Binder binder)
         newSetBinder(binder, qualifiedKey(HttpResourceBinding.class));
         newOptionalBinder(binder, qualifiedKey(SslContextFactory.Server.class));
         configBinder(binder).bindConfig(qualifiedKey(HttpServerConfig.class), HttpServerConfig.class, configPrefix);
+        configBinder(binder).bindConfig(TracingEnabledConfig.class);
```
findepi: Alternatively, the `TracingServletFilter` could be bound by the `TracingModule`?

cc @electrum

tracing/pom.xml
```diff
@@ -71,6 +71,12 @@
             <artifactId>okhttp-jvm</artifactId>
             <version>5.3.2</version>
             <scope>runtime</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.jetbrains</groupId>
+                    <artifactId>annotations</artifactId>
```
findepi, line 78: Is this avoidable? worth documenting?

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -959,14 +958,13 @@ private Request applyRequestFilters(Request request)
 
     private Span startSpan(Request request)
     {
-        String method = request.getMethod().toUpperCase(ENGLISH);
```
findepi: > It is already uppercased in the builder so we can avoid doing it once again

I don't see this in `io.airlift.http.client.Request.Builder#setMethod`
nor enforcement in `io.airlift.http.client.Request#Request`

what am i missing?

wendigo: https://github.com/airlift/airlift/pull/1828/changes/4dfb4f216aed0b86483716d8f6a9223404890e4f#diff-3acedd36258c75fca5179c7df7706f66da4b8a6043b0ac6c5b25ab14e61ff4adR271

http-server/src/main/java/io/airlift/http/server/HttpServerModule.java
wendigo: TracingModule doesn't depend on the http-server. I've made a binding conditional

http-server/pom.xml
wendigo: Done

tracing/pom.xml
wendigo: Documented. Can't really be avoided - this promotes annotations to runtime wherever client/server are used - this required a lot of changes in Trino. This makes it simpler to use this change.

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
wendigo: I've improved commit message

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
```diff
@@ -68,6 +68,7 @@ public void configure(Binder binder)
 
         // bind the configuration
         configBinder(binder).bindConfig(HttpClientConfig.class, annotation, name);
+        configBinder(binder).bindConfig(TracingEnabledConfig.class);
```
findepi, line 71: can we use use optional bindings here to avoid binding the config from a different module than it's normally used.

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
```diff
@@ -120,18 +119,6 @@ public void setNodeInfo(NodeInfo nodeInfo)
...
         }
 
-        @Inject(optional = true)
-        public void setOpenTelemetry(OpenTelemetry openTelemetry)
-        {
-            this.openTelemetry = openTelemetry;
-        }
-
-        @Inject(optional = true)
-        public void setTracer(Tracer tracer)
```
findepi, line 130: this is how we made the dependency optional?
or the `Tracer` is almost always bound, but not necessarily to a useful object (no-op)?

tracing/pom.xml
electrum: This is really a bug in `okhttp-jvm`. It should have declared the dependency as optional (non-transitive).

http-server/pom.xml
electrum: This commit wasn't updated with a blank line

http-client/pom.xml
```diff
@@ -70,6 +70,11 @@
             <artifactId>stats</artifactId>
         </dependency>
 
+        <dependency>
```
electrum, line 78: This drags in many dependencies just to avoid adding a copy of the trivial `TracingEnabledConfig`. We should simply have a copy of it here. This is a big ugly as it's part of the API, but consumers don't use the complex constructors directly, anyway.

http-client/src/main/java/io/airlift/http/client/Request.java
```diff
@@ -333,6 +347,13 @@ public Builder setSpanBuilder(SpanBuilder spanBuilder)
             return this;
         }
 
+        @CanIgnoreReturnValue
+        public Builder setCurrentSpan(Span currentSpan)
```
electrum: Make this package-private as it's internal to the client

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
```diff
@@ -149,12 +133,21 @@ public HttpClient get()
                     .addAll(injector.getInstance(Key.get(new TypeLiteral<Set<HttpRequestFilter>>() {}, annotation)))
                     .build();
 
+            if (injector.getInstance(TracingEnabledConfig.class).isEnabled()) {
+                ImmutableSet<HttpRequestFilter> filtersWithTracing = ImmutableSet.<HttpRequestFilter>builderWithExpectedSize(filters.size() + 1)
+                        .addAll(filters)
+                        .add(new TracingRequestFilter(name, injector.getInstance(OpenTelemetry.class), injector.getInstance(Tracer.class)))
+                        .build();
+
+                filters = filtersWithTracing;
```
electrum: Inline `filtersWithTracing`?

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
electrum: This optional dependency is still needed. Otherwise, if `TracingModule` is not installed but `tracing.enabled` is set at runtime, the `injector.getInstance()` calls below will fail.

http-client/pom.xml
findepi: if we copy, we should ensure they remain copies -- https://github.com/airlift/airlift/pull/1811#discussion_r2907803560

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
wendigo: This should fail if the config was swallowed but you are missing deps to construct

http-client/pom.xml
wendigo: @findepi how can I ensure that this is a copy without introducing a dependency on a tracing module?

findepi: test classpath?

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
electrum: The problem is that it will only fail conditionally, based on the user's config.

electrum (changes requested):
We can't have a silent failure based on config if the tracing module isn't installed.

http-client/pom.xml
electrum: This is a trivial class. It's fine to have copies.

http-client/src/main/java/io/airlift/http/client/HttpClientModule.java
wendigo: Ok so what is the expected behavior? Shoud I at least generate a warning that dependency is missing when tracing is enabled?

wendigo:
Replaced by https://github.com/airlift/airlift/pull/1976, https://github.com/airlift/airlift/pull/1979 and https://github.com/airlift/airlift/pull/1981

PR #1831, opened 2026-03-12
Enable FAIL_ON_TRAILING_TOKENS when deserializing

wendigo:

### DIFF
--- a/json/src/main/java/io/airlift/json/JsonCodec.java
+++ b/json/src/main/java/io/airlift/json/JsonCodec.java
@@ -15,10 +15,10 @@
  */
 package io.airlift.json;
 
-import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.ObjectWriter;
+import com.fasterxml.jackson.databind.exc.MismatchedInputException;
 import com.fasterxml.jackson.databind.json.JsonMapper;
 import com.google.common.base.Suppliers;
 import com.google.common.reflect.TypeParameter;
@@ -37,7 +37,6 @@ import java.util.Optional;
 import java.util.function.Supplier;
 
 import static com.fasterxml.jackson.databind.SerializationFeature.INDENT_OUTPUT;
-import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 
 @ThreadSafe
@@ -142,11 +141,11 @@ public class JsonCodec<T>
     public T fromJson(String json)
             throws IllegalArgumentException
     {
-        ObjectReader objectReader = reader.get();
-        try (JsonParser parser = objectReader.createParser(json)) {
-            T value = objectReader.readValue(parser);
-            checkArgument(parser.nextToken() == null, "Found characters after the expected end of input");
-            return value;
+        try {
+            return reader.get().readValue(json);
+        }
+        catch (MismatchedInputException e) {
+            throw mapMismatchedInputException(e, "string", type);
         }
         catch (IOException e) {
             throw new IllegalArgumentException("Invalid JSON string for %s".formatted(type), e);
@@ -204,11 +203,11 @@ public class JsonCodec<T>
     public T fromJson(byte[] json)
             throws IllegalArgumentException
     {
-        ObjectReader objectReader = reader.get();
-        try (JsonParser parser = objectReader.createParser(json)) {
-            T value = objectReader.readValue(parser);
-            checkArgument(parser.nextToken() == null, "Found characters after the expected end of input");
-            return value;
+        try {
+            return reader.get().readValue(json);
+        }
+        catch (MismatchedInputException e) {
+            throw mapMismatchedInputException(e, "bytes", type);
         }
         catch (IOException e) {
             throw new IllegalArgumentException("Invalid JSON bytes for " + type, e);
@@ -243,11 +242,11 @@ public class JsonCodec<T>
     public T fromJson(InputStream json)
             throws IllegalArgumentException
     {
-        ObjectReader objectReader = reader.get();
-        try (JsonParser parser = objectReader.createParser(json)) {
-            T value = objectReader.readValue(parser);
-            checkArgument(parser.nextToken() == null, "Found characters after the expected end of input");
-            return value;
+        try {
+            return reader.get().readValue(json);
+        }
+        catch (MismatchedInputException e) {
+            throw mapMismatchedInputException(e, "stream", type);
         }
         catch (IOException e) {
             throw new IllegalArgumentException("Invalid JSON bytes for " + type, e);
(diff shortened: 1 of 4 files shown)


http-server/pom.xml
```diff
@@ -200,6 +200,18 @@
...
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <scope>runtime</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.airlift</groupId>
+            <artifactId>json</artifactId>
+            <scope>runtime</scope>
```
findepi: Why runtime? document.

runtime scope is not checked by toolings, so it needs to be maintained by humans.

cc @electrum for new dep

api/src/test/java/io/airlift/api/servertests/ServerTestBase.java
```diff
@@ -92,7 +93,11 @@ private ServerTestBase(Optional<Class<?>> maybeServiceClass, Consumer<ApiModule.
                 .host("localhost")
                 .build();
 
-        jsonMapper = injector.getInstance(JsonMapper.class);
+        jsonMapper = injector.getInstance(JsonMapper.class)
+                .rebuild()
+                // https://github.com/OpenAPITools/openapi-generator/issues/23146
+                .disable(DeserializationFeature.FAIL_ON_TRAILING_TOKENS)
+                .build();
```
findepi: cc @Randgalt

api/src/test/java/io/airlift/api/servertests/ServerTestBase.java
```diff
@@ -92,7 +93,11 @@ private ServerTestBase(Optional<Class<?>> maybeServiceClass, Consumer<ApiModule.
                 .host("localhost")
                 .build();
 
-        jsonMapper = injector.getInstance(JsonMapper.class);
+        jsonMapper = injector.getInstance(JsonMapper.class)
+                .rebuild()
+                // https://github.com/OpenAPITools/openapi-generator/issues/23146
```
findepi: this should be a TODO note

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -263,17 +262,25 @@ public T fromJson(InputStream json)
...
         }
         catch (IOException e) {
             throw new IllegalArgumentException(format("Invalid JSON characters for %s", type), e);
         }
     }
 
+    private static IllegalArgumentException mapMismatchedInputException(MismatchedInputException e, String source, Type type)
+    {
+        if (e.getMessage() != null && e.getMessage().contains("TRAILING_TOKEN")) {
+            return new IllegalArgumentException("Found characters after the expected end of input");
```
findepi: Include `e` as cause, especially that we're matching e by its message. For example TRAILING_TOKEN could be part of input data

findepi (approved):

http-server/pom.xml
electrum: These should both be test scoped since they are only used in the tests

http-server/src/test/java/io/airlift/http/server/TestHttpServerModule.java
```diff
@@ -111,7 +113,10 @@ public void testCanConstructSingleServer()
                 new HttpServerModule(),
                 new TestingNodeModule(),
                 new TracingModule("airlift.http-server", "1.0"),
-                binder -> binder.bind(Servlet.class).to(DummyServlet.class));
+                binder -> {
+                    binder.bind(Servlet.class).to(DummyServlet.class);
+                    binder.bind(JsonMapper.class).toProvider(new JsonMapperProvider());
```
electrum: Why do we need to add this? It doesn't seem to be used

json/src/main/java/io/airlift/json/JsonSubType.java
```diff
@@ -264,7 +264,7 @@ public Deserializer(String propertyName, Supplier<Map<String, Class<?>>> subClas
         public B deserialize(JsonParser p, DeserializationContext ctxt)
                 throws IOException
         {
-            TreeNode treeNode = p.readValueAsTree();
+            TreeNode treeNode = ctxt.readTree(p);
```
electrum: Add a test that fails without this change

tracing/src/test/java/io/airlift/tracing/TestTracingModule.java
```diff
@@ -15,7 +17,8 @@ void testNoopTracing()
     {
         Injector injector = new Bootstrap(
                 new TestingNodeModule(),
-                new TracingModule("testService", "testVersion"))
+                new TracingModule("testService", "testVersion"),
+                binder -> binder.bind(JsonMapper.class).toProvider(new JsonMapperProvider()))
```
electrum: Why not use `JsonModule`

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -263,17 +262,25 @@ public T fromJson(InputStream json)
...
+            throw mapMismatchedInputException(e, "reader", type);
         }
         catch (IOException e) {
             throw new IllegalArgumentException(format("Invalid JSON characters for %s", type), e);
         }
     }
 
+    private static IllegalArgumentException mapMismatchedInputException(MismatchedInputException e, String source, Type type)
+    {
+        if (e.getMessage() != null && e.getMessage().contains("TRAILING_TOKEN")) {
```
electrum: Is there really no better way? Can we use a longer substring match?

json/src/main/java/io/airlift/json/JsonCodec.java
```diff
@@ -265,19 +256,19 @@ public T fromJson(Reader json)
...
+        catch (Exception e) {
+            throw mapException(e, "characters", type);
         }
     }
 
-    private static IllegalArgumentException mapMismatchedInputException(MismatchedInputException e, String source, Type type)
+    private static IllegalArgumentException mapException(Exception e, String source, Type type)
     {
-        if (e.getMessage() != null && e.getMessage().contains("TRAILING_TOKEN")) {
+        if (e instanceof MismatchedInputException mismatchedInputException && mismatchedInputException.getMessage() != null && mismatchedInputException.getMessage().contains("TRAILING_TOKEN")) {
```
electrum: I think this could be a switch expression with pattern match and guard

http-server/src/test/java/io/airlift/http/server/TestHttpServerModule.java
wendigo: It is used by the SpanDeserializer

http-server/pom.xml
wendigo: It is used transitively (tracing uses json to get JsonMapper to construct a map type)

electrum: Test scope is correct now, as you should know as you worked to get that fixed.

wendigo:
(rebased)

Issue #1858, opened 2026-03-23
[MCP] De-prioritize sessions and other tasks


Randgalt:
## Background

Sessions in MCP are going to change in the next version of the protocol. Further, it's likely that sessions will be removed altogether. Many of the features in the protocol are being refactored/revised to not require sessions.

Due to all of this, we should de-prioritize sessions in the Airlift implementation.

## Details

The MCP Protocol is about to go through major changes. These changes are targeted for the June 2026 release of the protocol spec. For the purposes of Airlift, the important changes are:

- De-prioritization of sessions (making them truly optional)
  - This affects many aspects of the current protocol: logging, initialization, elicitations, etc.
  - There is an active debate whether to completely remove sessions altogether or to merely make them completely optional.
- Important proposals to watch:
  - [Multi-Round Trip Requests](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2322)
  - [Remove sessions](https://github.com/modelcontextprotocol/transports-wg/pull/25)
  - [Explicit session management](https://github.com/modelcontextprotocol/transports-wg/pull/20)
  - [Transports working group meeting on sessions](https://github.com/modelcontextprotocol/transports-wg/blob/1645872ed794aa152acf1fcb179dcddb957a72eb/meetings/2026-03-18_weekly.md) - I'm an influencer!
  - [Lots more changes](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1442)

Due to all of this, the series of tasks below will de-prioritization sessions in the Airlift MCP implementation. However, all features of MCP will still be supported and for all clients backward and forward. The details, however, of how features are implemented will be determined internally by Airlift. For example:

- if Airlift is _not_ configured with sessions, many features will be disabled. 
- for clients that declare an older protocol, Airlift will emulate features that require newer protocols where possible. For example, MRTR and Tasks can be emulated.

Importantly: these changes may require some breaking changes to Airlift MCP client implementations.

Additionally, let's list miscellaneous changes/updates that need to be done.

## Work

- [x] https://github.com/airlift/airlift/pull/1859
- [x] https://github.com/airlift/airlift/pull/1860
- [x] https://github.com/airlift/airlift/pull/1861
- [ ] https://github.com/airlift/airlift/pull/1869
- [ ] Implement tasks without the use of sessions
- [ ] Update to new protocol when available
- [ ] https://modelcontextprotocol.io/seps/414-request-meta

Randgalt:
[...]

PR #1869, opened 2026-03-25
[MCP] WIP - support the mutli-roundtrip requests proposal

Randgalt:
### Remove server-to-client features in favor of MRTR

Standardize on the upcoming MRTR spec changes (preliminary). For
clients that use an older protocol, emulate MRTR.

### Models, etc. to support MRTR

Based on the preliminary spec. Likely to change in the future.

### Make CallToolResult a sealed interface

- In upcoming protocols, tool calls return a union model of many different types
- Make CallToolResult a sealed interface so that it can model any needed type
- Rename old model for `CallToolResult` to `ToolContent`

----

Part of https://github.com/airlift/airlift/issues/1858

### DIFF
--- a/mcp/src/main/java/io/airlift/mcp/McpEntities.java
+++ b/mcp/src/main/java/io/airlift/mcp/McpEntities.java
@@ -46,7 +46,7 @@ public interface McpEntities
 
     Optional<CompletionEntry> completionEntry(McpRequestContext requestContext, CompleteReference ref);
 
-    Optional<ReadResourceResult> readResourceContents(McpRequestContext requestContext, ReadResourceRequest readResourceRequest);
+    Optional<ReadResourceResult> readResourceContents(McpRequestContext requestContext, ReadResourceRequest readResourceRequest, boolean allowIncompleteResult);
 
     void addTool(Tool tool, ToolHandler toolHandler);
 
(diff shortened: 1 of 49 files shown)


wendigo:
Can you explain how MRTR replaces sessions?

Randgalt:
> Can you explain how MRTR replaces sessions?

It doesn't entirely. It removes the need for sessions for multiple round-trip request (MRTR) types. Currently there are three of these: elicitations, sampling and roots. In previous versions of the protocol, the tool would "return" one of these requests in the SSE stream and block until the response is received. The response would arrive as a complete separate HTTP POST. In a horizontally scaled topology that means it would likely go to a different server. So, sessions are required to correlate the response POST with the tool that is blocking waiting for it.

In MRTR, the request is returned as the result of the tool call and the tool ends. The client then calls the tool again with the response now embedded in the tool call's parameters. In the MRTR spec these are called InputRequests and InputResponses. So, the tool returns InputRequests when it needs it and when it sees a tool call arrive without any InputResponses.

wendigo:
@Randgalt so that's basically a cookie, am I wrong? ;-)

Randgalt:
The requirements have changed a lot. I will reimplement.

PR #1881, opened 2026-03-31
Support labeled stats

grantatspothero:
Smaller PR broken out from here: https://github.com/airlift/airlift/pull/1786

### DIFF
--- a/openmetrics/src/main/java/io/airlift/openmetrics/LabeledStatConfig.java
+++ b/openmetrics/src/main/java/io/airlift/openmetrics/LabeledStatConfig.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.airlift.openmetrics;
+
+import io.airlift.configuration.Config;
+import io.airlift.configuration.ConfigDescription;
+import jakarta.validation.constraints.Max;
+import jakarta.validation.constraints.Min;
+
+public class LabeledStatConfig
+{
+    private int labeledStatMaxCardinality = 10_000;
+
+    @Min(1)
+    @Max(1_000_000)
+    public int getLabeledStatMaxCardinality()
+    {
+        return labeledStatMaxCardinality;
+    }
+
+    @Config("labeled-stat.max-cardinality")
+    @ConfigDescription("After max cardinality for a single labeled stat is reached, emits warn logs and stops creating new labeled stat values for that labeled stat")
+    public LabeledStatConfig setLabeledStatMaxCardinality(int maxCardinality)
+    {
+        this.labeledStatMaxCardinality = maxCardinality;
+        return this;
+    }
+}
--- a/openmetrics/pom.xml
+++ b/openmetrics/pom.xml
@@ -52,6 +52,11 @@
         </dependency>
 
         <dependency>
+            <groupId>jakarta.validation</groupId>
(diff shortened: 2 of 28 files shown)


stats/src/main/java/io/airlift/stats/labeled/LabeledStatConfig.java
```diff
@@ -0,0 +1,39 @@
...
+    private int labeledStatMaxCardinality = 10_000;
+
+    @Min(0)
+    @Max(1_000_000)
+    public int getLabeledStatMaxCardinality()
+    {
+        return labeledStatMaxCardinality;
+    }
+
+    @Config("openmetrics.labeled-stat-max-cardinality")
```
wendigo: I don't follow. Why this belongs to module stats while config points to `openmetrics`? This is mixed up.

grantatspothero: Moved around the config and forgot to update config name, fixed.

openmetrics/src/main/java/io/airlift/openmetrics/LabeledStatConfig.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Min(0)
+    @Max(1_000_000)
+    public int getLabeledStatMaxCardinality()
+    {
+        return labeledStatMaxCardinality;
+    }
+
+    @Config("labeled-stat.max-cardinality")
+    @ConfigDescription("After max cardinality for a single labeled stat is reached, emits warn logs and stops creating new labeled stat values for that labeled stat")
```
grantatspothero, line 33: cc: @losipiuk from discussion in previous comment: https://github.com/airlift/airlift/pull/1786#discussion_r3011777185

Now we prevent high cardinality stats from creating more labeled stat values, which prevents a single bad stat from consuming lots of memory.

losipiuk: Nice

stats/src/main/java/io/airlift/stats/labeled/GlobalLabeledStatRegistry.java
```diff
@@ -0,0 +1,46 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+public class GlobalLabeledStatRegistry
+        implements LabeledStatRegistry
+{
+    private final MBeanExporter exporter;
+    private final int labeledStatMaxCardinality;
+
+    @Inject
```
wendigo: `stats` module is a library, so it shouldn't depends on guice and have configs. This doesn't belong here - I'd move labeled metrics somewhere else (openmetrics)

grantatspothero: Removed guice and required manual injection of constructor params. Still kept airlift config in stats module, does it make sense to also not depend on airlift config?

I also moved `LabeledStatModule` to openmetrics module (even though it is technically not specific to openmetrics, as can use labeled stats with just JMX). Helps remove some manual wiring in the common case, but doesn't have to be used.

wendigo: stats so far was a poor "utility" module. It didn't require any configuration or wiring

sample-server/src/main/java/io/airlift/sample/Main.java
```diff
@@ -51,6 +52,7 @@ public static void main(String[] args)
                 new MBeanModule(),
                 new JmxModule(),
                 new JmxHttpModule(),
+                new LabeledStatModule(),
```
wendigo: Why is it needed? Can it be optional?

grantatspothero: Yes it is not required here, wanted to show sample usage. I can remove.

stats/src/main/java/io/airlift/stats/labeled/GlobalLabeledStatRegistry.java
grantatspothero: I moved the airlift config into openmetrics, this removed the airlift config dependency from stats. 

This is not ideal because the config is not related to openmetrics, its related to airlift labeled stats. But I don't see any other option if we do not want the airlift config dependency in stats.

wendigo:
@grantatspothero can you rebase this?

PR #1900, opened 2026-04-27
[MCP] Improve JSON-schema generation

Randgalt:

Use the Victools lib to do generation instead of hand-coding. However,
we will shade it into our jar to avoid unneeded deps/conflicts.

### DIFF
--- a/mcp/pom.xml
+++ b/mcp/pom.xml
@@ -18,9 +18,15 @@
         <dep.docker-java.version>3.7.1</dep.docker-java.version>
         <dep.hikari.version>7.0.2</dep.hikari.version>
         <dep.postgresql.version>42.7.11</dep.postgresql.version>
+        <dep.jsonschema-generator.version>4.38.0</dep.jsonschema-generator.version>
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>com.fasterxml</groupId>
+            <artifactId>classmate</artifactId>
+        </dependency>
+
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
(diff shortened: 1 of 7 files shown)


mcp/pom.xml
```diff
@@ -250,7 +257,8 @@
                             <createDependencyReducedPom>false</createDependencyReducedPom>
                             <artifactSet>
                                 <includes>
-                                    <include>io.modelcontextprotocol.sdk:*</include>
+                                    <include>com.github.victools:*</include>
+                                    <include>com.fasterxml:classmate</include>
```
electrum: Why shade this library? It seems fine to have a direct dependency

mcp/pom.xml
```diff
@@ -31,6 +39,18 @@
             <artifactId>jackson-databind</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>com.github.victools</groupId>
+            <artifactId>jsonschema-generator</artifactId>
+            <version>${dep.jsonschema-generator.version}</version>
+            <exclusions>
+                <exclusion>
```
electrum: This dependency seems to be needed, so rather than excluding it, add it to `dependencyManagement` with the later version.

mcp/pom.xml
```diff
@@ -18,9 +18,17 @@
         <dep.docker-java.version>3.7.1</dep.docker-java.version>
         <dep.hikari.version>7.0.2</dep.hikari.version>
         <dep.postgresql.version>42.7.10</dep.postgresql.version>
+        <dep.jsonschema-generator.version>4.38.0</dep.jsonschema-generator.version>
```
electrum, line 21: Only add version properties if the version is needed in multiple places

electrum:
@coderabbitai full review

mcp/pom.xml
Randgalt: The `com.fasterxml:classmate` is the one I'm worried about. Any Jackson dep like this always causes conflicts. I figured that if I shade this I'd shade the other one. But, I can remove the shading if you like.

electrum: I've found that shading usually causes more issues, because you can't easily upgrade the dependency in downstream projects when you need to for security issues, etc. If it becomes a problem, we can pull it into airbase so the version is managed consistently. Also, now with dependabot, we keep everything up to date automatically.

electrum (approved):

PR #1918, opened 2026-05-11
[API] Allow Object element type via @ApiAllowObject

huberty89:
Add a TYPE_USE `@ApiAllowObject` annotation so record components of type `List<@ApiAllowObject Object>` or `Map<String, @ApiAllowObject Object>` are accepted by the API builder and exposed as free-form `type: object` elements in the generated OpenAPI schema. Unannotated `Object` fields remain rejected as before.

### DIFF
--- a/api/src/main/java/io/airlift/api/builders/ResourceBuilder.java
+++ b/api/src/main/java/io/airlift/api/builders/ResourceBuilder.java
@@ -46,6 +46,7 @@ import static io.airlift.api.internals.Generics.typeResolver;
 import static io.airlift.api.internals.Mappers.openApiName;
 import static io.airlift.api.model.ModelResourceModifier.HAS_RESOURCE_ID;
 import static io.airlift.api.model.ModelResourceModifier.HAS_VERSION;
+import static io.airlift.api.model.ModelResourceModifier.IS_ANY_OBJECT;
 import static io.airlift.api.model.ModelResourceModifier.IS_MULTIPART_FORM;
 import static io.airlift.api.model.ModelResourceModifier.IS_STREAMING_RESPONSE;
 import static io.airlift.api.model.ModelResourceModifier.IS_UNWRAPPED;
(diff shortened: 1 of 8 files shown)


api/src/main/java/io/airlift/api/ApiAllowObject.java
```diff
@@ -0,0 +1,12 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.TYPE_USE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface ApiAllowObject {}
```
Randgalt: I think this is too big a hole to put in the system. Why not simply allow `Map<String, Object>` and `List<Object>` always? Alternatively, we can add an `ApiServiceTrait` to allow these.

electrum: @huberty89 would allowing these two forms work for you?

@Randgalt is always allowing these two forms better than an annotation? My understanding is that you want to enforced properly typed APIs, which I agree with. But for some cases, such as a database query service, you really do need to return dynamically typed data. The thinking here is that the annotation is an acknowledgement that this requires dynamic data, and is a signal for reviewers or tools.

Randgalt: It feels contrived to add these annotations when there's no rule other than if the annotation exists it's allowed. I'm thinking that `Map<String, Object>` and `List<Object>` are useful enough that we should always allow it. Said another way, is there a reason to _not_ allow them?

Randgalt: An alternative is to create new types: `ApiObjectList` and `ApiObject` or something like that.

Randgalt:
Is this superseded by @11xor6's PR?

api/src/main/java/io/airlift/api/ApiAllowObject.java
huberty89: > @huberty89 would allowing these two forms work for you?

Yes

electrum:
@huberty89 Let's drop the annotation for now and simply allow `Map<String, Object>` and `List<Object>`. Based on what you said, this will solve our use case, and everyone is in agreement about it.

huberty89:
I push fixup with removing annotation - if this solution is fine I will later squash and rename the commit message.

I also made alternative PR https://github.com/airlift/airlift/pull/1935 where this behaviour is controller by `ApiServiceTrait` to limit where this applies.

Randgalt:
@huberty89 I'd prefer the Service trait. Thank you!

huberty89:
Superseded by https://github.com/airlift/airlift/pull/1935

PR #1924, opened 2026-05-18
[API] Support ApiJson API types

11xor6:
Summary

- Add an `ApiJson` wrapper base with `ApiJsonNode`, `ApiJsonObject`, and `ApiJsonList` API types that serialize as their wrapped JSON values.
- Teach API Builder resource modeling, validation, request handling, serialization defaults, and OpenAPI schema generation to support ApiJson fields, optionals, collections, maps, and request/response bodies.
- Reject ApiJson types where they would be exposed through paths, IDs, filters, order-by query parameters, or raw Jackson node declarations.

### DIFF
--- a/api/src/main/java/io/airlift/api/ApiJson.java
+++ b/api/src/main/java/io/airlift/api/ApiJson.java
@@ -1,0 +1,13 @@
+package io.airlift.api;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+import com.fasterxml.jackson.databind.JsonNode;
+
+public sealed interface ApiJson<T extends JsonNode>
+        permits ApiJsonList,
+                ApiJsonNode,
+                ApiJsonObject
+{
+    @JsonValue
+    T value();
+}
--- a/api/src/main/java/io/airlift/api/ApiJsonList.java
+++ b/api/src/main/java/io/airlift/api/ApiJsonList.java
@@ -1,0 +1,17 @@
+package io.airlift.api;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+
+import static com.fasterxml.jackson.annotation.JsonCreator.Mode.DELEGATING;
+import static java.util.Objects.requireNonNull;
+
+public record ApiJsonList(ArrayNode value)
+        implements ApiJson<ArrayNode>
+{
+    @JsonCreator(mode = DELEGATING)
+    public ApiJsonList
+    {
+        requireNonNull(value, "value is null");
+    }
+}
--- a/api/src/main/java/io/airlift/api/ApiJsonNode.java
+++ b/api/src/main/java/io/airlift/api/ApiJsonNode.java
@@ -1,0 +1,17 @@
+package io.airlift.api;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.databind.JsonNode;
+
+import static com.fasterxml.jackson.annotation.JsonCreator.Mode.DELEGATING;
+import static java.util.Objects.requireNonNull;
+
+public record ApiJsonNode(JsonNode value)
+        implements ApiJson<JsonNode>
+{
+    @JsonCreator(mode = DELEGATING)
+    public ApiJsonNode
+    {
+        requireNonNull(value, "value is null");
+    }
+}
--- a/api/src/main/java/io/airlift/api/ApiJsonObject.java
+++ b/api/src/main/java/io/airlift/api/ApiJsonObject.java
@@ -1,0 +1,17 @@
+package io.airlift.api;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import static com.fasterxml.jackson.annotation.JsonCreator.Mode.DELEGATING;
+import static java.util.Objects.requireNonNull;
+
+public record ApiJsonObject(ObjectNode value)
+        implements ApiJson<ObjectNode>
+{
+    @JsonCreator(mode = DELEGATING)
+    public ApiJsonObject
+    {
+        requireNonNull(value, "value is null");
+    }
+}
--- a/api/src/main/java/io/airlift/api/builders/ResourceBuilder.java
+++ b/api/src/main/java/io/airlift/api/builders/ResourceBuilder.java
@@ -41,6 +41,9 @@ import java.util.stream.Stream;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static io.airlift.api.internals.ApiJsonTypes.apiJsonResourceDescription;
+import static io.airlift.api.internals.ApiJsonTypes.apiJsonResourceName;
+import static io.airlift.api.internals.ApiJsonTypes.isApiJsonType;
 import static io.airlift.api.internals.Generics.extractGenericParameter;
 import static io.airlift.api.internals.Generics.typeResolver;
 import static io.airlift.api.internals.Mappers.openApiName;
@@ -172,7 +175,12 @@ public class ResourceBuilder
         if (typeToken.isSubtypeOf(Map.class)) {
             Type keyType = extractGenericParameter(typeToken.getType(), 0);
             Type valueType = extractGenericParameter(typeToken.getType(), 1);
-            if (String.class.equals(keyType) && Object.class.equals(valueType)) {
+            if (keyType.equals(String.class) && (valueType.equals(String.class) || isApiJsonType(valueType))) {
+                return internalBuildAndAdd(componentName, valueType)
+                        .asResourceType(ModelResourceType.MAP)
+                        .withContainerType(resource);
+            }
+            if (keyType.equals(String.class) && valueType.equals(Object.class)) {
                 return anyObjectLeafResource()
                         .asResourceType(ModelResourceType.MAP)
                         .withContainerType(resource);
@@ -190,6 +198,15 @@ public class ResourceBuilder
                 }
             }
             return modelResource;
+        }
+
+        if (isApiJsonType(typeToken.getType())) {
+            return new ModelResource(
+                    typeToken.getRawType(),
+                    apiJsonResourceName(typeToken.getType()),
+                    apiJsonResourceDescription(typeToken.getType()),
+                    ImmutableList.of(),
+                    ModelResourceType.JSON);
         }
 
         if (typeToken.isSubtypeOf(ApiStreamResponse.class)) {
--- a/api/src/main/java/io/airlift/api/internals/ApiJsonTypes.java
+++ b/api/src/main/java/io/airlift/api/internals/ApiJsonTypes.java
@@ -1,0 +1,80 @@
+package io.airlift.api.internals;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.reflect.TypeToken;
+import io.airlift.api.ApiJson;
+import io.airlift.api.ApiJsonList;
+import io.airlift.api.ApiJsonNode;
+import io.airlift.api.ApiJsonObject;
+
+import java.lang.reflect.Type;
+
+import static io.airlift.api.internals.Generics.typeResolver;
+
+public final class ApiJsonTypes
+{
+    private ApiJsonTypes() {}
+
+    public static boolean isApiJsonType(Type type)
+    {
+        Class<?> clazz = rawClass(type);
+        return ApiJson.class.isAssignableFrom(clazz);
+    }
+
+    public static String apiJsonResourceName(Type type)
+    {
+        Class<?> clazz = rawClass(type);
+        if (ApiJsonNode.class.isAssignableFrom(clazz)) {
+            return "json";
+        }
+        if (ApiJsonObject.class.isAssignableFrom(clazz)) {
+            return "jsonObject";
+        }
+        if (ApiJsonList.class.isAssignableFrom(clazz)) {
+            return "jsonList";
+        }
+        throw unsupportedType(clazz);
+    }
+
+    public static String apiJsonResourceDescription(Type type)
+    {
+        Class<?> clazz = rawClass(type);
+        if (ApiJsonNode.class.isAssignableFrom(clazz)) {
+            return "An arbitrary JSON value.";
+        }
+        if (ApiJsonObject.class.isAssignableFrom(clazz)) {
+            return "A JSON object.";
+        }
+        if (ApiJsonList.class.isAssignableFrom(clazz)) {
+            return "A JSON array.";
+        }
+        throw unsupportedType(clazz);
+    }
+
+    public static Class<? extends JsonNode> jacksonJsonType(Type type)
+    {
+        Class<?> clazz = rawClass(type);
+        if (ApiJsonNode.class.isAssignableFrom(clazz)) {
+            return JsonNode.class;
+        }
+        if (ApiJsonObject.class.isAssignableFrom(clazz)) {
+            return ObjectNode.class;
+        }
+        if (ApiJsonList.class.isAssignableFrom(clazz)) {
+            return ArrayNode.class;
+        }
+        throw unsupportedType(clazz);
+    }
+
+    private static IllegalArgumentException unsupportedType(Class<?> clazz)
+    {
+        return new IllegalArgumentException("Unsupported ApiJson type: " + clazz.getName());
+    }
+
+    private static Class<?> rawClass(Type type)
+    {
+        return TypeToken.of(typeResolver.resolveType(type)).getRawType();
+    }
+}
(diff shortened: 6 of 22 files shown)


Randgalt:
@coderabbitai full review

Randgalt:
@11xor6 is the goal here to support JSON fields/parameters? I'd prefer to handle that with something like a `ApiJson` type (with overloads for the Jackson objects or something like that). Did you consider something like that?

Also, how does this relate to https://github.com/airlift/airlift/pull/1918? cc @huberty89

11xor6:
> @11xor6 is the goal here to support JSON fields/parameters? I'd prefer to handle that with something like a `ApiJson` type (with overloads for the Jackson objects or something like that). Did you consider something like that?

@Randgalt Yeah we have APIs, like the query endpoint, which have responses which are not consistently structured. There are also requests where the UI generates JSON which we store in the DB. So both for requests and responses, but definitely not anything in the URL/query params.  And no, I didn't consider an `ApiJson` type, but I could convert to something like that. At first glance I don't see any existing class which could be a good pattern to base that off of. Maybe something like `ApiId` with implementations for `JsonNode`, `JsonList`, and `JsonObject`?

Randgalt:
> Maybe something like ApiId with implementations for JsonNode, JsonList, and JsonObject?

I think that will be cleaner and more inline with the rest of the library.

api/src/main/java/io/airlift/api/ApiJson.java
```diff
@@ -0,0 +1,42 @@
+package io.airlift.api;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+import java.util.Objects;
+
+import static java.util.Objects.requireNonNull;
+
+@SuppressWarnings("unused")
+public abstract class ApiJson<T extends com.fasterxml.jackson.databind.JsonNode>
```
Randgalt: Just `JsonNode`. FQPN isn't needed

api/src/main/java/io/airlift/api/ApiJson.java
```diff
@@ -0,0 +1,42 @@
+package io.airlift.api;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+import java.util.Objects;
+
+import static java.util.Objects.requireNonNull;
+
+@SuppressWarnings("unused")
```
Randgalt: not needed

api/src/main/java/io/airlift/api/ApiJson.java
Randgalt: Let's change this to a sealed interface:

```
public sealed interface ApiJson<T extends JsonNode>
        permits ApiJsonList,
                ApiJsonNode,
                ApiJsonObject
{
    @JsonValue
    T value();
}
```

api/src/main/java/io/airlift/api/ApiJsonList.java
```diff
@@ -0,0 +1,14 @@
...
+
+public final class ApiJsonList
+        extends ApiJson<ArrayNode>
+{
+    @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
+    public ApiJsonList(ArrayNode value)
+    {
+        super(value);
+    }
+}
```
Randgalt, line 17:
```
public record ApiJsonList(ArrayNode value)
        implements ApiJson<ArrayNode>
{
    @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
    public static ApiJsonList from(ArrayNode value)
    {
        return new ApiJsonList(value);
    }
}
```

api/src/main/java/io/airlift/api/ApiJsonNode.java
```diff
@@ -0,0 +1,13 @@
...
+
+public final class ApiJsonNode
+        extends ApiJson<com.fasterxml.jackson.databind.JsonNode>
+{
+    @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
+    public ApiJsonNode(com.fasterxml.jackson.databind.JsonNode value)
+    {
+        super(value);
+    }
+}
```
Randgalt:
```
public record ApiJsonNode(JsonNode value)
        implements ApiJson<JsonNode>
{
    @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
    public static ApiJsonNode build(JsonNode value)
    {
        return new ApiJsonNode(value);
    }
}
```

api/src/main/java/io/airlift/api/ApiJsonObject.java
```diff
@@ -0,0 +1,14 @@
...
+
+public final class ApiJsonObject
+        extends ApiJson<ObjectNode>
+{
+    @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
+    public ApiJsonObject(ObjectNode value)
+    {
+        super(value);
+    }
+}
```
Randgalt:
```
public record ApiJsonObject(ObjectNode value)
        implements ApiJson<ObjectNode>
{
    @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
    public static ApiJsonObject build(ObjectNode value)
    {
        return new ApiJsonObject(value);
    }
}
```

api/src/main/java/io/airlift/api/internals/ApiJsonTypes.java
```diff
@@ -0,0 +1,75 @@
...
+public final class ApiJsonTypes
+{
+    private ApiJsonTypes() {}
+
+    public static boolean isApiJsonType(Type type)
+    {
+        Class<?> clazz = rawClass(type);
+        return clazz.equals(ApiJsonNode.class) ||
+                clazz.equals(ApiJsonObject.class) ||
+                clazz.equals(ApiJsonList.class);
```
Randgalt: Given the other changes, this can now be:

```
        Class<?> clazz = rawClass(type);
        return ApiJson.class.isAssignableFrom(clazz);
```

api/src/main/java/io/airlift/api/internals/ApiJsonTypes.java
```diff
@@ -0,0 +1,75 @@
...
+        if (clazz.equals(ApiJsonObject.class)) {
+            return "A JSON object.";
+        }
+        if (clazz.equals(ApiJsonList.class)) {
+            return "A JSON array.";
+        }
+        throw unsupportedType(clazz);
+    }
+
+    public static Class<? extends com.fasterxml.jackson.databind.JsonNode> jacksonJsonType(Type type)
```
Randgalt: Just `JsonNode`. Don't need FQPN

api/src/main/java/io/airlift/api/validation/ResourceSerializationValidator.java
```diff
@@ -166,6 +172,9 @@ private Object internalGetDefaultValue(ValidationContext validationContext, Clas
         if (UUID.class.isAssignableFrom(clazz)) {
             return new UUID(0L, 0L);
         }
+        if (isApiJsonType(clazz)) {
```
Randgalt: This will need to move to `getDefaultRecord()` just after the `if (!clazz.isRecord())` check

Randgalt:
`api/docs/enforcement.md` should be updated to include doc about this.

Randgalt:
@huberty89 can you confirm that this solves your use case too please?

electrum:
Needing to convert to these specialized objects is more complicated and less efficient. We should go with the solution in #1918 of allowing `Object`, at least in specific places.

dain:
I'm concerned that forcing a transit through JsonNode is expensive and complex.  Only the real simple way to deal with this, with a complex tree, would be to convert the JsonNode back to JSON text and then just read it in using an object. Which is just way expensive. Otherwise, you have to write new code to walk the JSON tree and convert it into the object format you want. 

Instead, I think we should just allow Object. My understanding is that the main objection to object is that a dev may use that inappropriately. Not that this is using Object is always wrong. In that case, I think adding an explicit flag/annotation/comment that says, "Hey, I know this is fround upon, but I know what I'm doing, and this is the right approach for me."

11xor6:
@coderabbitai review

11xor6:
@coderabbitai full review

api/src/main/java/io/airlift/api/builders/ResourceBuilder.java
```diff
@@ -177,6 +180,11 @@ private ModelResource internalBuild(Optional<String> componentName, Type resourc
                         .asResourceType(ModelResourceType.MAP)
                         .withContainerType(resource);
             }
+            if (String.class.equals(keyType) && (valueType.equals(String.class) || isApiJsonType(valueType))) {
```
electrum: Flip the `.equals` for `keyType`

api/src/main/java/io/airlift/api/builders/ResourceBuilder.java
```diff
@@ -192,6 +200,10 @@ private ModelResource internalBuild(Optional<String> componentName, Type resourc
             return modelResource;
         }
 
+        if (isApiJsonType(typeToken.getType())) {
+            return new ModelResource(typeToken.getRawType(), apiJsonResourceName(typeToken.getType()), apiJsonResourceDescription(typeToken.getType()), ImmutableList.of(), ModelResourceType.JSON);
```
electrum: Wrap parameters

api/src/main/java/io/airlift/api/internals/Generics.java
```diff
@@ -38,8 +40,8 @@ static void validateMap(Type type)
 
         Type keyType = extractGenericParameter(typeToken.getType(), 0);
         Type valueType = extractGenericParameter(typeToken.getType(), 1);
-        if (!(keyType.equals(String.class)) || !(valueType.equals(String.class))) {
-            throw new ValidatorException("Maps in resources must be Map<String, String>. %s is not".formatted(typeToken.getType()));
+        if (!(keyType.equals(String.class)) || !(valueType.equals(String.class) || isApiJsonType(valueType))) {
```
electrum: No need for parens on the expression

electrum: This might be easier to read if you switch `||` to `&&`

api/src/main/java/io/airlift/api/internals/ApiJsonTypes.java
```diff
@@ -0,0 +1,75 @@
...
+import io.airlift.api.ApiJson;
+import io.airlift.api.ApiJsonList;
+import io.airlift.api.ApiJsonNode;
+import io.airlift.api.ApiJsonObject;
+
+import java.lang.reflect.Type;
+
+import static io.airlift.api.internals.Generics.typeResolver;
+
+public final class ApiJsonTypes
```
electrum, line 16: Add tests for these methods. It's hard to tell what they do, which makes the code using them hard to read.

api/src/main/java/io/airlift/api/openapi/SchemaBuilder.java
```diff
@@ -343,6 +351,35 @@ private Optional<Schema<?>> buildBasicSchema(ModelResource modelResource)
...
+        return Optional.of(buildApiJsonSchema(modelResource));
+    }
+
+    private Schema<?> buildApiJsonSchema(ModelResource modelResource)
+    {
+        Schema<?> schema = buildApiJsonSchema(TypeToken.of(modelResource.type()).getRawType());
+        return schema.description(adjustedDescription(modelResource));
+    }
+
+    private Schema<?> buildApiJsonSchema(Class<?> clazz)
```
electrum: I think this can be static

api/src/main/java/io/airlift/api/openapi/SchemaBuilder.java
```diff
@@ -343,6 +351,35 @@ private Optional<Schema<?>> buildBasicSchema(ModelResource modelResource)
         return schema;
     }
 
+    private Optional<Schema<?>> buildApiJsonSchemaIfPresent(ModelResource modelResource)
+    {
+        if (!isApiJsonType(modelResource.type())) {
+            return Optional.empty();
+        }
+        return Optional.of(buildApiJsonSchema(modelResource));
```
electrum:
```java
return Optional.of(modelResource)
        .filter(resource -> isApiJsonType(resource.type()))
        .map(this::buildApiJsonSchema);

File: api/src/main/java/io/airlift/api/openapi/SchemaBuilder.java [hunk 15]
```diff
@@ -343,6 +351,35 @@ private Optional<Schema<?>> buildBasicSchema(ModelResource modelResource)
...
+    private Schema<?> buildApiJsonSchema(ModelResource modelResource)
+    {
+        Schema<?> schema = buildApiJsonSchema(TypeToken.of(modelResource.type()).getRawType());
+        return schema.description(adjustedDescription(modelResource));
+    }
+
+    private Schema<?> buildApiJsonSchema(Class<?> clazz)
+    {
+        Class<?> jsonType = jacksonJsonType(clazz);
+        if (jsonType.equals(ObjectNode.class)) {
```
electrum: Use switch expression

api/src/main/java/io/airlift/api/validation/MethodValidator.java
```diff
@@ -105,9 +111,65 @@ else if ((context == null) && (suspended == null)) {
...
+    {
+        Set<String> apiJsonResourceNames = Stream.concat(Stream.of(modelMethod.returnType()), modelMethod.requestBody().stream())
+                .flatMap(resource -> apiJsonResourceNames(resource, new HashSet<>()))
+                .collect(toUnmodifiableSet());
+        if (apiJsonResourceNames.isEmpty()) {
+            return;
+        }
+
+        Set<String> queryParameterNames = modelMethod.optionalParameters()
+                .stream()
```
electrum: Previous line

api/src/main/java/io/airlift/api/validation/MethodValidator.java
```diff
@@ -105,9 +111,65 @@ else if ((context == null) && (suspended == null)) {
...
+                (((modelResource.resourceType() == ModelResourceType.LIST) || (modelResource.resourceType() == ModelResourceType.MAP)) && isApiJsonType(modelResource.type()))) {
+            return Stream.of(modelResource.name());
+        }
+        if (!visited.add(modelResource.type())) {
+            return Stream.empty();
+        }
+
+        Stream<ModelResource> directComponents = modelResource.components().stream();
+        Stream<ModelResource> polySubResources = modelResource.polyResource()
+                .stream()
```
electrum: Previous line

api/src/main/java/io/airlift/api/validation/MethodValidator.java
```diff
@@ -105,9 +111,65 @@ else if ((context == null) && (suspended == null)) {
...
+                .findFirst()
+                .ifPresent(name -> {
+                    throw new ValidatorException("ApiJson field %s cannot be used as an API query parameter".formatted(name));
+                });
+    }
+
+    private static Stream<String> apiJsonResourceNames(ModelResource modelResource, Set<Type> visited)
+    {
+        if ((modelResource.resourceType() == JSON) ||
+                (((modelResource.resourceType() == ModelResourceType.LIST) || (modelResource.resourceType() == ModelResourceType.MAP)) && isApiJsonType(modelResource.type()))) {
```
electrum: Add wrapping

api/src/main/java/io/airlift/api/validation/ResourceSerializationValidator.java
```diff
@@ -218,4 +227,18 @@ private Object internalGetDefaultValue(ValidationContext validationContext, Clas
         }
         throw new ValidatorException("Could not create default value for %s".formatted(clazz.getName()));
     }
+
+    private ApiJson<?> getDefaultApiJson(Class<?> clazz)
+    {
+        if (clazz.equals(ApiJsonNode.class)) {
```
electrum: Use switch expression

api/src/main/java/io/airlift/api/ApiJsonList.java
```diff
@@ -0,0 +1,21 @@
...
+
+public record ApiJsonList(ArrayNode value)
+        implements ApiJson<ArrayNode>
+{
+    public ApiJsonList
+    {
+        requireNonNull(value, "value is null");
+    }
+
+    @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
```
electrum: Why do we need a separate factory method? Same comment for the others.

electrum (approved):

api/src/main/java/io/airlift/api/openapi/SchemaBuilder.java
11xor6: Switching on `Class<?>` is pretty ugly:
```java
switch (clazz) {
    case Class<?> apiJsonClass when apiJsonClass == ApiJsonNode.class -> "An arbitrary JSON value.";
    case Class<?> apiJsonClass when apiJsonClass == ApiJsonObject.class -> "A JSON object.";
    case Class<?> apiJsonClass when apiJsonClass == ApiJsonList.class -> "A JSON array.";
    default -> throw unsupportedType(clazz);
};
```

api/src/main/java/io/airlift/api/validation/ResourceSerializationValidator.java
11xor6: See other comment, the switches aren't pretty.

electrum (approved):

PR #1926, opened 2026-05-18
[API] Add ApiUuidId

11xor6:
## Summary

- Add `ApiUuidId` as a UUID-backed `ApiId` helper with a zero-UUID default constructor.
- Teach API generic validation helpers to recognize UUID-backed IDs.
- Cover UUID ID conversion and validation in the API conforming tests.

### DIFF
--- a/api/src/main/java/io/airlift/api/ApiUuidId.java
+++ b/api/src/main/java/io/airlift/api/ApiUuidId.java
@@ -1,0 +1,43 @@
+package io.airlift.api;
+
+import com.google.common.collect.ImmutableList;
+
+import java.util.UUID;
+
+import static io.airlift.api.responses.ApiException.badRequest;
+import static java.util.Objects.requireNonNull;
+
+public abstract class ApiUuidId<RESOURCE>
+        extends ApiId<RESOURCE, UUID>
+{
+    protected ApiUuidId(UUID id)
+    {
+        super(requireNonNull(id, "id is null"));
+    }
+
+    protected ApiUuidId(String id)
+    {
+        super(id);
+    }
+
+    public UUID value()
+    {
+        return toInternal();
+    }
+
+    @Override
+    public UUID toInternal()
+    {
+        try {
+            return UUID.fromString(id);
+        }
+        catch (IllegalArgumentException e) {
+            throw badRequest("Invalid id: " + id, ImmutableList.of("id"));
+        }
+    }
+
+    protected static UUID defaultUuidValue()
+    {
+        return new UUID(0L, 0L);
+    }
+}
--- a/api/src/main/java/io/airlift/api/validation/ValidationContext.java
+++ b/api/src/main/java/io/airlift/api/validation/ValidationContext.java
@@ -13,6 +13,7 @@ import io.airlift.api.ApiStreamResponse.ApiByteStreamResponse;
 import io.airlift.api.ApiStreamResponse.ApiOutputStreamResponse;
 import io.airlift.api.ApiStreamResponse.ApiTextStreamResponse;
 import io.airlift.api.ApiStringId;
+import io.airlift.api.ApiUuidId;
 import io.airlift.api.model.ModelResource;
 import io.airlift.log.Logger;
 import jakarta.ws.rs.core.MediaType;
@@ -25,6 +26,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.UUID;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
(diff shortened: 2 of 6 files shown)


api/src/main/java/io/airlift/api/validation/ValidationContext.java
```diff
@@ -213,7 +218,7 @@ public void validateId(Class<?> idClass)
             throw new ValidatorException("ID's resource type parameter is not a valid resource type: %s".formatted(resourceType));
         }
 
-        if (!ApiEnumId.class.isAssignableFrom(idClass)) {
+        if (!ApiEnumId.class.isAssignableFrom(idClass) && !ApiUuidId.class.isAssignableFrom(idClass)) {
```
Randgalt: Do me a favor and add ` && !ApiStringId.class.isAssignableFrom(idClass)`. It's missing. It's just coincidence that it's working now.

api/src/main/java/io/airlift/api/ApiUuidId.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.google.common.collect.ImmutableList;
+
+import java.util.UUID;
+
+import static io.airlift.api.responses.ApiException.badRequest;
+import static java.util.Objects.requireNonNull;
+
+public abstract class ApiUuidId<RESOURCE>
+        extends ApiId<RESOURCE, UUID>
```
Randgalt, line 11: `ApiStringId` uses a wrapper for the internal type. tbh I don't remember why. I'm mentioning it in case it comes up in the future.

api/src/main/java/io/airlift/api/ApiUuidId.java
```diff
@@ -0,0 +1,43 @@
...
+{
+    public ApiUuidId()
```
Randgalt: This breaks the contract of `ApiId` which throws `UnsupportedOperationException` on the no-arg ctor. Please make this a static factory instead.

api/src/test/java/io/airlift/api/validation/ResourceWithAllTypes.java
```diff
@@ -44,6 +45,7 @@ public record ResourceWithAllTypes(
         @ApiDescription("A type") Set<SimpleResource> sSimpleResource,
         @ApiDescription("A type") Optional<SimpleResource> oSimpleResource,
         @ApiDescription("A type") Optional<UUID> oUuid,
+        @ApiDescription("A type") SimpleUuidId nameId,
```
Randgalt: This is missing the other types: optional, list, optional list, etc. You'll need to create a new test resource type due to naming checks.

Randgalt:
@coderabbitai full review

api/src/main/java/io/airlift/api/ApiUuidId.java
11xor6: Oh yeah, I remember figuring out why that was, but I now also can't remember; however I do remember that it wasn't an issue for this class.

api/src/main/java/io/airlift/api/ApiUuidId.java
11xor6: I did this to provide the required example value, so is there something I'm missing here? I changed it so that this throws the same error as `ApiId`, and created a method for the default value, but I'm unsure if that's what you're asking.

api/src/test/java/io/airlift/api/validation/ResourceWithAllTypes.java
11xor6: I think I got this sorted. Please make sure, also there wasn't a straight `SimpleId` entry so I added that as well.

Randgalt (approved):

api/src/main/java/io/airlift/api/validation/ValidationContext.java
```diff
@@ -213,7 +218,7 @@ public void validateId(Class<?> idClass)
             throw new ValidatorException("ID's resource type parameter is not a valid resource type: %s".formatted(resourceType));
         }
 
-        if (!ApiEnumId.class.isAssignableFrom(idClass)) {
+        if (!ApiEnumId.class.isAssignableFrom(idClass) && ApiStringId.class.isAssignableFrom(idClass) && !ApiUuidId.class.isAssignableFrom(idClass)) {
```
Randgalt: `&& !ApiStringId.class.isAssignableFrom(idClass)`

Randgalt (changes requested):

api/src/main/java/io/airlift/api/validation/ValidationContext.java
11xor6: 🤦

PR #1929, opened 2026-05-19
[MCP] Fix MCP optional parameter schemas

ibodrov-iceguard:
Generating schemas from Optional<T> exposes a synthetic object with a value field. Clients follow that schema and send wrapper objects that the method binding does not expect.

Build method parameter schemas from the Optional payload type so clients send the same shape that the runtime accepts.

### DIFF
--- a/mcp/src/main/java/io/airlift/mcp/model/JsonSchemaBuilder.java
+++ b/mcp/src/main/java/io/airlift/mcp/model/JsonSchemaBuilder.java
@@ -32,6 +32,7 @@ import java.util.function.BiConsumer;
 import java.util.stream.Stream;
 
 import static io.airlift.mcp.reflection.ReflectionHelper.listArgument;
+import static io.airlift.mcp.reflection.ReflectionHelper.schemaType;
 import static java.util.Objects.requireNonNull;
 
 public class JsonSchemaBuilder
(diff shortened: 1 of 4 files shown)


electrum:
@coderabbitai full review

electrum (approved):

mcp/src/main/java/io/airlift/mcp/model/JsonSchemaBuilder.java
```diff
@@ -127,6 +128,15 @@ public ObjectNode build(Optional<String> description, List<MethodParameter> para
                 }));
     }
 
+    private static Type schemaType(ObjectParameter objectParameter)
```
Randgalt: nit: this method might be better in `ReflectionHelper` as:

```
static Type schemaType(Class<?> type)
{
  ...
}
```

mcp/src/main/java/io/airlift/mcp/model/JsonSchemaBuilder.java
```diff
@@ -127,6 +128,15 @@ public ObjectNode build(Optional<String> description, List<MethodParameter> para
...
+    {
+        if (Optional.class.equals(objectParameter.rawType())) {
+            return optionalArgument(objectParameter.genericType()).orElse(objectParameter.genericType());
```
Randgalt: We should assert `!objectParameter.required()`

Randgalt (approved):

PR #1932, opened 2026-05-20
[MCP] procotol 2026/refactor and prep

Randgalt:
Major changes are coming for the MCP protocol next month. Let's begin prepping for them:

- Roots are being removed from the protocol
- Remove server-to-client requests as this will radically change


Randgalt:
cc @codyzwief

Randgalt:
TODO - I need to update the docs

mcp/src/main/java/io/airlift/mcp/operations/LegacyOperations.java
```diff
@@ -384,7 +388,7 @@ private void handleEventStreaming(HttpServletRequest request, HttpServletRespons
         Stopwatch timeoutStopwatch = Stopwatch.createStarted();
         Stopwatch pingStopwatch = Stopwatch.createStarted();
 
-        MessageWriterImpl messageWriter = new MessageWriterImpl(response);
+        ResumableMessageWriter messageWriter = MessageWriter.newResumableMessageWriter(response);
```
codyzwief: nit: static import

mcp/src/main/java/io/airlift/mcp/operations/SessionlessOperations.java
```diff
@@ -105,7 +106,7 @@ public void handleRpcRequest(HttpServletRequest request, HttpServletResponse res
 
         log.debug("Processing MCP request: %s, session: %s", method, request.getHeader(HEADER_SESSION_ID));
 
-        MessageWriterImpl messageWriter = new MessageWriterImpl(response);
+        MessageWriter messageWriter = MessageWriter.newMessageWriter(response);
```
codyzwief: static import

mcp/src/main/java/io/airlift/mcp/messages/MessageWriterImpl.java
```diff
@@ -0,0 +1,180 @@
...
+
+    private void internalWriteMessage(String messageId, String data)
+    {
+        state.updateAndGet(current -> switch (current) {
+            case LATENT -> {
+                response.setContentType("text/event-stream");
+                yield State.SSE;
+            }
+
+            case HTTP -> throw new IllegalStateException("Output is set as an HTTP output stream and cannot be upgraded to SSE");
```
codyzwief, line 161: Should this be an `McpException`, with an internal error code, just in case there's a bug somewhere and this path gets hit in the future?

mcp/src/main/java/io/airlift/mcp/messages/MessageWriterImpl.java
```diff
@@ -0,0 +1,180 @@
...
+
+    MessageWriterImpl(HttpServletResponse response, boolean resumable)
+    {
+        this(ResponseFacade.build(response), resumable);
+    }
+
+    @VisibleForTesting
+    MessageWriterImpl(ResponseFacade response, boolean resumable)
+    {
+        this.response = response;
```
codyzwief: nit: `requireNonNull`

mcp/src/main/java/io/airlift/mcp/messages/MessageWriterImpl.java
```diff
@@ -0,0 +1,180 @@
...
+        this.sentMessages = resumable ? Optional.of(new CopyOnWriteArrayList<>()) : Optional.empty();
+    }
+
+    @Override
+    public void write(String data)
+    {
+        try {
+            PrintWriter writer = response.getWriter();
```
codyzwief: nit: the writer is only used in the else block on 89-92. Perhaps it could be moved there?

mcp/src/main/java/io/airlift/mcp/messages/MessageWriterImpl.java
```diff
@@ -0,0 +1,180 @@
...
+                }
+                return State.HTTP;
+            });
+            if (currentState == State.SSE) {
+                writeMessage(data);
+            }
+            else {
+                // no messages were sent, keep it as a standard JSON response
+                writer.write(data);
+            }
```
codyzwief, line 93: Do we need to consider setting the `application/json` content type header here like we do when the stream gets upgraded to SSE?

codyzwief (approved):
LGTM! A couple nits and one question.

mcp/src/main/java/io/airlift/mcp/messages/MessageWriterImpl.java
Randgalt: No, I don't think so. If this happens, it implies a coding error not an MCP error.

mcp/src/main/java/io/airlift/mcp/messages/MessageWriterImpl.java
Randgalt: It's a subtle thing, I'll add a comment. Calling `response.getWriter();` commits the response inside of Jetty. We need to do this here.

mcp/src/main/java/io/airlift/mcp/messages/MessageWriterImpl.java
Randgalt: That's actually done elsewhere and we don't want to change what might be set.

PR #1943, opened 2026-05-25
Apply max response header size to Jetty maxResponseHeaderSize

### LINKED ISSUE #1942: `http-server.max-response-header-size` has no effect on Jetty 12.1+
## Summary

On Airlift versions that depend on Jetty 12.1.x, setting `http-server.max-response-header-size` to a value larger than 16 KiB has no effect. The HTTP generator caps response headers at the Jetty default of 16 KiB and throws:
```
org.eclipse.jetty.http.HttpException$RuntimeException: 500: Response Header Fields Too Large
    at org.eclipse.jetty.server.internal.HttpConnection$SendCallback.process(HttpConnection.java:880)
```

even though the user-provided value (e.g. 4 MB) is parsed, validated, and visible in `HttpServerConfig`.

The same configuration works as expected with Airlift versions that depend on Jetty 12.0.x.

## Affected versions

- Airlift 361+ (transitively via airbase 305+, which pins Jetty 12.1.1)
- Confirmed broken on Airlift 361 (Jetty 12.1.1) and Airlift 386 (Jetty 12.1.5)
- Working on Airlift 336 (Jetty 12.0.22) and earlier

Found while debugging Trino 477/479, where prepared statements are echoed back in the `X-Trino-Added-Prepare` response header. For sufficiently large prepared statements the header exceeds 16 KiB and the request fails. There is no workaround via Airlift configuration.

## Root cause

Jetty 12.1.0 split `HttpConfiguration` into two fields:

- `responseHeaderSize` (default 8 KiB): size of the initial header buffer.
- `maxResponseHeaderSize` (default 16 KiB): hard cap enforced by the HTTP generator via `setMaxHeaderBytes(...)`.

In Jetty 12.0.x, `_maxResponseHeaderSize` defaulted to `-1`, in which case the generator falls back to `responseHeaderSize`. With that fallback,
setting `responseHeaderSize` alone is sufficient. The fallback was removed in 12.1.

Airlift's `http-server.max-response-header-size` is wired up to call only `HttpConfiguration.setResponseHeaderSize(...)`. It never calls `setMaxResponseHeaderSize(...)`.

Result on Jetty 12.1+: the buffer is sized for e.g. 4 MB, but the generator cap remains at 16 KiB, and large responses fail.

dervoeti:
Fixes https://github.com/airlift/airlift/issues/1942

### DIFF
--- a/http-server/src/main/java/io/airlift/http/server/HttpServer.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpServer.java
@@ -204,9 +204,7 @@ public class HttpServer
             baseHttpConfiguration.setRequestHeaderSize(toIntExact(config.getMaxRequestHeaderSize().toBytes()));
         }
         if (config.getMaxResponseHeaderSize() != null) {
-            int maxResponseHeaderSizeBytes = toIntExact(config.getMaxResponseHeaderSize().toBytes());
-            baseHttpConfiguration.setResponseHeaderSize(maxResponseHeaderSizeBytes);
-            baseHttpConfiguration.setMaxResponseHeaderSize(maxResponseHeaderSizeBytes);
+            baseHttpConfiguration.setMaxResponseHeaderSize(toIntExact(config.getMaxResponseHeaderSize().toBytes()));
         }
         if (config.getOutputBufferSize() != null) {
             baseHttpConfiguration.setOutputBufferSize(toIntExact(config.getOutputBufferSize().toBytes()));
--- a/http-server/src/main/java/io/airlift/http/server/HttpServer.java
+++ b/http-server/src/main/java/io/airlift/http/server/HttpServer.java
@@ -204,7 +204,9 @@ public class HttpServer
             baseHttpConfiguration.setRequestHeaderSize(toIntExact(config.getMaxRequestHeaderSize().toBytes()));
         }
         if (config.getMaxResponseHeaderSize() != null) {
-            baseHttpConfiguration.setResponseHeaderSize(toIntExact(config.getMaxResponseHeaderSize().toBytes()));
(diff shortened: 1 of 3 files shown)


http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -204,7 +204,9 @@ public HttpServer(
             baseHttpConfiguration.setRequestHeaderSize(toIntExact(config.getMaxRequestHeaderSize().toBytes()));
         }
         if (config.getMaxResponseHeaderSize() != null) {
-            baseHttpConfiguration.setResponseHeaderSize(toIntExact(config.getMaxResponseHeaderSize().toBytes()));
+            int maxResponseHeaderSizeBytes = toIntExact(config.getMaxResponseHeaderSize().toBytes());
```
electrum: Add `@MaxDataSize("1GB")` to `HttpServerConfig#getMaxResponseHeaderSize` to ensure that `toIntExact` can't fail here.

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -204,7 +204,9 @@ public HttpServer(
...
+            baseHttpConfiguration.setResponseHeaderSize(maxResponseHeaderSizeBytes);
```
electrum: I think we should only call `baseHttpConfiguration.setMaxResponseHeaderSize` so that we don't allocate a large buffer for every request.

electrum:
Add a test in `TestHttpServerProvider` to verify that the server respects this config. Send a response header that is larger than the 16kB default.

electrum (approved):

wendigo:
This lacks negative test that shows that headers larger than configured size will cause a response to fail. This tests happy path only @electrum

PR #1956, opened 2026-06-01
Support JsonValue resource schemas

11xor6:
## Summary
- model record resources with active Jackson `@JsonValue` as their delegated wire shape in OpenAPI
- preserve API Builder resource identity, paths, operation IDs, descriptions, tags, and alternate schema names
- add validation for ambiguous or unsupported `@JsonValue` resource declarations
- preserve the outer array layer for `@ApiUnwrapped List<T>` when `T` is a `@JsonValue` resource

## Stacking
This PR is currently stacked on #1955 and should be reviewed/merged after #1955.

### DIFF
--- a/api/src/main/java/io/airlift/api/model/ModelResource.java
+++ b/api/src/main/java/io/airlift/api/model/ModelResource.java
@@ -8,12 +8,12 @@ import com.google.common.collect.ImmutableSet;
 import java.lang.reflect.Type;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
-import static com.google.common.collect.ImmutableList.toImmutableList;
 import static io.airlift.api.model.ModelResourceModifier.IS_UNWRAPPED;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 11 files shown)


api/src/main/java/io/airlift/api/model/ModelResource.java
```diff
@@ -48,65 +50,70 @@ public record ModelResource(
 
     public ModelResource(Type type, String name, String description, List<ModelResource> components, ModelResourceType resourceType)
```
dain: you guys should consider switching to a builder... the number of constructors and with methods is pretty bulky

api/src/main/java/io/airlift/api/openapi/SchemaBuilder.java
```diff
@@ -330,10 +330,41 @@ private Schema<?> buildResourceSchema(ModelResource modelResource, BuildSchemaMo
...
+            return schema;
+        }
+
+        schemas.put(new SchemaKey(Optional.empty(), modelResource.containerType(), modelResource.resourceType(), mode), schema);
+        return asRef(schema);
+    }
+
+    private Schema<?> asNamedJsonValueSchema(Schema<?> valueSchema, String name, String description)
+    {
+        if (valueSchema.get$ref() != null) {
```
dain, line 344: calling a method with a `$` in the name has some code smell.  Is that really what you want to do here?

dain (commented):
looking good

dain (approved):

api/src/main/java/io/airlift/api/openapi/SchemaBuilder.java
11xor6: `$ref` is a specific field in OpenAPI, not some special internal field.

dain (approved):

PR #1964, opened 2026-06-04
Support qualified API JAX-RS bindings

ibodrov-iceguard:
## Why

`ApiModule` currently registers generated API Builder JAX-RS resources with the default `JaxrsBinder`. Applications that run multiple JAX-RS servers or enforce explicit server qualifiers need those generated API resources to target the same qualifier as the rest of the application, without bridge modules or validator exemptions for Airlift internals.

## Example

An application with a public HTTP server can already bind regular resources with `jaxrsBinder(binder, ForPublic.class)`. API Builder resources should be able to use the same server binding:

```java
install(ApiModule.builder()
        .addApi(api -> api.add(ChatApi.class))
        .withJaxrsQualifier(ForPublic.class)
        .build());
```

With this change, generated API resources and OpenAPI resources are registered only in the `ForPublic` JAX-RS resource set.


wendigo:
Already implemented

PR #1974, opened 2026-06-10
Improve decay counter performance

wendigo:
By avoiding 3 monitor acquisitions, 3 nanoTime() and 3 Math.exp calls.

synchronized monitor → StampedLock: add() now takes the read lock, so concurrent adders no longer serialize against each other. The write lock is only needed for the rare operations that replace the accumulated count - rescale (once per 50 seconds), merge, reset, resetTo.

### DIFF
--- a/stats/src/main/java/io/airlift/stats/DecayCounter.java
+++ b/stats/src/main/java/io/airlift/stats/DecayCounter.java
@@ -5,6 +5,8 @@ import com.google.errorprone.annotations.ThreadSafe;
 import org.weakref.jmx.Managed;
 
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.DoubleAdder;
+import java.util.concurrent.locks.StampedLock;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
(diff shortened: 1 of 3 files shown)


wendigo:
@coderabbitai full review

stats/src/main/java/io/airlift/stats/DecayCounter.java
```diff
@@ -41,91 +51,174 @@ public DecayCounter(double alpha, Ticker ticker)
...
     {
-        long nowInSeconds = getTickInSeconds();
+        add(value, getTickInSeconds());
+    }
 
-        if (nowInSeconds - landmarkInSeconds >= RESCALE_THRESHOLD_SECONDS) {
-            rescaleToNewLandmark(nowInSeconds);
+    void add(long value, long nowInSeconds)
+    {
+        long stamp = lock.readLock();
```
findepi, line 81: There are two changes combined

- tick can be read once and shared
- synchronized → RW Lock

I like the former. 
I doubt the latter is a win. RW Locks are often slower that synchronized.

PR #1976, opened 2026-06-10
Short-circuit tracing injector in http-client

wendigo:
Do not copy request if there are no fields to be injected into headers

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
+++ b/http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
@@ -968,6 +968,10 @@ public class JettyHttpClient
     @SuppressWarnings("DataFlowIssue")
     private Request injectTracing(Request request, Span span)
     {
+        // a propagator that injects no headers (e.g. noop telemetry) does not need the request rebuilt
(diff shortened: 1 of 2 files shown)


http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
```diff
@@ -968,6 +968,10 @@ private Span startSpan(Request request)
     @SuppressWarnings("DataFlowIssue")
     private Request injectTracing(Request request, Span span)
     {
+        // a propagator that injects no headers (e.g. noop telemetry) does not need the request rebuilt
+        if (propagator.fields().isEmpty()) {
+            return request;
```
findepi, line 973: Why is it worth it?

findepi (approved):
LGTM % question

http-client/src/main/java/io/airlift/http/client/jetty/JettyHttpClient.java
wendigo: This avoid creating a new request (through the builder). Basically every request is copied despite the fact that it's not changed.

PR #1978, opened 2026-06-10
Optimize HttpUriBuilder.encode by avoiding set lookup

wendigo:
Use static precomputed table of allowed chars

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
@@ -4,11 +4,9 @@ import com.google.common.base.Splitter;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.ListMultimap;
 import com.google.common.net.HostAndPort;
-import com.google.common.primitives.Bytes;
 
 import java.net.URI;
 import java.net.URLDecoder;
-import java.nio.ByteBuffer;
 import java.util.Iterator;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
```diff
@@ -258,15 +256,30 @@ private ListMultimap<String, String> parseParams(String query)
         return result;
     }
 
-    private String encode(String input, byte... allowed)
+    private static boolean[] allowedTable(byte[] chars, char... extra)
```
findepi: `updateSet` ?

http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
```diff
@@ -296,4 +308,16 @@ private static String bracketedHostString(HostAndPort hostAndPort)
         }
         return host;
     }
+
+    private static boolean[] updateSet(byte[] chars, char... extra)
```
findepi: i proposed `updateSet` name because i misread the semantics. i see it was a stupid suggestion

- PCHAR `byte[]` → `char[]` (can we give PCHAR a more meaningful name?)
- `boolean[] membership(char[] chars, char... moreChars)`  (maybe you can find a better name)

http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
```diff
@@ -258,15 +256,18 @@ private ListMultimap<String, String> parseParams(String query)
...
-        StringBuilder builder = new StringBuilder();
+        if (!needsEncoding(input, allowed)) {
+            return input;
+        }
 
-        ByteBuffer buffer = UTF_8.encode(input);
-        while (buffer.remaining() > 0) {
-            byte b = buffer.get();
+        byte[] bytes = input.getBytes(UTF_8);
+        StringBuilder builder = new StringBuilder(bytes.length + 16); // add some small headroom
```
findepi: given `needsEncoding` already scans input characters, we could size it exactly.
i am not worried about 16 being too big, but the case where it's too small implies reallocation.

findepi (approved):

http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
```diff
@@ -296,4 +309,16 @@ private static String bracketedHostString(HostAndPort hostAndPort)
         }
         return host;
     }
+
+    private static boolean[] concatenate(byte[] chars, char... extra)
```
findepi: i don't think it concatenates , given it takes byte[], char[] and... doesn't concatenate them :)

PR #1979, opened 2026-06-10
Disable trace propagation if noop tracer is used

wendigo:

### DIFF
--- a/http-server/src/main/java/io/airlift/http/server/tracing/TracingServletFilter.java
+++ b/http-server/src/main/java/io/airlift/http/server/tracing/TracingServletFilter.java
@@ -9,6 +9,7 @@ import io.opentelemetry.api.trace.SpanBuilder;
 import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.api.trace.StatusCode;
 import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.api.trace.TracerProvider;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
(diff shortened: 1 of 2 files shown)


http-server/src/main/java/io/airlift/http/server/tracing/TracingServletFilter.java
```diff
@@ -49,12 +50,19 @@ public final class TracingServletFilter
...
     public TracingServletFilter(OpenTelemetry openTelemetry, Tracer tracer)
     {
         this.propagator = openTelemetry.getPropagators().getTextMapPropagator();
         this.tracer = requireNonNull(tracer, "tracer is null");
+        // A noop tracer combined with a propagator that carries no fields cannot produce spans,
+        // attributes, or extracted contexts, so all per-request work can be skipped. The noop
+        // TracerProvider returns a singleton tracer, which is how both OpenTelemetry.noop() and
+        // io.airlift.tracing.Tracing.noopTracer() create tracers; if a different noop
+        // implementation is bound, the filter simply stays active.
+        this.tracingDisabled = tracer == TracerProvider.noop().get("noop") && propagator.fields().isEmpty();
```
Randgalt, line 65: Where does that constant `"noop"` come from?

wendigo: it doesn't matter as the method is:

```
@Override
  public Tracer get(String instrumentationScopeName) {
    return DefaultTracer.getInstance();
  }
```

Randgalt (approved):

PR #1980, opened 2026-06-10
Stream OpenMetrics exposition avoiding materialization 

wendigo:

### DIFF
--- a/openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
+++ b/openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
@@ -15,25 +15,23 @@ package io.airlift.openmetrics;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.inject.Inject;
-import io.airlift.openmetrics.types.CompositeMetric;
 import io.airlift.openmetrics.types.Metric;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.QueryParam;
+import jakarta.ws.rs.core.StreamingOutput;
 
-import java.util.LinkedHashMap;
+import java.io.BufferedWriter;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
-import java.util.stream.Stream;
 
-import static com.google.common.base.Preconditions.checkState;
-import static com.google.common.collect.ImmutableList.toImmutableList;
-import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static io.airlift.openmetrics.MetricsUtils.groupMetricFamilies;
+import static io.airlift.openmetrics.MetricsUtils.writeMetricFamilies;
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Objects.requireNonNull;
-import static java.util.stream.Collectors.groupingBy;
-import static java.util.stream.Collectors.toList;
 
 @Path("/metrics")
 public class MetricsResource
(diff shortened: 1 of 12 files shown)


openmetrics/src/main/java/io/airlift/openmetrics/types/Labels.java
```diff
@@ -0,0 +1,64 @@
...
+
+import static java.util.Collections.synchronizedMap;
+import static java.util.Map.Entry.comparingByKey;
+import static java.util.stream.Collectors.joining;
+
+final class Labels
+{
+    // Labels are typically the constant node annotations shared by every metric on every scrape,
+    // so the rendered form is memoized. Keys are held weakly to avoid pinning maps owned by callers.
+    private static final Map<Map<String, String>, String> RENDERED_CACHE = synchronizedMap(new WeakHashMap<>());
```
Randgalt: Why not Guava? You'd get better concurrency:

```
Map<Map<String, String>, String> RENDERED_CACHE = CacheBuilder.newBuilder()
    .weakKeys()
    .build()
    .asMap();
```

wendigo: makes sense

wendigo: Done

openmetrics/src/main/java/io/airlift/openmetrics/types/Metric.java
```diff
@@ -28,13 +31,45 @@ public sealed interface Metric
 {
     String metricName();
 
-    String getMetricExposition();
+    void writeMetricExposition(Writer writer)
```
findepi, line 48: Is the contract that implementation must append some text and `\n`, or without `\n`?

wendigo: Yes. It needs to write a full line

findepi: is it worth documenting the contract?

wendigo: Ok

openmetrics/src/main/java/io/airlift/openmetrics/types/Metric.java
```diff
@@ -27,13 +31,53 @@ public sealed interface Metric
...
+     * form, escaped or raw. A stray {@code \n} or {@code \r} corrupts not just this
+     * metric but the parse of every line that follows.
+     *
+     * @param writer destination for the exposition; buffered by the caller, so
+     *         implementations may issue many small writes
+     */
+    void writeMetricExposition(Writer writer)
+            throws IOException;
+
+    default String getMetricExposition()
```
grantatspothero: This is preserved for backwards compatibility, but callers should prefer `writeMetricExposition` right?

Document that.

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
```diff
@@ -50,13 +58,17 @@ public MetricsResource(OpenMetricsCollector collector)
 
     @GET
     @Produces(OPENMETRICS_CONTENT_TYPE)
-    public String getMetrics(@QueryParam("name[]") List<String> filter)
+    public StreamingOutput getMetrics(@QueryParam("name[]") List<String> filter)
```
grantatspothero, line 51: Thanks for doing this, much better than the previous approach of repeatedly building intermediate strings in memory 🙌

openmetrics/src/main/java/io/airlift/openmetrics/types/Metric.java
```diff
@@ -27,13 +31,53 @@ public sealed interface Metric
...
+        try {
+            writeMetricExposition(writer);
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+        return writer.toString();
+    }
 
     String getMetricDescriptor();
```
grantatspothero: Any reason to not implement `void writeMetricDescriptor` in a similar fashion as `void writeMetricExposition`?

openmetrics/src/main/java/io/airlift/openmetrics/types/Labels.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.common.cache.CacheBuilder;
+
+import java.util.Map;
+
+import static java.util.Map.Entry.comparingByKey;
+import static java.util.stream.Collectors.joining;
+
+final class Labels
+{
+    // Labels are typically the constant node annotations shared by every metric on every scrape,
```
grantatspothero: I introduced a similar abstraction in this PR: https://github.com/airlift/airlift/pull/1881/changes/7f3d65c59dac1bd1f0efd2b89eb2282e2f001f28#diff-7b07b53e9a01ee7d53ed38c2d3f517931ebe2fd81b98089bbd74e4e9b2b2f6c5

The main difference is Labels is a rendering cache (using reference equality on Map to cache toString) while LabelSet is intended to have value semantics (using equals on a canonical sorted string representation).

If LabelSet existed, you could cache the string representation at construction time and that gives you a rendering cache for free. No need for a map of weak keys.

openmetrics/src/main/java/io/airlift/openmetrics/types/Metric.java
wendigo: Dropped

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
wendigo: Yes. These metrics can be actually pretty large.

openmetrics/src/main/java/io/airlift/openmetrics/types/Metric.java
wendigo: Done

wendigo:
In Trino /metrics yield 1MB output on the default configuration

openmetrics/src/main/java/io/airlift/openmetrics/types/Labels.java
```diff
@@ -0,0 +1,67 @@
...
+    // so the rendered form is memoized. Keys are held weakly to avoid pinning maps owned by callers.
+    private static final Map<Map<String, String>, String> RENDERED_CACHE = CacheBuilder.newBuilder()
```
losipiuk: Not super convinced this is beneficial given key comparison is costly anyway

openmetrics/src/main/java/io/airlift/openmetrics/types/BigCounter.java
```diff
@@ -31,14 +35,16 @@ public record BigCounter(String metricName, BigInteger value, Map<String, String
     }
 
     @Override
-    public String getMetricExposition()
+    public void writeMetricExposition(Writer writer)
```
losipiuk, line 38: you are removing old methods. Are you sure those are not used outside of airlfit?

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
losipiuk, line 51: Why not pass `OutputStream` as argument and return `void`? This semantics is not obvious when you just look at signature

openmetrics/src/main/java/io/airlift/openmetrics/types/BigCounter.java
wendigo: These are not used outside of airlift. Already checked build with Trino

openmetrics/src/main/java/io/airlift/openmetrics/MetricsResource.java
wendigo: https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/StreamingOutput.html

openmetrics/src/main/java/io/airlift/openmetrics/types/Labels.java
wendigo: I'm dropping this commit

openmetrics/src/main/java/io/airlift/openmetrics/types/Metric.java
```diff
@@ -28,56 +26,28 @@ public sealed interface Metric
...
+     * form, escaped or raw. A stray {@code \n} or {@code \r} corrupts not just this
+     * metric but the parse of every line that follows.
+     *
+     * @param writer destination for the exposition; buffered by the caller, so
+     *         implementations may issue many small writes
+     */
+    void writeMetricExposition(Writer writer)
+            throws IOException;
+
+    void writeMetricDescription(Writer writer)
```
grantatspothero: nit: naming `writeMetricDescriptor`, openmetrics uses this terminology.

grantatspothero (approved):
Looks good [non-binding]

Randgalt (approved):

PR #1983, opened 2026-06-10
Add single-quantile fast path to TDigest.valueAt

wendigo:
This avoids allocating a single element double[] array which is frequently called while scraping JMX stats


findepi:
Many new code lines. Why is it worth adding them?

PR #1989, opened 2026-06-10
Add typed API filters with schema-aware binding

huberty89:
This is improvement proposition.

Adds `TypedApiFilter<T>` and `TypedApiFilterList<T>` as typed alternatives to the existing string-based API filters.

This improves type safety for API query filters: the generic type is now used for both OpenAPI schema generation and runtime request binding, so the documented filter type and the value received by service code stay aligned. That should help prevent mistakes after API changes, since filter types are no longer described by hand in OpenAPI while parsed separately at runtime.

Supported filter types are `Boolean`, `Integer`, `Long`, `Double`, `String`, `Instant`, `UUID`, and concrete enums. `Instant` filters are exposed as `string` / `date-time` in OpenAPI and parsed from ISO-8601 strings at runtime. Enum filters use the existing enum naming behavior, including `ApiEnumNamingFormat`.


Randgalt:
@huberty89 what about just changing `ApiFilter` to be typed. Current uses of `ApiFilter` would be mapped to `ApiFilter<Object>` using the raw class. Current usages can update as needed but wouldn't break. Same with `TypedApiOrderBy`

huberty89:
> @huberty89 what about just changing `ApiFilter` to be typed. Current uses of `ApiFilter` would be mapped to `ApiFilter<Object>` using the raw class. Current usages can update as needed but wouldn't break. Same with `TypedApiOrderBy`

I agree it would be nice to have single choice (not both `ApiFilter` and `TypedApiFilter`) but what about those existing methods in `ApiFilter` and `ApiFilterList` - it won't be a bit confusing to have those in case of generic type?

Randgalt:
> I agree it would be nice to have single choice (not both `ApiFilter` and `TypedApiFilter`) but what about those existing methods in `ApiFilter` and `ApiFilterList` - it won't be a bit confusing to have those in case of generic type?

I believe that existing implementations would just show up as using raw versions, like use `List` without type arguments. They should still compile.

api/src/main/java/io/airlift/api/internals/Mappers.java
```diff
@@ -208,44 +206,44 @@ static ApiHeader buildHeader(HttpHeaders httpHeaders, String name)
...
     }
 
-    static ApiFilter buildFilter(UriInfo uriInfo, String name)
+    static ApiFilter<Object> buildFilter(UriInfo uriInfo, String name)
     {
         return buildFilter(Optional.ofNullable(uriInfo.getQueryParameters().getFirst(name)));
     }
 
-    static ApiFilter buildFilter(Optional<Object> queryParameter)
+    static ApiFilter<Object> buildFilter(Optional<Object> queryParameter)
```
Randgalt, line 214: Can be:

```
static <T> ApiFilter<T> buildFilter(Optional<T> queryParameter)
```

api/src/main/java/io/airlift/api/internals/Mappers.java
```diff
@@ -208,44 +206,44 @@ static ApiHeader buildHeader(HttpHeaders httpHeaders, String name)
...
-    static ApiFilterList buildFilterList(UriInfo uriInfo, String name)
+    static ApiFilterList<Object> buildFilterList(UriInfo uriInfo, String name)
     {
         List<String> strings = uriInfo.getQueryParameters().get(name);
         List<Object> values = (strings != null) ? strings.stream().collect(toImmutableList()) : ImmutableList.of();
         return buildFilterList(values);
     }
 
-    static ApiFilterList buildFilterList(List<Object> values)
+    static ApiFilterList<Object> buildFilterList(List<Object> values)
```
Randgalt, line 226: Can be:

```
static <T> ApiFilterList<T> buildFilterList(List<T> values)
```

api/src/test/java/io/airlift/api/servertests/integration/testingserver/external/PublicWidgetApi.java
```diff
@@ -51,9 +51,9 @@ public PublicWidgetApi(InternalController controller, ApiQuotaController quotaCo
     }
 
     @ApiList(description = "List all widgets")
-    public ApiPaginatedResult<ExternalWidget> listWidgets(@ApiParameter ApiPagination pagination, @ApiParameter(allowedValues = {"id", "name", "size"}) ApiOrderBy ordering, @ApiParameter ApiFilter name, @ApiParameter ApiFilterList size)
+    public ApiPaginatedResult<ExternalWidget> listWidgets(@ApiParameter ApiPagination pagination, @ApiParameter(allowedValues = {"id", "name", "size"}) ApiOrderBy ordering, @ApiParameter ApiFilter<Object> name, @ApiParameter ApiFilterList<Object> size)
     {
-        return controller.list(pagination.withOrdering(ordering), name, size).map(InternalWidget::map);
+        return controller.list(pagination.withOrdering(ordering), name, size).map(widget -> widget.map());
```
Randgalt, line 56: Can be method ref

PR #1991, opened 2026-06-11
Add API request cancellation

dain:
## Why
- Expensive API handlers need a framework-supported way to stop backend work when clients abandon requests.
- Leaving cancellation to each resource implementation would make callers responsible for preserving request context, tracing, executor handoff, and lifecycle details.
- API Builder should keep its typed request/response model without taking a direct dependency on Jetty internals.
- Long-lived keep-alive connections must not retain completed request cancellation state.

## Approach
- Add a request-scoped `ApiCancellation` interface injectable with `@Context`.
- Run API methods that use `ApiCancellation` on the configured cancellable request executor.
- Add an opt-in `REQUEST_CANCELLATION` HTTP server feature that installs the low-level Jetty cancellation filter.
- Have the HTTP server publish a generic request cancellation future, and have API Builder adapt that into `ApiCancellation`.
- Require cancellation listeners to provide their own executor for cleanup work.
- Document the API and HTTP server setup, including executor sizing expectations.

### DIFF
--- a/api/src/main/java/io/airlift/api/binding/RequestApiCancellation.java
+++ b/api/src/main/java/io/airlift/api/binding/RequestApiCancellation.java
@@ -1,0 +1,71 @@
+package io.airlift.api.binding;
+
+import io.airlift.api.ApiCancellation;
+import io.airlift.log.Logger;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+
+import static java.util.Objects.requireNonNull;
+
+class RequestApiCancellation
+        implements ApiCancellation
+{
+    private static final Logger log = Logger.get(RequestApiCancellation.class);
+    private static final ListenerRegistration NOOP_REGISTRATION = () -> {};
+
+    private final CompletableFuture<?> cancellation;
+
+    RequestApiCancellation(CompletableFuture<?> cancellation)
+    {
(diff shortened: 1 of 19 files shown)


dain:
@coderabbitai reviw

api/src/main/java/io/airlift/api/binding/RequestApiCancellation.java
```diff
@@ -0,0 +1,71 @@
...
+                return;
+            }
+
+            try {
+                executor.execute(() -> {
+                    try {
+                        listener.run();
+                    }
+                    catch (Throwable e) {
+                        log.warn(e, "API cancellation listener failed");
```
nineinchnick, line 56: Should this log an error? If the cancel callback fails, some state might need to be manually cleaned up, like database records or objects in persistent storage.

nineinchnick (approved):

api/src/main/java/io/airlift/api/binding/RequestApiCancellation.java
dain: An error level seems like it would be a bit too high for this.  Basically, we got a callback from the user, and it threw an exception. There's really nothing we can do about it because this is an async code base, so the best we can do is log it. But if they have a bug like that, it's just going to flood the log with error messages.

PR #1994, opened 2026-06-12
Handle missing HTTP server keystore

nineinchnick:
## Why
- Jetty allows HTTP servers to use an externally supplied SSLContext, and that setup does not expose a KeyStore through SslContextFactory.
- The certificate-expiration JMX/OpenMetrics path treats the KeyStore as always present, so metric collection can fail repeatedly with a NullPointerException.
- Servers without a readable KeyStore cannot report certificate expiration from this path, so the metric should be absent instead of breaking collection.

## Approach
- Treat a missing SslContextFactory KeyStore the same as having no discoverable certificates.
- Cover the injected SSLContext path with a provider test.


nineinchnick:
@dain PTAL

dain (approved):

PR #1997, opened 2026-06-15
Cache forward-decay weight calculation

wendigo:
This avoids recalculating weight (using expensive Math.exp) on every add.

### DIFF
--- a/stats/src/main/java/io/airlift/stats/DecayCounter.java
+++ b/stats/src/main/java/io/airlift/stats/DecayCounter.java
@@ -31,6 +31,12 @@ public final class DecayCounter
     private long landmarkInSeconds;
     private double count;
 
+    // The forward-decay weight is constant for a given (second, landmark) pair, so cache it and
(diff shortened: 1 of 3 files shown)


wendigo:
@findepi this is similar to a previous PR but focused only on caching weight calculation

stats/src/main/java/io/airlift/stats/DecayCounter.java
martint, line 1: I think it may be cleaner to make ExponentialDecay a stateful object and keeping the cached value there + appropriate API to set the landmark and get current weight.

wendigo: Makes sense. Let me implement this

wendigo: https://github.com/airlift/airlift/pull/2003

wendigo:
Superseded by https://github.com/airlift/airlift/pull/2003

PR #2001, opened 2026-06-15
Fix merging decay t-digests with different landmarks

wendigo:
Newly added test fails without the fix

### DIFF
--- a/stats/src/main/java/io/airlift/stats/DecayTDigest.java
+++ b/stats/src/main/java/io/airlift/stats/DecayTDigest.java
@@ -19,6 +19,7 @@ import com.google.common.base.Ticker;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static io.airlift.stats.ExponentialDecay.weight;
 
 public class DecayTDigest
@@ -195,6 +196,7 @@ public class DecayTDigest
     public void merge(DecayTDigest other)
     {
         DecayTDigest copy = other;
+        checkArgument(other.alpha == alpha, "Expected DecayTDigest to have alpha %s, but was %s", alpha, other.alpha);
         // Both digests store forward-decayed weights relative to their own landmark, so the raw
         // centroid weights are only comparable once both share a landmark. Align to the newer
         // landmark (rescaling reduces weights, which is numerically safe) before combining them,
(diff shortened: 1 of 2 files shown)


wendigo:
This is correctness issue that can produce invalid metrics for merged decay t-digests. This affects Distribution and TimeDistribution

stats/src/main/java/io/airlift/stats/DecayTDigest.java
```diff
@@ -194,6 +194,19 @@ public DecayTDigest duplicate()
...
+        // centroid weights are only comparable once both share a landmark. Align to the newer
+        // landmark (rescaling reduces weights, which is numerically safe) before combining them,
+        // mirroring DecayCounter.merge. Rescaling the older operand is harmless here because the
+        // only caller discards the merged-in digest immediately afterwards.
+        if (alpha > 0.0) {
+            if (landmarkInSeconds < other.landmarkInSeconds) {
+                rescale(other.landmarkInSeconds);
+            }
+            else if (landmarkInSeconds > other.landmarkInSeconds) {
+                other.rescale(landmarkInSeconds);
```
losipiuk: I do not think we should modify `other` in this method

wendigo: What do you propose instead?

wendigo:
@coderabbitai full review

stats/src/main/java/io/airlift/stats/DecayTDigest.java
```diff
@@ -194,6 +195,22 @@ public DecayTDigest duplicate()
...
+        // Both digests store forward-decayed weights relative to their own landmark, so the raw
+        // centroid weights are only comparable once both share a landmark. Align to the newer
+        // landmark (rescaling reduces weights, which is numerically safe) before combining them,
+        // mirroring DecayCounter.merge.
+        if (alpha > 0.0 && landmarkInSeconds < other.landmarkInSeconds) {
+            rescale(other.landmarkInSeconds);
+        }
+        else if (alpha > 0.0 && landmarkInSeconds > other.landmarkInSeconds) {
+            // Rescale a copy of the other digest so that this method does not mutate its argument;
+            // callers may retain a reference to it.
```
martint, line 209: It should be safe to rescale the other digest. The change just affects internal representation, but it's fundamentally the same data structure.

wendigo: That was my initial implementation but @losipiuk objected. I can revert

wendigo: I'll rescale once https://github.com/airlift/airlift/pull/2003 lands

PR #2002, opened 2026-06-15
Stop warning on missing API docs links

nineinchnick:
ApiService documentation links default to an empty list, and the model only appends the links to the OpenAPI description when they are present. Treating the default value as warning-worthy makes services with no external documentation look suspicious even though validation does not require those links.

URI parsing still rejects malformed link values when annotations provide them.


nineinchnick:
@Randgalt I don't have much context around this, just noticed these warnings in the logs, and they don't look actionable.

Randgalt (approved):
Yeah - this seems right. I don't remember why this check is here. Maybe this used to be optional.

nineinchnick:
@electrum @dain PTAL

PR #2003, opened 2026-06-15
Extract exponential decay into DecayConfig and DecayState

wendigo:
Bundle the decay factor, landmark, and ticker into ExponentialDecay, which now also caches the most recently computed weight. DecayCounter, DecayTDigest, and QuantileDigest delegate their landmark/weight/rescale logic to it instead of each duplicating the (alpha, ticker, landmark) trio. The decay presets now build ExponentialDecay instances, and containers (Distribution, TimeDistribution) fork an independent timeline per sub-structure.


wendigo:
@coderabbitai full review

wendigo:
Supersedes https://github.com/airlift/airlift/pull/1997

wendigo:
@coderabbitai full review

wendigo:
@coderabbitai full review

stats/src/main/java/io/airlift/stats/DecayConfig.java
```diff
@@ -0,0 +1,191 @@
...
+        @Override
+        public long read()
+        {
+            return 0;
+        }
+    };
+
+    public DecayConfig
+    {
+        checkArgument(alpha >= 0 && alpha < 1, "alpha must be in range [0, 1)");
```
martint: alpha == 0 should not be allowed. It's meaningless for an exponential decay setup.

stats/src/main/java/io/airlift/stats/DecayCounter.java
```diff
@@ -22,79 +19,82 @@
...
     public DecayCounter(double alpha, Ticker ticker)
     {
-        this(0, alpha, ticker, TimeUnit.NANOSECONDS.toSeconds(ticker.read()));
+        this(DecayConfig.of(alpha, ticker));
     }
 
-    private DecayCounter(double count, double alpha, Ticker ticker, long landmarkInSeconds)
+    public DecayCounter(DecayConfig config)
+    {
+        this(0, requireNonNull(config, "config is null").newState());
```
martint: requireNonNull is unnecessary

stats/src/main/java/io/airlift/stats/DecayConfig.java
wendigo: alpha == 0 means no decay. It's accepted in the existing code

martint: That's in the DecayCounter, and it's because we didn't have access to anything like OptionalInt at the time. However, now that we're encapsulating the exponential decay logic in a dedicated class, that class should not allow alpha == 0, and instead, it should be handled one level above it by not keeping an exponential decay object around when it's disabled.

wendigo: Do you want me to use Optional<DecayState> or just use null to signal "no decay" ?

wendigo:
@martint DecayState requires alpha > 0 now

wendigo:
CleanShot 2026-06-17 at 13 50 32@2x

Last recent two commits improve TDigest insert/merge further - by around 30% for both operations

wendigo:
PTAL @martint

martint (approved):
The decay counter changes look good. Split the rest to a separate PR.

PR #2005, opened 2026-06-16
Small optimizations

wendigo:

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/FullJsonResponseHandler.java
+++ b/http-client/src/main/java/io/airlift/http/client/FullJsonResponseHandler.java
@@ -76,6 +76,7 @@ public class FullJsonResponseHandler<T>
     {
         private final int statusCode;
         private final ListMultimap<HeaderName, String> headers;
+        private final Charset charset;
         private final boolean hasValue;
         private final byte[] jsonBytes;
         private final byte[] responseBytes;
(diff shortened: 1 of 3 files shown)


http-client/src/main/java/io/airlift/http/client/FullJsonResponseHandler.java
```diff
@@ -86,6 +87,7 @@ public JsonResponse(int statusCode, ListMultimap<HeaderName, String> headers, by
         {
             this.statusCode = statusCode;
             this.headers = ImmutableListMultimap.copyOf(headers);
+            this.charset = getCharset();
```
findepi, line 90: Calling instance methods in constructor is generally error-prone.
One has to go over 4 calls to verify the call reads only those fields that were already assigned before the call happened.

Furthermore, `getCharset` can throw-and-suppress, which means it can be costly, and the charset information is not always needed (eg. `getJson` and `getJsonBytes` don't seem to use it).

http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
```diff
@@ -28,6 +29,12 @@ public class HttpUriBuilder
     private String path = ""; // decoded path
     private final ListMultimap<String, String> params = LinkedListMultimap.create(); // decoded query params
 
+    private static final Splitter QUERY_PARAM_SPLITTER = Splitter.on("&")
+            .omitEmptyStrings();
+
+    private static final Splitter QUERY_PARAM_VALUE_SPLITTER = Splitter.on("=")
```
findepi, line 35: `com.google.common.base.Splitter#withKeyValueSeparator(char)`

wendigo: Won't work. This will return a Map that rejects repeated keys (which are fine in the URI)

PR #2008, opened 2026-06-17
Use Splitters in the HttpUriBuilder

wendigo:

### DIFF
--- a/http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
+++ b/http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
@@ -8,6 +8,7 @@ import com.google.common.net.HostAndPort;
 import java.net.URI;
 import java.net.URLDecoder;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 
 import static com.google.common.base.Preconditions.checkArgument;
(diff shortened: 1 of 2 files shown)


wendigo:
@findepi ptal

electrum (approved):

http-client/src/main/java/io/airlift/http/client/HttpUriBuilder.java
```diff
@@ -244,13 +248,11 @@ private ListMultimap<String, String> parseParams(String query)
...
-                    .omitEmptyStrings()
-                    .split(query);
-
-            for (String pair : pairs) {
-                String[] parts = pair.split("=", 2);
-                result.put(percentDecode(parts[0]), percentDecode(parts[1]));
+            for (String pair : QUERY_SPLITTER.split(query)) {
+                List<String> parts = QUERY_PAIR_SPLITTER.splitToList(pair);
+                String key = percentDecode(parts.get(0));
+                String value = parts.size() > 1 ? percentDecode(parts.get(1)) : null;
```
electrum, line 254: I think the old code would fail if there was value, which means we don't have any tests for this case. Add a test to verify that using null here is safe and won't break elsewhere.

PR #2009, opened 2026-06-17
Add a compact serialization format and auto-select it when smaller

wendigo:
TDigest serialized every centroid as two fixed-width 8-byte doubles, so size grew at 37 + 16·n bytes regardless of the data. Add a second, lossless wire format that exploits the structure of a merged digest:

- Means are stored ascending, so each is mapped to a sortable long (bits ^ ((bits >> 63) & Long.MAX_VALUE) — monotonic across the full double range and self-inverse) and delta-encoded as a varint. The first delta is zig-zag encoded since it may be negative; subsequent deltas are non-negative. Clustered means produce small deltas.
- Weights are written as varints when all integral (the common count/add(value) case), guarded by a flag bit; otherwise they fall back to fixed-width doubles (fractional/decay weights).
- The header (min/max/compression/totalWeight) stays fixed-width; centroidCount becomes a varint.

### DIFF
--- a/stats/src/main/java/io/airlift/stats/TDigest.java
+++ b/stats/src/main/java/io/airlift/stats/TDigest.java
@@ -15,6 +15,7 @@ package io.airlift.stats;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.primitives.Doubles;
+import io.airlift.slice.DynamicSliceOutput;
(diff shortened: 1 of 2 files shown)


wendigo:
For a single tpch query that I was running locally, original format size was 13,669 bytes, while new format was 6,668 bytes which is 48.78% of the original size

PR #1625, opened 2025-11-03
Fixups/etc for API module

Randgalt:
Various fixups/etc. needed after attempting to port an existing project to the api module.

# Airlift contribution check list

 - [X] Git commit messages follow https://cbea.ms/git-commit/.
 - [X] All automated tests are passing.
 - [x] Pull request was categorized using one of the existing labels.

PR #2016, opened 2026-06-24
Ignore 5xx API status validation

nineinchnick:
## Why
- Server error responses are not endpoint-specific API contract outcomes; they can be produced by exception paths, filters, infrastructure, or framework-level failures.
- Treating undeclared 5xx statuses as missing response annotations makes DEBUG mode replace the original failure.
- Warning on undeclared 5xx statuses still suggests every endpoint should document generic server errors, which adds annotation noise without improving client-visible contracts.

## Approach
- Skip undeclared-status validation for server-error response families while leaving existing validation for non-server-error responses intact.

### DIFF
--- a/api/src/main/java/io/airlift/api/binding/JaxrsStatusValidator.java
+++ b/api/src/main/java/io/airlift/api/binding/JaxrsStatusValidator.java
@@ -34,10 +34,13 @@ public class JaxrsStatusValidator
     @Override
     public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext)
     {
-        if ((responseContext.getStatusInfo().getFamily() != Response.Status.Family.SUCCESSFUL)) {
-            findApiServiceMethod(requestContext, modelServices).ifPresent(modelMethod ->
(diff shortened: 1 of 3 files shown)


nineinchnick:
I checked the PR that added this filter, and it didn't include any reasoning for the filter: https://github.com/airlift/airlift/pull/1625

electrum (approved):
Looks good to me, but I'd like @Randgalt to review

api/src/main/java/io/airlift/api/binding/JaxrsStatusValidator.java
```diff
@@ -42,6 +42,11 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
 
     private void validateApiError(ModelMethod modelMethod, String httpMethod, int statusCode, URI uri)
     {
+        boolean isServerError = (Response.Status.Family.familyOf(statusCode) == Response.Status.Family.SERVER_ERROR);
```
Randgalt: I think is fine. I'd move this to the `if ((responseContext.getStatusInfo().getFamily() != Response.Status.Family.SUCCESSFUL)) {` check in the `filter()` method.

Randgalt (approved):

PR #2019, opened 2026-06-26
Split HTTP and HTTPS server protocol configs

dain:
## Why
- Disabling HTTP should also remove the plain HTTP configuration surface, so an HTTPS-only server does not accept HTTP-only properties or bind insecure sockets/connectors.
- For typical operators this should be mostly transparent: HTTP remains enabled by default, the HTTP port still has the same default, and HTTP-enabled deployments can continue setting the existing HTTP properties.
- The intentional compatibility break is limited to configs that disable a protocol while still setting that protocol's properties, plus developer code that directly uses the affected config classes or constructors.

## Approach
- Move HTTP-only port and connector settings into `HttpConfig`, which is only bound when HTTP is enabled.
- Move HTTPS connector settings into `HttpsConfig`, which is only bound when HTTPS is enabled.
- Treat accept queue size as a per-socket setting by exposing `http-server.http.accept-queue-size` and `http-server.https.accept-queue-size`, while keeping `http-server.accept-queue-size` as a legacy alias on both protocol configs for migration.
- Require at least one protocol to remain enabled so the server cannot be configured without any listener.
- Avoid creating the HTTP socket channel, Jetty connector, and `@http` virtual host when HTTP is disabled.
- Update direct construction paths and tests for the new protocol-specific config objects.

## Backward Incompatibility
- Operator configs that disable HTTP and still set `http-server.http.port`, `http-server.http.accept-queue-size`, `http-server.http.acceptor-threads`, or `http-server.http.selector-threads` will now fail because those properties are intentionally unused when HTTP is disabled.
- Operator configs that disable HTTPS and still set `http-server.https.accept-queue-size`, `http-server.https.acceptor-threads`, or `http-server.https.selector-threads` will now fail because those properties are intentionally unused when HTTPS is disabled.
- Operator configs that disable both HTTP and HTTPS will now fail because the server must have at least one listener.
- Operator configs should migrate from the legacy shared `http-server.accept-queue-size` property to `http-server.http.accept-queue-size` and/or `http-server.https.accept-queue-size` when they need protocol-specific socket backlog tuning.
- Developer code using `HttpServerConfig` directly must move HTTP port/thread access to `HttpConfig` and HTTPS accept queue/thread access to `HttpsConfig`.
- Developer code using `HttpServerConfig#getHttpAcceptQueueSize()` or `setHttpAcceptQueueSize(...)` must use `HttpConfig#getAcceptQueueSize()`/`setAcceptQueueSize(...)` or `HttpsConfig#getAcceptQueueSize()`/`setAcceptQueueSize(...)`.
- Developer code constructing `HttpServerInfo` directly must now pass `Optional<HttpConfig>` in addition to `Optional<HttpsConfig>`.


electrum:
Wrap commit descriptions

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -133,17 +133,50 @@ public HttpServer(
...
+                servlet,
+                filters,
+                resources,
+                serverFeatures,
+                clientCertificate,
+                mBeanServer,
+                maybeSslContextFactory);
+    }
+
+    public HttpServer(
```
electrum, line 123: Do we need the compatibility constructor? Or make the old one `@Deprecated`?

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -133,17 +133,50 @@ public HttpServer(
...
         requireNonNull(nodeInfo, "nodeInfo is null");
         requireNonNull(config, "config is null");
+        requireNonNull(maybeHttpConfig, "httpConfig is null");
         requireNonNull(maybeHttpsConfig, "httpsConfig is null");
         requireNonNull(servlet, "servlet is null");
         requireNonNull(maybeSslContextFactory, "maybeSslContextFactory is null");
         requireNonNull(clientCertificate, "clientCertificate is null");
         requireNonNull(mBeanServer, "mBeanServer is null");
 
+        checkArgument(!config.isHttpEnabled() || maybeHttpConfig.isPresent(), "httpConfig must be present when HTTP is enabled");
```
electrum: Could assign this to `httpConfig` here with `orElseThrow()`

Also same cleanup for `httpsConfig`

http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -296,7 +330,14 @@ public HttpServer(
          */
         StatisticsHandler statsHandler = new StatisticsHandler();
 
-        ServletContextHandler servletContext = createServletContext(servlet, resources, filters, Set.of("http", "https"), showStackTrace, serverFeatures.contains(LEGACY_URI_COMPLIANCE));
+        ImmutableSet.Builder<String> connectorNames = ImmutableSet.builder();
```
electrum: Can just use `HashSet` here

http-server/src/main/java/io/airlift/http/server/HttpServerConfig.java
```diff
@@ -63,9 +63,6 @@
 public class HttpServerConfig
 {
     private boolean httpEnabled = true;
```
electrum, line 65: Do we need a validator that fails if both HTTP and HTTPS are disabled?

http-server/src/main/java/io/airlift/http/server/HttpServerInfo.java
```diff
@@ -40,16 +40,13 @@ public class HttpServerInfo
...
-        this(config, Optional.empty(), nodeInfo);
-    }
-
     @Inject
-    public HttpServerInfo(HttpServerConfig config, Optional<HttpsConfig> httpsConfig, NodeInfo nodeInfo)
+    public HttpServerInfo(HttpServerConfig config, Optional<HttpConfig> httpConfig, Optional<HttpsConfig> httpsConfig, NodeInfo nodeInfo)
     {
         if (config.isHttpEnabled()) {
-            httpChannel = createChannel(nodeInfo.getBindIp(), config.getHttpPort(), config.getHttpAcceptQueueSize());
+            checkArgument(httpConfig.isPresent(), "httpConfig must be present when HTTP is enabled");
```
electrum:
```java
HttpConfig http = httpConfig.orElseThrow(() -> ...)

File: http-server/src/main/java/io/airlift/http/server/HttpServerInfo.java [hunk 6]
```diff
@@ -61,7 +58,8 @@ public HttpServerInfo(HttpServerConfig config, Optional<HttpsConfig> httpsConfig
 
         if (config.isHttpsEnabled()) {
             checkArgument(httpsConfig.isPresent(), "httpsConfig must be present when HTTPS is enabled");
-            httpsChannel = createChannel(nodeInfo.getBindIp(), httpsConfig.orElseThrow().getHttpsPort(), config.getHttpAcceptQueueSize());
+            HttpsConfig https = httpsConfig.orElseThrow();
```
electrum:
```java
HttpsConfig https = httpsConfig.orElseThrow(() -> ...)

File: http-server/src/test/java/io/airlift/http/server/TestHttpsConfig.java [hunk 7]
```diff
@@ -93,6 +102,22 @@ public void testExplicitPropertyMappings()
         assertFullMapping(properties, expected);
     }
 
+    @Test
+    public void testLegacyPropertyMapping()
+    {
+        Map<String, String> currentProperties = new ImmutableMap.Builder<String, String>()
+                .put("http-server.https.accept-queue-size", "2048")
+                .put("http-server.https.automatic-shared-secret", "automatic-secret")
```
electrum: Why is this in here? It's not deprecated and has the same value in both maps

electrum (approved):

http-server/src/main/java/io/airlift/http/server/HttpServerInfoProvider.java
```diff
@@ -32,9 +32,10 @@ public void setInjector(Injector injector)
     @Override
     public HttpServerInfo get()
     {
-        HttpServerConfig httpConfig = injector.getInstance(qualifiedKey(qualifier, HttpServerConfig.class));
+        HttpServerConfig serverConfig = injector.getInstance(qualifiedKey(qualifier, HttpServerConfig.class));
+        Optional<HttpConfig> httpConfig = injector.getInstance(qualifiedKey(qualifier, new TypeLiteral<>() {}));
         Optional<HttpsConfig> httpsConfig = injector.getInstance(qualifiedKey(qualifier, new TypeLiteral<>() {}));
```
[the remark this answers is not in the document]
dain: You're wrong. You may need to update to modern Java, but this advice is ancient, and Java properly does type inference.

http-server/src/main/java/io/airlift/http/server/HttpServerProvider.java
```diff
@@ -73,6 +73,7 @@ public HttpServer get()
                     injector.getInstance(NodeInfo.class),
                     injector.getInstance(qualifiedKey(qualifier, HttpServerConfig.class)),
                     injector.getInstance(qualifiedKey(qualifier, new TypeLiteral<>() {})),
+                    injector.getInstance(qualifiedKey(qualifier, new TypeLiteral<>() {})),
                     injector.getInstance(qualifiedKey(qualifier, Servlet.class)),
                     injector.getInstance(qualifiedKey(qualifier, new TypeLiteral<>() {})),
                     injector.getInstance(qualifiedKey(qualifier, new TypeLiteral<>() {})),
```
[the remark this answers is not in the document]
dain: You're wrong. You may need to update to modern Java, but this advice is ancient, and Java properly does type inference.

PR #2030, opened 2026-06-30
Handle raw JMX metric edge cases

dain:
- Preserve raw JMX boolean and `TabularData` attributes at the shared metrics collection boundary.
- Route raw JMX `TabularData` through the existing OpenMetrics composite renderer.
- Skip raw JMX collection for final `ObjectName`s already exported through jmxutils, using a single snapshot of managed exports per collection.


electrum (approved):
Make sure to wrap commit descriptions

PR #2037, opened 2026-07-04
Add native OTLP log export

dain:
Add an Airlift logging backend that publishes application logs directly to OpenTelemetry using OTLP. The backend uses the existing OpenTelemetry exporter configuration and SDK log pipeline so applications can configure the same collector transport used by other OpenTelemetry signals.

The logging backend also supports early initialization by adding a bootstrap logging hook that runs after Airlift logging is configured and before Guice is invoked. This lets applications install the OTLP handler early while keeping bootstrap and log-manager independent of OpenTelemetry.

### DIFF
--- a/opentelemetry/src/main/java/io/airlift/opentelemetry/OpenTelemetryLoggingModule.java
+++ b/opentelemetry/src/main/java/io/airlift/opentelemetry/OpenTelemetryLoggingModule.java
@@ -1,0 +1,106 @@
+package io.airlift.opentelemetry;
+
+import com.google.inject.Binder;
+import com.google.inject.Module;
+import io.airlift.bootstrap.LoggingBootstrap;
+import io.airlift.bootstrap.LoggingBootstrapContext;
+import io.airlift.configuration.ConfigurationFactory;
+import io.airlift.node.NodeConfig;
+import io.airlift.node.NodeInfo;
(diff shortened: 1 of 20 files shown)


opentelemetry/src/main/java/io/airlift/opentelemetry/OpenTelemetryLoggingModule.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        OpenTelemetryExporterConfig exporterConfig = configurationFactory.build(OpenTelemetryExporterConfig.class);
+        NodeConfig nodeConfig = configurationFactory.build(NodeConfig.class);
+        if (nodeConfig.getNodeId() == null) {
+            throw new IllegalStateException("log.otlp.enabled=true requires node.id to be configured");
+        }
+        NodeInfo nodeInfo = new NodeInfo(nodeConfig);
+        Resource resource = OpenTelemetryModule.createResource(serviceName, serviceVersion, nodeInfo);
+        LogRecordProcessor logRecordProcessor = logRecordProcessorFactory.apply(exporterConfig, MeterProvider.noop());
```
nineinchnick: Can we pass a real meter provider? I think it's used to track errors.

nineinchnick (approved):

PR #2050, opened 2026-07-13
Harden automatic mTLS with a derived CA and per-node keys

wendigo:
Derive the automatic-mTLS certificate authority deterministically from the shared secret using salted PBKDF2 key stretching instead of an unsalted SHA1PRNG seed, and have each node generate its own random leaf key signed by that CA. This keeps the private key that terminates each TLS connection unique per node and unrecoverable from the shared secret, while resisting offline brute-force of a weak secret. CertificateBuilder gains CA-signing support and no longer emits an empty SubjectAltName extension, and bcprov moves to compile scope for deterministic EC arithmetic. Shared secrets shorter than 32 characters are now rejected.

### DIFF
--- a/http-server/src/test/java/io/airlift/http/server/TestHttpServerProvider.java
+++ b/http-server/src/test/java/io/airlift/http/server/TestHttpServerProvider.java
@@ -329,7 +329,7 @@ public class TestHttpServerProvider
                 .setKeystorePath(getResource("test.keystore.with.two.passwords").getPath())
                 .setKeystorePassword("airlift")
                 .setKeyManagerPassword("airliftkey")
-                .setAutomaticHttpsSharedSecret("shared-secret");
+                .setAutomaticHttpsSharedSecret("shared-secret-with-sufficient-entropy-1234");
 
         createAndStartServer();
         assertThat(server.getHttpConnectionStats()).isNull();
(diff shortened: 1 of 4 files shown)


electrum:
Bouncy Castle has historically been a CVE nightmare. I don't think we want to bring it in as a dependency.

wendigo:
@electrum what about conscrypt? Recent version is supported on macos/aarch and linux/aarch

electrum (dismissed):
This seems to be combining multiple changes. Could we split this up into multiple commits or PRs for easier review?

* Fix empty SAN handling.
* Add generic CA-signed/leaf support to `CertificateBuilder`.
* Switch automatic mTLS to a derived CA with per-node leaf keys and PKIX.
* Enforce the minimum shared-secret length.
* Replace the raw-secret `SHA1PRNG` seed with a versioned, environment-salted PBKDF2 derivation.
* Separately evaluate whether CA derivation must be identical across JCA providers and, if so, how to implement the required EC point multiplication without introducing an undesirable runtime dependency.
* Add caching as a separate performance optimization.

electrum:
We should use whatever is built into the JDK. If not, there needs to be a very strong reason why. We don't want to be in the business of shipping cryptography libraries (other than the JDK).

electrum:
Also, given Google's track record for maintaining open source Java libraries, I don't have confidence that it would be well maintained in the future.

wendigo:
@electrum we do that in Trino already (there is a bouncycastle dependency)

electrum:
It's not in the Trino tarball. Looks like it's only a test dependency, brought in by `io.minio:minio`.

wendigo:
@electrum PR splitted. Bouncycastle is gone

electrum (commented):
This seems fine to me. @dain should give the final approval

wendigo:
@dain PTAL

PR #2052, opened 2026-07-18
Add HTTP client interceptors and exchange recording

jvanzyl:
## Why
- Airlift HTTP clients can transform requests or observe status codes, but cannot surround a complete synchronous transport exchange.
- Applications need an OkHttp-style extension point for request rewriting, short circuits, response wrapping, and observability without replacing the Jetty transport.
- Bounded, sanitized recordings make real exchanges reusable as fixtures for tests and mock-server implementations.

## Approach
- Add a minimal `Request`/`StreamingResponse` interceptor chain with deterministic global and named-client registration.
- Adapt existing request filters into the chain while preserving their public API, ordering, tracing, and synchronous failure lifecycle.
- Add a versioned recorder that tees caller-consumed response bytes, reports incomplete or truncated capture, and sanitizes every exchange before it reaches a sink.
- Keep retry policy outside the SPI so generated clients record each physical attempt independently.
- Treat this as the useful first part of the OkHttp model: async interception, separate application/network registries, transport-internal follow-ups, and response-to-request association are not added.

### DIFF
--- a/http-client-generator/src/test/java/io/airlift/http/client/generator/AirliftHttpClientCodegenIntegrationTest.java
+++ b/http-client-generator/src/test/java/io/airlift/http/client/generator/AirliftHttpClientCodegenIntegrationTest.java
@@ -13,6 +13,13 @@
  */
 package io.airlift.http.client.generator;
 
+import com.sun.net.httpserver.HttpExchange;
+import com.sun.net.httpserver.HttpServer;
+import io.airlift.http.client.HttpClientConfig;
+import io.airlift.http.client.RecordedExchange;
+import io.airlift.http.client.RecordedExchangeSanitizer;
+import io.airlift.http.client.RecordingHttpClientInterceptor;
+import io.airlift.http.client.jetty.JettyHttpClient;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 import org.openapitools.codegen.ClientOptInput;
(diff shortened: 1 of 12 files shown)


jvanzyl:
## Reviewer Notes

Review order:
1. Start with `HttpClientInterceptor` and the `JettyHttpClient` chain/lifecycle.
2. Check `HttpClientBinder` and `HttpClientModule` for legacy filter compatibility and global-before-named ordering.
3. Review `RecordedExchange` and `RecordingHttpClientInterceptor` for bounded teeing, close propagation, and sanitize-before-sink behavior.
4. Finish with the focused interceptor, recorder, binder, and generated-client tests.

Non-goals:
- Async interception, separate application/network interceptor registries, redirect or authentication follow-up interception inside Jetty, and retry metadata in the SPI.
- A richer response type that carries its originating request; `StreamingResponse` remains unchanged.

Risks:

| Risk | Check | Inspection Target | Verified Evidence |
|---|---|---|---|
| Existing clients change execution path | Request filters still run once before tracing and user interceptors, with unchanged public bindings and async behavior | `JettyHttpClient`, `HttpClientModule`, binder tests | focused interceptor and binder tests pass 29/29 |
| Spans end before synchronous failures are recorded | Handler, exception-handler, and response-close failures are recorded before the span ends; handled transport errors preserve prior status | tracing lifecycle in `JettyHttpClient` | focused failure-span tests pass |
| Recording consumes or leaks response bodies | Capture copies only bytes consumed by the caller and delegates close exactly once | `RecordingHttpClientInterceptor` | recorder tests pass 11/11 |
| Retry becomes coupled to transport interception | No retry types enter production interceptor sources; each physical generated-client call records independently | interceptor sources and generated-client integration test | generator tests pass 6/6 |
| Default reactor produces intermittent transport failures | Confirm failures are reproducible before attributing them to this diff | legacy proxy, HTTP/2, and async tests | every earlier reported method passed in isolation, a concurrency-disabled cohort passed 184 tests, and the unchanged candidate passed the final default reactor |

Verification:
- `./mvnw -pl http-client -Dtest=TestJettyHttpClientInterceptor,TestHttpClientBinder test`: 29 tests passed.
- `./mvnw -pl http-client -Dtest=TestRecordingHttpClientInterceptor test`: 11 tests passed.
- `./mvnw -pl http-client-generator test`: 6 tests passed, including nested Maven and generated 429-to-200 recording.
- production-source retry-coupling search: no matches.
- `git diff --check origin/master...HEAD`: passed.
- four repeatedly affected transport classes with JUnit concurrency disabled: 184 tests passed.
- `./mvnw -pl http-client,http-client-generator -am test`: all 15 selected modules passed with default concurrency, including 773 HTTP-client tests with 13 skips and 6 generator tests.
- earlier combined-reactor failures: classified as intermittent concurrent transport-test harness behavior because different proxy, HTTP/2, or async methods failed, every reported method passed in isolation, and the unchanged candidate passed the final default reactor.
- `api-maven-plugin` nested-reactor issue: classified as a pre-existing non-hermetic bootstrap defect; the fixture passes when `440-SNAPSHOT` is installed locally, while ordinary `-am test` does not provision that sibling plugin in a clean repository.

PR #2067, opened 2026-07-24
Don't color the log message to ensure visibility against all backgrounds

huw0:
# Summary

Sometimes Airlift log messages are not visible with a white terminal background.

# Detail

Presently the Airlift log stdout format always specifically colours the message part of the log in white.

In most cases this is ok, since terminals with a white background usually remap white to an alternative colour, such as light grey (e.g macOS).

However this is not guaranteed. Recent versions of OpenShift default to a light theme and don't do this remap. As a consequence the message part of Trino log lines is invisible without selecting the text or viewing the log via another means (e.g k8s CLI).

# Solution

The solution is simple, the message part of the log line should not have an explicit colour set. Instead this should use the terminal default which will guarantee that it is always visible.

### DIFF
--- a/log-manager/src/main/java/io/airlift/log/StaticFormatter.java
+++ b/log-manager/src/main/java/io/airlift/log/StaticFormatter.java
@@ -23,7 +23,6 @@ import static io.airlift.log.TerminalColors.Color.BRIGHT_BLACK;
 import static io.airlift.log.TerminalColors.Color.CYAN;
 import static io.airlift.log.TerminalColors.Color.GREEN;
 import static io.airlift.log.TerminalColors.Color.PURPLE;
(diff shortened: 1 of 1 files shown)


wendigo:
@huw0 NO_COLOR

huw0:
`NO_COLOR` seems like a hack in this instance?

The colours on the message are useful. When scanning the log it's useful to have WARNs etc appropriately highlighted.

It shouldn't be the case that it's a choice between no highlighting or highlighting where the message is only visible in dark mode terminal (...or terminals that work around this behaviour).

wendigo:
@huw0 WHITE doesn't mean "white color" it means "white color from a palette"

wendigo:
https://github.com/airlift/airlift/pull/2068

PR #2079, opened 2026-07-29
[MCP] MRTR Protocol

Randgalt:
- Support new MRTR protocol
- Emulate MRTR for older protocols

See: https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/mrtr


mcp/src/main/java/io/airlift/mcp/model/CallToolResult.java
```diff
@@ -7,19 +7,58 @@
...
+                        Optional.empty(),
+                        Optional.empty(),
+                        Optional.empty(),
+                        Optional.of(INPUT_REQUIRED),
+                        requestState,
+                        Optional.of(inputRequests.buildOrThrow()),
+                        Optional.empty());
+            }
+        };
+    }
```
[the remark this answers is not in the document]
Randgalt: resolved

mcp/src/main/java/io/airlift/mcp/operations/legacy/MrtrEmulator.java
```diff
@@ -0,0 +1,62 @@
...
+    MrtrEmulator(McpConfig config)
+    {
+        timeout = config.getEventStreamingTimeout().toJavaTime();
+    }
+
+    <T extends InputRequests> T emulate(LegacyRequestContextImpl requestContext, T result, BiFunction<Optional<String>, Map<String, Object>, T> handler)
+    {
+        while (result.resultType().orElse(COMPLETE) == INPUT_REQUIRED) {
+            Map<String, InputRequest> inputRequests = result.inputRequests().orElseThrow(() -> new IllegalArgumentException("Input requests may not be empty"));
+            Map<String, Object> inputResponses = new HashMap<>();
```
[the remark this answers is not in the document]
Randgalt: addressed

mcp/src/main/java/io/airlift/mcp/model/GetPromptRequest.java
```diff
@@ -23,9 +25,19 @@ public GetPromptRequest(String name, Map<String, Object> arguments)
...
     @Override
     public GetPromptRequest withMeta(Map<String, Object> meta)
     {
         return new GetPromptRequest(name, arguments, Optional.of(meta));
     }
+
+    public GetPromptRequest withInputResponses(Optional<String> requestState, Map<String, Object> inputResponses)
+    {
+        return new GetPromptRequest(name, arguments, requestState, Optional.of(inputResponses), meta);
+    }
```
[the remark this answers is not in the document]
Randgalt: fixed

mcp/src/test/java/io/airlift/mcp/ConformanceEndpoints.java
```diff
@@ -274,4 +286,196 @@ public void testTriggerToolChange(McpRequestContext requestContext)
...
+            builder.add("user_name", METHOD_ELICITATION_CREATE, elicitRequestForm);
+        }
+
+        if (requestContext.clientCapabilities().sampling().isPresent()) {
+            CreateMessageRequest createMessageRequest = new CreateMessageRequest(Role.USER, new TextContent("Generate a greeting"), 100);
+            builder.add("greeting", METHOD_SAMPLING_CREATE_MESSAGE, createMessageRequest);
+        }
+
+        return builder.build();
+    }
```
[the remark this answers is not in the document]
Randgalt: It's OK - this is just a test

mcp/src/main/java/io/airlift/mcp/operations/legacy/MrtrEmulator.java
```diff
@@ -0,0 +1,66 @@
...
+    <T extends InputRequests<T>> T emulate(LegacyRequestContextImpl requestContext, T result, BiFunction<Optional<String>, Map<String, Object>, T> handler)
+    {
+        while (result.resultType().orElse(COMPLETE) == INPUT_REQUIRED) {
+            Map<String, InputRequest> inputRequests = result.inputRequests().orElseGet(ImmutableMap::of);
+            checkArgument(!inputRequests.isEmpty(), "Input requests may not be empty");
+
+            Map<String, Object> inputResponses = new HashMap<>();
+            inputRequests.forEach((key, inputRequest) -> {
+                try {
+                    JsonRpcResponse<Object> response = requestContext.serverToClientRequest(inputRequest.method(), inputRequest.params(), Object.class, timeout, POLL_INTERVAL);
```
Randgalt, line 47: Maybe `timeout` should be adjusted per call so that timeout is total timeout for all input requests?

wendigo:
@Randgalt can you address coderabbit comments?

Randgalt:
> @Randgalt can you address coderabbit comments?

done

mcp/src/main/java/io/airlift/mcp/McpException.java
```diff
@@ -20,24 +20,36 @@ public class McpException
 
     public McpException(JsonRpcErrorDetail errorDetail)
     {
-        this.errorDetail = requireNonNull(errorDetail, "errorDetail is null");
+        requireNonNull(errorDetail, "errorDetail is null");
+        super(errorDetail.message());
```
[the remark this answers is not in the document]
Randgalt: That's no longer true in Java 25

mcp/src/test/java/io/airlift/mcp/ConformanceEndpoints.java
```diff
@@ -274,4 +286,196 @@ public void testTriggerToolChange(McpRequestContext requestContext)
...
+    {
+        String hashedValue = Hashing.sha256().hashString("corned beef", UTF_8)
+                .toString();
+
+        return inputResponses.requestState().map(requestState -> {
+                    if (requestState.equals(hashedValue)) {
+                        return new CallToolResult(new TextContent("It worked"));
+                    }
+                    throw new Error("RequestState mismatch");
+                })
```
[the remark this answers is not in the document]
claymccoy: I agree. It seems like Error should not be thrown, or the reasoning for it should be documented.

mcp/src/main/java/io/airlift/mcp/model/CallToolRequest.java
```diff
@@ -8,29 +8,37 @@
...
+public record CallToolRequest(String name, Map<String, Object> arguments, Optional<String> requestState, Optional<Map<String, Object>> inputResponses, Optional<Map<String, Object>> meta)
+        implements Meta<CallToolRequest>, InputResponses<CallToolRequest>
 {
     public CallToolRequest
     {
         requireNonNull(name, "name is null");
         arguments = ImmutableMap.copyOf(requireNonNullElse(arguments, ImmutableMap.of()));
-        meta = requireNonNullElse(meta, Optional.<Map<String, Object>>empty()).map(ImmutableMap::copyOf);
+        requestState = requireNonNullElse(requestState, Optional.empty());
+        inputResponses = requireNonNullElse(inputResponses, Optional.empty());
```
claymccoy, line 19: Should this `normalize(inputResponses)` like ReadResourceRequest does? I'm assuming that `normanlize` copies the externally passed in mutable map whose state could be changed externally.

mcp/src/main/java/io/airlift/mcp/model/GetPromptRequest.java
```diff
@@ -9,13 +9,15 @@
...
-        implements Meta<GetPromptRequest>
+public record GetPromptRequest(String name, Map<String, Object> arguments, Optional<String> requestState, Optional<Map<String, Object>> inputResponses, Optional<Map<String, Object>> meta)
+        implements Meta<GetPromptRequest>, InputResponses<GetPromptRequest>
 {
     public GetPromptRequest
     {
         requireNonNull(name, "name is null");
         arguments = ImmutableMap.copyOf(requireNonNullElse(arguments, ImmutableMap.of()));
+        requestState = requireNonNullElse(requestState, Optional.empty());
+        inputResponses = requireNonNullElse(inputResponses, Optional.empty());
```
claymccoy, line 20: Same question about using `normalize(inputResponses)`.

mcp/src/main/java/io/airlift/mcp/McpClientException.java
```diff
@@ -9,6 +9,8 @@ public class McpClientException
 
     public McpClientException(McpException mcpException)
     {
+        super(mcpException.getMessage(), mcpException);
+
```
claymccoy, line 13: Should these lines be switched? Line 12 will NPE and never get to line 14.

mcp/src/main/java/io/airlift/mcp/model/InputRequests.java
```diff
@@ -0,0 +1,68 @@
...
+    @JsonProperty
+    Optional<ResultType> resultType();
+
+    @JsonProperty
+    Optional<String> requestState();
+
+    @JsonProperty
+    Optional<Map<String, InputRequest>> inputRequests();
+
+    T withInputRequests(Optional<ResultType> resultType, Optional<String> requestState, Optional<Map<String, InputRequest>> inputRequests);
```
claymccoy, line 26: This is a really cool usage of sealed classes, generics, and builders. But seeing this method with the 3 optional params makes me wonder if this would all be cleaner if resultType, requestState, and inputRequests were just a simple record themselves and used via composition rather than inheritence. That said, I am sure there is a reason that it was done this way. It did take me a minute to really understand what was going on due to the complexity.
Perhaps the same for *Results.

mcp/src/main/java/io/airlift/mcp/McpException.java
```diff
@@ -68,10 +86,10 @@ public static McpException exception(String message)
 
     public static McpException exception(int code, String message)
     {
-        return exception(code, message, Optional.empty());
+        return exceptionWithData(code, message, Optional.empty());
```
claymccoy, line 89: Does anything external use this? It is public, so that is a possibility. Perhaps it should be deprecated for one release?

claymccoy (approved):
I read through the "Multi Round-Trip Requests" and then looked at this. I'm not an expert on the new round trip functionality, so my feedback may be a bit superficial. The tests help a lot with understanding how it is supposed to work.
The NPE risk, normalize, an throwing Errors would be easy fixes and seem worthwhile. The other stuff is more subjective.

mcp/src/main/java/io/airlift/mcp/model/InputRequests.java
```diff
@@ -0,0 +1,68 @@
...
+import static java.util.Objects.requireNonNull;
+
+public sealed interface InputRequests<T extends InputRequests<T>>
+        permits CallToolResult,
+                GetPromptResult,
+                ReadResourceResult
+{
+    @JsonProperty
+    Optional<ResultType> resultType();
+
```
codyzwief: From what I can tell in the [schema](https://modelcontextprotocol.io/specification/2026-07-28/schema#inputrequiredresult) `ResultType` is a required property, but here it's Optional. Why not make it required and default it to `COMPLETE`?

mcp/src/main/java/io/airlift/mcp/model/CallToolResult.java
```diff
@@ -36,10 +71,21 @@ public CallToolResult(List<Content> content)
...
-        return new CallToolResult(content, structuredContent, isError, Optional.of(meta));
+        return new CallToolResult(content, structuredContent, isError, resultType, requestState, inputRequests, Optional.of(meta));
+    }
+
+    public CallToolResult withResultType(ResultType resultType)
+    {
+        return new CallToolResult(content, structuredContent, isError, Optional.of(resultType), requestState, inputRequests, meta);
     }
 
     public static CallToolResult forError(McpClientException mcpClientException)
```
codyzwief, line 76: `forError` should probably include the `COMPLETE` ResultType (or some other one). ISTM that field is being omitted in the error case.

wendigo:
@Randgalt I did some changes, ptal

wendigo (approved):

mcp/src/main/java/io/airlift/mcp/operations/legacy/MrtrEmulator.java
```diff
@@ -0,0 +1,67 @@
...
+    MrtrEmulator(McpConfig config)
+    {
+        timeout = config.getEventStreamingTimeout().toJavaTime();
+    }
+
+    <T extends InputRequests<T>> T emulate(LegacyRequestContextImpl requestContext, T result, BiFunction<Optional<String>, Map<String, Object>, T> handler)
+    {
+        while (result.resultType().orElse(COMPLETE) == INPUT_REQUIRED) {
+            Map<String, InputRequest> inputRequests = result.inputRequests().orElseGet(ImmutableMap::of);
+            checkArgument(!inputRequests.isEmpty(), "Input requests may not be empty");
```
codyzwief: Do we need to check here (or somewhere else) that the client supports the InputRequests the server is attempting to send? From Servers MUST NOT send an inputRequests that the client has not declared support for in its capabilities. For example, if a client does not declare support for elicitation, the server MUST NOT include any elicitation/create requests in the inputRequests field.:

> Servers MUST NOT send an inputRequests that the client has not declared support for in its capabilities. For example, if a client does not declare support for elicitation, the server MUST NOT include any elicitation/create requests in the inputRequests field.

mcp/src/main/java/io/airlift/mcp/model/InputResponses.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @JsonProperty
+    Optional<Map<String, Object>> inputResponses();
+
+    @JsonIgnore
+    default Map<String, Object> inputResponsesMap()
+    {
+        return inputResponses().orElseGet(ImmutableMap::of);
+    }
```
codyzwief, line 31: Do we want to consider making this typed in some fashion so individual implementations don't have to run a `jsonMapper.convertValue()`? If the typing doesn't match up with what the tool expects then using the JSON mapper will throw an exception, while I wonder if we could possibly catch that earlier and convert type mismatches to `INVALID_PARAMS` or something similar. WDYT?

mcp/src/main/java/io/airlift/mcp/model/CallToolRequest.java
```diff
@@ -8,29 +8,37 @@
...
+        meta = requireNonNullElse(meta, Optional.empty());
```
codyzwief: It looks like this is no longer copying `meta` to a new ImmutableMap. Do we want to reinstate that (and also do the same for `inputResponses` too)?

mcp/src/main/java/io/airlift/mcp/model/CallToolRequest.java
codyzwief: +1

mcp/src/main/java/io/airlift/mcp/model/InputResponses.java
```diff
@@ -0,0 +1,57 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Map;
+import java.util.Optional;
+
+@SuppressWarnings("rawtypes")
+public interface InputResponses<T extends InputResponses<T>>
```
codyzwief: Why is `InputRequests` sealed but `InputResponses` isn't? IMO we should pick one or the other for both if possible, since they're pretty similar.

mcp/src/main/java/io/airlift/mcp/model/GetPromptResult.java
```diff
@@ -25,18 +46,38 @@ public record PromptMessage(Role role, Content content)
     public GetPromptResult
     {
         description = requireNonNullElse(description, Optional.empty());
-        messages = ImmutableList.copyOf(messages);
+        messages = requireNonNullElse(messages, Optional.<List<PromptMessage>>empty()).map(ImmutableList::copyOf);
+        resultType = requireNonNullElse(resultType, Optional.empty());
+        requestState = requireNonNullElse(requestState, Optional.empty());
+        inputRequests = requireNonNullElse(inputRequests, Optional.<Map<String, InputRequest>>empty()).map(ImmutableMap::copyOf);
+        meta = requireNonNullElse(meta, Optional.empty());
         meta = normalize(meta);
```
codyzwief: nit: remove line 53, `normalize()` handles nulls

mcp/src/main/java/io/airlift/mcp/operations/legacy/MrtrEmulator.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Inject
+    MrtrEmulator(McpConfig config)
+    {
+        timeout = config.getEventStreamingTimeout().toJavaTime();
+    }
+
+    <T extends InputRequests<T>> T emulate(LegacyRequestContextImpl requestContext, T result, BiFunction<Optional<String>, Map<String, Object>, T> handler)
+    {
+        while (result.resultType().orElse(COMPLETE) == INPUT_REQUIRED) {
```
codyzwief: Do we want to consider any limit or maximum to the number of iterations we'll allow here?

mcp/src/main/java/io/airlift/mcp/operations/legacy/MrtrEmulator.java
codyzwief: According to the spec, `InputRequests` doesn't have to be specified:

> Servers MUST include at least one of inputRequests or requestState in every InputRequiredResult response.

Should we remove this `checkArgument` and just pass through to an empty `inputResponses`?

mcp/src/main/java/io/airlift/mcp/operations/legacy/MrtrEmulator.java
```diff
@@ -0,0 +1,67 @@
...
+                    }
+                    response.result().ifPresent(value -> inputResponses.put(key, value));
+                }
+                catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                    throw exception(INTERNAL_ERROR, e, "Interrupted while waiting for server-to-client request");
+                }
+                catch (TimeoutException e) {
+                    throw exception(REQUEST_TIMEOUT, e);
+                }
```
codyzwief, line 62: I don't have any good ideas here but how could we gracefully handle some timeout or communication blip if there are multiple input requests and all succeed except one?

mcp/src/main/java/io/airlift/mcp/operations/legacy/MrtrEmulator.java
```diff
@@ -0,0 +1,67 @@
...
+            });
+
+            result = handler.apply(result.requestState(), inputResponses);
```
codyzwief: nit: might want to `ImmutableMap.copyOf(inputResponses)`

codyzwief (approved):

mcp/src/main/java/io/airlift/mcp/operations/legacy/MrtrEmulator.java
```diff
@@ -41,27 +61,50 @@ <T extends InputRequests<T>> T emulate(LegacyRequestContextImpl requestContext,
             Map<String, InputRequest> inputRequests = result.inputRequests().orElseGet(ImmutableMap::of);
             checkArgument(!inputRequests.isEmpty(), "Input requests may not be empty");
 
+            Map<String, Future<Optional<Object>>> futures = new HashMap<>();
+            inputRequests.forEach((key, inputRequest) -> futures.put(key, executorService.submit(() -> resolve(requestContext, inputRequest))));
+
             Map<String, Object> inputResponses = new HashMap<>();
-            inputRequests.forEach((key, inputRequest) -> {
+            futures.forEach((key, future) -> {
```
Randgalt: What's the benefit of this thread? The emulated MRTR needs to run in the same thread as the request. Why create a thread here just to block on it?

mcp/src/main/java/io/airlift/mcp/model/CallToolResult.java
```diff
@@ -90,6 +91,7 @@ public CallToolResult withResultType(ResultType resultType)
...
     {
-        return new CallToolResult(ImmutableList.of(new Content.TextContent(mcpClientException.unwrap().errorDetail().message())), Optional.empty(), true, Optional.empty());
+        return new CallToolResult(ImmutableList.of(new Content.TextContent(mcpClientException.unwrap().errorDetail().message())), Optional.empty(), true, Optional.empty())
```
Randgalt: It's not clear to me that this is correct. For legacy operations, clients will not expect the result type field which is why it's optional. I'm -1 on this change (and related changes).

mcp/src/main/java/io/airlift/mcp/model/InputRequests.java
Randgalt: It's optional because of legacy operations. This is a new field and we might break old clients that aren't expecting this field.

mcp/src/main/java/io/airlift/mcp/model/CallToolResult.java
Randgalt: Again, ResultType is new and we don't want to break old clients.

mcp/src/main/java/io/airlift/mcp/model/InputResponses.java
Randgalt: We can add a utility for that. I have one I've been working on for the draft client.

mcp/src/main/java/io/airlift/mcp/operations/legacy/MrtrEmulator.java
Randgalt: I considered that but my belief is that this is outside the scope of the library. The only thing the server could do is throw an exception. Further, we can only really check for elicitation and sampling. All other input requests are opaque. Maybe a separate Elicitation utility of some kind as that's the only pre-defined input request that will be supported in future protocols.

mcp/src/main/java/io/airlift/mcp/operations/legacy/MrtrEmulator.java
Randgalt: I added that on the draft client side. Maybe we can add it later. This PR is getting old and stale.

mcp/src/main/java/io/airlift/mcp/model/CallToolResult.java
Randgalt: Result type is handling in a different, more compatible manner now.

mcp/src/main/java/io/airlift/mcp/model/CallToolResult.java
Randgalt: Result type is handling in a different, more compatible manner now.

mcp/src/main/java/io/airlift/mcp/model/InputRequests.java
Randgalt: Result type is handling in a different, more compatible manner now.

mcp/src/test/java/io/airlift/mcp/ConformanceEndpoints.java
```diff
@@ -274,4 +290,199 @@ public void testTriggerToolChange(McpRequestContext requestContext)
...
+            ElicitRequestForm elicitRequestForm = new ElicitRequestForm("What is your name?", schema);
+            builder.add("user_name", METHOD_ELICITATION_CREATE, elicitRequestForm);
+        }
+
+        if (requestContext.clientCapabilities().sampling().isPresent()) {
+            CreateMessageRequest createMessageRequest = new CreateMessageRequest(Role.USER, new TextContent("Generate a greeting"), 100);
+            builder.add("greeting", METHOD_SAMPLING_CREATE_MESSAGE, createMessageRequest);
+        }
+
+        return builder.build();
```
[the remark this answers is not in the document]
Randgalt: This is as designed

mcp/src/main/java/io/airlift/mcp/McpException.java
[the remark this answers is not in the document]
Randgalt: This is false, as we keep telling you.

mcp/src/main/java/io/airlift/mcp/operations/OperationsImpl.java
```diff
@@ -203,14 +211,21 @@ public static ReadResourceResult readResources(McpEntities entities, McpRequestC
...
+                    return new McpClientException(exceptionWithData(INVALID_PARAMS, "Resource not found: " + readResourceRequest.uri(), data));
                 });
+    }
 
-        return new ReadResourceResult(resourceContents);
+    public static boolean isInputRequired(Object result)
+    {
+        if (result instanceof InputRequests inputRequests) {
+            return inputRequests.inputRequests().map(map -> !map.isEmpty())
+                    .orElse(false);
```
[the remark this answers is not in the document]
Randgalt: Disagree

mcp/src/test/java/io/airlift/mcp/ConformanceEndpoints.java
```diff
@@ -274,4 +290,199 @@ public void testTriggerToolChange(McpRequestContext requestContext)
...
+    }
+
+    public record PromptContext(String context) {}
+
+    @McpPrompt(name = "test_input_required_result_prompt", description = "MRTR tests")
+    public GetPromptResult testInputRequiredResultPrompt(InputResponses<?> inputResponses)
+    {
+        Optional<ElicitResult> elicitResult = inputResponses.getInputResponse("user_context")
+                .map(value -> jsonMapper.convertValue(value, ElicitResult.class));
+        if (elicitResult.isPresent() && (elicitResult.orElseThrow().action() == ElicitResult.Action.ACCEPT) && inputResponses.requestState().orElse("").equals("beep-bop")) {
```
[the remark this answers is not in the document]
Randgalt: Disagree

mcp/src/main/java/io/airlift/mcp/operations/legacy/MrtrEmulator.java
Randgalt: This is technically true but it poses a problem. If we allow empty inputRequests we'd get a spinning busy loop as there would be no server to client requests posted. So, we need to prohibit this.

mcp/src/main/java/io/airlift/mcp/operations/OperationsImpl.java
```diff
@@ -203,14 +210,21 @@ public static ReadResourceResult readResources(McpEntities entities, McpRequestC
...
                 });
+    }
 
-        return new ReadResourceResult(resourceContents);
+    public static boolean isInputRequired(Object result)
+    {
+        if (result instanceof InputRequests inputRequests) {
+            return inputRequests.requestState().isPresent()
+                    || inputRequests.inputRequests().map(map -> !map.isEmpty()).orElse(false);
+        }
```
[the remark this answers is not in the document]
Randgalt: Per the MRTR spec, this is correct. The MRTR emulator handles empty `inputRequests` separately.

wendigo (approved):

mcp/src/main/java/io/airlift/mcp/operations/legacy/MrtrEmulator.java
```diff
@@ -0,0 +1,69 @@
...
+    {
+        timeout = config.getEventStreamingTimeout().toJavaTime();
+    }
+
+    <T extends InputRequests> T emulate(LegacyRequestContextImpl requestContext, T result, BiFunction<Optional<String>, Map<String, Object>, T> handler)
+    {
+        while (isInputRequired(result)) {
+            Map<String, InputRequest> inputRequests = result.inputRequests().orElseGet(ImmutableMap::of);
+            checkArgument(!inputRequests.isEmpty(), "Input requests may not be empty for MRTR emulation");
+
```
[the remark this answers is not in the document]
Randgalt: `IllegalArgumentException` is what we want

PR #2087, opened 2026-08-03
Support reload-safe application lifecycles

jvanzyl:
## Summary

- add process-wide observation of successful and failed bootstrap attempts
- clean up lifecycle-managed resources when injector creation fails partway through startup
- fully destroy retired Jetty servers so exported MBeans do not survive shutdown

## Motivation

Development tools can start and retire multiple application generations in the same JVM. They need to observe bootstrap completion without changing application assembly, and every retired or rejected generation must release the resources it created.

Previously, a failure during eager singleton construction could leave lifecycle-managed instances running because bootstrap returned no injector to the caller. Jetty shutdown also stopped the server without destroying it, leaving stale MBean registrations behind for later generations.

This change makes those lifecycle boundaries explicit while preserving ordinary application startup behavior. Bootstrap success is reported only after lifecycle startup completes, cleanup failures are suppressed onto the original startup failure, and Jetty destruction remains paired with scheduled-executor cleanup.

## Validation

- bootstrap success and failure observation coverage
- reverse-order cleanup after partial startup failure
- listener failure isolation coverage
- Jetty MBean removal after shutdown
- repeated in-process application generation exercise

### DIFF
--- a/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
@@ -48,6 +48,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
@@ -72,6 +73,8 @@ import static java.util.Objects.requireNonNull;
  */
 public class Bootstrap
 {
+    private static final CopyOnWriteArrayList<BootstrapListener> listeners = new CopyOnWriteArrayList<>();
+
     private final String name;
     private final Logger log;
 
@@ -115,6 +118,23 @@ public class Bootstrap
         this.name = requireNonNull(name, "name is null");
         this.modules = ImmutableList.copyOf(modules);
         this.log = Logger.get(name);
+    }
+
+    /**
+     * Adds a class-loader-wide listener for bootstrap attempts that begin after registration.
+     */
+    public static void addListener(BootstrapListener listener)
+    {
+        listeners.add(requireNonNull(listener, "listener is null"));
+    }
+
+    /**
+     * Removes a class-loader-wide listener from bootstrap attempts that begin after removal.
+     * An attempt already in progress continues using its initial listener snapshot.
+     */
+    public static void removeListener(BootstrapListener listener)
+    {
+        listeners.remove(requireNonNull(listener, "listener is null"));
     }
 
     public Bootstrap setRequiredConfigurationProperty(String key, String value)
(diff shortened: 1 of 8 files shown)


http-server/src/main/java/io/airlift/http/server/HttpServer.java
```diff
@@ -556,15 +558,48 @@ public void stop()
...
+        catch (Throwable shutdownFailure) {
+            failure = addFailure(failure, shutdownFailure);
+        }
 
-        if (scheduledExecutorService != null) {
-            scheduledExecutorService.shutdown();
+        if (failure != null) {
+            throwIfInstanceOf(failure, Exception.class);
+            throwIfUnchecked(failure);
+            throw new AssertionError(failure);
```
wendigo: RuntimeException

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -117,6 +120,16 @@ public Bootstrap(String name, Iterable<? extends Module> modules)
         this.log = Logger.get(name);
     }
 
+    public static void addListener(BootstrapListener listener)
+    {
+        listeners.add(requireNonNull(listener, "listener is null"));
```
nineinchnick, line 128: It should not be possible to add or remove listeners if the state is not `UNINITIALIZED`

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -343,33 +356,65 @@ public Set<ConfigPropertyMetadata> configure()
...
-        Injector injector = Guice.createInjector(Stage.PRODUCTION, moduleList.build());
+            // create the injector
+            Injector injector = Guice.createInjector(Stage.PRODUCTION, moduleList.build());
 
-        injector.getInstance(LifeCycleManager.class).start();
+            injector.getInstance(LifeCycleManager.class).start();
+            notifyInitialized(injector);
+            return injector;
+        }
+        catch (RuntimeException | Error failure) {
```
nineinchnick: Do not catch `Error`. From [the JDK docs](https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html):
```
An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch.
```

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -343,33 +356,65 @@ public Set<ConfigPropertyMetadata> configure()
...
+            moduleList.addAll(modules);
 
-        // create the injector
-        Injector injector = Guice.createInjector(Stage.PRODUCTION, moduleList.build());
+            // create the injector
+            Injector injector = Guice.createInjector(Stage.PRODUCTION, moduleList.build());
 
-        injector.getInstance(LifeCycleManager.class).start();
+            injector.getInstance(LifeCycleManager.class).start();
+            notifyInitialized(injector);
```
nineinchnick: How are listeners different from the `LifeCycleManager`? This should be documented somewhere so users of the Bootstrap module know when to use either one of them.

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
```diff
@@ -386,6 +387,12 @@ public Injector initialize()
             return injector;
         }
         catch (RuntimeException | Error failure) {
+            try {
+                lifeCycleModule.stopAfterInitializationFailure();
+            }
+            catch (RuntimeException | Error cleanupFailure) {
```
nineinchnick: Don't catch `Error`

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleManager.java
```diff
@@ -178,7 +190,22 @@ public void stop()
...
     {
-        if (!state.compareAndSet(State.STARTED, State.STOPPING)) {
+        stop(handler, false);
+    }
+
+    private void stop(LifeCycleStopFailureHandler handler, boolean stopBeforeStarted)
+    {
+        if (stopBeforeStarted) {
+            State currentState;
+            do {
```
nineinchnick, line 202: Why is this loop needed here? Consider adding a code comment.

bootstrap/src/main/java/io/airlift/bootstrap/LifeCycleModule.java
```diff
@@ -89,11 +96,22 @@ private <T> void provision(ProvisionInvocation<T> provision)
     @Singleton
     public LifeCycleManager getLifeCycleManager()
     {
-        LifeCycleManager lifeCycleManager = new LifeCycleManager(name, injectedInstances, lifeCycleMethodsMap);
+        LifeCycleManager lifeCycleManager = new LifeCycleManager(name, List.of(), lifeCycleMethodsMap);
         this.lifeCycleManager.set(lifeCycleManager);
+        for (Object instance : injectedInstances) {
```
nineinchnick, line 102: Why is this change needed? The `LifeCycleManager` constructor does this.

bootstrap/src/test/java/io/airlift/bootstrap/TestBootstrap.java
```diff
@@ -497,7 +497,7 @@ public void stop()
     public static class FailingInstance
     {
         @Inject
-        public FailingInstance(StartedInstance startedInstance)
+        public FailingInstance(@SuppressWarnings("UnusedVariable") StartedInstance startedInstance)
```
nineinchnick, line 568: Should this be squashed with the previous commit, so every commit compiles?

nineinchnick (approved):

bootstrap/src/main/java/io/airlift/bootstrap/Bootstrap.java
jvanzyl: These listeners are intentionally class-loader-wide rather than owned by an individual Bootstrap instance, so there is no single instance state against which registration can be gated. Each initialize() attempt snapshots the registry at entry: additions and removals affect later attempts only and cannot mutate delivery for an attempt already in progress. The Javadocs and deterministic module-configuration mutation test now define and verify that contract.

nineinchnick: I missed that this method is static, thanks!

PR #2100, opened 2026-08-17
Support SecretKeySpec keystore entries in KeystoreSecretProvider

niteshy:
KeystoreSecretProvider previously read only PBE-protected SecretKeyEntry values via getEntry. Some credential tooling writes secrets as UTF-8 bytes in a SecretKeySpec using setKeyEntry; read those via getKey before falling back to the existing PBE path.

Also add keystore-entry-password config, normalize aliases to lowercase, and extract shared read logic into KeystoreSecretUtils.

Trino (and other Airlift consumers) can use this to resolve credentials from JCEKS files produced by legacy Java KeyStore credential stores.

### DIFF
--- a/secrets-keystore-plugin/src/main/java/io/airlift/secrets/keystore/KeystoreSecretProvider.java
+++ b/secrets-keystore-plugin/src/main/java/io/airlift/secrets/keystore/KeystoreSecretProvider.java
@@ -16,26 +16,35 @@ package io.airlift.secrets.keystore;
 import com.google.inject.Inject;
 import io.airlift.spi.secrets.SecretProvider;
 
-import javax.crypto.SecretKeyFactory;
-import javax.crypto.spec.PBEKeySpec;
-
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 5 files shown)


wendigo:
I don't think that we want to support "legacy" use cases.

secrets-keystore-plugin/src/main/java/io/airlift/secrets/keystore/KeystoreSecretProvider.java
```diff
@@ -16,26 +16,35 @@
...
 
     @Inject
     public KeystoreSecretProvider(KeystoreSecretProviderConfig config)
             throws GeneralSecurityException, IOException
     {
-        keystorePassword = config.getKeyStorePassword().toCharArray();
+        requireNonNull(config, "config is null");
+        String configuredKeystorePassword = config.getKeyStorePassword();
+        keystorePassword = configuredKeystorePassword.toCharArray();
+        String configuredEntryPassword = config.getKeyStoreEntryPassword() != null
```
electrum: Can use requireNonNullElse

PR #2133, opened 2026-09-13
Add OpenID Connect  token exhange secrets provider and Secret to SPI

gertjanal:
# Add dynamic (runtime) secrets provider to OpenID Connect token-exchange

This PR is the successor of this PR: https://github.com/trinodb/trino/pull/30944, where the goal is to allow oauth2 token exchange per Identity. The current secrets provider implementation does not seem dynamic (per user request), this PR makes that possible including the support of oidc.

How does it work?

The secretsResolver can be asked for a Secret, which can be more than just a simple resolved String. In this PR, the `HttpHeadersSecret` is added to return a class that can tell the http client which headers to use for authorization.

The `BiFunction<String, Optional<Object>> contextResolver` argument can be used to ask the caller (trino plugin) for additional context, in my case the oauth access token of the current user.

# TODO
[ ] Discuss if this SPI change is what we want (introduction of `Secret` class)
[ ] Discuss if we like the httpclient request builder `setHeaders(Secret)` (I'm not fully convinced, although the code for using trino plugins looks clean)

_Note: When ready I will add the linked Trino PR to this description_


http-client/src/main/java/io/airlift/http/client/Request.java
```diff
@@ -303,6 +304,13 @@ public Builder setHeader(HeaderName name, String value)
             return this;
         }
 
+        @CanIgnoreReturnValue
+        public Builder setHeaders(HttpHeadersSecret headersSecret)
```
gertjanal, line 308: pom change is needed, not sure if we want this but the usage in the trino code is very clean:

```
HttpHeadersSecret secret = secretsResolver.resolveSecret(config.getSecretName(), "access_token", HttpHeadersSecret.class, () -> ...);

preparePost().setUri(...).setHeaders(secret)...;
```

secrets-oidc-plugin/src/main/java/io/airlift/secrets/oidc/OidcSecretProvider.java
```diff
@@ -0,0 +1,118 @@
...
+                                .setUri(config.getTokenUrl())
+                                .setMethod("POST")
+                                .setHeader(CONTENT_TYPE, "application/x-www-form-urlencoded")
+                                .setBodyGenerator(new FormDataBodyBuilder()
+                                        .addField("client_id", requireNonNull(config.getClientId(), "clientId is null"))
+                                        .addField("client_secret", requireNonNull(config.getClientSecret(), "clientSecret is null"))
+                                        .addField("grant_type", "urn:ietf:params:oauth:grant-type:token-exchange")
+                                        .addField("subject_token", subjectToken)
+                                        .addField("subject_token_type", "urn:ietf:params:oauth:token-type:access_token")
+                                        .addField("scope", config.getScopes())
```
gertjanal, line 68: we might even want the contextResolver to provide the scope and audience (later PR?)

configuration/src/main/java/io/airlift/configuration/secrets/SecretsResolver.java
```diff
@@ -78,4 +80,11 @@ public String resolveSecret(String secretProviderName, String keyName)
         checkArgument(secretProvider != null, "No secret provider for key '%s'", secretProviderName);
         return secretProvider.resolveSecretValue(keyName);
     }
+
+    public <T extends Secret> Optional<T> resolveSecret(String secretProviderName, String keyName, Class<T> type, Function<String, Optional<Object>> contextResolver)
```
gertjanal, line 84: Note that contextResolver can be replaced by a `Map<String, Object>`, but asking for things might be better than including values that might not be used

secrets-oidc-plugin/src/main/java/io/airlift/secrets/oidc/OidcSecretProviderConfig.java
```diff
@@ -0,0 +1,123 @@
...
+    private String clientSecret;
+    private URI tokenUrl;
+    private String audience;
+    private String scopes;
+    private Duration cacheExpiry = Duration.valueOf("1h");
+
+    @AssertTrue
+    public boolean isHttpsEnabled()
+    {
+        return tokenUrl != null && (tokenUrl.getScheme().equals("https") || tokenUrl.getHost().equals("localhost") || tokenUrl.getHost().endsWith(".local"));
```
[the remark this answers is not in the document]
gertjanal: TODO remove local checks before merge (disables local setup poc instructions)

secrets-oidc-plugin/src/main/java/io/airlift/secrets/oidc/OidcSecretProvider.java
```diff
@@ -0,0 +1,101 @@
...
+                                        .build())
+                                .build();
+
+                        StringResponseHandler.StringResponse response = httpClient.execute(request, createStringResponseHandler());
+                        if (response.getStatusCode() != 200) {
+                            throw new IllegalStateException(config.getTokenUrl() + " returned " + response.getStatusCode() + ": " + response.getBody());
+                        }
+
+                        TokenExchangeResponse exchangeResponse = TOKEN_EXCHANGE_RESPONSE_JSON_CODEC.fromJson(response.getBody());
+                        return new BearerTokenCredential(exchangeResponse.accessToken(), Instant.now().plusSeconds(exchangeResponse.expiresIn()).minusSeconds(30));
```
gertjanal, line 79: Fail when expiry is shorter than configured cache expiry?

configuration/src/main/java/io/airlift/configuration/secrets/SecretsResolver.java
kokosing, line 84: I don't follow this. Secret provider is a plugin framework. There could be many configured or none at once. In place where you ask for a secret resolution, caller has no clue how a secret provider is configured so passing some information here specific for a given plugin seems odd. I believe if anything we should pass some generic information that plugin make later use in a way its like. 

For example one can pass a "session" object that then internally is used to create http headers that can be used to retrieve given header.

